PR #1, opened 2015-06-22
Add "Gir.toml" with default params

EPashkin:


gkoz:
This is a nice direction but seems unfinished, am I correct?

src/main.rs
```diff
@@ -1,21 +1,50 @@
...
+    
+    let girs_dir = match args.get_str("-d") {
+        "" => toml.lookup("options.girs_dir")
+                  .unwrap_or_else(|| panic!("No options.girs_dir in config"))
+                  .as_str().unwrap(),
+        a => a
+    };
+    //println!("Dir: {:?}", girs_dir);
+
+    let gir = match args.get_str("<gir>") {
```
gkoz: Calling the library we're interested in currently just "gir" seems confusing, not specific enough.

gkoz:
Don't we want a command line parameter to point to a config file (they might be several of them)?

src/main.rs
EPashkin: You called it that, gi_rust (as analog of grust)?

EPashkin:
Yes it only first step.
Currenly loading single gtk30.gir takes 30sec, IMHO too long :(
I disabled loading others for debug purposes (not commit that)

src/main.rs
EPashkin: Sorry, misreaded (think about repository name)
I think gir will be universal converter in future.

EPashkin: or may be stay with <lib>

EPashkin:
Not sure about config file location will be necessary. May be in future.

gkoz:
> Currenly loading single gtk30.gir takes 30sec, IMHO too long

The debug build is slow but if you build with `--release` (and run `target/release/gir`) the speed is not bad.

src/main.rs
gkoz: The value of this argument is the name of a specific library (Gtk, Gdk, etc) we want to work with.

EPashkin:
Yes, with release it 3sec or (5sec with all libs). Much better.

PR #2, opened 2015-06-22
Add object list in config

EPashkin:


src/main.rs
```diff
@@ -1,21 +1,54 @@
...
+Usage: gir [-d <girs_dir>] [<gir>]
+
+Options:
+    -d PATH            Directory for girs
 ";
 
 fn main() {
     let args = Docopt::new(USAGE).unwrap()
         .parse().unwrap_or_else(|e| e.exit());
+    //println!("{:?}", args);
```
gkoz: Do these debugging lines still serve any purpose?

src/main.rs
```diff
@@ -1,21 +1,54 @@
...
+    //println!("Dir: {}", args.get_str("-d"));
+    //println!("File: {}", args.get_str("<gir>"));
+    
+    let toml = read_toml("Gir.toml");
+    // println!("Toml: {:?}", toml);
+    
+    let girs_dir = match args.get_str("-d") {
+        "" => toml.lookup("options.girs_dir")
```
gkoz: I still believe this needs a more specific name. What's a "gir"? `gir = "Gtk-3.0"`? Could be `library`, `namespace`, something with a meaning.

src/main.rs
```diff
@@ -1,21 +1,54 @@
...
+                  .unwrap_or_else(|| panic!("No options.girs_dir in config"))
+                  .as_str().unwrap(),
+        a => a
+    };
+    //println!("Dir: {:?}", girs_dir);
+
+    let gir = match args.get_str("<gir>") {
+        "" => toml.lookup("options.gir")
```
gkoz: I still believe this needs a more specific name. What's a "gir"? gir = "Gtk-3.0"? Could be library, namespace, something with a meaning.

EPashkin: Parameter `<library>` with `let library_name = match ...` ?

gkoz: Seems okay.

EPashkin:
Removed all debug prints, except 'objects'

gkoz:
Thanks.

Consider that the generator is capable of knowing if something is a `GObject`, a `GtkWidget`, etc so putting that into the config seems like wasted effort.

EPashkin:
Yes. Don't looked at `library` at moment of writing.
I fix this and name (to "Gtk.Widget") in next PR.

PR #4, opened 2015-06-23
Create empty widget file

EPashkin:
Preview PR, don't merge now.

### DIFF
--- a/src/chunk.rs
+++ b/src/chunk.rs
@@ -1,0 +1,77 @@
+use std::iter::*;
+use std::fmt::Debug;
+
+/// Chunk of code text
+#[derive(Debug)]
+pub enum Chunk {
+    Text(String),
+}
+
+impl ToString for Chunk {
+    #[inline]
+    fn to_string(&self) -> String {
+        match self {
+            &Chunk::Text(ref str) => str.clone(),
+        }
+    }
+}
+
+pub trait IntoChunk {
+    fn into_chunk(&self) -> Chunk;
+}
+
+impl IntoChunk for String {
+    #[inline]
+    fn into_chunk(&self) -> Chunk {
+        Chunk::Text(self.clone())
+    }
+}
+
+impl<'a> IntoChunk for &'a str {
+    #[inline]
+    fn into_chunk(&self) -> Chunk {
+        Chunk::Text(self.to_string())
+    }
+}
+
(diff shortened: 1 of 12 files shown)


src/chunk.rs
```diff
@@ -0,0 +1,105 @@
+use std::iter::*;
+use std::fmt::Debug;
+
+/// Chunk of code text
+#[derive(Debug)]
+pub enum Chunk {
+    Text(String),
+    EndLine,
+}
```
gkoz: If the `Chunk`'s purpose is to deal with newlines, consider a wrapper around `File` that would just do `s/\n/\r\n/g`, should be simpler.

EPashkin: Chunk's purpose to separate unchunkable part of string.
Ex. for `    fn add<T: Upcast<Widget>>(&self, widget: &T) {`
will be these chunks `Indent(1)`, `Text("fn add<T: Upcast<Widget>>(")`, `Text("&self,")`,
`Separator`, `Text("widget: &T")`, `Text(") {")`
Formater will split long result lines by `Separator`.
For line width 40, `Text("widget: &T")` will go to new string with `Indent(2)` instead of `Separator` before him.
You know better solution?

EPashkin: And CRLF varianted for normal diff under Windows.

gkoz: I'd hate to have another Rust formatter here. There are two of them already, both unfinished.

EPashkin: You about https://github.com/nrc/rustfmt?
Ok. Then it will be even easier.

gkoz: They may not be good at handling the long lines currently, but I believe it more productive to improve the existing formatters.

EPashkin: Splitting existing code more difficult than doing it on generation.

EPashkin:
Final commit for this PR.
Uncunking done later.

gkoz:
Not creating a backup if one already exists is unconventional. If we're using version control, backups may not even be an issue, but creating them once and then only keeping the stale ones seems like a footgun.

EPashkin:
Fix

gkoz:
Great!
Should I merge now?

EPashkin:
Yes

gkoz:
Thanks for working on this!

PR #8, opened 2015-06-25
Add analysis and general codegens

EPashkin:


EPashkin:
Not sure about name `Library::find_type_unwraped`

src/analysis/widget.rs
```diff
@@ -0,0 +1,33 @@
...
+        type_
+    }
+}
+
+pub fn new(env: &Env, obj: &GObject) -> Info {
+    let full_name = obj.name.clone();
+
+    let class_id = env.library.find_type_unwraped(0, &full_name, "Class");
+
+    let type_name = split_namespace_name(&full_name).1.into();
```
gkoz: This name may be confusing because in the xmls, `type-name` has the prefix (e.g. `GtkWidget`, without a dot). And possibly it's `Class::glib_name` that should've been named `type_name`.

EPashkin: `rust_type_name`?

gkoz: It can be argued that the short name is just `name`, same as in the xmls.

EPashkin: If you prefer change Class::glib_name to glib_type_name, i do it after this merge

EPashkin: IMHO anayzing look from rust side, so only glib_ needed for differentiate

EPashkin: And it know nothing about xml

gkoz:
Fine, I'll merge and tweak the names afterwards.

PR #9, opened 2015-06-25
Generate static type

EPashkin:
Check parser changes

### DIFF
--- a/src/library.rs
+++ b/src/library.rs
@@ -159,6 +159,7 @@ pub struct Interface {
 pub struct Class {
     pub name: String,
     pub glib_name: String,
+    pub get_type_func_name: String,
     pub functions: Vec<Function>,
     pub parent: Option<TypeId>,
     pub implements: Vec<TypeId>,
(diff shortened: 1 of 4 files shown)


src/library.rs
```diff
@@ -159,6 +159,7 @@ pub struct Interface {
 pub struct Class {
     pub name: String,
     pub glib_name: String,
+    pub get_type_func_name: String,
```
gkoz: Hmm, maybe it's time to make these two names consistent with the xmls.

src/parser.rs
```diff
@@ -165,11 +165,13 @@ impl Library {
         }
 
         let glib_name = attrs.get("type-name").unwrap_or(name);
+        let get_type_func_name = attrs.get("get-type").unwrap_or("None");
```
gkoz: This should use the same error handling as `"name"`:
```rust
let get_type_func_name = try!(
    attrs.get("get-type").ok_or_else(|| error!("Missing get-type attribute", parser)));
```
and located above the loop so the `parser` would report the correct position.

EPashkin: Ok, do it after #8 merged

src/library.rs
EPashkin: glib_type_name, glib_get_type ?

EPashkin:
Fixed parser, change class names leave to you

PR #12, opened 2015-06-25
Change default output directory

EPashkin:
Found strange bug, sometimes it analyzing 1-2 files (always including RadioButton).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 target
+gtk/
--- a/Gir.toml
+++ b/Gir.toml
@@ -1,7 +1,7 @@
 [options]
 girs_dir = "../gir-files"
 library = "Gtk-3.0"
-target_path = "../gtk"
+target_path = "./gtk"
 
 [[object]]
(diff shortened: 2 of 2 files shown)


gkoz:
> sometimes it analyzing 1-2 files (always including RadioButton)

Fixed in 97726b9.

EPashkin:
Yes, tnx, seems fixed. But why in skipps items that before `Ignored`?

gkoz:
The order of the toml items is not preserved, there's a `HashMap` somewhere I guess. This surprised me too.

EPashkin:
Yes, it random-based hashed, and I used it in GObjects :(
Started change to Vec, as i may need sort order by added https://github.com/EPashkin/gir/tree/change_GObjects_to_Vec.
But now i think is ok.

PS. Can I comment `show(&library);` in next PR?

gkoz:
> Can I comment show(&library); in next PR?

Yes, it's pretty useless.

PR #22, opened 2015-07-01
Declarations return values

EPashkin:


EPashkin:
Can you give me example code for function, returning Records (`gtk_container_get_path_for_child`) ?
Then I add it records this PR.

gkoz:
I don't have a complete plan for the "boxed" types yet. The function would simply call `from_glib_full` on the pointer but we can't assume that any type has a `FromGlibPtr` implementation. Might have to use a whitelist at first.

EPashkin:
Then no Records for now, these functions will be commented
Ready for merge

EPashkin:
Or I can add commit, that changes
```
//fn get_path_for_child(TODO: Params) -> /*Unknown kind*/Unknows rust type: "WidgetPath";
```
to something like
```
//fn get_path_for_child(TODO: Params) -> /*TODO: Record*/WidgetPath;
```

gkoz:
> Or I can add commit, that changes

This might matter if we want to copy-paste some code from the generated files into the manually maintained ones, having to remove these notes would be annoying.

EPashkin:
Fixed type in `Unknows rust type`

EPashkin:
For `this` you refer current line or offered?

gkoz:
I'm fine with both versions right now, it's just if we were to copy these fragments somewhere else the comments might be in the way. But that can be fixed when it actually becomes an issue.

EPashkin:
So I write commit or you just merge?

gkoz:
I'll merge.

PR #25, opened 2015-07-02
Generate parameters values

EPashkin:
Not all. Without trait constraint.

### DIFF
--- a/src/analysis/rust_type.rs
+++ b/src/analysis/rust_type.rs
@@ -21,6 +21,11 @@ impl ToRustType for library::Fundamental {
             &Int => "i32".into(),      //maybe dependent on target system
             &UInt => "i32".into(),     //maybe dependent on target system
 
+            &Float => "f32".into(),
+            &Double => "f64".into(),
+
+            &Utf8 => "String".into(),
+
             &Type => "Type".into(),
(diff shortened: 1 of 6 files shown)


src/analysis/rust_type.rs
```diff
@@ -21,6 +21,11 @@ impl ToRustType for library::Fundamental {
             &Int => "i32".into(),      //maybe dependent on target system
             &UInt => "i32".into(),     //maybe dependent on target system
 
+            &Float => "f32".into(),
+            &Double => "f64".into(),
+
+            &Utf8 => "String".into(),
```
gkoz: We want `String` for return values but `&str` for parameters.

src/codegen/parameter.rs
```diff
@@ -0,0 +1,36 @@
...
+    fn to_parameter(&self, library: &library::Library) -> String;
+}
+
+impl ToParameter for library::Parameter {
+    fn to_parameter(&self, library: &library::Library) -> String {
+        if self.instance_parameter {
+            "&self".into()
+        } else {
+            //TODO: Out parameters. Ex. gtk_range_get_range_rect
+            //TODO: change Utf8 to &str for inputs
```
EPashkin: Here TODO about string

PR #27, opened 2015-07-03
Remake rust type

EPashkin:
Not sure that it good way, but it helps comment all functions with unsupported parameters.
Also I do not like class names, but nothing normal could not come up

### DIFF
--- a/src/analysis/rust_type.rs
+++ b/src/analysis/rust_type.rs
@@ -1,49 +1,144 @@
+use std::result;
+
 use library;
 
+pub type Result = result::Result<String, String>;
+
+pub trait AsStr {
+    fn as_str(&self) -> &str;
+}
+
+impl AsStr for Result {
+    #[inline]
+    fn as_str(&self) -> &str {
+        self.as_ref().unwrap_or_else(|s| s)
+    }
+}
+
 pub trait ToRustType {
-    fn to_rust_type(&self) -> String;
+    fn to_rust_type(&self) -> Result;
 }
 
 impl ToRustType for library::Fundamental {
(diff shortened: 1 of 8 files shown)


src/analysis/rust_type.rs
```diff
@@ -1,49 +1,152 @@
+use std::result;
+
 use library;
 
+pub struct Result(result::Result<String, String>);
```
gkoz: Can't this be achieved with a type alias?

EPashkin: IMHO only with trait and trait impl

gkoz: Isn't the `as_str` method equivalent to `unwrap_or_else(|s| s)`?

EPashkin: Yes missed this.

gkoz: So it's a trade off between using the more verbose standard functions in place of `as_str` and wrapping values in this `Result` in 10 places?

EPashkin: Fixed

EPashkin: So you think, that better convert to trait now? String::as_str still unstable.

EPashkin: Ignore about String::as_str

gkoz: A trait makes more sense to me -- you just import it in modules where this sugar is needed instead of paying the price everywhere such a `Result` is created.

EPashkin:
Traited

gkoz:
Thanks!

Issue #21, opened 2015-06-30
Generate uses code


EPashkin:
It is necessary to generate a lines with "use xxx;".
I believe that it is better to add some mutable state to avoid duplication.
Not sure it will be added to `env::Env` via `RefCell` or to separate struct.
Now I see the following sources of necessary types:
- Common to all `glib::translate: *, glib::types, ffi`
- `object::{Object, Upcast}` with `DownCast` when constructors present
- parent types, that will be added in `impl_parents`
- interface types from currently not implemented function `impl_interfaces`
- functions parameters

gkoz:
We could avoid importing most of the types by taking advantage of reexports in the root namespace.
```rust
unsafe impl <Upcast<::Container>> for Bin { }
```

EPashkin:
It works if all usable type will reexported.
If some still needed use (Ex. `c_double`) then task still ever 
more difficul, we need do check on both "sides"

gkoz:
Maybe we do need to cache the lines before writing to the file, this would allow is to gather the list of types that must be imported.

EPashkin:
Yes, i proposed this from beginning. May be I was not clear enough.
I try add uses (with interfaces) tomorrow, as I say in beginning of last PR, I prefer get compilable code sooner.

EPashkin:
Closed with #34

PR #28, opened 2015-07-04
Generate function body

EPashkin:
Almost full functions generation.
Last "trait constraint" and "return out parameters", but before I prefer get compilable gtk code.
If you don't start #21 then say, I try do it.

### DIFF
--- a/src/codegen/function.rs
+++ b/src/codegen/function.rs
@@ -1,12 +1,12 @@
 use std::io::{Result, Write};
-use std::fmt;
 
 use analysis;
 use env::Env;
 use library;
-use super::return_value::ToReturnValue;
-use super::parameter::ToParameter;
+use super::function_body::Builder;
 use super::general::tabs;
+use super::parameter::ToParameter;
+use super::return_value::ToReturnValue;
 
 pub fn generate<W: Write>(w: &mut W, env: &Env, analysis: &analysis::functions::Info,
     in_trait: bool, only_declaration: bool, indent: i32) -> Result<()> {
@@ -50,17 +50,9 @@ pub fn declaration(library: &library::Library, analysis: &analysis::functions::I
     format!("fn {}({}){}", analysis.name, param_str, return_str)
 }
 
-macro_rules! write_to_vec(
-    ($dst:expr, $($arg:tt)*) => (
-        $dst.push(fmt::format(format_args!($($arg)*)))
-    )
-);
-
 pub fn body(analysis: &analysis::functions::Info) -> Vec<String> {
-    let mut v: Vec<String> = Vec::new();
-    //TODO: real generation
-    write_to_vec!(v, "unsafe {{");
-    write_to_vec!(v, "{}TODO: call ffi:{}()", tabs(1), analysis.glib_name);
-    write_to_vec!(v, "}}");
-    v
+    let mut builder = Builder::new();
+    builder.glib_name(&analysis.glib_name);
+
+    builder.generate()
 }
--- a/src/codegen/function.rs
+++ b/src/codegen/function.rs
@@ -7,6 +7,7 @@ use super::function_body::Builder;
 use super::general::tabs;
 use super::parameter::ToParameter;
 use super::return_value::ToReturnValue;
+use super::translate_from_glib::TranslateFromGlib;
 
 pub fn generate<W: Write>(w: &mut W, env: &Env, analysis: &analysis::functions::Info,
     in_trait: bool, only_declaration: bool, indent: i32) -> Result<()> {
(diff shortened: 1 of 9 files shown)


src/codegen/function_body.rs
```diff
@@ -0,0 +1,58 @@
...
+        format!("{}ffi::{}({}){}", self.from_glib_prefix,
+            self.glib_name, param_str, self.from_glib_suffix)
+    }
+    fn generate_parameters(&self) -> String {
+        let mut param_str = String::with_capacity(500);
+
+        for (pos, par) in self.parameters.iter().enumerate() {
+            if pos > 0 { param_str.push_str(", ") }
+            param_str.push_str(&par);
+        }
```
gkoz: Can this be done with `self.parameters.connect(", ")`?

src/codegen/function_body.rs
```diff
@@ -0,0 +1,58 @@
...
+    from_glib_prefix: String,
+    from_glib_suffix: String,
+    parameters: Vec<String>,
+}
+
+impl Builder {
+    pub fn new() -> Builder {
+        Default::default()
+    }
+    pub fn glib_name(mut self, name: &str) -> Builder {
```
gkoz: I don't think it matters here but FYI they [recommend against](https://aturon.github.io/ownership/builders.html#non-consuming-builders-(preferred):) consuming builders.

src/codegen/function_body.rs
EPashkin: Thank, don't know about it.

src/codegen/function.rs
```diff
@@ -50,17 +52,17 @@ pub fn declaration(library: &library::Library, analysis: &analysis::functions::I
...
-    let mut v: Vec<String> = Vec::new();
-    //TODO: real generation
-    write_to_vec!(v, "unsafe {{");
-    write_to_vec!(v, "{}TODO: call ffi:{}()", tabs(1), analysis.glib_name);
-    write_to_vec!(v, "}}");
-    v
+    //TODO: change to map on parameters with pass Vec<String> to builder
+    for par in &analysis.parameters {
+        let s = par.to_parameter_body(library, in_trait);
+        builder = builder.parameter(s);
```
gkoz: I guess this is some of the motivation for non-consuming builders: you'd just do `builder.parameter(s);` here.

src/codegen/function_body.rs
EPashkin: Thank again, forgot which is preferable

src/codegen/function.rs
```diff
@@ -50,17 +52,17 @@ pub fn declaration(library: &library::Library, analysis: &analysis::functions::I
...
-macro_rules! write_to_vec(
-    ($dst:expr, $($arg:tt)*) => (
-        $dst.push(fmt::format(format_args!($($arg)*)))
-    )
-);
+pub fn body(library: &library::Library, analysis: &analysis::functions::Info,
+    in_trait: bool) -> Vec<String> {
+    let mut builder = Builder::new()
+        .glib_name(&analysis.glib_name)
+        .from_glib(analysis.ret.to_return_value_body(&library, &analysis));
```
gkoz: Could you give a bit more thought to the names here? Wouldn't the `from_glib` and `to_return_value_body` methods have related names? Right now the only way to make sense of what happens here is to read a few function bodies.

EPashkin: I do not like just 'from_glib' instead `to_return_value_body` for trait.
But for Builder It really means "conversion from glib".
Nothing better not come to mind.

EPashkin:
Fixed builder and parameters joining.
Not changed `from_glib` and `to_return_value_body`

gkoz:
Could `to_return_value_body` be though of as `translate_from_glib` or `to_glue_from_glib` because it generates the translation glue? Conversely the input parameters would be handled by `translate_to_glib`.

EPashkin:
Only one "problem" translate_from_glib return "function" string and "translate_to_glib" - "method", 
may be needed some suffix?

src/codegen/function_body.rs
```diff
@@ -0,0 +1,49 @@
...
+    from_glib_prefix: String,
+    from_glib_suffix: String,
+    parameters: Vec<String>,
+}
+
+impl Builder {
+    pub fn new() -> Builder {
+        Default::default()
+    }
+    pub fn glib_name<'a>(&'a mut self, name: &str) -> &'a mut Builder {
```
gkoz: Did lifetime elision not handle this correctly? Reading the [TRPL](https://doc.rust-lang.org/book/lifetimes.html#lifetime-elision), it seems the return value would get the same lifetime as `self`.

gkoz:
> Only one "problem" `translate_from_glib` return "function" string and `translate_to_glib` - "method",
may be needed some suffix?

Generally, it's possible that `translate_to_glib` may need to support the "function" syntax as well, although I can't think of an example currently.

EPashkin:
Then it is better to make a difference now. `translate_from_glib_as_function` for method (don't changing trait name)?

gkoz:
The method name seems okay. Not sure if there's point to having the trait name be different, shorter maybe.

EPashkin:
Renamed. What about #21?

EPashkin:
IMHO ready to merge

gkoz:
Thanks.

PR #31, opened 2015-07-05
Combine ToRustType's implementations

EPashkin:
Your advice https://github.com/gkoz/gir/issues/30#issuecomment-118555191 is really reduced code in this case. I'll try it for other Traits.

IMHO Tuple want be named (`LibrarysTypeId` ?) and moved to library.

### DIFF
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -1,6 +1,6 @@
 use std::vec::Vec;
 
-use analysis::rust_type::{Result, ToParameterRustType, ToRustType};
+use analysis::rust_type::{Result, parameter_rust_type, rust_type};
 use env::Env;
 use library;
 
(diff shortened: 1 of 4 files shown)


src/analysis/functions.rs
```diff
@@ -26,21 +26,19 @@ pub fn analyze(env: &Env, type_: &library::Class, class_tid: library::TypeId) ->
...
 fn analyze_function(env: &Env, type_: &library::Function, class_tid: library::TypeId) -> Info {
-    let klass = env.type_(class_tid);
-
     let mut commented = false;
 
     let ret = if type_.ret.typ == Default::default() { None } else { Some(type_.ret.clone()) };
 
     if let Some(ref ret_) = ret {
-        let rust_type = env.type_(ret_.typ).to_parameter_rust_type(ret_.direction);
+        let rust_type = (&env.library, ret_.typ).to_parameter_rust_type(ret_.direction);
```
gkoz: I'm a little worried that we might be fooling ourselves here and this is no different than the conservative
```rust
let rust_type = rust_type::to_parameter(&env.library, ret_.typ, ret_.direction);
```

So the idea of implementing traits for tuples and even giving those tuples names may not have been a good advice after all.

src/analysis/rust_type.rs
```diff
@@ -19,67 +19,45 @@ pub trait ToRustType {
     fn to_rust_type(&self) -> Result;
 }
 
-impl ToRustType for library::Fundamental {
```
gkoz: This impl seems good as is, it doesn't depend on the library and namespace and you're right to be concerned about the method below growing too big.

gkoz:
I guess this experiment convinces me that adding either the namespace name or `ns_id` field to the types is the lesser evil.

src/analysis/functions.rs
EPashkin: No it was good, it allow remove library for trait function parameters, mostly unneeded.

src/analysis/rust_type.rs
EPashkin: Now the this file is significantly reduced.

EPashkin:
Currently `ns_id` not needed, merge this.

EPashkin:
Question: is not it better to call this tuple, as is will be used many times.

EPashkin:
Tuple better idea when not with combining. For rust_type.rs  you right, it not different for function call, and better be replaced. So don't merge, it now, I change traits to functions.

gkoz:
Maybe the implications of having such a composite type are even bigger. Could it replace `TypeId` in most places? What if there was a `TypeRef { library: &Library, tid: TypeId }` and that's what many methods on library would deal with?

EPashkin:
Strange, currently (after unpulled PR)  i used it only in rust_type.rs, in other places while refactoring, i removed library at all or used function. So now `TypeRef` uneeded.

gkoz:
Nice :)

EPashkin:
:( Missed one usage in ToTypeKind, will be converted to function, still unneeded.

EPashkin:
Done.

gkoz:
Thanks!

PR #32, opened 2015-07-05
Different names for types from other libs

### LINKED ISSUE #30: Problem with types from other libs
Now types from other libs shows wrong way. As example `gtk_scale_get_layout` uses `Pango.Layout`, but currently shows as 'Layout'.
To temporary fix problem I start branch https://github.com/EPashkin/gir/tree/comment_types_from_other_libs, but faced the problem: library's Class don't know, what namespace it belong.
Can I add ns:u16 to almost all types?
Final solution for this issue must be using `pango::Layout` string as type?

Also, can I delete row
```
library.namespace_mut(INTERNAL_NAMESPACE).name = INTERNAL_NAMESPACE_NAME.into();
```
and it current replacement as unnecessary?

EPashkin:
Fix #30

### DIFF
--- a/Gir.toml
+++ b/Gir.toml
@@ -1,6 +1,7 @@
 [options]
 girs_dir = "../gir-files"
-library = "Gtk-3.0"
+library = "Gtk"
+version = "3.0"
 target_path = "./gtk"
 
 [[object]]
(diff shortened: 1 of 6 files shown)


EPashkin:
Only question about private ns_id in TypeId, I added getter in #32, but maybe better make it pub

gkoz:
Not publishing the fields probably was a mistake.

EPashkin:
Removed ns_id() and rebased to master

PR #33, opened 2015-07-05
Generalization of widgets

EPashkin:
Not sure about good name for 'TypeKind::of()',  prefer 'for' but it keyword :(
Also about 'special_type'

### DIFF
--- a/Gir.toml
+++ b/Gir.toml
@@ -3,6 +3,7 @@ girs_dir = "../gir-files"
 library = "Gtk"
 version = "3.0"
 target_path = "./gtk"
+special_type = "Widget"
 
 [[object]]
 name = "Gtk.Widget"
--- a/src/analysis/general.rs
+++ b/src/analysis/general.rs
@@ -7,19 +7,19 @@ pub struct StatusedTypeId{
     pub status: GStatus,
 }
 
-pub trait IsSpecialType {
-    fn is_special_type(&self) -> bool;
+pub trait IsChildOfSpecialType {
+    fn is_child_of_special_type(&self) -> bool;
 }
 
-impl IsSpecialType for Class {
-    fn is_special_type(&self) -> bool {
-        self.glib_type_name == "GtkWidget" || self.parents.contains(&SPECIAL_TYPE_ID)
+impl IsChildOfSpecialType for Class {
+    fn is_child_of_special_type(&self) -> bool {
+        self.parents.contains(&SPECIAL_TYPE_ID)
     }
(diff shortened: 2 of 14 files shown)


EPashkin:
Forgot update test. Fixed.

gkoz:
I'm not aware of any library other than GTK that pulls those shenanigans with constructors returning a superclass. But is there anything special about `Widget` really? If we see the that a `Foo`'s constructor return `Widget` instead of `Foo` isn't this enough to know it needs to be downcasted?

EPashkin:
Main problem for me was that `IsWidget` (new `IsChildOfSpecialType`) needed library to check if it widget, so `ToTypeKind` needed library also. Also check for widget always call `find_type`.

EPashkin:
And don't forget about "Widget:from_glib_none", that appears in all constructors.

gkoz:
> And don't forget about `Widget:from_glib_none`, that appears in all constructors.

That's what I was thinking of: if the return type `X` doesn't match the constructor's `Y` -- just do `X::from_...().downcast_unchecked()`

EPashkin:
You right. Then maybe TypeKind::Special not needed at all. I try it.

EPashkin:
Tried https://github.com/gkoz/gir/pull/33#issuecomment-118631437 for constructor and it works. Maybe special kind (and all kinds) don't needed. I close this PR after full check.

EPashkin:
Really don't needed. Closing.

gkoz:
Lying about the types isn't limited to constructors, BTW. In `gtk_dialog_get_content_area` they tell us that the type is `Box` but the C type is `GtkWidget*` so supposedly that's what we're meant to put in the ffi definitions. That's going to be fun.
```xml
        <return-value transfer-ownership="none">
          <doc xml:space="preserve">the content area #GtkBox.</doc>
          <type name="Box" c:type="GtkWidget*"/>
        </return-value>
```

EPashkin:
Thanks, added TODO to `TranslateFromGlib`

gkoz:
Unfortunately the same happens in the other direction as well.
In the `gtk_file_chooser_button_new_with_dialog`:
```xml
          <parameter name="dialog" transfer-ownership="none">
            <doc xml:space="preserve">the widget to use as dialog</doc>
            <type name="Dialog" c:type="GtkWidget*"/>
          </parameter>
```

EPashkin:
IMHO with incoming parameters it not problem, only problem with return types

PR #34, opened 2015-07-05
Generate uses

EPashkin:
Current only common and parents

### DIFF
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -1,6 +1,7 @@
+use std::collections::HashSet;
 use std::vec::Vec;
 
-use analysis::rust_type::{Result, parameter_rust_type, rust_type};
+use analysis::rust_type::*;
 use env::Env;
 use library;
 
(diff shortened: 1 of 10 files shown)


gkoz:
Hmm, I don't see why the `Info` can't be borrowed mutably.

EPashkin:
I think you can pass &mut to many subfunction, but did not checked.

EPashkin:
Removed RefCell.
Don't merge now, i add interfaces in hour or two

EPashkin:
Now only build errors in Gtk is `pango::Layout`, and 36 unresolved `ffi::*` :smile:

EPashkin:
Ready to merge

src/analysis/functions.rs
```diff
@@ -14,21 +15,27 @@ pub struct Info {
...
 
-fn analyze_function(env: &Env, type_: &library::Function, class_tid: library::TypeId) -> Info {
+fn analyze_function(env: &Env, type_: &library::Function, class_tid: library::TypeId,
+    used_types: &mut HashSet<String>) -> Info {
     let mut commented = false;
 
-    let ret = if type_.ret.typ == Default::default() { None } else { Some(type_.ret.clone()) };
+    let ret = if type_.ret.typ == Default::default() { None } else {
+        let _ = used_rust_type(&env.library, type_.ret.typ)
+            .map(|s| used_types.insert(s));
```
gkoz: `Result::ok` seems a good fit for this: `used_rust_type(&env.library, type_.ret.typ).ok().map(|s| used_types.insert(s));`

EPashkin: Yes, i fix that

EPashkin:
Done.

EPashkin:
Also 35 waiting

EPashkin:
Don't understand why no "unused result" warning for `Option<()>`

gkoz:
Only the `Result` causes the "unused" warning. Using `map` here seems a bit of abuse anyway because the value isn't actually mapped, so a good old `if` could be better. But that's not too important.

PR #35, opened 2015-07-05
Remove typekind widget

EPashkin:
Still not sure about good name for 'TypeKind::of()', prefer 'for' but it keyword :(


gkoz:
Thanks!

PR #38, opened 2015-07-07
Change external types rule

EPashkin:
External types (`pango::Layout` in Scale) now count as ignored, if not set as "manual" in `Gir.toml`


EPashkin:
Autogenerated files will go to 'object_reform' branch in gtk or you create subbranch for them?

gkoz:
A separate branch makes more sense. I'll play with that tomorrow.

PR #40, opened 2015-07-09
Generate ffi for functions of classes

EPashkin:


EPashkin:
Next PRs will be structs for Classes and Interfaces, may be enums

src/codegen/sys/functions.rs
```diff
@@ -0,0 +1,98 @@
...
+    if func.ret.typ == Default::default() { return (false, String::new()) }
+    let ffi_type = ffi_type(env, func.ret.typ);
+    let commented = ffi_type.is_err();
+    (commented, format!(" -> {}", ffi_type.as_str()))
+}
+
+fn function_parameter(env: &Env, par: &library::Parameter) -> (bool, String) {
+    let ffi_type = ffi_type(env, par.typ);
+    let commented = ffi_type.is_err();
+    (commented, format!("{}: {}", general::fix_parameter_name(&par.name), ffi_type.as_str()))
```
gkoz: I'd prefer we get the `ffi_type` value from the `c:type` attribute to let the compiler catch any translation mistakes and be consistent with the C headers.

src/codegen/sys/statics.rs
```diff
@@ -0,0 +1,362 @@
+use std::io::{Result, Write};
+
+pub fn generate<W: Write>(w: &mut W) -> Result<()>{
```
gkoz: Does this file belong in some other branch?

src/codegen/sys/ffi_type.rs
```diff
@@ -0,0 +1,68 @@
...
+        _ => Err(format!("Unknown rust type: {:?}", type_.get_name() )),
+        //TODO: check usage library::Type::get_name() when no _ in this
+    }
+}
+
+fn fix_external_name(env: &Env, type_id: library::TypeId, name: &str) -> Result {
+    if type_id.ns_id == library::MAIN_NAMESPACE || type_id.ns_id == library::INTERNAL_NAMESPACE {
+        Ok(name.into())
+    } else {
+        let name_with_prefix = format!("{}_ffi::{}",
```
gkoz: Since we're in the FFI land anyway the `_ffi` suffix seems redundant.

gkoz: But we'll need a map of namespaces to prefixes because out `gdk` crate contains both `Gdk` and `GdkPixbuf` and the `glib` crate also has `gio` and `gobject`.

src/codegen/sys/functions.rs
EPashkin: Will done in next PR with cheking for `gtk_dialog_get_content_area` return type mismatch

src/codegen/sys/ffi_type.rs
EPashkin: `extern crate gdk_sys as gdk_ffi;` so in map will do 'gdk_ffi'

src/codegen/sys/statics.rs
EPashkin: No this file from this branch, for compiling.
File will greatly reduced when generated structs for Class and Interface, but still last here.

src/codegen/sys/functions.rs
gkoz: To be clear, we can use our knowledge about the type of course, e.g. if we have
```xml
<type name="utf8" c:type="gchar*">
```
we can first replace "gchar" or "char" (depending on what they've put in there) with "c_char" and process the `*` and `const` if any.

In
```xml
<type name="cairo.Pattern" c:type="cairo_pattern_t**">
```
we can replace "cairo_pattern_t" with "cairo::Pattern" and go from there.

PR #43, opened 2015-07-09
Replace unwrap_or_else(panic) with expect

EPashkin:


src/analysis/object.rs
```diff
@@ -27,7 +27,7 @@ impl Info {
     //TODO: add test in tests/ for panic
     pub fn type_<'a>(&self, library: &'a library::Library) -> &'a library::Class {
         let type_ = library.type_(self.class_tid).as_class()
-            .unwrap_or_else(|| panic!("{} is not a class.", self.full_name));
+            .expect(&format!("{} is not a class.", self.full_name));
```
gkoz: `unwrap_or_else` has the benefit of only calling `format` when you need it.

EPashkin: Ok. I change formated back.

EPashkin:
Fixed

PR #44, opened 2015-07-09
Generate sys structs

EPashkin:


EPashkin:
Don't merge, next is records

EPashkin:
Ok, I fix for.
About Records, normal if I replace `GtkTreeIter{...}` and other two with just `struct`?

gkoz:
Records need to have their fields, most of them are unusable as empty structs.

EPashkin:
Ok, then Records will be manual (or later).
Look to last commit, can I remove GAppLaunchContext and others external?

EPashkin:
About enums: I won't do it in next two days, and as I see, you start some preparation, so maybe you do them.

gkoz:
> Look to last commit, can I remove GAppLaunchContext and others external?

Yeah, they are out of place in the gtk crate.

> About enums: I won't do it in next two days, and as I see, you start some preparation, so maybe you do them.

Okay.

EPashkin:
Ready to merge

gkoz:
Thanks!

PR #45, opened 2015-07-12
Out parameters as return

EPashkin:
Not checked with pointer types and caller-allocates="1".

### DIFF
--- a/src/codegen/function_body.rs
+++ b/src/codegen/function_body.rs
@@ -13,7 +13,8 @@ pub struct Builder {
     glib_name: String,
     from_glib_prefix: String,
     from_glib_suffix: String,
-    parameters: Vec<String>,
+    parameters: Vec<(String, bool)>,
+    outs_as_return: bool,
 }
 
 impl Builder {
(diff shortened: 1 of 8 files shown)


src/codegen/function_body.rs
```diff
@@ -29,21 +30,64 @@ impl Builder {
...
+        for (pos, &(ref par, out_as_returns)) in self.parameters.iter().enumerate() {
+            if pos > 0 { param_str.push_str(", ") }
+            let s = par;
+            if out_as_returns { param_str.push_str("&") }
+            param_str.push_str(&s);
+        }
+        param_str
+    }
+    fn get_outs(&self) -> Vec<&String> {
+        let outs: Vec<&String> = self.parameters.iter()
```
gkoz: Doesn't the compiler infer the correct `collect` from the function signature without the `outs` variable?

src/codegen/function_body.rs
```diff
@@ -29,21 +30,64 @@ impl Builder {
...
+    }
+    fn generate_out_return(&self) -> String {
+        let outs = self.get_outs();
+        let (prefix, suffix) = if outs.len() > 1 { ("(", ")") } else { ("", "") };
+        let mut ret_str = String::with_capacity(100);
+        ret_str.push_str(prefix);
+        for (pos, ref par) in outs.iter().enumerate() {
+            if pos > 0 { ret_str.push_str(", ") }
+            let s = par;
+            ret_str.push_str(&s);
```
gkoz: You probably need `&*par` here.

src/codegen/function_body.rs
EPashkin: yes

src/codegen/function_body.rs
EPashkin: Strange, worked `par`, `&par`, `&*par`, `*par`. What better?

EPashkin: and `&&par` also worked

gkoz: AFAIK the idiomatic syntax to coerce owned stuff to slices is `&*`

gkoz: Also passing around `&String`s and `&Vec`s is unusual, because the borrowed types are `&str` and `&[T]`.

EPashkin: Fixed `fn get_outs(&self) -> Vec<&str>`
Can't fix `fn write_out_variables(&self, v: &mut Vec<String>)` as it mutable.

src/codegen/function_body.rs
```diff
@@ -29,21 +30,62 @@ impl Builder {
...
+        }
+    }
+    fn generate_out_return(&self) -> String {
+        let outs = self.get_outs();
+        let (prefix, suffix) = if outs.len() > 1 { ("(", ")") } else { ("", "") };
+        let mut ret_str = String::with_capacity(100);
+        ret_str.push_str(prefix);
+        for (pos, ref par) in outs.iter().enumerate() {
+            if pos > 0 { ret_str.push_str(", ") }
+            ret_str.push_str(&*par);
```
gkoz: Hmm, now that the vector contains `&str`s, this can probably be reduced to
```rust
for (pos, par) in ... {
    ...
    ret_str.push_str(par);
```

src/codegen/function_body.rs
gkoz: Yes, the mutable case is fine.

src/codegen/function_body.rs
EPashkin: Oops, forgot that.

EPashkin:
Done

src/codegen/function_body.rs
```diff
@@ -29,21 +30,62 @@ impl Builder {
...
-        let param_str = self.parameters.connect(", ");
+        let param_str = self.generate_func_parameters();
         format!("{}ffi::{}({}){}", self.from_glib_prefix,
             self.glib_name, param_str, self.from_glib_suffix)
     }
+    fn generate_func_parameters(&self) -> String {
+        let mut param_str = String::with_capacity(100);
+        for (pos, &(ref par, out_as_returns)) in self.parameters.iter().enumerate() {
+            if pos > 0 { param_str.push_str(", ") }
+            let s = par;
```
gkoz: This binding seems unnecessary too.

EPashkin: Fixed

gkoz:
Thanks

PR #46, opened 2015-07-12
Generate sys interface functions

EPashkin:
WIP, don't merge.
Just to inform about.

### DIFF
--- a/src/codegen/general.rs
+++ b/src/codegen/general.rs
@@ -44,7 +44,7 @@ pub fn objects_child_type<W: Write>(w: &mut W, type_name: &str, glib_name: &str)
     Ok(())
 }
 
-pub fn impl_parents<W: Write>(w: &mut W, type_name: &str, parents: &Vec<StatusedTypeId>) -> Result<()>{
+pub fn impl_parents<W: Write>(w: &mut W, type_name: &str, parents: &[StatusedTypeId]) -> Result<()>{
     try!(writeln!(w, ""));
     for stid in parents {
         //TODO: don't generate for parents without traits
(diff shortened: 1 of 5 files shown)


gkoz:
I'm deep into breaking the parser to allow anonymous unions and functions. Hopefully there won't be many conflicts :) I'm making these `Function` fields optional but most of the time unwrapping them should be safe:
```rust
    pub name: Option<String>,
    pub c_identifier: Option<String>,
```

gkoz:
Actually, the `name` might not become optional, I could've gotten carried away.

EPashkin:
Ok, i try don't touch library.
Can you rename `c_identifier` to `glib_type_name`?

EPashkin:
Ignore last.

EPashkin:
Ready to merge
I thought to do:
1) merge generate_classes_xxx, but only for 2 cases too many work
2) use library::*, but don't like mix with this PR

EPashkin:
3) Add trait to library GlibName { fn glib_name(&self) -> &str; } (this also allow merge `generate_classes_structs`), but it better be in library

gkoz:
>  merge generate_classes_xxx, but only for 2 cases too many work

Not sure which ones these are, if it's `generate_classes_structs` and `generate_interfaces_structs`, they seem too small to bother.

The trait might be useful.. if every implementing type is able to return a slice.

gkoz:
Ah, I think I see know. `generate_interface_funcs` is it? Can't the function take a `&[Function]` instead of the `Interface` or `Class`?

EPashkin:
Ok, it simpler, that add traits or macros, I do that.

EPashkin:
Done

EPashkin:
Still leave class name for better search

gkoz:
Thanks!

PR #47, opened 2015-07-13
Change ignoring rule for sys crate

EPashkin:
For fully compilable we need fix really unimplemented types in Gir.toml.
I started adding, but have no time until the evening. 48 errors to go.
Also I noticed an unpleasant thing: when changing Gir.toml, `cargo run` recompile the project, and not just launches.
```
[[object]]
name = "Gdk.ModifierType"
status = "ignore"

[[object]]
name = "Gdk.Gravity"
status = "ignore"

[[object]]
name = "Gio.File"
status = "ignore"

[[object]]
name = "Gio.Icon"
status = "ignore"

[[object]]
name = "Pango.FontFace"
status = "ignore"
```

### DIFF
--- a/src/codegen/sys/ffi_type.rs
+++ b/src/codegen/sys/ffi_type.rs
@@ -95,7 +95,7 @@ fn fix_external_name(env: &Env, type_id: library::TypeId, name: &str) -> Result 
     } else {
         let name_with_prefix = format!("{}_ffi::{}",
             fix_namespace(env, type_id), name);
-        if env.type_status(&type_id.full_name(&env.library)) == GStatus::Ignore {
+        if env.type_status_sys(&type_id.full_name(&env.library)) == GStatus::Ignore {
             Err(name_with_prefix.into())
         } else {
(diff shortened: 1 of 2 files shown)


gkoz:
> I started adding, but have no time until the evening. 48 errors to go.

If the sys crate generator gets completed soon, we could be replace the handmade files and the cross-crate ignores might not be needed then.

> Also I noticed an unpleasant thing: when changing Gir.toml, cargo run recompile the project, and not just launches.

This could be because of the existence of the build script. May be fixable with an [exclude](http://doc.crates.io/manifest.html#the-exclude-and-include-fields-(optional)) option.

gkoz:
We might need a trait `Ignored` (`SysIgnored`) that would be implemented recursively by all types so we'd know if a type has a problematic dependency even if it's in another crate.

EPashkin:
`Git.toml` not included, so exclude IMHO not work (not tested).
May be Version creater must not overwrite file with same content?

PR #51, opened 2015-07-13
library: change glib_type_name to c_type for Enumeration and Bitfield

gkoz:
...and make the attribute mandatory. Unlike glib:type-name it's always present.


gkoz:
@homu r+

PR #52, opened 2015-07-13
parser: read the functions' `throws` attribute, insert GError parameter

gkoz:


gkoz:
@EPashkin you might not like doing this in the parser, I'm open to alternatives. This seems like the only place that can easily see every function. Doing the same in the analysis modules will be more complicated.

EPashkin:
No problem with this, it really parameter.
But may be better add as `ParameterDirection::Error` ?

gkoz:
Would replacing `instance_parameter` with `kind: enum ParameterKind { Normal, Instance, Error }` work?

EPashkin:
Maybe. Not sure what we needed function generation.
So now merge as is, when we needed code with errors, then change.

EPashkin:
No effect on current generated widgets.

gkoz:
@homu r+

PR #55, opened 2015-07-14
Preparing for glib 1

EPashkin:
WIP.
Questionable:
1)  Change crate name from snake to lowercase (don't like change current names like `glib` to `g_lib`). Later, there will be problems with names like `gdkpixbuf`, but currently we have rename it anyway.
2) Change generate folder to `..\gtk`, as we needed 5 libraries, and i very don't like add it all to `gir/.gitignore` and have it in work tree.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,7 +6,7 @@ authors = [
 	"Gleb Kozyrev <gleb@gkoz.com>",
 ]
 build = "build.rs"
-exclude = ["Gir.toml"]
+exclude = ["Gir*.toml"]
 
 [dependencies]
 docopt = "*"
--- a/Gir_GLib.toml
+++ b/Gir_GLib.toml
@@ -1,0 +1,6 @@
+[options]
+work_mode = "sys"
+girs_dir = "../gir-files"
+library = "GLib"
+version = "2.0"
+#target_path = "./glib"
--- a/Gir_Gtk.toml
+++ b/Gir_Gtk.toml
@@ -4,6 +4,11 @@ girs_dir = "../gir-files"
 library = "Gtk"
 version = "3.0"
 #target_path = "./gtk"
+external_libraries = [
+   "GLib",
+   "Gdk",
+   "Pango", 
+]
 
 [[object]]
 name = "Gtk.Widget"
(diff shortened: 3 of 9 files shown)


EPashkin:
As result this PR will be generated only `lib.rs` with `glib` by empty command line.

EPashkin:
Problem with enums:
- some of them (`NormalizeMode`, `SpawnError`) has many values on one number;
- `ThreadError` has only one value

EPashkin:
Ready for merge

EPashkin:
For second PR I plan add options.other_libraries = ["GObject", "Gio"] and start doing normal tree.

gkoz:
> 2) Change generate folder to ..\gtk, as we needed 5 libraries, and i very don't like add it all to gir/.gitignore and have it in work tree.

This can be alleviated with a single ignored `out` directory in the repo root. I dislike both surprising behaviors of clobbering `../$SOME_DIR` and the generator knowing too much about directory structure (e.g. if the sys generator generates a single `${LIB}.rs`, it has no business putting it into `${LIB}-sys/src`, that's for the invoker to manage).

EPashkin:
One `out` don't work for me, as i "need" check for all generated sys, and check for normal `gtk` too.
But i can do it with 3 (current) command lines.
I change target_path back to ./gtk. And in next PR remove it from config, so it always needed.

gkoz:
You know what... I might've overlooked the big picture. It seems possible to generate complete sys crates where a human would only need to touch `Cargo.toml`. In that case the generator doesn't only make one or more modules, but also the build script.

But still the Gir files being in the `gir` repo root is temporary, they should either live in the `tests` directory or in the appropriate bindings repos (`glib`, `gtk` etc).

EPashkin:
Done.
In `normal` mode `gir` still need to know, that widgets go to `src/widgets` ?

gkoz:
As long as the normal mode produces a flat set of modules, it doesn't need that, strictly speaking. But I'm not sure how this is going to evolve, so changing now might be pointless.

There're reasons for the sys mode to know about `src` but less so about `glib-sys` (`gtk-sys`).

EPashkin:
ok

gkoz:
@homu r+

PR #60, opened 2015-07-14
Compile-test the produced glib-sys crate

gkoz:


gkoz:
@EPashkin here's a possible way to organize the integration tests and the accompanying toml files.

gkoz:
@homu r+

PR #61, opened 2015-07-15
Testing gobject-sys

EPashkin:
Tried to add gobject-sys, but there it also define `struct GObject`,
may be `struct GObject` generation better skipped?

### DIFF
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -156,7 +156,9 @@ fn generate_unions<W: Write>(w: &mut W, items: &[&library::Union])
     try!(writeln!(w, ""));
     for item in items {
         try!(writeln!(w, "pub type {} = c_void;", item.name));
-        try!(writeln!(w, "pub type {} = {};", item.c_type.as_ref().unwrap(), item.name));
+        if let Some(ref c_type) = item.c_type {
+            try!(writeln!(w, "pub type {} = {};", c_type, item.name));
+        }
         try!(writeln!(w, ""));
     }
 
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -43,10 +43,12 @@ fn generate_lib<W: Write>(w: &mut W, env: &Env) -> Result<()>{
     let classes = prepare(ns);
     let interfaces = prepare(ns);
 
+    try!(generate_aliases(w, env, &prepare(ns)));
     try!(generate_enums(w, &ns.name, &prepare(ns)));
     try!(generate_bitfields(w, &ns.name, &prepare(ns)));
+    try!(generate_unions(w, &prepare(ns)));
     try!(functions::generate_callbacks(w, env, &prepare(ns)));
-    try!(generate_records(w, env, &prepare_records(ns)));
+    try!(generate_records(w, env, &prepare(ns)));
     try!(generate_classes_structs(w, &classes));
     try!(generate_interfaces_structs(w, &interfaces));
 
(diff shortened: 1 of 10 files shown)


EPashkin:
Same problem with `GValue`

EPashkin:
IMHO `use_glib_ffi`, `only_for_gtk` etc.  must be moved to Gir.toml for module as `static_code`

src/codegen/sys/lib_.rs
```diff
@@ -182,6 +183,7 @@ fn generate_interfaces_structs<W: Write>(w: &mut W, interfaces: &[&library::Inte
 fn generate_records<W: Write>(w: &mut W, env: &Env, records: &[&library::Record]) -> Result<()> {
     try!(writeln!(w, ""));
     for record in records {
+        if record.c_type == "GValue" { continue; }
```
gkoz: Is there a reason to have `GObject` and `GValue` defined in `glib-sys` contrary to the gir definitions?

gkoz:
> IMHO use_glib_ffi, only_for_gtk etc. must be moved to Gir.toml for module as static_code

I wouldn't want to put any code in the configs. And the only discrepancy I expect is the predefined types in `glib-sys`, that isn't that big a deal and we could try to minimize the diff.

src/codegen/sys/lib_.rs
EPashkin: Its used in glib, interdependency :cry
I'm wrong?

gkoz: I don't think it is. Removing these two definitions doesn't break `glib-sys` for me.

gkoz:
It seems the hardcoded typedefs can actually be limited to `gboolean`, `gpointer` and `gconstpointer`, so an extra module is overkill.

gkoz:
`only_for_gtk` is down to constants so it won't be needed at all.

gkoz:
BTW, I've update the gir-files repo, don't forget to pull.

gkoz:
`Gio` references some `*Private` structs too so the simplest approach will be to not blacklist any structs at all.

EPashkin:
This union in GObject-2.0.gir has no `c:type` so `generate_unions` panics on unwrap.
`<union name="_Value__data__union">`
Write `pub type {} = {}` only for Some?

gkoz:
Meh. Yes please.

EPashkin:
No `GValue` now, maybe from
`WARN:gir::codegen::sys::ffi_type: [Missing glib_name of _Value__data__union, can't match != gpointer]`

EPashkin:
It present but commented

EPashkin:
I merged latest master

gkoz:
I guess I've fooled myself with that `ffi_type` skipping checks deal. If that's the only error, we could just merge despite failing tests.

EPashkin:
After i manual uncomment GValue and add `, GList, GSList` to statics `use glib_ffi::`
We have commented GVaClosureMarshal for va_list and after uncomment last only warning
```
src\lib.rs:59:36: 59:46 warning: literal out of range for i32, #[warn(overflowing_literals)] on by default
src\lib.rs:59         const G_PARAM_DEPRECATED = 2147483648,
```

EPashkin:
Seems ready to merge

gkoz:
Oh right, this PR might need to be rebased onto master and reopened as such..

EPashkin:
After you merge your tests  to master?

EPashkin:
Also I don't like removing `Gir_` prefix from toml.

EPashkin:
Oh, missed, that you merged tests already

gkoz:
> Also I don't like removing Gir_ prefix from toml.

Okay, let's put it back. The capitalization annoys me a bit :)

PR #67, opened 2015-07-15
Tests gobject

EPashkin:


gkoz:
Thanks!
Let's see if I've fixed `ffi_type` sufficiently.
@homu r+

EPashkin:
Only last va_list problem and warning

gkoz:
This'll be easier to fix after merging.

gkoz:
Uh-oh. We simply can't support `va_list` so it's blacklisting time.

EPashkin:
IMHO good way to blacklisting add
```
[[objects]
name = "GObject.SignalCVaMarshaller"
status = "ignore"
```
and check status in `generate_aliases`.
PS. Commenting `pub type GSignalCVaMarshaller = GVaClosureMarshal;` fixed build

EPashkin:
Also we can check all non-internal namespaces in `fix_external_name`

gkoz:
This doesn't seem appropriate under `[[objects]]`. Also for the purpose of sys modules I'd like something less verbose, like a single array with blacklisted identifiers, e.g.
```toml
ignore = [
    "SignalCVaMarshaller",
    "SomethingElse",
]
```

gkoz:
Yeah, bad types from external namespaces could be a problem...

EPashkin:
We can just rename [[objects]] to something else [[identifiers]]

EPashkin:
This way we don't need add second check.

gkoz:
toml allows this BTW
```toml
[[GObject.SignalCVaMarshaller]]
status = "ignore"
```
Could save some typing.

EPashkin:
But how to detect that it is `objects` ?

gkoz:
If the generator went through the objects (and other items) it finds in the library, it would know and just check them against the config.

EPashkin:
So we need store toml in environment?

EPashkin:
And do parsing on every request?

gkoz:
I assumed it's parsed once and held in memory :)

EPashkin:
Yes, it parsed as toml file, but we needed parse "status=ignore" on every request, as we can't "combine" all old [[objects]] to one category and parse to own format.

gkoz:
Asking the `toml::Table` if it has a key `status` inside a table `Widget` inside a table `Gtk` should not involve parsing. (I should've put single brackets in that example: `[Gtk.Widget]`)

EPashkin:
But we needed parse `status`

gkoz:
I see. Okay, currently all items are uniform pairs of name and status so we actually can parse them once. If some items' configs get distinct keys in the future, we surely can manage that somehow (the library knows which is which after all).

gkoz:
But explicitness could be a virtue too I guess.

EPashkin:
We can count `ignore` as simply alias for [[objects]]name=...\status='ignore'

EPashkin:
The main thing from my point of view, that is stored in name => Status. With the test of the dubbing, etc.

gkoz:
An `ignore` shorthand is okay.

Issue #29, opened 2015-07-04
Generate sys module


EPashkin:
It you don't start it, i can begin do it (about  1 day) in folowing way:
Add in Gir.toml `options.sys_subpath = "gtk-sys"`
Generate funcs.rs with `status = "generate"` widgets functions, that will be included and reexported in lib.rs.
All sys-depended code will go to `sys` subdirs of `analyze` and `codegen`

EPashkin:
Also "return out parameters" little depends on this.

gkoz:
I do plan to start the sys crate generation but why should anything else depend on that?

EPashkin:
You still planing, or already started?
In "return out parameters" I planing add `rets:Vec<Parameters>` along with old `ret: Option<Parameter>` to `object::Info`.
I want this issue act as warning remover, therefore, I propose do it myself

gkoz:
No, I have not started that. But I didn't want to couple the sys crate generation with other modes too much, for instance, I don't see why it should care about which widgets and methods are generated.

EPashkin:
If you want generate all function, then no dependence.

EPashkin:
gkoz: if you don't begin generate sys module, can i generate functions for all classes?

gkoz:
Alright, go ahead.

EPashkin:
gkoz: can you add to gtk-sys file funcs.rs where move `GtkAboutDialog` functions?
I have problem with this.

gkoz:
I don't see the point of splitting the sys crates into modules especially if they're autogenerated.

EPashkin:
Then where i write functions?

EPashkin:
Or you want autogenerate all sys module from begining?

gkoz:
Yes, I could be wrong but this task seems simple enough to generate everything at once.

EPashkin:
Ok. Then in first version i just rewrite begining of lib.rs.

gkoz:
:+1:

gkoz:
I think we're almost feature-complete vs the handmade sys crates so it's time to eliminate the gtk-related hardcoded stuff and accomodate generating all of `glib-sys`, `gdk-sys`, (possbily `pango-sys`), and `gtk-sys`.

EPashkin:
Any hardcode in sys?
IMHO better start with `glib-sys`. we need generate `lib.rs, glib.rs, gobject.rs, gio.rs`?

EPashkin:
[...]

PR #68, opened 2015-07-15
Small fixes in sys

gkoz:


EPashkin:
Alias `GType` defined in glib and gobject, I don't like `ignore` it in gobject, as it comment all `*_get_type` in normal mode. Seems this need be hardcoded.

EPashkin:
And for `gdk_pixbuf` not sure we needed rename, it may needed apply `fix_namespace` in too many places (at minumum in package naming). But we can stay with this until really problem occurs.

EPashkin:
Or we can just change `crate_name` to apply `snake_case` after lowering first 2-3 chars.

EPashkin:
About `GType`: may be we just needed new status `dont_declare`

gkoz:
It seems we can ignore `GType` in `glib-sys` and either add the `gobject` import in other crates or always qualify it in hardcoded references. It likely doesn't belong in `Fundamental`.

gkoz:
Didn't notice that `GLib` actually references `GType` as well. So it's not that simple.

EPashkin:
Can you do it in this PR with rebasing?

gkoz:
On it.

EPashkin:
Don't forget about `crate_name` it you accept https://github.com/gkoz/gir/pull/68#issuecomment-121814536

gkoz:
> Don't forget about `crate_name` it you accept #68 (comment)

Here, not sure if this change conflicts with the other use of `crate_name`.

EPashkin:
Can you add test for crate_name("GdkPixbuf")?

gkoz:
@homu r+

EPashkin:
Don't see bad changes in gtk generation. Compiling stops only on gdk-sys, seems this better fixed before
```
D:\eap\rust\0\gdk\gdk-sys\src\lib.rs:16:40: 16:45 error: unresolved import `glib_ffi::gsize`. There is no `gsize` in `glib_ffi`
D:\eap\rust\0\gdk\gdk-sys\src\lib.rs:16 pub use glib_ffi::{gboolean, gpointer, gsize, GType};
```

EPashkin:
IMHO better abort Homu, and fix `gsize`

EPashkin:
Seems gsize just needed include in static::only_for_glib

EPashkin:
Was wrong, seems it currently replaced to `size_t` by generator.
Then no problem

gkoz:
I was away :) We should add the other libs to the tests.

EPashkin:
I currently do gio

EPashkin:
You get message about changes in https://github.com/gkoz/gir/issues/29 header?

gkoz:
Building every crate is wasteful, there should be a test crate that depends on the others.

EPashkin:
So build only crates, that last in build order? At end it will be only gtk

EPashkin:
Then we need change test.sh to 2 setting, LIBS and BUILD?

gkoz:
> You get message about changes in #29 header?

No, missed that.

PR #70, opened 2015-07-16
Shorthand for status="ignore" in toml

EPashkin:
Problems, that I don't solved fully:
1. iterate all GStatuses,
2. place `ignore` not in `options`
3. `parse_status_shorthand`: `for` in `map`


EPashkin:
Problems with G_PARAM_DEPRECATED, if I change generate_bitfields to u32, then it start error in glib `G_LOG_LEVEL_MASK = -4`

gkoz:
Thanks.
@homu r+

I'm not sure about the enums yet, by setting a variant to `INT_MAX + 1` they may be forcing it to become 64-bit in size.

PR #71, opened 2015-07-16
Add gio to tests

EPashkin:
5 errors

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,5 @@ target
 /src/version.rs
 tests/gir-files
 tests/sys/*-sys/**
+tests/sys/sys_build/Cargo.lock
 !tests/sys/*-sys/Cargo.toml
--- a/src/codegen/sys/ffi_type.rs
+++ b/src/codegen/sys/ffi_type.rs
@@ -70,7 +70,10 @@ fn ffi_inner(env: &Env, tid: library::TypeId, inner: String) -> Result {
                 Utf8 => "c_char",
                 Filename => "c_char",
                 Type => "GType",
-                Pointer => return Ok(inner),
+                Pointer => match &inner[..]  {
+                    "void" => "c_void",
+                    _ => return Ok(inner),
+                },
                 Unsupported => return Err(format!("[Unsupported type {}]", inner)),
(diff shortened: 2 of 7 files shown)


EPashkin:
No namespaces in records, and `void`

EPashkin:
Fixed void*

EPashkin:
About array, we can add it to `use glib_ffi::{..` but IMHO must be better fix

gkoz:
We'll still want a `test` crate that depends on others because it makes sense to add some small libraries that aren't `gtk` dependencies (`libsecret`, `libvte`, other `g*` ones), and that crate could also do some sanity checks (ideally, it would compare Rust and C `sizeof` for some types).

gkoz:
I'll have to think about the `Array` types, didn't expect them.

EPashkin:
Ok, now i understand, agreed with that.

EPashkin:
I add `sys_test` module with empty lib.rs here while you thinking.
May be better add array fix to master and rebase current.

EPashkin:
Ok, i change fix void*

gkoz:
> Ok, i change fix void*

Not just yet... :)

EPashkin:
Added crate, moved fix void* to last

.gitignore
```diff
@@ -3,4 +3,5 @@ target
 /src/version.rs
 tests/gir-files
 tests/sys/*-sys/**
+tests/sys/sys_build/**
 !tests/sys/*-sys/Cargo.toml
```
gkoz: This line probably has to become
```
!tests/sys/*/Cargo.toml
```
then.

EPashkin: Not needed: git show changes in added files

gkoz: No, why even ignore the whole crate if you only have to ignore its `Cargo.lock` and `target`.

gkoz: I mean this crate isn't autogenerated after all.

EPashkin:
Updated .gitignore, ffi_type.rs

gkoz:
I think I'll merge this to work on the arrays with more coverage.

gkoz:
Thanks

PR #72, opened 2015-07-16
Arrays and bitflags improvements

gkoz:
Add GArray, GByteArray and GPtrArray support.
Make enums and bitflags c_int and c_uint; convert negative bitflags members to positive repr.
Add pango to tests.

### DIFF
--- a/src/codegen/sys/ffi_type.rs
+++ b/src/codegen/sys/ffi_type.rs
@@ -89,7 +89,10 @@ fn ffi_inner(env: &Env, tid: library::TypeId, inner: String) -> Result {
                 Err(s) => Err(format!("[{}; {}]", s, size)),
             }
         }
-        Type::List(..) | Type::SList(..) => Ok(inner),
+        Type::Array(..) | Type::PtrArray(..)
+                | Type::List(..) | Type::SList(..) | Type::HashTable(..) => {
+            fix_name(env, tid, &inner)
+        }
         _ => {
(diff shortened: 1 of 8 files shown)


gkoz:
So `G_PARAM_DEPRECATED` is... deprecated and useless. Since we're defining it as a constant, might use the blacklisting facility.

EPashkin:
You sure about this? IMHO it marks parameter as deprecated. But may be really useless for us.

gkoz:
Bah! You're right. I'll continue investigating the enum sizes then

EPashkin:
By way can you change test.sh that it scans *.toml? So we don't need update LIBS

EPashkin:
And about bitfields, IMHO it always unsigned, so better just convert all to u32

gkoz:
> By way can you change test.sh that it scans *.toml? So we don't need update LIBS

Good idea but won't the generator fail because of missing `src` directories? It may need to create them on demand.

gkoz:
Unsigned bitflags seems logical. I guess we'll have to convert rare negative values (`G_LOG_LEVEL_MASK`) to positive representations.

gkoz:
@homu r+

EPashkin:
You create `src` in script, creating also can be added to gir, if you know how to check file existence. I found no stable method.

gkoz:
If the script just globs over the toml files, it won't know which directories to create.

`PathExt::exists` seems unstable but we still have options. For example always create the directory ignoring the `AlreadyExists` error.

PR #73, opened 2015-07-16
Create src directory

EPashkin:


EPashkin:
Its good time to add `Gir_` prefix (or `gir_`) to tomls, if you allow.

tests/sys/test.sh
```diff
@@ -1,15 +1,13 @@
...
-LIBS="glib gobject gio pango"
 export RUST_LOG="gir=warn"
 
 cd "`dirname $0`"
 cargo build --release || exit 1
 
-for LIB in $LIBS; do
-	mkdir -p ${LIB}-sys/src
-	$GIR -c ${LIB}.toml || exit 2
+for TOML in `ls *.toml`; do
```
gkoz: `for TOML in *.toml` should work too

src/codegen/sys/lib_.rs
```diff
@@ -18,9 +19,12 @@ pub fn generate(env: &Env) {
 
     let path =  PathBuf::from(&env.config.target_path)
         .join(file_name_sys("lib"));
-    println!("Generating file {:?}", path);
 
-    save_to_file(path, &mut |w| generate_lib(w, env));
+    let parent = path.parent().unwrap();
+    fs::create_dir(parent).ok();
```
gkoz: Haha, this is a novel way of ignoring errors. `let _ = ...` might be more obvious.

EPashkin:
Really works. I tried `for TOML in "*.toml"` and it not work.

gkoz:
Yes, renaming is fine. Consider `-` instead of `_` if it's not against your taste ;)

src/codegen/sys/lib_.rs
EPashkin: You yourself have taught me that

EPashkin:
Only problem with `-` it mixed symbols in `gir-gdk_pixbuf'.

EPashkin:
But we already have directory like `gio-sys` so be `gir-gdk-pixbuf.toml`

gkoz:
Thanks!
@homu r+

PR #74, opened 2015-07-16
tests: add libsecret

gkoz:


gkoz:
@homu r+

gkoz:
Now we know that `version` is not always "x.y" but can just be "1".

EPashkin:
Yes and this not very good.

PR #75, opened 2015-07-17
Tests gdk and gdk-pixbuf

EPashkin:


EPashkin:
`Gdk` uses manual version of `cairo`
Also has problem with GdkEventKey.is_modifier so some `GdkEvent*` commented, may be also for `pub state: [TypeId { ns_id: 1, id: 16 }]`

EPashkin:
Used manual version of `cairo`, as generated contains nothing (may be due no includes in `cairo-1.0.gir`)

EPashkin:
bad field in GdkEvent* now `pub state: [TypeId { ns_id: 1, id: 16 } Gdk.ModifierType]`

EPashkin:
This field contains no `c:type` in gir. Seems this case needed fixed specially.
Can you fix it?

EPashkin:
Possible problem: `GdkPixbuf-2.0.gir` includes `GModule` but we don't need it.

EPashkin:
For `_truncated_record_marker` IMHO better add field name in comments.

gkoz:
Thanks!
@homu r+

gkoz:
Generating `cairo` is problematic, yes. First, we'll need to prepare its GIRs ourselves, it doesn't come with them included. And generating a good GIR is not easy. But then is the issue of it having a number of non-glib dependencies.

I'll look into `GdkEvent` and adding a comment to truncated records.

PR #77, opened 2015-07-17
Wrap extern fns in Option

gkoz:
An fn type is not nullable so the more faithful representation of the C type is wrapping in an Option.


EPashkin:
If it still compiles, then agree.

gkoz:
@homu r+

PR #78, opened 2015-07-17
Add gtk-sys to tests

EPashkin:
WIP

### DIFF
--- a/tests/sys/atk-sys/Cargo.toml
+++ b/tests/sys/atk-sys/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "atk-sys"
+version = "0.0.1"
+
+[lib]
+name = "atk_sys"
+
+[dependencies]
+bitflags = "^0.3"
+libc = "^0.1"
+
+[dependencies.glib-sys]
+path = "../glib-sys"
+
+[dependencies.gobject-sys]
+path = "../gobject-sys"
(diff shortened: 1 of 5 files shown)


EPashkin:
xlib-sys contains only struct and types and bad lines
```
pub type XEvent = c_void;
pub type XEvent = XEvent;
```

EPashkin:
IMHO ready for merge

EPashkin:
`xlib.Window` only used in `GtkPlug` and `GtkSocket`

EPashkin:
Updated info about `gmodule` and `cairo` in https://github.com/gkoz/gir/issues/29

EPashkin:
As all girs.toml contains path to sys, started to think, that its need suffixed with `-sys`

gkoz:
> `pub type XEvent = XEvent;`

I think, we're going to stop cutting of the prefixes anyway.
But we'll need some solution to avoid creating `xlib-sys`, `fontconfig-sys` and so on.

gkoz:
And ideally the `cfg(os = "unix")` stuff could be marked appropriately.

EPashkin:
When we has more info about system dependence we can add statuses that converted to `ignore` on wrong system.

EPashkin:
Was wrong with last.

EPashkin:
We can add status `generate_on_linux` for `GtkPlug` and `GtkSocket` but don't know how to remove `use xlib_sys as xlib_ffi` in correct way.

gkoz:
Thanks.
@homu r+

`gmodule` and what else comes with glib, `gthread`, I think, it makes sense to have them too in case someone wants them.

gkoz:
Oh and I don't mind renaming the gir files.

PR #79, opened 2015-07-17
Stop stripping prefixes from enums and bitfields

gkoz:


EPashkin:
Not sure that we really need do that.
If you do it because xlib::xevent, then it xlib-2.0.gir problem. It has line
`<union name="XEvent" c:type="XEvent"/>`

gkoz:
More because it's inconsistent to strip some identifiers but not others, I added stripping imitating the manual crates but that's silly. The front end crates will have to reexport any symbols they want under different names.

EPashkin:
Then ok.

gkoz:
@homu r+

PR #80, opened 2015-07-18
Generate other functions

EPashkin:
WIP

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -491,6 +491,8 @@ impl Library {
 
     fn read_global_function(&mut self, parser: &mut Reader, ns_id: u16,
                             attrs: &Attributes) -> Result<(), Error> {
+        let moved_to = attrs.get("moved-to").is_some();
+        if moved_to { return ignore_element(parser); }
         let pos = parser.position();
         let func = try!(self.read_function(parser, ns_id, "global", attrs));
         if func.c_identifier.is_none() {
--- a/src/codegen/sys/functions.rs
+++ b/src/codegen/sys/functions.rs
@@ -31,7 +31,7 @@ pub fn generate_other_funcs<W: Write>(w: &mut W, env: &Env, functions: &[library
 fn generate_object_funcs<W: Write>(w: &mut W, env: &Env, c_type: &str,
     glib_get_type: &str, functions: &[library::Function]) -> Result<()> {
     let write_get_type = glib_get_type != "intern";
-    if write_get_type && !functions.is_empty() {
+    if write_get_type || !functions.is_empty() {
         try!(writeln!(w, ""));
         try!(writeln!(w, "    //========================================================================="));
         try!(writeln!(w, "    // {}", c_type));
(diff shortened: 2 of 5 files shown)


EPashkin:
Done

EPashkin:
IMHO `cairo` now will be not empty, I try it.
And next I try features.

src/parser.rs
```diff
@@ -496,7 +496,11 @@ impl Library {
         if func.c_identifier.is_none() {
             return Err(mk_error!("Missing c:identifier attribute", &pos));
         }
-        self.add_function(ns_id, func);
+        let already_added = self.namespace(ns_id).classes_function_cidentifiers
+            .contains(func.c_identifier.as_ref().unwrap());
+        if !already_added {
```
gkoz: There are duplicate function definitions in girs?

EPashkin:
In `Gio-2.0.gir` two `g_icon_new_for_string` and 30 others

gkoz:
Do all of them have the `moved-to` attribute? We could probably skip those in the codegen.

EPashkin:
Oh missed that, then it can just skipped in simply way. I check

EPashkin:
Can i change result of `Parser::read_function` to `Result<Option<Function>, Error>` or there better way ignore function?

gkoz:
I think we can check the `moved-to` attribute upstream (e.g. in `read_class`) and instead of `read_function` call `ignore_element` then.

gkoz:
To avoid duplication this could be done in a wrapper e.g. `read_function_if_not_moved`

EPashkin:
IMHO we needed this check only for global functions, so i just check `moved-to` beforehand as you say

gkoz:
Makes sense.

EPashkin:
Done
If you think that we don't need feature, we can call GuillaumeGomez

src/codegen/sys/ffi_type.rs
```diff
@@ -72,6 +72,7 @@ fn ffi_inner(env: &Env, tid: library::TypeId, inner: String) -> Result {
                 Type => "GType",
                 Pointer => match &inner[..]  {
                     "void" => "c_void",
+                    "FILE" => return Err(inner),
```
gkoz: Looks like `libc` provides `FILE` so we could add it to imports instead.

gkoz:
This still doesn't generate functions on records (e.g.`Array.append_vals` in `GLib`), does it?

gkoz:
About the features, I'll open a PR against rust-gnome to remove the redundant ones.

EPashkin:
Yes, forgot about `g_array_append_vals` and others

EPashkin:
Fixed FILE
About features: we easy can add it to functions.

EPashkin:
I started do record functions

EPashkin:
IMHO Ready for merge

gkoz:
Thanks.
@homu r+

PR #81, opened 2015-07-18
Tests cairo

EPashkin:


EPashkin:
Seems you was right about not full cairo-1.0.gir.
Just in case I check for build other syses with this.

gkoz:
It's possible to generate a complete definition but we probably won't need that for a while.

EPashkin:
It allow `gtk` compile, but not contain functions, used in original `cairo`.
Not sure merge this or stay open or close.

EPashkin:
If you want merge, wait for rebase after previous PR.

EPashkin:
Rebased

gkoz:
I'll let this sit until we have a better idea of what to do with cairo. Merging now could hide some conflicts with the existing crate.

EPashkin:
It too old and seems unneeded

PR #84, opened 2015-07-19
Generate consts

EPashkin:
`c:type` parsed but unused, `name` used only for check for ignoring.
On build some warning about overflowing `pub const G_MININT8:i8 = 128;` etc.
Not sure about names checker functions for `GStatus`.
May be better remove `only_for_gtk` as it contains only `GTK_ENTRY_BUFFER_MAX_SIZE`

### DIFF
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -7,7 +7,6 @@ use case::CaseExt;
 use analysis::rust_type::parameter_rust_type;
 use env::Env;
 use file_saver::*;
-use gobjects::GStatus;
 use library;
 use nameutil::*;
 use super::ffi_type::ffi_type;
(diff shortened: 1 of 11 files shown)


src/codegen/sys/lib_.rs
```diff
@@ -119,6 +121,43 @@ fn generate_bitfields<W: Write>(w: &mut W, items: &[&library::Bitfield])
...
+            Err(x) => ("//", x),
+        };
+        if env.type_status_sys(&format!("{}.{}", env.config.library_name,
+            constant.name)).ignored() {
+            comment = "//";
+        }
+        let mut value = constant.value.clone();
+        if type_ == "&str" {
+            type_ = "&'static str".into();
+            value = format!("\"{}\"", escape_string(&value));
```
gkoz: You likely can avoid escaping by using the [raw string literal syntax](https://doc.rust-lang.org/reference.html#raw-string-literals).

EPashkin:
Fixed escaping

gkoz:
Thanks!
@homu r+

PR #86, opened 2015-07-20
Parse versions

EPashkin:
WIP


EPashkin:
Stuck on:
1. Don't found way to early returns from .map() so `from_str` has manual iteration
2. Maybe same in `read_function`, don't have time to fully test.
3. Doc tests not compiling, not found why.

EPashkin:
4. Found error in gir: `Version: 3.0.`. May be it same error for you add `.filter(!s.is_empty())` in gtk

EPashkin:
Ignore fixing part.
But i still not fully understand for what you do it.

gkoz:
I used `take_while` to cut of the iterator on a parsing error.
`filter(!is_empty())` in main crates' build scripts is needed to avoid returning an empty `cfg` that causes an ICE in rustc.

Regarding the version numbers complexity. Do you think there's any harm in using a tuple and defaulting failed elements to 0?

Then there're some strategies of acting on the versions to choose from: 
 * we may get away with ignoring the "patch" version (third element) and just leaving the door open to adding it if there's demand;
 * collect the list of x.y.z versions that are present in such attributes and put the list into the build scripts, so e.g. the build script would enumerate `pango_1_30`, `pango_1_31_0`, `pango_1_32`, `pango_1_32_4`, `pango_1_34`, `pango_1_36`, `pango_1_36_7`.

src/parser.rs
```diff
@@ -593,6 +593,13 @@ impl Library {
         let name = try!(attrs.get("name").ok_or_else(|| mk_error!("Missing function name", parser)));
         let c_identifier = attrs.get("identifier").or_else(|| attrs.get("type"));
         let kind = try!(FunctionKind::from_str(kind_str).map_err(|why| mk_error!(why, parser)));
+        let version = match attrs.get("version") {
+            Some(v) => {
+                println!("Version: {}", v);
+                Some(try!(v.parse().map_err(|why| mk_error!(why, parser))))
+                },
```
gkoz: The Rust style asks us to align the closing brace one tab to the left (probably a typo?) and, surprisingly, not add the trailing comma.

EPashkin: align - typo,
thinked it better always add comma beetwen cases.

EPashkin:
> filter(!is_empty()) in main crates' build scripts is needed to avoid returning an empty cfg that causes an ICE in rustc.

Thinked that empty part of `split` result can be only from empty string which better check directly. Now see other option ("3.0.").

> Regarding the version numbers complexity. Do you think there's any harm in using a tuple and defaulting failed elements to 0?

I plan replace empty parts with 0. `"3.1." => Version([3,1,0])` `"3..2" => Version([3,0,2])`.
Don't thinked about tupples.
Possible problems:
- comparing different length (2,3) and (1,4,2),
- iterating (mostly on comparing)

gkoz:
If we decide to care about the patch version part, iterating blindly doesn't seem possible, we'll have to use the actual set of versions encountered when parsing the library (have `Library` maintain that set).

A fixed length tuple that defaults to zero seems an easy compromise. To rephrase the linked example,
```rust
    let mut parts = lib.version.splitn(4, '.')
        .map(|s| s.parse())
        .take_while(Result::is_ok)
        .map(Result::unwrap);
    let version: (u16, u16, u16) = (parts.next().unwrap_or(0), parts.next().unwrap_or(0), parts.next().unwrap_or(0));
```
If some of the parts fails to parse the tuple is filled with zeroes (`"3.1."` => `(3, 1, 0)`, `"2.a.1"` => `(2, 0, 0)`). Implementing `Ord` is straightforward, formatting `cfg` directives is manageable.

EPashkin:
Queston, we need about add patch version in cfgs or not?

gkoz:
Let's only add them if they're non-zero: `pango_1_32_4`, `pango_1_34`.

EPashkin:
Ok.

EPashkin:
And add warning to log this.

gkoz:
The more specific versions don't seem to have any impact that would justify a warning.

EPashkin:
But we know, that in library some function has really 3digits version, and adjust decision about iterating in build.rs

EPashkin:
Remake Version to tuple

src/versions.rs
```diff
@@ -0,0 +1,39 @@
+use std::str::FromStr;
+
+#[derive(Debug, Clone, PartialEq)]
+// major, minor, patch
+pub struct Version(pub (u16,u16,u16));
```
gkoz: Could this be `Version(pub u16, pub u16, pub u16)`?

EPashkin:
We have 1.32.4 in pango and some odd minored in others. Seems we really need store version in Namespace, and use it when generating build.rs or include to it  module with versions.

src/versions.rs
EPashkin: Yes :(

src/main.rs
```diff
@@ -24,6 +24,7 @@ mod nameutil;
 mod parser;
 mod traits;
 mod version;
+mod versions;
```
gkoz: I'd prefer a useful module use the name `version` and the former `version` renamed to something else.

EPashkin:
Why u16, not u32?

EPashkin:
like `gir_version` ?

gkoz:
> Why u16, not u32?

`u16` seemed exceedingly big for any sane version numbers.

> like `gir_version` ?

Sounds good.

src/versions.rs
```diff
@@ -0,0 +1,39 @@
...
+impl FromStr for Version {
+    type Err = String;
+
+    /// Parse a `Version` from a string.
+    fn from_str(s: &str) -> Result<Version, String> {
+        let mut parts = s.splitn(4, '.')
+            .map(|s| s.parse())
+            .take_while(Result::is_ok)
+            .map(Result::unwrap);
+        let version: (u16, u16, u16) = (parts.next().unwrap_or(0),
```
gkoz: Now you can stick the name `Version` in here and avoid enumerating the fields below.

EPashkin:
Renamed files

src/version.rs
```diff
@@ -0,0 +1,38 @@
+use std::str::FromStr;
+
+#[derive(Debug, Clone, PartialEq)]
```
gkoz: The `check_version` signature got me thinking: `Copy` should go here too.

gkoz:
Looks good. Do you want to merge or build on this more?

EPashkin:
As classes and interfaces seems not versioned, I only add deprecated-version for functions

gkoz:
We might also be interested in enumeration and bitfield members' versions.

EPashkin:
And constants.
IMHO better add config check to function before this, so ready for merge.
Warnings about versions I have not remove, `check_version` will changed to store versions to namespace. But it can be done only after generating main code to prevent desynchronization.
Or you prefer remove `check_version` now?

EPashkin:
With version autodetection we have problems if generate config check on sys before do it on full generated main :cry:

gkoz:
> Or you prefer remove check_version now?

It's fine for now.

Re. desynchronization, do I understand it correctly that your concern is about compatibility between the 'sys' and 'normal' modes because the 'normal' mode doesn't know about cfg versions? The objective of replacing manual sys crates needs the existence of cfg versions but doesn't require any conditional compilation in sys crates.

gkoz:
@homu r+

EPashkin:
Seems I don't understand something.
We have sys with autodetected `cfg_xxx` and function
```
#[cfg(gtk_3_12)]
 pub fn GTK_POPOVER(widget: *mut ffi::GtkWidget) -> *mut ffi::GtkPopover {
     unsafe { ffi::cast_GtkPopover(widget) }
 }
```
its current analog `gtk_popover_get_type` with same cfg. Or `gtk_popover_new`.
normal gtk get cfg value through DEP_GTK_CFG.
My concern is that:
if `pub use self::popover::Popover;` in gtk main.rs don't have same `#[cfg(gtk_3_12)]` for some reason,
then it not detected on CI as it use latest version.

gkoz:
`main.rs` suggests an application? Testing applications for compatibility with older libraries could be difficult.
Our CI currently builds with both GTK 3.6 (travis and appveyor) and 3.14 (appveyor only), so missing a `cfg` should get caught.

EPashkin:
Oh, forgot that gtk default is 3.6 - almost oldest supported version.

EPashkin:
Then no obvious problems, but still don't like that condition on sys may added before same condition on normal.

PR #88, opened 2015-07-24
Generate the build script

gkoz:


EPashkin:
IMHO better add separate mode `sys_build` for call `build::generate`.
`check_version` better be renamed to something like `save_version` with removing log messages.

EPashkin:
We need parameter in config `options.min_version`. For gtk it `3.4` and ignore too old version.
Not sure if it needed also in this PR (then we simply not add old version in `check_version`) or used only in "normal" mode.

EPashkin:
We can change *.toml for full test.
```
build = "build.rs"

[build-dependencies]
pkg-config = "0.3"
gcc = "0.3"
```
But then we need install `libsecret` and others to build machine.
So not sure if that really needed.

gkoz:
> IMHO better add separate mode sys_build for call build::generate.

Won't this complicate the use too much? I'm also more inclined to do some `Cargo.toml` manipulation, 'cause doing it manually is too annoying.

> check_version better be renamed to something like save_version with removing log messages.

I'll do that.

> We need parameter in config options.min_version.

This looks like it can be done separately.

> We can change *.toml for full test. But then we need install libsecret and others to build machine.

It's achievable but not clear if we want this.

EPashkin:
Ok, no `sys_build`
IMHO better rebase this to #89 and use `options.min_cfg_version`. That way all functionality be in one PR

gkoz:
Makes sense.

gkoz:
@homu r+

PR #91, opened 2015-07-25
Cargo.toml generation and manipulation

gkoz:
Turns out the `toml-rs` library can't preserve the order of the items but at least it's deterministic.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,8 @@
 language: rust
+addons:
+  apt:
+    packages:
+    - libgtk-3-dev
 script:
   - cargo build --release
   - cargo test --release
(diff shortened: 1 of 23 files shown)


EPashkin:
Can you add `min_cfg_version = "3.4"` to `gir-gtk.toml` in this PR?
Not sure what value in other crates.

gkoz:
I'll add some

EPashkin:
Problems with gio: build.rs checks wrong package (at mininumum not builds on Windows).
```
const LIBRARY_NAME: &'static str = "gio";
const PACKAGE_NAME: &'static str = "gio-unix-2.0";
```

gkoz:
Hmm, I must've botched the package name logic in a previous PR

gkoz:
Yeah, it should've kept the first package it found, not the last one. Hang on

EPashkin:
Thanks.

gkoz:
@homu r+

PR #96, opened 2015-07-27
Some comsetic changes

gkoz:
It seems redundant to use the `_ffi` suffix in the sys crates when referring to sibling crates. All of this is FFI already.


EPashkin:
Will not confusion between sys and normal version without suffix?

EPashkin:
But if all compiled, then no problem.

gkoz:
The normal crates still pick the names they like (sometimes it's just `ffi`) but these crates constitute a self-sufficient layer.

gkoz:
@homu r+

PR #97, opened 2015-07-28
Don't make backups

EPashkin:
Can make instead conditional compilation or config options

### DIFF
--- a/src/codegen/sys/build.rs
+++ b/src/codegen/sys/build.rs
@@ -19,7 +19,8 @@ pub fn generate(env: &Env) {
     let _ = fs::create_dir_all(parent);
 
     println!("Generating file {:?}", path);
-    save_to_file(&path, &mut |w| generate_build_script(w, env));
+    save_to_file(&path, env.config.make_backup,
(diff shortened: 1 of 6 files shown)


gkoz:
We could add a command-line option but with git everywhere the backups just seem redundant and slightly annoying.

EPashkin:
Ok, I make command line options -b, --make_backup

EPashkin:
Done

EPashkin:
What better do with classes functions?
Ex. `gtk_window_get_default_icon_name`?

gkoz:
> What better do with classes functions?
Ex. gtk_window_get_default_icon_name?

Static methods probably. If I understand the question right.

gkoz:
@homu r+

EPashkin:
> Static methods probably. If I understand the question right.

```
impl Window {
    pub fn get_default_icon_name() -> String {
        unsafe { from_glib_none(ffi::gtk_window_get_default_icon_name()) }
    }
}
```
Ok.

gkoz:
Looks good.

PR #101, opened 2015-07-30
Downcasts

EPashkin:
WIP

### DIFF
--- a/Gir_Gtk.toml
+++ b/Gir_Gtk.toml
@@ -52,6 +52,10 @@ name = "Gtk.RadioButton"
 status = "generate"
 
 [[object]]
+name = "Gtk.AboutDialog"
+status = "generate"
+
+[[object]]
 name = "Gtk.Actionable"
 status = "manual"
 
--- a/src/analysis/c_type.rs
+++ b/src/analysis/c_type.rs
@@ -1,0 +1,50 @@
+pub fn rustify_pointers(c_type: &str) -> (String, String) {
+    let mut input = c_type.trim();
+    let leading_const = input.starts_with("const ");
+    if leading_const {
+        input = &input[6..];
+    }
+    let end = [
+        input.find(" const"),
+        input.find("*const"),
+        input.find("*"),
+        Some(input.len()),
+    ].iter().filter_map(|&x| x).min().unwrap();
+    let inner = input[..end].trim().into();
+
+    let mut ptrs: Vec<_> = input[end..].rsplit('*').skip(1)
+        .map(|s| if s.contains("const") { "*const" } else { "*mut" }).collect();
+    if let (true, Some(p)) = (leading_const, ptrs.last_mut()) {
+        *p = "*const";
+    }
+
+    let res = (ptrs.connect(" "), inner);
+    trace!("rustify `{}` -> `{}` `{}`", c_type, res.0, res.1);
+    res
+}
+
+#[cfg(test)]
+mod tests {
+    use super::rustify_pointers as rustify_ptr;
+
+    fn s(x: &str, y: &str) -> (String, String) {
+        (x.into(), y.into())
+    }
+
+    #[test]
+    fn rustify_pointers() {
+        assert_eq!(rustify_ptr("char"), s("", "char"));
+        assert_eq!(rustify_ptr("char*"), s("*mut", "char"));
+        assert_eq!(rustify_ptr("const char*"), s("*const", "char"));
+        assert_eq!(rustify_ptr("char const*"), s("*const", "char"));
+        assert_eq!(rustify_ptr("char const *"), s("*const", "char"));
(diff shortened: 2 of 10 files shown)


EPashkin:
Now generated next code
```
fn get_content_area(&self) -> Option<Box> {
    unsafe {
        Widget::from_glib_none(ffi::gtk_dialog_get_content_area(self.upcast().to_glib_none().0)).downcast_unchecked()
    }
}
```
What correct conversion instead `downcast_unchecked()` for `Option<>`?
`downcast().ok()`?

gkoz:
Hmm. I wrote this as
```rust
    fn get_content_area(&self) -> widgets::box_::Box {
        unsafe {
            Widget::from_glib_none(ffi::gtk_dialog_get_content_area(self.upcast().to_glib_none().0))
                .downcast_unchecked()
        }
    }
```
so apparently I didn't consider this nullable and `Option` isn't good there.

Generally in a nullable case I'd probably make it
```rust
            Widget::from_glib_none(ffi::gtk_dialog_get_content_area(self.upcast().to_glib_none().0))
                .map(|w| w.downcast_unchecked())
```

EPashkin:
IMHO Result.Map still return result, not Option.

gkoz:
Maybe `Downcast` should just be implemented for `Option`.

gkoz:
Here's an actual working example:
```rust
    fn get_header_bar(&self) -> Option<HeaderBar> {
        unsafe {
            Option::<Widget>::from_glib_none(
                ffi::gtk_dialog_get_header_bar(self.upcast().to_glib_none().0))
                .map(Downcast::downcast_unchecked)
        }
    }
```

EPashkin:
Made
```
fn get_content_area(&self) -> Option<Box> {
    unsafe {
        Option::<Widget>::from_glib_none(ffi::gtk_dialog_get_content_area(self.upcast().to_glib_none().0)).map(Downcast::downcast_unchecked)
    }
}
```
Found only ~30 `<return-value nullable="1" ` in gtk, none with nullable="0".
Seen some functions without nullable but NULL in doc (Ex. `gtk_dialog_get_widget_for_response`). 
So currently count all pointer result (calculated in `can_be_nullable_return`) except constructors as nullable.
If you find better solution, it can be changed later.

gkoz:
The nullable issue seems to be impossible do handle without maintaining our own metadata. I've realized today that we don't have to enter all of the metadata by hand though, the generator could have an interactive mode and ask us about ambiguous functions then save the answers into the config.

EPashkin:
Also we need solve possible conflict with std::boxed::Box, but it IMHO better done when we can compile gtk

EPashkin:
Better no interactive mode in console generator.
And i don't like check all ambiguous functions in one time.

EPashkin:
Also don't like rewrite toml, as it lost ordering and all comments

gkoz:
Not in one time, more like `git add -p` that can be rerun many times and only saves any concrete answers you give.

The toml rewriting deficiency is unfortunate yes, someone was working on fixing that but it's not clear if we'll get that soon. That kind of metadata could live in a separate file though.

EPashkin:
Agree that some `nullable_returns = ["gtk_dialog_get_widget_for_response"]` in config needed

EPashkin:
IMHO implementation of interactive mode takes too much time and not give virtually no acceleration in comparison with the simple adding to configuration. As fast variant I can add prints or `warn!` for possible nullables.

gkoz:
> Agree that some `nullable_returns = ["gtk_dialog_get_widget_for_response"]` in config needed

The default assumption should be a nullable return value (except from a constructor).

You're probably right about the interactive mode overhead.

EPashkin:
> The default assumption should be a nullable return value (except from a constructor).

Its current mode (see `analysis\return_value`)

But you say that `gtk_dialog_get_content_area` not nullable in your opinion, for these exception we need some metadata, in config or not.
Not sure what cases is larger: not marked nullables, or not nullables in your mode.

EPashkin:
IMHO nullable returns need better thinking and discussions, may be in Issue.
If current mode is good, and no mistakes in code, then PR ready to merge.

gkoz:
Looks like `create_dir_all` is in the wrong place, should've been in `save_to_file`...

EPashkin:
Agree, just waited for working directory check before "refactor" it.
Move it now?

gkoz:
No, it's not urgent.

gkoz:
Just one thing, could you `s/ctype/c_type/` for consistency?

EPashkin:
Yes, it all different: `glib_name`, `c_type` :(

gkoz:
There's less of `glib_name` now, we may be able to remove it. It's just `c_type` won't always be appropriate.

EPashkin:
Renamed all including module

gkoz:
Thanks!
@homu r+

EPashkin:
This solves part of #102

PR #107, opened 2015-08-01
Fix equality comparison for sh

### LINKED ISSUE #106: gir-secret not generated
Problem both in Windows7 and Ubuntu.
On windows7 I don't have `libsecret` installed, so don't worry about it.
But when I tried build on Ubuntu it still not generated with other error.
```
Running `/gir/target/release/gir -c gir-secret.toml`
thread '<main>' panicked at 'No such file or directory (os error 2) - ../gir-files/Secret-1.gir', /gir/src/parser.rs:37
```
Checked gir-files and don't found Secret-1.gir not in directory, not in git history.

EPashkin:
Closes #106

### DIFF
--- a/tests/sys/test.sh
+++ b/tests/sys/test.sh
@@ -1,8 +1,8 @@
 #!/bin/sh
 
 GIR="cargo run --release --"
-FEATURES=
-if [ "$1" == "full" ]; then
+FEATURES=""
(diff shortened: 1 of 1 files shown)


gkoz:
Thanks!
@homu r+

PR #110, opened 2015-08-13
sys: rewrite some gdk-pixbuf function names for Windows

gkoz:


EPashkin:
Convert functions will be added in same file?

gkoz:
I'm conflicted about adding the shims here.

On the one hand the sys crate only expose the available symbols, not some unified API and it's the "normal" crate's job to call the correct functions. On the other hand, dealing with that when generating the normal crates could be difficult or at least require duplicating this rewriting logic.

Right now my preference is to only add this bit and adjust the gdk crate to call the right functions depending on `cfg(windows)`.

EPashkin:
Also not sure about better way do this.

gkoz:
@homu r+

PR #112, opened 2015-09-15
non_nullable override and some fixes

gkoz:


src/gobjects.rs
```diff
@@ -1,5 +1,5 @@
 use std::ascii::AsciiExt;
-use std::collections::HashMap;
+use std::collections::{HashMap};
```
EPashkin: Typo?

EPashkin:
Seems good. Thanks

src/codegen/function.rs
```diff
@@ -81,7 +81,9 @@ pub fn body(env: &Env, analysis: &analysis::functions::Info,
 
     //TODO: change to map on parameters with pass Vec<String> to builder
     for par in &analysis.parameters {
-        let s = par.translate_to_glib(&env.library, in_trait);
+        let upcast = analysis.upcasts.iter().any(|&(ref name, _, _)| name == &par.name)
+            || in_trait && par.instance_parameter;
```
EPashkin: May be it compiler-optimized, but I prefer check simply condition first.

src/gobjects.rs
gkoz: There was `HashSet` there but I replaced it with a sorted vector later. I'll clean this up.

src/codegen/function.rs
gkoz: You're right, thanks for spotting this.

gkoz:
@homu r+

PR #114, opened 2015-09-21
Partially implement container handling

gkoz:
`GList` return value is translated to `Vec` correctly now.
Somewhat generalized passing input parameters by reference.

```patch
-    //fn get_children(&self) -> /*Unknown kind*/Unknown rust type: "List TypeId { ns_id: 1, id: 3 }";
-    //fn get_focus_chain(&self, focusable_widgets: /*Unknown kind*/Unknown rust type: "List TypeId { ns_id: 1, id: 3 }") -> bool;
+    fn get_children(&self) -> Vec<Widget>;
+    //fn get_focus_chain(&self, focusable_widgets: /*Unimplemented*/Vec<Widget>) -> bool;

-    //fn set_focus_chain(&self, focusable_widgets: /*Unknown kind*/Unknown rust type: "List TypeId { ns_id: 1, id: 3 }");
+    //fn set_focus_chain(&self, focusable_widgets: /*Unimplemented*/&[Widget]);

-    //fn set_icon(&self, icon: Option<gdk_pixbuf::Pixbuf>);
-    //fn set_icon_from_file(&self, filename: Fundamental: Filename, error: Option</*Unknown kind*/Unknown rust type: "Error">) -> bool;
-    //fn set_icon_list(&self, list: /*Unknown kind*/Unknown rust type: "List TypeId { ns_id: 7, id: 2 }");
+    //fn set_icon(&self, icon: Option<&gdk_pixbuf::Pixbuf>);
+    //fn set_icon_from_file(&self, filename: Fundamental: Filename, error: /*Unknown kind*/Unknown rust type: "Error") -> bool;
+    //fn set_icon_list(&self, list: /*Unimplemented*/&[gdk_pixbuf::Pixbuf]);
```

### DIFF
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -7,7 +7,7 @@ use analysis::return_value;
 use analysis::rust_type::*;
 use analysis::upcasts::Upcasts;
 use env::Env;
-use library;
+use library::{self, Nullable};
 use traits::*;
 use version::Version;
 
@@ -55,8 +55,9 @@ fn analyze_function(env: &Env, func: &library::Function, class_tid: library::Typ
                 panic!("Too many parameters upcasts for {}", func.c_identifier.as_ref().unwrap())
             }
         }
-        if parameter_rust_type(env, par.typ, par.direction)
-            .is_err() { commented = true; }
+        if parameter_rust_type(env, par.typ, par.direction, Nullable(false)).is_err() {
+            commented = true;
+        }
     }
 
     let outs = out_parameters::analyze(func);
--- a/src/analysis/return_value.rs
+++ b/src/analysis/return_value.rs
@@ -18,9 +18,12 @@ pub fn analyze(env: &Env, func: &library::Function, class_tid: library::TypeId,
         used_rust_type(env, func.ret.typ).ok().map(|s| used_types.insert(s));
         // Since GIRs are bad at specifying return value nullability, assume
         // any returned pointer is nullable unless overridden by the config.
-        let mut nullable = func.ret.nullable || can_be_nullable_return(env, func.ret.typ);
-        if non_nullable_overrides.binary_search(&func.name).is_ok() {
-            nullable = false;
+        let mut nullable = func.ret.nullable;
+        if !*nullable && can_be_nullable_return(env, func.ret.typ) {
+            *nullable = true;
+        }
+        if *nullable && non_nullable_overrides.binary_search(&func.name).is_ok() {
+            *nullable = false;
         }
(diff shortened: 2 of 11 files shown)


EPashkin:
Don't understand reason for Nullable, but look very good.

gkoz:
Apparently it's not considered a good style to use boolean parameters. A named type adds explicitness and type safety.

EPashkin:
Then not better make Nullable enum{NotNullable, CanBeNull} ? (can't find good names)

gkoz:
It seems the good names might be `Nullable { Yes, No }` and this isn't much different from `Nullable(pub bool)`.

EPashkin:
Yes :(
And long names useful only with `use library::Nullable::NotNullable` that make more difficult change used value (from NotNullable to CanBeNull).
But use `Nullable(false)` don't seems right for me.

gkoz:
I'll probably have to make it an enum then. Do you happen to have some Rust examples where such yes/no enums were used effectively?

gkoz:
A hazard of the enum approach is that `if nullable == Nullable::Nullable` could break if someone added another variant later. Avoiding ifs would not be reliable as well.

EPashkin:
Yes. And safe usage (only one that I found) also don't look good
```
if  !skip_option {
  match nullable {
    Yes => rust_type = rust_type.map_any(|s| format!("Option<{}>", s)),
    No => ()
  }
}
```
So maybe it better just stay plain bool.

EPashkin:
About "it's not considered a good style to use boolean parameters", can you provide some links?
I understand about values that linked with scope (number with unit of measure), but don't hear about just bool.

gkoz:
There's [this bit in Rust style guidelines](http://doc.rust-lang.org/style/features/types/README.html#use-custom-types-to-imbue-meaning;-do-not-abuse-bool,-option-or-other-core-types.-%3Cstrong%3E[fixme:-needs-rfc]%3C/strong%3E)

This kind of advice is all over the internet applying to Java, C++, any language that doesn't support named parameters.

I've seen Rust code with commented annotations like `/* nullable */ false` which seem to just ask for a newtype.

No one has bothered me about [this use of a bool newtype](https://github.com/gtk-rs/gtk/blob/master/src/signal.rs#L57-L59) so far.

EPashkin:
Thanks, I missed this.
Then just don't forget nullable in Parameter.

gkoz:
Switched `Parameter::nullable` to `Nullable`.

EPashkin:
Thanks

gkoz:
@homu r+

PR #115, opened 2015-09-22
Working toward object_reform

gkoz:
* Move the generated files files to `mod auto` and generate `auto/mod.rs`
* Mangle module, function, parameter names if they conflict with Rust keywords
* Generate non-widget objects
* Add `Bitfield` and `Filename` support
* Minor internal and non-functional output tweaks

### DIFF
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -38,17 +38,20 @@ pub fn analyze(env: &Env, klass: &library::Class, class_tid: library::TypeId,
 }
 
 fn analyze_function(env: &Env, func: &library::Function, class_tid: library::TypeId,
-    non_nullable_overrides: &[String], used_types: &mut HashSet<String>) -> Info {
+    non_nullable_overrides: &[String], all_used_types: &mut HashSet<String>) -> Info {
     let mut commented = false;
     let mut upcasts: Upcasts = Default::default();
+    let mut used_types: Vec<String> = Vec::with_capacity(4);
 
-    let ret = return_value::analyze(env, func, class_tid, non_nullable_overrides, used_types);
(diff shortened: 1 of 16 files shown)


EPashkin:
Maybe add Popover (or other) as example of widget version?
```
[[object]]
name = "Gtk.Popover"
status = "generate"
```

gkoz:
I've add a lot of widgets to my local copy in a branch of the `gtk` repo, just not sure what should or shouldn't go into the config here...

gkoz:
I could push one that build cleanly on the object_reform branch, or one that exposes some generator bugs.

EPashkin:
Understand

src/analysis/functions.rs
```diff
@@ -60,6 +63,12 @@ fn analyze_function(env: &Env, func: &library::Function, class_tid: library::Typ
         }
     }
 
+    if !commented {
+        for s in used_types {
+            all_used_types.insert(s);
+        }
+    }
```
gkoz: This  might need to be revisited when we encounter unused import warnings depending on the set cfg versions.

An alternative is putting something like this in modules that have version cfgs:
```
#![cfg_attr(not(gtk_3_10), allow(unused_imports))]
```

gkoz:
The "Analyzing... Generating..." output to stdout is getting out of hand, I'd prefer to switch to logging macros, so e.g. ignoring a class would have higher severity than just generating one and easier to notice.

EPashkin:
Agree

gkoz:
I've addressed the "generate mod.rs in separate function" comment.

Going to merge this.

EPashkin:
Thanks, i rebase my PR after merging (it will have conflict)

gkoz:
@homu r+

PR #116, opened 2015-09-23
sys: fix up c_type when the actual type is more specific

gkoz:
```patch
-    pub fn gtk_menu_item_set_submenu(menu_item: *mut GtkMenuItem, submenu: *mut GtkWidget);
+    pub fn gtk_menu_item_set_submenu(menu_item: *mut GtkMenuItem, submenu: *mut GtkMenu);
```
This will avoid silly upcast workarounds for cases like
```rust
    pub fn set_submenu<T: Upcast<Menu>>(&self, submenu: Option<&T>) {
```
not knowing how to upcast to `Widget`.


EPashkin:
Good idea

gkoz:
@homu r+

Issue #117, opened 2015-09-24
Fix out parameters as return for TypeKind other of Direct


EPashkin:
Currently out parameters treated same way that result code like this
```
    fn query_child_packing<T: Upcast<Widget>>(&self, child: &T) -> (bool, bool, u32, PackType) {
        let mut expand.to_glib() = Default::default();
        let mut fill.to_glib() = Default::default();
        let mut padding = Default::default();
        let mut pack_type = Default::default();
        unsafe {
            ffi::gtk_box_query_child_packing(self.upcast().to_glib_none().0, child.upcast().to_glib_none().0, &mut expand.to_glib(), &mut fill.to_glib(), &mut padding, &mut pack_type);
        }
        (expand.to_glib(), fill.to_glib(), padding, pack_type)
    }
```
List of ffi functions which must return out parameters
- [x] gtk_alignment_get_padding
- [x] gtk_box_query_child_packing
- [x] gtk_button_get_alignment
- [x] gtk_calendar_get_date
- [x] gtk_frame_get_label_align
- [x] gtk_label_get_layout_offsets
- [x] gtk_layout_get_size
- [x] gtk_misc_get_alignment
- [x] gtk_misc_get_padding

gkoz:
I think I might be fine with postponing all kinds of clever processing, but it's still difficult to settle on a uniform solution.

Simply using `&mut T` (`Option<&mut T>` for `optional`* ones) doesn't work well for returning objects and boxed types (nothing to initialize the variable with at the call site). We could use `&mut Option<T>` and `Option<&mut Option<T>>` for that at the cost to ergonomics and wait for `&out` to become available. This is something that will work in all cases (we'll still have to add temporary variables when conversions are needed). 

We could just put the actual return type and all the outparams in a tuple and return that but what about the optional outparams?

EPashkin:
It might be solution.
Only it is not clear whether or not to restrict the typekinds of return values.

EPashkin:
There functions with mixed outparams (optional and not)?
Any way to detect when parameters optional (besides #119) ?

gkoz:
By optional I mean `optional="1"`, you are allowed to pass `NULL` to ignore the value. Sometimes this has performance impact -- the function will take extra time to produce the value, but only if you actually want it. Can't find an example right now.

BTW optional seems to be the majority.

I don't know if there are combinations of optional and non-optional.

EPashkin:
[...]

PR #118, opened 2015-09-24
Limit out_as_return to unconverted types

EPashkin:
First part of #117

### DIFF
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -83,7 +83,7 @@ fn analyze_function(env: &Env, func: &library::Function, class_tid: library::Typ
         }
     }
 
-    let outs = out_parameters::analyze(func);
+    let outs = out_parameters::analyze(env, func);
 
     Info {
         name: nameutil::mangle_keywords(&*func.name).into_owned(),
--- a/src/analysis/out_parameters.rs
+++ b/src/analysis/out_parameters.rs
@@ -1,18 +1,30 @@
 use std::vec::Vec;
 
+use env::Env;
 use library::*;
+use super::type_kind::TypeKind;
 
 pub type Info = Vec<Parameter>;
 
-pub fn analyze(type_: &Function) -> Info {
+pub fn analyze(env: &Env, type_: &Function) -> Info {
     let mut outs = Info::new();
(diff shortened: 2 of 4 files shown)


gkoz:
Will `TypeKind::Enumeration` and the assumption that enums are passed directly become an issue when the converted enums appear? We already know `IconSize` will have to be a distinct enum on the Rust side and thus converted.

EPashkin:
Then now only allow `TypeKind::Direct` ?
Or just make note, that we need divide `Library::Enumeration` to `TypeKind::Enumeration` and `TypeKind::Converted`?

gkoz:
Maybe allowing overriding the type kind will suffice. Say, the `TypeKind::Enumeration` is still direct but we add a `TypeKind::ForceConversion` and apply that through config to `IconSize`, maybe `GType` too, etc.

`IconSize` is just ridiculous though because three(?) functions will take `Option<IconSize>` and I don't think handling this case in the code is worth the trouble.

EPashkin:
Rebased

gkoz:
@homu r+

gkoz:
Okay, I had some doubts about this change but thought I just didn't understand it. Surely we'd want to comment out a function that has outparams we can't handle? `query_child_packing` still is wrong.

EPashkin:
Yes, seems need comment function, if unsupported out params present.
Or there current way convert `&mut bool` to `*mut gboolean` ?
If not commenting function will be next stage.

On stage after it when we accept `TypeKind::Converted` function must look like this?
```
    fn query_child_packing<T: Upcast<Widget>>(&self, child: &T) -> (bool, bool, u32, PackType) {
        let mut expand = Default::default();
        let mut fill = Default::default();
        let mut padding = Default::default();
        let mut pack_type = Default::default();
        unsafe {
            ffi::gtk_box_query_child_packing(self.upcast().to_glib_none().0, child.upcast().to_glib_none().0, &mut expand, &mut fill, &mut padding, &mut pack_type);
        }
        (from_glib(expand), from_glib(fill), padding, pack_type)
    }
```

gkoz:
Can't just convert a pointer, because the types have different sizes.
This code looks good.

PR #121, opened 2015-09-24
Comment functions with unsupported outs

EPashkin:
Part of #117


gkoz:
Thanks.
@homu r+

EPashkin:
Actually this commit cover not all cases, IMHO `gtk_widget_translate_coordinates` stay with `&mut` arguments as it has return type.
Not sure about right way do fix it. Always check all parameters in `out_parameters::analyze` don't seems good for me.

PR #122, opened 2015-09-24
Allow converted outs

EPashkin:


EPashkin:
Part of #117

gkoz:
Turns out this doesn't work either
```rust
let mut pack_type = Default::default();
```

Since these are outparams we could get away with just `mem::uninitialized()`.

EPashkin:
IMHO better don't do it on current stage.
Also currently none of `gtk/object_reform` branches compiles, so I can't check function correctness.

gkoz:
Sorry about that. I've updated `glib/object_reform` and `gtk/object_reform_gir` and they seem to build with no other overrides now.
You need at least a beta compiler, because the glob import bug is only fixed in 1.4.

gkoz:
I've fixed the appveyor issue, now it fails to build for a legitimate reason. We'll definitely have to redo the `used_types` thing to keep track of versions...

EPashkin:
Thanks, then I build current nightly and try build again.

EPashkin:
GTK builds on current nightly, thanks.

EPashkin:
Minor problem about `mem::uninitialized` that it need use statement.
We can use as `std::mem::uninitialized` and use `std` always,
or add `std::mem` to `used_types` in `analysis::functions::analyze_function`

I plan call function in this way
```
let mut padding_top = unsafe{ mem::uninitialized() }; 
```

gkoz:
Adding to `used_types` seems fine. Maybe rename them to `imports`.

Minimizing the `unsafe` blocks is not a good idea, I'd prefer the assignments and the function call to be in a single `unsafe` block.

EPashkin:
Changed outs initializing.
Only one minor problem, that use looks like this:
```
use object::*;
use Bin;
use Buildable;
use Container;
use Widget;
use std::mem;
```

gkoz:
BTW there's another reason to make the `unsafe` block bigger (possibly just make them contain the whole fn body): `from_glib_*` functions are `unsafe` and when we start putting objects in the tuple it will need to be moved inside the `unsafe` block.

The imports... looks like we might throw all kinds of stuff in there and make them look nicer in a centralized manner.

EPashkin:
Make it look like this?
```
    fn query_child_packing<T: Upcast<Widget>>(&self, child: &T) -> (bool, bool, u32, PackType) {
        unsafe {
            let mut expand = mem::uninitialized();
            let mut fill = mem::uninitialized();
            let mut padding = mem::uninitialized();
            let mut pack_type = mem::uninitialized();
            ffi::gtk_box_query_child_packing(self.upcast().to_glib_none().0, child.upcast().to_glib_none().0, &mut expand, &mut fill, &mut padding, &mut pack_type);
            (from_glib(expand), from_glib(fill), padding, pack_type)
        }
    }
```

gkoz:
Yes, this will be more reliable.

EPashkin:
Changed

gkoz:
Nice, thanks!
@homu r+

PR #127, opened 2015-09-27
Add better panicking for not found Gir.toml

GuillaumeGomez:


src/config.rs
```diff
@@ -142,9 +142,14 @@ impl Config {
 
 fn read_toml(filename: &str) -> toml::Value {
     let mut input = String::new();
-    File::open(filename).and_then(|mut f| {
+    match File::open(filename).and_then(|mut f| {
```
EPashkin: `if xxx.is_err() { panic }`
will be clearer?

src/config.rs
```diff
@@ -142,9 +142,14 @@ impl Config {
...
         f.read_to_string(&mut input)
-    }).unwrap();
+    }) {
+        Ok(_) => {}
+        Err(_) => {
+            panic!("\"{}\": No such file", filename);
```
EPashkin: Error can appear from `open` or from `read_to_string` so "No such file" IMHO don't cover all cases.

gkoz:
Indeed, why lose information, when you can say "failed to read {filename} due to {error}".

But more generally would the idiomatic way be to bubble up any errors from `config` and `parser` into `main`? How much boilerplate would that require?

EPashkin:
As this function can produce 2 types of errors: io::result and long text from toml parsing IMHO better allow `read_toml` panic

gkoz:
Idiomatic error handling is supposed to cope with varying error types, the simplest solution being `Result<Config, Box<Error>>`. I suspect the `Error` trait object would not show the file name though. So a little bit of boilerplate might be needed.

GuillaumeGomez:
So what solution do you want me to implement ?

gkoz:
The ideal would be to return `Result` and do the handling in the `main`. That could take a bit of work but we all might learn from that.

If you just want to improve the error message here, don't throw away the inside of `Err(..)` use it in the error message.

GuillaumeGomez:
Updated, should be a lot better now.

EPashkin:
IMHO returning just string not best way return error,
but in case toml error we need return vector of errors with its position,
thus need add custom struct to store it (too much work)
so string will do if @gkoz accept

gkoz:
If you can't use `try!` and have to resort to `Err(e) => return Err(e)`, you doing `Result` wrong. The ideal case I meant is returning a type that actually implements `Error` (maybe just an enum with `IoError`, `TomlError`, `XmlError` variants). It might involve implementing `From<io::Error>` and some getting used to but that's the learning part I'm talking about.

I agree that the "Errors in config" thing would need more details.

EPashkin:
IMHO with just panic function was better.

gkoz:
My hypothesis is that proper use of `Result` should not make the functions uglier, as the conversion between different errors is taken outside. But when reporting the file name comes into the picture, we'll just have to see. This could inform the future development of Rust error handling ;)

GuillaumeGomez:
"the location info should be put the TomlError as well"

src/config.rs
```diff
@@ -28,6 +34,42 @@ impl FromStr for WorkMode {
...
+
+impl Display for ConfigError {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        match self {
+            &ConfigError::Io(ref e, ref filename) => {
+                write!(f, "\"{}\": ", filename);
+                e.fmt(f)
+            }
+            &ConfigError::TomlError(ref err, ref filename) => {
+                write!(f, "Toml parsing error on \"{}\": error: {}",
```
gkoz: I'd probably put this "Error parsing `{}`: {}"

src/config.rs
```diff
@@ -28,6 +34,42 @@ impl FromStr for WorkMode {
...
+                         filename, err.desc)
```
gkoz: After the issue with the location below is settled, `err.desc` would just be `err`

src/config.rs
```diff
@@ -140,22 +181,23 @@ impl Config {
...
         None => {
-            for err in &parser.errors {
-                let (loline, locol) = parser.to_linecol(err.lo);
-                let (hiline, hicol) = parser.to_linecol(err.hi);
-                println!("{}:{}:{}-{}:{} error: {}",
-                         filename, loline, locol, hiline, hicol, err.desc);
-            }
-            panic!("Errors in config")
+            let tmp = parser.errors.get(0).unwrap();
+            Err(ConfigError::TomlError(toml::ParserError{lo:0,hi:0,desc:tmp.desc.clone()}, filename.to_owned()))
```
gkoz: I'd rather `remove` the element but retain all its information (and possibly file a bug about this struct not deriving `Clone`). And rereading the docs maybe the element we want is the last one not the first. The toml-rs library is just weird.

src/config.rs
```diff
@@ -28,6 +34,41 @@ impl FromStr for WorkMode {
...
+#[derive(Debug)]
+pub enum ConfigError {
+    Io(IoError, String),
+    TomlError(String, String),
+}
+
+impl Error for ConfigError {
+    fn description(&self) -> &str {
+        match self {
+            &ConfigError::Io(ref e, _) => e.description(),
```
gkoz: It's nicer to match `*self` and not put the `&`s here.

src/config.rs
```diff
@@ -28,6 +34,41 @@ impl FromStr for WorkMode {
...
+            &ConfigError::TomlError(ref s, _) => s,
+        }
+    }
+}
+
+impl Display for ConfigError {
+    fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        match self {
+            &ConfigError::Io(ref e, ref filename) => {
```
gkoz: Same with `&` and could the `err` bind be named consistently?

src/config.rs
```diff
@@ -140,22 +180,27 @@ impl Config {
...
-fn read_toml(filename: &str) -> toml::Value {
+fn read_toml(filename: &str) -> Result<toml::Value, ConfigError> {
     let mut input = String::new();
-    File::open(filename).and_then(|mut f| {
+    match File::open(filename).and_then(|mut f| {
         f.read_to_string(&mut input)
-    }).unwrap();
+    }) {
+        Ok(_) => {}
+        Err(e) => {
```
gkoz: I think Rust style would be
```rust
        Ok(_) => (),
        Err(e) => return Err(ConfigError::Io(e, filename.to_owned())),
```

src/config.rs
```diff
@@ -140,22 +180,27 @@ impl Config {
...
-                println!("{}:{}:{}-{}:{} error: {}",
-                         filename, loline, locol, hiline, hicol, err.desc);
-            }
-            panic!("Errors in config")
+            let err = &parser.errors[parser.errors.len() - 1];
+            let (loline, locol) = parser.to_linecol(err.lo);
+            let (hiline, hicol) = parser.to_linecol(err.hi);
+            Err(ConfigError::TomlError(format!("{}:{}-{}:{} error: {}",
+                                               loline, locol, hiline, hicol, err.desc),
+                                       filename.to_owned()))
```
gkoz: It seems `let s = format!(...);` will be shorter and avoid rightward drift.

src/config.rs
```diff
@@ -28,6 +34,41 @@ impl FromStr for WorkMode {
...
+            }
+        }
+    }
+}
+
+impl From<IoError> for ConfigError {
+    fn from(e: IoError) -> ConfigError {
+        ConfigError::Io(e, String::new())
+    }
+}
```
gkoz: It looks like, as in the other PR, this impl is of no use :(

GuillaumeGomez:
Updated !

src/config.rs
```diff
@@ -140,22 +173,24 @@ impl Config {
...
 
-fn read_toml(filename: &str) -> toml::Value {
+fn read_toml(filename: &str) -> Result<toml::Value, ConfigError> {
     let mut input = String::new();
-    File::open(filename).and_then(|mut f| {
+    match File::open(filename).and_then(|mut f| {
         f.read_to_string(&mut input)
-    }).unwrap();
+    }) {
+        Ok(_) => {}
```
gkoz: One-line variants should have trailing commas.

src/main.rs
```diff
@@ -32,7 +32,13 @@ mod version;
 fn main() {
     env_logger::init().unwrap_or_else(|e| panic!("{}", e));
 
-    let cfg = config::Config::new();
+    let cfg = match config::Config::new() {
+        Ok(c) => c,
+        Err(e) => {
+            println!("{}", e);
```
gkoz: I think I'd like this to say "Failed to read config {}", because currently the error message still is somewhat cryptic.

GuillaumeGomez: You asked me to remove my previous error message (just to remind you haha).

gkoz: I don't think I did. It was phrased suboptimally once but after that I was surprised you removed all the text.

GuillaumeGomez: Then I misunderstood. My bad. I re-add it !

src/config.rs
GuillaumeGomez: First time I head about this but yes sir !

GuillaumeGomez:
Updated !

src/config.rs
```diff
@@ -1,6 +1,11 @@
+#![allow(unused_must_use)]
```
gkoz: I don't think this is warranted.
Sorry I hadn't noticed this earlier. Everything else is good!

GuillaumeGomez: Actually it is because of line 55.

gkoz: `try!` harder ;)

GuillaumeGomez: Oh come on !

gkoz:
Great! Sorry this has been such a drag, I hope we'll improve.
@homu r+

Issue #119, opened 2015-09-24
Support optional outparam return types


gkoz:
Some functions return `bool` in addition to outparams with the understanding that returning `FALSE` means the outparams haven't been initialized, this should translate to `None`.

Example: `gtk_widget_translate_coordinates`.

EPashkin:
Full list of GTK functions can be received by XPath `//method[return-value[type[@name="gboolean"]] and parameters/parameter[@direction="out"]]`
Ex. `gtk_container_get_focus_chain`

gkoz:
Nice idea, thanks!

gkoz:
All of the command line tools I've tried fail to do anything useful even with just `//method`. Looks like my XML/XPath skills have degraded too much :/

gkoz:
Looks like there're 67 functions of this kind in Gtk, 21 in Gdk. Most seem to fit the rule.

PR #128, opened 2015-09-27
rust_type: only consider Option for Pointer, Object and Interface types

gkoz:


EPashkin:
As I see it only affect gesture_drag.rs whose 2 function has `f64` out parameters with `bool` return and must be fixed in #119, so I not sure about this way.

EPashkin:
If you will do any changes in this PR, swap commits please

gkoz:
> As I see it only affect gesture_drag.rs whose 2 function has f64 out parameters with bool return and must be fixed in #119, so I not sure about this way.

Yes and no. The previous change aiming to avoid `Option` for containers wasn't entirely correct. It's less error prone to just say: only pointers have a notion of nullability.

Updated.

EPashkin:
By way, as we don't plan pass NULL to out parameters, may be also change parser to set all outs nonnullable?

EPashkin:
> Yes and no. The previous change aiming to avoid Option for containers wasn't entirely correct. It's less error prone to just say: only pointers have a notion of nullability.

Understand. No problem then.

gkoz:
> By way, as we don't plan pass NULL to out parameters, may be also change parser to set all outs nonnullable?

Correction: `nullable` describes the returned value, it's `optional` that allows passing a `NULL` pointer in. I don't see the benefit of doing this.

EPashkin:
For `gtk_gesture_drag_get_offset` both out parameters marked as `nullable` in GIR

gkoz:
That might be a mistake. See https://wiki.gnome.org/action/show/Projects/GObjectIntrospection/Annotations#Type_signature

EPashkin:
> Indicates that NULL may be a valid value for a parameter (in, out, inout)

Don't it mean that if you don't wan't `x` then can pass NULL pointer as value?

gkoz:
Together with 
>  (optional) For (out) parameters, NULL signifies that the output value is to be ignored.

I consider "value" to be the actual data returned in the outparameter and `nullable` being strictly orthogonal to `optional`. The majority of the annotations are consistent with this interpretation.

EPashkin:
`//parameter[@direction="out" and @nullable]` shows 13 parameters
`//parameter[@direction="out" and @optional]` shows many parameters,
may be you right and it just mistake (as gdouble can't be NULL).

gkoz:
It's valid for a value to be `nullable` independently from being `optional` and I have the impression that many items that should be tagged `nullable` are not, just as the usual return values.

gkoz:
@homu r+

PR #129, opened 2015-09-28
Outs as optional return

### LINKED ISSUE #119: Support optional outparam return types

EPashkin:
Closes #119


EPashkin:
Don't found good name for `*_parts` functions

gkoz:
Good job! Might call those "parts" "leaves" if that makes it any clearer. There's a typo in "tuple" in several places.

EPashkin:
Fixed "tuple". Don't change to "leaves" as don't understand its meaning.

gkoz:
Thanks!
@homu r+

PR #131, opened 2015-09-29
Ignore GInitiallyUnowned when checking for ignored parents

gkoz:
I'm not sure why the presence of ignored parents requires ignoring a class. Isn't skipping an Upcast implementation enough?


EPashkin:
Yes skipping ignoreds Upcast may be good way.
Was done when parent check stops on Widgets, so may be just include all parents as `manual` also will do normal.

PR #133, opened 2015-10-01
Update example command in README.md

EPashkin:


README.md
```diff
@@ -2,6 +2,7 @@
 
 Running:
 ```shell
-cargo run --release -- -d ../gir-files Gtk-3.0
+cargo run --release -- -c Gir_Gtk.toml -d ../gir-files -o ./gtk
```
gkoz: Just `gtk` might be nicer looking than `./gtk`, also maybe just call it something neutral like `out_dir`?

README.md
```diff
@@ -2,6 +2,7 @@
...
 ```
 where `gir-files` contains the [GIR definitions](https://github.com/gkoz/gir-files).
+In `./gtk` will be placed generated files.
```
gkoz: This would be "The generated files will be placed in ..."

gkoz:
Thanks for the improvements. There're some nits but it's a good direction overall.

EPashkin:
`out_dir` to general and and it ask for `-c name.toml` so it will be not command that just copy-paste-run

gkoz:
Thanks.
@homu r+

PR #135, opened 2015-10-03
Make full cover of errors in Config::new

### LINKED ISSUE #134: Config::new error processing
Currently `Config::new` returns `Result<Config, ConfigError>`, but contains many `panic!` and even `exit`.
Tried to expand ConfigError cases to parsing options in https://github.com/EPashkin/gir/commits/error_case, but code look ugly as needed internal macros (which can be changed to new trait on Toml) and don't cover `min_cfg_version` and starting `Docopt`.

May be better just make `Config::new` return `Config` and panic on `read_toml` result instead passing it to `main`?

EPashkin:
Closes #134.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -5,6 +5,7 @@ use std::error::Error;
 use std::io::Error as IoError;
 use std::fmt::{self, Display, Formatter};
 use docopt::Docopt;
+use docopt::Error as DocoptError;
 use toml;
 
 use gobjects;
@@ -33,30 +34,75 @@ impl FromStr for WorkMode {
 
 #[derive(Debug)]
 pub enum ConfigError {
+    CommandLine(DocoptError),
     Io(IoError, String),
     TomlError(String, String),
+    Options(String, String),
 }
 
 impl Error for ConfigError {
     fn description(&self) -> &str {
+        use self::ConfigError::*;
         match *self {
-            ConfigError::Io(ref e, _) => e.description(),
-            ConfigError::TomlError(ref s, _) => s,
+            CommandLine(ref e) => e.description(),
+            Io(ref e, _) => e.description(),
+            TomlError(ref s, _) => s,
+            Options(ref s, _) => s,
         }
     }
 }
 
 impl Display for ConfigError {
     fn fmt(&self, f: &mut Formatter) -> fmt::Result {
+        use self::ConfigError::*;
         match *self {
-            ConfigError::Io(ref err, ref filename) => {
-                try!(write!(f, "\"{}\": ", filename));
+            CommandLine(ref err) => err.fmt(f),
+            Io(ref err, ref filename) => {
+                try!(write!(f, "Failed to read config \"{}\": ", filename));
                 err.fmt(f)
             }
-            ConfigError::TomlError(ref err, ref filename) => {
+            TomlError(ref err, ref filename) => {
+                write!(f, "\"{}\": {}", filename, err)
+            }
+            Options(ref err, ref filename) => {
                 write!(f, "\"{}\": {}", filename, err)
             }
         }
+    }
+}
+
+impl<'a> From<DocoptError> for ConfigError {
+    fn from(e: DocoptError) -> ConfigError {
+        ConfigError::CommandLine(e)
+    }
+}
+
+impl<'a> From<(IoError, &'a str)> for ConfigError {
+    fn from(e: (IoError, &'a str)) -> ConfigError {
+        ConfigError::Io(e.0, e.1.into())
+    }
+}
+
+impl<'a, 'b> From<(&'a str, &'b str)> for ConfigError {
+    fn from(e: (&'a str, &'b str)) -> ConfigError {
+        ConfigError::Options(e.0.into(), e.1.into())
+    }
+}
+
+impl<'a> From<(String, &'a str)> for ConfigError {
+    fn from(e: (String, &'a str)) -> ConfigError {
+        ConfigError::Options(e.0, e.1.into())
+    }
+}
+
+trait LookupStr {
+    fn lookup_str<'a>(&'a self, option: &'a str, err: &str, config_file: &str) -> Result<&'a str, ConfigError>;
+}
+
+impl LookupStr for toml::Value {
+    fn lookup_str<'a>(&'a self, option: &'a str, err: &str, config_file: &str) -> Result<&'a str, ConfigError> {
+        let value = try!(self.lookup(option).ok_or((err, config_file)));
+        Ok(value.as_str().unwrap())
     }
 }
 
@@ -88,9 +134,8 @@ pub struct Config {
 
 impl Config {
     pub fn new() -> Result<Config, ConfigError> {
-        let args = Docopt::new(USAGE)
-            .and_then(|dopt| dopt.parse())
-            .unwrap_or_else(|e| e.exit());
+        let args = try!(Docopt::new(USAGE)
+            .and_then(|dopt| dopt.parse()));
 
         let config_file = match args.get_str("-c") {
             "" => "Gir.toml",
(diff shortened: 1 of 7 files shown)


EPashkin:
Adding trait next

src/main.rs
```diff
@@ -34,9 +34,10 @@ fn main() {
 
     let cfg = match config::Config::new() {
         Ok(c) => c,
+        Err(config::ConfigError::CommandLine(e)) => e.exit(),
         Err(e) => {
-            println!("Failed to read config: {}", e);
-            return;
+            error!("Failed to read config: {}\n", e);
```
gkoz: This will make more sense together with forcing the logger to always print errors. Perhaps, warnings as well.

gkoz: The logger output isn't very pleasant tough so it might make sense to retain a limited number of prints: for fatal errors and announcing the generator mode and possibly the library name.

EPashkin:
Restored `prinltn`.

EPashkin:
`::std::process::exit(1)` caused message `An unknown error occurred` :(
I thought that good for scripting, now restore it also.

gkoz:
> ::std::process::exit(1) caused message An unknown error occurred :(

I don't see why this should happen, but since it just kills the process, the proper use would probably be
```rust
fn main() {
    exit(work());
}

fn work() -> i32 {
    // ...
}
```

EPashkin:
`An unknown error occurred` comes from cargo, no this message if run program directly.
Currently stay without `fn work` as don't think that we need call `exit` on normal cases (not sure about this).
Also removed `Failed to read config` text and just print error.

gkoz:
> Also removed Failed to read config text and just print error.

This is suboptimal for the first time users who have no idea what `Git.toml` is and why it can't be read. It's reasonable to say what kind of action lead to the error.

EPashkin:
Trait almost nice as macro.
Next plan replace `config::ConfigError` => `config::Error` and `TomlError` => `Toml`

EPashkin:
Then `Failed to read config` must go only when IoError.

EPashkin:
May be with comment, `Forgot -c option?`   ?

gkoz:
Ah right, the errors are very different. Maybe the `Display` trait impls should print the full explanation then.

EPashkin:
Expanded message for IoError

EPashkin:
Renamed. Don't like `s` in `Options` but not sure.

src/config.rs
```diff
@@ -32,34 +33,79 @@ impl FromStr for WorkMode {
...
+    }
+}
+
+impl<'a> From<(IoError, &'a str)> for Error {
+    fn from(e: (IoError, &'a str)) -> Error {
+        Error::Io(e.0, e.1.into())
+    }
+}
+
+impl<'a, 'b> From<(&'a str, &'b str)> for Error {
```
gkoz: Variance makes putting different lifetime parameters here unnecessary: a single `'a` will work out to be the shortest of the given borrows.

src/config.rs
```diff
@@ -32,34 +33,79 @@ impl FromStr for WorkMode {
...
+}
+
+trait LookupStr {
+    fn lookup_str<'a>(&'a self, option: &'a str, err: &str, config_file: &str) -> Result<&'a str, Error>;
+}
+
+impl LookupStr for toml::Value {
+    fn lookup_str<'a>(&'a self, option: &'a str, err: &str, config_file: &str) -> Result<&'a str, Error> {
+        let value = try!(self.lookup(option).ok_or((err, config_file)));
+        Ok(value.as_str().unwrap())
```
gkoz: Would it make sense to handle this `unwrap` here as well?

EPashkin: With message `"Option {} not string", option` ?

src/config.rs
EPashkin: Ok.

src/config.rs
EPashkin: But then we need also handle `min_cfg_version` and `external_libraries` and it troublesome, as `unwrap` in clousure.

gkoz: Perhaps `` "Invalid `{}` value, expected a string, found {}", option, value.type_str() ``
(BTW 'option' is a misleading name for a toml entry, or field, anyway)

EPashkin:
Made `min_cfg_version` and `external_libraries` safe.

EPashkin:
As PR still don't merged make PathBuf change here.
Will be at least one commit for PathBufs in Config

EPashkin:
Done. If you want I can leave last two commits for fixing #137

src/config.rs
```diff
@@ -87,78 +148,71 @@ pub struct Config {
...
             .unwrap_or_else(|| Default::default());
         gobjects::parse_status_shorthands(&mut objects, &toml);
 
-        let external_libraries = toml.lookup("options.external_libraries")
-            .map(|a| a.as_slice().unwrap().iter()
-                .filter_map(|v|
+        let external_libraries = match toml.lookup("options.external_libraries") {
+            Some(a) => try!(a.as_result_slice("options.external_libraries", &config_file))
+                .iter().filter_map(|v|
                     if let &toml::Value::String(ref s) = v { Some(s.clone()) } else { None } )
```
gkoz: Is this equivalent to `filter_map(|v| v.as_str().map(String::from))`?

src/config.rs
```diff
@@ -87,78 +148,71 @@ pub struct Config {
...
+                .collect(),
+            None => Vec::new(),
+        };
 
-        let min_cfg_version = toml.lookup("options.min_cfg_version")
-            .map_or_else(|| Ok(Default::default()), |t| t.as_str().unwrap().parse())
-            .unwrap_or_else(|e| panic!(e));
+        let min_cfg_version = match toml.lookup("options.min_cfg_version") {
+            Some(v) => try!(try!(v.as_result_str("options.min_cfg_version", &config_file))
+                            .parse().map_err(|e: String| (e, config_file))),
```
gkoz: This should probably be unrolled into a block with two `try!` lines, it's too difficult to parse.

src/config.rs
```diff
@@ -32,34 +34,93 @@ impl FromStr for WorkMode {
...
 
+impl<'a> From<DocoptError> for Error {
+    fn from(e: DocoptError) -> Error {
+        Error::CommandLine(e)
+    }
+}
+
+impl<P: AsRef<Path>> From<(IoError, P)> for Error {
+    fn from(e: (IoError, P)) -> Error {
+        Error::Io(e.0, e.1.as_ref().into())
```
gkoz: `as_ref().into()` was bothering me so I come up with this:
```rust
impl<P: AsRef<OsStr> + AsRef<Path>> From<(IoError, P)> for Error {
    fn from(e: (IoError, P)) -> Error {
        Error::Io(e.0, PathBuf::from(&e.1))
    }
}
```
The downside is having to require `P: AsRef<OsStr> + AsRef<Path>` everywhere but it should be harmless because `Path` and `OsStr` essentially are interchangeable.

src/config.rs
gkoz: BTW, generally multiline match arms should be put in blocks.

src/config.rs
gkoz: Or, to be precise, if the actual functions took `P: AsRef<OsStr> + AsRef<Path>`, these impls would only need `AsRef<OsStr>`.

src/config.rs
EPashkin: Yes. Thanks

src/config.rs
EPashkin: Ok, and in block as it multiline

EPashkin:
Fixed.

gkoz:
Great work!
@homu r+

gkoz:
> ::std::process::exit(1) caused message An unknown error occurred :(

Since the latest nightlies `cargo` will instead print
```
Process didn't exit successfully: `/path/to/gir/target/release/gir` (exit code: 1)
```

EPashkin:
Good, it clearer

PR #138, opened 2015-10-04
Relative paths in config

### LINKED ISSUE #137: Relative paths in the config should be appended to the config's parent
It's confusing for a path taken from the config to change depending on the current working directory. The only way to ensure predictability seems to start from the directory which contains the config.

EPashkin:
Fix #137

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -158,20 +158,30 @@ impl Config {
             a => a,
         }.into();
 
+        let config_dir = match config_file.parent() {
+            Some(path) => path.into(),
+            None => PathBuf::new(),
+        };
+
         //TODO: add check file existence when stable std::fs::PathExt
         let toml = try!(read_toml(&config_file));
 
         let work_mode_str = match args.get_str("-m") {
-            "" => try!(toml.lookup_str("options.work_mode", "No options.work_mode in config", &config_file)),
+            "" => try!(toml.lookup_str("options.work_mode",
+               "No options.work_mode in config", &config_file)),
             a => a,
         };
         let work_mode = WorkMode::from_str(work_mode_str)
             .unwrap_or_else(|e| panic!(e));
 
         let girs_dir: PathBuf = match args.get_str("-d") {
-            "" => try!(toml.lookup_str("options.girs_dir", "No options.girs_dir in config", &config_file)),
-            a => a
-        }.into();
+            "" => {
+                let path = try!(toml.lookup_str("options.girs_dir",
+                    "No options.girs_dir in config", &config_file));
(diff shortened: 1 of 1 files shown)


EPashkin:
Rebased

src/pathutil.rs
```diff
@@ -0,0 +1,107 @@
...
+        tp!(Some("/c"), "./a/b.txt", "/c/a/b.txt");
+        tp!(Some("/c"), "../a/b.txt", "/c/../a/b.txt");
+    }
+
+    #[test]
+    #[cfg(windows)]
+    fn test_make_base_on_windows() {
+        let none: Option<&str> = None;
+        tp!(none, "a/b.txt", "a/b.txt");
+        tp!(Some("/c"), "a/b.txt", "\\c\\a\\b.txt");
```
gkoz: `r"\c\a\b.txt"`

EPashkin: Oh, forgot

EPashkin:
Fixed

gkoz:
This feels too complicated for the task.

If `config_file` is absolute (will this mess up the error message BTW?) why should its parent be `Option`? Why make `path_based_on` (I'd call it `rebase`) handle the `Option`?

EPashkin:
Don't though that I can safe unwrap parent of absolute path.

EPashkin:
You prefer `config_file` stay unchanged and just extract path from this?

gkoz:
I'm not a fan of making changing the relative path into absolute in the first place. If `parent` returns `None` if there's an empty parent ("Gir.toml") or an invalid input ("C:\" with no file). The first can be solved with prepending "." and the second by doing this after opening the file (and reading "C:\" will return an error).

EPashkin:
It works only if we don't plan changing current folder.

gkoz:
I believe codegen already assumes that we don't play with `current_dir`.

EPashkin:
Then code really simply.

EPashkin:
Strange, `PathBuf::from(".") == PathBuf::new()`, but seems works normally

gkoz:
Oh, that's nice, `PathBuf::new()` is better than an explicit "`.`".

EPashkin:
Done

gkoz:
Looks like some lines over 100 chars now?

EPashkin:
Yes, many. But too many tabs in `library_name` for example, that look bad anyway.

EPashkin:
Fixed some

gkoz:
Thanks! There seem to be opportunities to reducing boilerplate but perhaps some other time.
@homu r+

PR #140, opened 2015-10-13
Various small tweaks

gkoz:
I'm hoping to convince cargo to print a meaningful message instead of "unknown error" when the exit code is non-zero.


src/main.rs
```diff
@@ -30,53 +33,36 @@ mod version;
...
-            ::std::process::exit(1);
+        Err(config::Error::CommandLine(err)) => {
+            if err.fatal() {
+                return Err(Box::new(err));
+            }
+            else {
+                return Ok(())
+            }
         }
+        x => try!(x),
```
EPashkin: IMHO separate Err(err) and Ok(c) will be clearer

gkoz:
Yes, turns out it also was possible to make the `match` smaller.

EPashkin:
`codegen-units` really speed up build.
Thanks.

gkoz:
@homu r+

gkoz:
@homu r+

src/main.rs
```diff
@@ -30,53 +33,30 @@ mod version;
...
 
     let cfg = match config::Config::new() {
-        Ok(c) => c,
-        Err(config::Error::CommandLine(e)) => e.exit(),
-        Err(e) => {
-            println!("{}", e);
-            ::std::process::exit(1);
-        }
+        Ok(cfg) => cfg,
+        Err(config::Error::CommandLine(ref err)) if !err.fatal() => return Ok(()),
```
EPashkin: Missed that `gir --help` (not fatal) don't print anything :(

PR #141, opened 2015-10-13
Fix `--help` handling

gkoz:


gkoz:
@EPashkin thanks for pointing that out

gkoz:
@homu r+

PR #146, opened 2015-10-13
Continuation of #116: use the proper type for objects when c:type is gpointer

gkoz:
A case handled by this PR is
```xml
<type name="StatusIcon" c:type="gpointer"/>
```

### DIFF
--- a/src/codegen/sys/ffi_type.rs
+++ b/src/codegen/sys/ffi_type.rs
@@ -15,12 +15,21 @@ pub fn ffi_type(env: &Env, tid: library::TypeId, c_type: &str) -> Result {
         if let Some(c_tid) = env.library.find_type(0, c_type) {
             // Fast track plain fundamental types avoiding some checks
             if env.library.type_(c_tid).maybe_ref_as::<Fundamental>().is_some() {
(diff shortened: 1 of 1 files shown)


src/codegen/sys/ffi_type.rs
```diff
@@ -15,12 +15,21 @@ pub fn ffi_type(env: &Env, tid: library::TypeId, c_type: &str) -> Result {
         if let Some(c_tid) = env.library.find_type(0, c_type) {
             // Fast track plain fundamental types avoiding some checks
             if env.library.type_(c_tid).maybe_ref_as::<Fundamental>().is_some() {
```
EPashkin: IMHO this not allow Classes in.

gkoz: This tests up if `c_type` names a fundamental.
Updated the PR descr.

gkoz:
I admit this `ffi_type` has become convoluted, confusing and somewhat redundant. I've already started rewriting it but that is what prompted the desire to alter the architecture.

EPashkin:
Ok. Thanks.

gkoz:
@homu r+

Issue #144, opened 2015-10-13
Object returning functions with outs


EPashkin:
Some function has outs and return something, they must return (ret, outs)

(Copy from https://github.com/gkoz/gir/issues/117#issuecomment-143057384)
gtk_cell_area_get_cell_at_position -> CellRenderer,
(throws)gtk_icon_info_load_symbolic -> GdkPixbuf.Pixbuf,
(throws)gtk_icon_info_load_symbolic_finish -> GdkPixbuf.Pixbuf,
(throws)gtk_icon_info_load_symbolic_for_context -> GdkPixbuf.Pixbuf,
(throws)gtk_icon_info_load_symbolic_for_context_finish -> GdkPixbuf.Pixbuf,
(throws, deprecated3.0)gtk_icon_info_load_symbolic_for_style -> GdkPixbuf.Pixbuf,
gtk_tree_selection_get_selected_rows -> GLib.List

All this functions return objects, so it can be converted by XXX::From_glib_none.

EPashkin:
Seems only 2 functions not covered #130

EPashkin:
Closing

Issue #130, opened 2015-09-28
Return Result for errors


EPashkin:
For throwable function return value must return `Result` (see discussion beginning at https://github.com/gkoz/gir/issues/117#issuecomment-143070962).

2 types of throwable functions:
1) without other outs that will returns `Result<(), glib::Error>`
1.1) Examples of FFI function return guint:
gtk_builder_add_objects_from_string
1.2) Examples of FFI function return gboolean:
gtk_css_provider_load_from_path,
gtk_file_chooser_add_shortcut_folder,
gtk_file_chooser_set_current_folder_file,
gtk_page_setup_load_file,
gtk_print_settings_to_file
1.3) Examples of FFI functions return something usable
gtk_icon_info_load_icon,
gtk_recent_info_create_app_info,
gtk_recent_manager_lookup_item
2) with other outs that will returns tuple(or single value) instead () on good case
Examples:
gtk_builder_value_from_string,
gtk_icon_info_load_symbolic

~~Blocked by #103~~

gkoz:
Additionally, if a throwing function returns `Boolean`, it can just be ignored making it `Result<(), glib::Error>`.

A list of throwing functions, return types and annotations: https://gist.github.com/gkoz/17f2ad8a05447b235b27

EPashkin:
You right about return value.
Thanks for list.

gkoz:
`Error` has conversions impls now, solving this will bring us pretty close to complete coverage of existing `gtk` API.

EPashkin:
Good, I try do it tomorrow

Issue #103, opened 2015-07-30
Records in normal mode


EPashkin:
Part of #102
Seems records will more difficult that I think.
- Its needed wrapper (analog of `Object` but maybe without inheritance).
- Also needed autogenerated file with wrapping ffi types. Not sure better do one file for all records in `gtk_sys` or separate some (with functions) or all.
- when record act as `in` parameter then it passed by reference and translated with `to_glib_none`.
- act as return then it optioned and translated with `from_glib_none()`

gkoz:
We seem to be going down the one struct per file path so it makes sense to do the same with records.

Wrapping them is a bit more difficult mostly because some of them are refcounted (`ref`/`unref` methods) and others (let's call them owned) are not (`copy`/`free`). Can't say if using generics is worth it. Possibly. We obviously don't want to expose any of (`ref`, `unref`, `copy`, `free`, and possibly more).

The owned records have to be taken as arguments by `&mut` not `&`. And I haven't really explored the implications for the `translate` API, but I expect it to work.

The kind of ownership transfer (none/full) we know.

gkoz:
> The owned records have to be taken as arguments by `&mut` not `&`.

Unless the function takes a `const Foo*`.

gkoz:
Theoretically, if an owned record is passed as an argument with `transfer: full` we could take the wrapper by value (consume). No idea how many such cases there actually are.

gkoz:
Some "Boxed" types support [has been added](https://github.com/gtk-rs/glib/pull/63). A completely opaque struct can be represented now. Other kinds need more thinking.

EPashkin:
6 owned(boxed) records without fields: `IconSource`(deprecated 3.10), `PaperSize`, `SelectionData`(no constructor), `TreePath`, `TreeRowReference`(no external usage), `WidgetPath`(has copy, free, ref, unref).
5 with fields: `Border`, `Requisition`, `TargetEntry`, `TextIter`, `TreeIter`.
1 without `glib:get-type`: `StockItem` (fields, deprecated 3.10).

Some have only copy, ref, unref: `IconSet`(deprecated 3.10), `TextAttributes`, so it don't count as boxed.
I try write generation for first 6 with `PaperSize` as start.

gkoz:
While the owned and refcounted records need slightly different wrappers, the difference probably doesn't matter for the conversions code.

The `caller-allocates` attribute, that is currently ignored, impacts the outparams records conversions.

[...]

PR #147, opened 2015-10-13
Add StyleContext to Gir_Gtk.toml. It has many issues

gkoz:


EPashkin:
Other candidates:
for #130 file_chooser (file_chooser_widged already added) or builder,
for #144 cell_area or tree_selection

EPashkin:
Also with StyleConcext better add StyleProvider(interface) and Border(record with methods)

EPashkin:
Forget about my add proposal for Border as records currently not realized #103

EPashkin:
`StyleContext` already added in `gtk`.
This PR can be merged after fix or closed.

gkoz:
It still has some problems though, mostly because of records handling.

`gdk::RGBA` is a 'plain old data' record. I'm reluctant to implement `From/ToGlibPtr` wrapper fot it but then the generator needs to distinguish POD records from wrapped ones.

`WidgetPath` doesn't seem to be supported currently.

EPashkin:
About `gdk::RGBA`: we can add `RecordType::POD` with assigned in same case as `RecordType::Direct` if some parameter set in config.
For that type we can: a) don't generate `.rs` and define it manually, b) use conversion specific for POD records

gkoz:
Okay, distinguishing between how `GdkRGBA` and `GdkRectangle` are handled seems too much trouble. Let's just have a single `RecordType` for this case and either assume the From/To conversions exist or generate them if possible. This way we can avoid reexporting any of them from the sys crates.

I'm not sure I understand the `Direct` variant. It seems to imply passing by value. There is a sort of by-value case with arrays of `GtkPageRange`.

EPashkin:
@gkoz Can we close this? `StyleContext` was added to gtk long ago

gkoz:
Okay.

PR #149, opened 2015-10-15
Replace type kind

EPashkin:
Changes in resulted code must be only "Unknown kind" => "Unknown conversion"


src/analysis/rust_type.rs
```diff
@@ -79,11 +79,15 @@ fn rust_type_full(env: &Env, type_id: library::TypeId, nullable: Nullable, by_re
...
-        match TypeKind::of(&env.library, type_id) {
-            TypeKind::Pointer |
-                TypeKind::Object |
-                TypeKind::Interface => rust_type = rust_type.map_any(|s| format!("Option<{}>", s)),
+        use analysis::conversion_type::ConversionType::*;
+        match ConversionType::of(&env.library, type_id) {
+            Owned => {
+                // ignore nullability of containers, they don't require wrapping in an Option
+                match *type_ {
+                    library::Type::List(..) => (),
```
EPashkin: May be better again add flag `skip_option` and do checks sooner.

src/codegen/translate_from_glib.rs
```diff
@@ -12,19 +12,18 @@ pub trait TranslateFromGlib {
...
+            Owned => {
                 let trans = from_glib_xxx(self.transfer);
-                (format!("FromGlibPtrContainer::{}", trans.0), trans.1)
+                match *env.type_(self.typ) {
+                    library::Type::List(..) => (format!("FromGlibPtrContainer::{}", trans.0), trans.1),
+                    _ => trans,
+                }
             }
-            _ => (format!("TODO {:?}:", kind), String::new()),
+            Unknown => ("/*Unknown conversion*/".into(), String::new()),
```
EPashkin: We losing some info here, but no sample code reach that.

src/analysis/rust_type.rs
gkoz: I've probably forgotten something but now it seems the conversion type doesn't matter here: it's sufficient to check if this is a container (only `List` here because I didn't feel like enumerating all containers).

EPashkin: Sorry, don't understand. I added `Type::List` as it only exception before 5962614835f7bd7c2ad016f2f4382864ea2fda00. Currently don't found any other usable containers.
I want restore `skip_option` that removed by your commit, as it simplified this condition.

gkoz:
Things might become clearer if `Simply`, `Owned` are changed to `Scalar`, `Pointer` because that seems the principal distinction between different conversions. In the future though `Pointer` might have to be split into `Refcounted` and `Boxed` because only the latter need mutable borrows (`to_glib_none_mut(&mut self)`).

src/analysis/rust_type.rs
gkoz: Okay. I don't mind putting `skip_option` back.

EPashkin:
Renamed cases. Restored `skip_option`.

EPashkin:
Now I understand what you say. Really checking for conversion type in `rust_type.rs` unneeded.
Removed it.

src/analysis/rust_type.rs
```diff
@@ -78,14 +79,8 @@ fn rust_type_full(env: &Env, type_id: library::TypeId, nullable: Nullable, by_re
...
-    if *nullable {
-        // ignore nullability of containers, they don't require wrapping in an Option
-        match TypeKind::of(&env.library, type_id) {
-            TypeKind::Pointer |
-                TypeKind::Object |
-                TypeKind::Interface => rust_type = rust_type.map_any(|s| format!("Option<{}>", s)),
-            _ => (),
-        }
+    if *nullable && !skip_option {
+        rust_type = rust_type.map_any(|s| format!("Option<{}>", s));
```
gkoz: I'm starting to recall that some scalar typed parapeters might be erroneously marked as nullable (e.g. an outparam is nullable instead of optional) so the check for `ConversionType::Pointer` might indeed be needed. I guess I should've put a better justification in that commit.

EPashkin: Currently no change on sample set without conversion check.

EPashkin: I can restore check if you want.

gkoz: Let's restore it to be safe.

EPashkin:
Restored conversion check in `rust_type.rs`

gkoz:
Great. Thanks for taking care of this enum.
@homu r+

Issue #150, opened 2015-10-20
Introduce code chunks


EPashkin:
Possible preparation for #136.

Code chunks is objects that can convert yourself into code and forms treelike hierarchy.
For example `Block` contains list of other chunks, `Function definition` contains list of argument chunks and return value.
Distant goal: generate `function definition` and `ffi_function_definition` on analyze, do some transformation on them (ex. move output parameters to return values, group functions to `Impl` or `Trait` chunks).
On code generation phase generate function description (based on `function definition`) and body chunks (based on `function definition` and `ffi_function_definition`) and write them to file.

Nearest goal: make `function_body::Builder` generate body chunks and write them without change resulting code.

Tasks for nearest goal
- [x] Silly(just text) chunks for commented function body
- [x] Silly chunks for body with direct call with fallback to current Builder for 
more advanced cases.
- [x] Chunks with `name`, `TypeId` e.t.c. for direct call.
- [x] Chunks for other body cases and remove current Builder.

Next can be done realization throws and other out cases on chunks (#130, #144)
After this need check and remove formatters that can be replaced with use `rustfmt`.

gkoz:
Not sure what "sully" means here.

gkoz:
`rustfmt` is more mature now but it chronically can't deal with lines that are too long. We might forgo any attempts at formatting, make sure each line is below 100 and let rustfmt do the rest.

EPashkin:
At first Parameters chunk for calling FFI just be `CallParam(String)` and call `translate_to_glib` made by chunk creator (same as current Builder).
Next chunk will be changed to `CallParam{name: String, typ: TypeId, ...}` and call `translate_to_glib` yourself on writing chunk code.

EPashkin:
I forgot about `rustfmt`, will try it after nearest goal, as IMHO all needed formating already present.

EPashkin:
Checked `rustfmt`. Now it builds for me (with some failing self-tests). Bad that it only nightly, so it will only external tools until `libsyntax` stabilized.
As it formats entire project from start file it good as external tool.

gkoz:
> Checked rustfmt. Now it builds for me (with some failing self-tests). Bad that it only nightly, so it will only external tools until libsyntax stabilized.

They've just merged a PR to make it work on stable.

PR #151, opened 2015-10-20
Chunks for commented function body

EPashkin:
Part #150 
WIP.


EPashkin:
Almost done.
I thought about move `chunk` and `writer` to root and move `writer::defines` to parent or reexport it.

EPashkin:
Seems `mod to_code` needed be `pub` to stable.

EPashkin:
Rebased to master. All done in this PR.

src/codegen/sys/lib_.rs
```diff
@@ -106,12 +107,12 @@ fn generate_bitfields<W: Write>(w: &mut W, items: &[&Bitfield])
     try!(writeln!(w, ""));
     for item in items {
         try!(writeln!(w, "bitflags! {{\n{}#[repr(C)]\n{0}flags {}: c_uint {{",
-                      tabs(1), item.c_type));
+                      TAB, item.c_type));
         for member in &item.members {
             let val: i64 = member.value.parse().unwrap();
-            try!(writeln!(w, "{}const {} = {},", tabs(2), member.c_identifier, val as u32));
+            try!(writeln!(w, "{}{0}const {} = {},", TAB, member.c_identifier, val as u32));
```
gkoz: We could use named parameters for tabs to make this less confusing:
```rust
try!(writeln!(w, "{tab}{tab}const {} = {},", member.c_identifier, val as u32, tab = TAB));
```
Or perhaps for all arguments.

src/writer/mod.rs
```diff
@@ -0,0 +1,6 @@
+mod defines;
+pub mod primitives;
+pub mod to_code; //TODO:remove pub
```
gkoz: Since the trait is reexported below the `pub` is indeed not necessary?

src/writer/to_code.rs
```diff
@@ -0,0 +1,31 @@
...
+    }
+}
+
+impl ToCode for Chunks {
+    fn to_code(&self) -> Strings {
+        let mut v = Strings::new();
+        for ch in self {
+            let s = ch.to_code();
+            //TODO: append
+            s.iter().map(|s| v.push(s.clone())).count();
```
gkoz: It's not recommended to use `map` for side-effects. A `for` loop here would be more idiomatic.

gkoz:
I don't see the point of aliasing `Vec<String>` to `Strings`: verbosity savings are small, semantics exactly the same. Also some functions seem to consume `Strings` without a need.

src/writer/mod.rs
EPashkin: See https://travis-ci.org/gkoz/gir/builds/86699127 stable failed without this.

src/codegen/sys/lib_.rs
EPashkin: Only for cases with tabs(2)? Make it for all arguments or for all cases IMHO make it more confusing.

src/writer/to_code.rs
EPashkin: Ok. I change it.

EPashkin:
I thought that some formatters will need consume so make it all consuming.
As main formating go to `rustfmt` and all current can receive `&[String]`, seems I was wrong.
One more reason remove `Strings`, as it better use slice as input parameter. I remove `Chunks` also.

src/codegen/sys/lib_.rs
gkoz: To me it would silghtly improve readability even when there's just one tab but that would be a lot of work for little gain.

We could just use `"\t"` and replace it with `"    "` in a writer adapter. Many editors will highlight `\t` in a string.

```rust
try!(writeln!(w, "\t\tconst {} = {},", member.c_identifier, val as u32));
```

EPashkin: Then we needed adapter, than it future will be replaced with nested Chunks.
IMHO if this version is confusing better leave `tabs(2)` and and replace with TAB only `tabs(1)` or remove this commit at all.

gkoz: My order of preference currently is
- `\t\t`
- `{tab}{tab}`
- `{}`, `tabs(2)`
- `{}{0}`, `TAB`

but any style should be used consistently. Since all writing goes through `save_to_file` it seems feasible to plug `\t` replacement there.

EPashkin:
Replaced `map` with `for`. Removed `String` and `Chunks`.
Removed commit with TABs replace, I will try make `\t\t` later.

gkoz:
`\t` and `tabs(x)` would probably coexist then.
Thanks!
@homu r+

PR #152, opened 2015-10-22
Rewrite files with CRLF line terminators

EPashkin:


gkoz:
Hmm, putting in CRLFs? I thought git should be taking care of that with something like
```
git config --global core.autocrlf true
```

gkoz:
Or I'm reading the diff backwards and you're removing them :)

gkoz:
@homu r+

EPashkin:
core.autocrlf=true marks toml.lock always changed, as it restored with CRLF, but recreated by cargo with CR.
Also some version `sh` that (Cygwin) that don't undershand script with CRLF, so I have autoclf=false by default.
Now I try change IDE to emacs, so sometimes make mistake with creating new file.

gkoz:
Will putting this `.gitattributes` help?
```
Cargo.lock eol-lf
```

If that helps I'd consider teaching `cargo new` to add that line. You could also put that in your global `gitattributes` file.

EPashkin:
`.gitattributes` with
```
Cargo.lock eol=lf
*.sh eol=lf
```
seems solves problem.
Global file also works fine.

PR #154, opened 2015-10-24
Add Untabber and replace tabs() with '\t' in impl_static_type

EPashkin:


src/writer/untabber.rs
```diff
@@ -0,0 +1,49 @@
...
+    fn write(&mut self, buf: &[u8]) -> Result<usize> {
+        let v = buf.into();
+        match String::from_utf8(v) {
+            Ok(s) => {
+                let bytes = untab(&s).into_bytes();
+                let delta_len = bytes.len() - buf.len();
+                match self.orig.write(&bytes) {
+                    Ok(n) if n < delta_len => Ok(n),
+                    //hack for return normal len
+                    Ok(n) => Ok(n - delta_len),
```
EPashkin: Without this I get
```
thread '<main>' panicked at 'assertion failed: index.start <= index.end', ../rust/src/libcore\slice.rs:572
```

EPashkin:
Seems I go wrong way for task replace '\t' with spaces on `save_to_file`.
But it worked somehow.

EPashkin:
If you accept that way, I replace other constant `tabs` call with `\t`

EPashkin:
Also minor problem related to #153
Seems `Cargo.toml eol=lf` also need be added for result Sys generate.

src/codegen/function.rs
```diff
@@ -13,7 +13,7 @@ use super::translate_to_glib::TranslateToGlib;
 use writer::primitives::{format_block, tabs};
 use writer::ToCode;
 
-pub fn generate<W: Write>(w: &mut W, env: &Env, analysis: &analysis::functions::Info,
+pub fn generate(w: &mut Write, env: &Env, analysis: &analysis::functions::Info,
```
gkoz: It seems making `Untabber` be generic over the inner writer type would avoid the need to change the function signatures.

src/writer/untabber.rs
gkoz: I wonder if the whole could be done without strings at all. Something like
```rust
let mut chunks = buf.split(|b| b == b'\t').peekable();
loop {
    let chunk = chunks.next();
    try!(self.orig.write(chunk));
    if chunks.peek().is_some() {
        try!(self.orig.write(TAB));
    else {
        break;
    }
}

File: src/codegen/function.rs (hunk 2 above)
Comment (EPashkin) [inline src/codegen/function.rs]: Then we need make function accept Untabber<W> so it seems totally not generic

Comment (EPashkin) [inline src/codegen/function.rs]: I has problem with `save_to_file`.
Tried `W: Write, F: FnMut(&mut W) -> Result<()>` but then rust can't detect right type for closure yourself, so I decided change function to accept trait directly.

File: src/writer/untabber.rs (hunk 1 above)
Comment (EPashkin) [inline src/writer/untabber.rs]: Possible problems:
1) that '\t' aka 0x09 can be part Utf8 character.
2) need check for result internal write and if it return 0, do something

Comment (gkoz) [inline src/writer/untabber.rs]: Code points longer than one byte have the highest bit set so we can be certain any `<128` characters are what they seem to be.

I realize now i should've used `write_all` there (and avoid writes if `chunk.is_empty()`) but since we're writing to files I believe we can ignore any exotic corner cases here and can just return `Ok(buf.len())` on success.

Comment (EPashkin) [inline src/writer/untabber.rs]: I agree that as we using only English code no these character appears.
Function modification by your example.
Only question, can I combine last 2 cases in last match?
```
    fn write(&mut self, buf: &[u8]) -> Result<usize> {
        let mut chunks = buf.split(|b| b == &b'\t').peekable();
        let mut len = 0;
        let tab_bytes = TAB.to_string().into_bytes();
        loop {
            match chunks.next() {
                Some(chunk) => match try!(self.orig.write(chunk)) {
                    0 if chunk.len() > 0 => return Ok(len),
                    n => len = len + n,
                },
                None => break,
            };
            if chunks.peek().is_some() {
                match try!(self.orig.write(&tab_bytes)) {
                    TAB_SIZE => len = len + 1,
                    0 => return Ok(len),
                    _ => unreachable!(),
                };
            } else {
                break;
            }
        }
        Ok(len)
    }
```

File: src/file_saver.rs [hunk 3]
```diff
@@ -1,10 +1,10 @@
-use std::fs;
-use std::fs::File;
-use std::io::Result;
+use std::fs::{self, File};
+use std::io::{Result, Write};
 use std::path::Path;
+use writer::untabber::Untabber;
 
 pub fn save_to_file<P, F>(path: P, make_backup: bool, closure: &mut F) where
```
gkoz: The closure discussion reminds me of a question I've had for a while now: is there a reason to not consume the closure?

EPashkin: Seems no reason for that, just done it automatically long ago.

src/writer/untabber.rs
gkoz: In my opinion using `write` is pointless, `write_all` was made specifically to take care of incomplete writes and stuff. There should be a `BufWriter` between this adapter and `File` anyway.

src/file_saver.rs
EPashkin: There was reason:
`src\file_saver.rs:19:5: 19:12 error: cannot borrow immutable local variable `closure` as mutable`

gkoz: Ah yes, a popular confusion point. The signature is not just signature, but also a pattern match. You can put `mut closure: F` there. Or rebind in the body:
```
let mut closure = closure;
```

EPashkin: Forgot `mut` in function defininuon `pub fn save_to_file<P, F>(path: P, make_backup: bool, mut closure: F) where`
not really no reason

src/writer/untabber.rs
EPashkin: Last version of write:
```
    fn write(&mut self, buf: &[u8]) -> Result<usize> {
        let mut chunks = buf.split(|b| b == &b'\t').peekable();
        loop {
            match chunks.next() {
                Some(chunk) => try!(self.orig.write_all(chunk)),
                None => break,
            };
            if chunks.peek().is_some() {
                try!(self.orig.write_all(&TAB_BYTES));
            } else {
                break;
            }
        }
        Ok(buf.len())
    }
```

EPashkin:
Fixed `write` and consume closure.

src/writer/untabber.rs
```diff
@@ -0,0 +1,37 @@
...
+impl Write for Untabber {
+    fn write(&mut self, buf: &[u8]) -> Result<usize> {
+        let mut chunks = buf.split(|b| b == &b'\t').peekable();
+        loop {
+            match chunks.next() {
+                Some(chunk) => try!(self.orig.write_all(chunk)),
+                None => break,
+            };
+            if chunks.peek().is_some() {
+                try!(self.orig.write_all(&TAB_BYTES));
```
gkoz: `TAB.as_bytes()` might be good enough here.

src/file_saver.rs
```diff
@@ -13,9 +13,10 @@ pub fn save_to_file<P, F>(path: P, make_backup: bool, closure: &mut F) where
         let _backuped = create_backup(&path)
             .unwrap_or_else(|why| panic!("couldn't create backup for {:?}: {:?}", path.as_ref(), why));
     }
-    let mut file = File::create(&path)
-        .unwrap_or_else(|why| panic!("couldn't create {:?}: {}", path.as_ref(), why));
-    closure(&mut file)
+    let file = File::create(&path)
+         .unwrap_or_else(|why| panic!("couldn't create {:?}: {}", path.as_ref(), why));
+    let mut untabber = Untabber::new(Box::new(file));
```
gkoz: We should probably put a `BufWriter` here or risk issuing very short writes.

src/writer/untabber.rs
EPashkin: Yes.

src/file_saver.rs
EPashkin: Not sure if it really needed for files, but add it.

EPashkin:
Added `BufWriter` and removed TAB_BYTES.

gkoz:
Thanks!
@homu r+

EPashkin:
Still not solved dynamic `Write` problem :cry:

EPashkin:
Can you add `Cargo.toml eol=lf` to .gitattributes without PR?

EPashkin:
And don't changed other `tabs`

gkoz:
Sorry. :)
I don't mind passing around `Write` as a trait object really. This could come up if we wanted to create an untabber crate.
Didn't feel like blocking this on having all tabs converted.

PR #156, opened 2015-10-25
Add chunks for functions body with direct call

EPashkin:
Part of #150


EPashkin:
Ping

gkoz:
I've not comprehended the big picture yet but that could be because it's difficult to switch from the ffi analysis rewrite I'm doing.

@homu r+

PR #157, opened 2015-10-28
Factor out analysis::namespaces

gkoz:


gkoz:
Uh-oh, looks like it breaks the old ffi generator a little bit :(

src/codegen/sys/ffi_type.rs
```diff
@@ -161,11 +160,11 @@ fn fix_name(env: &Env, type_id: library::TypeId, name: &str) -> Result {
         match *env.library.type_(type_id) {
             Type::Array(..) | Type::PtrArray(..)
                     | Type::List(..) | Type::SList(..) | Type::HashTable(..) => {
-                if Some(MAIN_NAMESPACE) == env.library.find_namespace("GLib") {
+                if env.namespaces.glib_is_main {
                     Ok(name.into())
                 }
                 else {
-                    Ok(format!("{}::{}", crate_name("GLib"), name))
+                    Ok(format!("{}::{}", &env.namespaces[type_id.ns_id].crate_name, name))
```
EPashkin: If it really right change ( `env.library.find_namespace("Glib") == type_id.ns_is` ) then not better use same as in lines 173-177 (`let name_with_prefix = ...`) ?

EPashkin: Yes in not right change (`*::GList` in build error). Possible solution save glib namespace id instead `glib_is_main` and check it
```
if env.namespaces.glib_ns_id == library::MAIN_NAMESPACE {
   Ok(name.into())
} else {
   Ok(format!("{}::{}", &env.namespaces[env.namespaces.glib_ns_id].crate_name, name))
}
```

gkoz: Yes, this looks like the correct solution.

EPashkin:
When you will rebase `ffi_ng` to this, please, add file `abi_tests.rs` when it first used.

gkoz:
Okay

gkoz:
@homu r+

PR #160, opened 2015-11-04
Records owned

EPashkin:
Part of #103

WIP.


src/analysis/functions.rs
```diff
@@ -19,7 +19,7 @@ pub struct Info {
     pub glib_name: String,
     pub kind: library::FunctionKind,
     pub comented: bool,
-    pub class_name: Result,
+    pub owner_name: Result,
```
gkoz: `owner` would probably be clearer as `receiver` (as in receiver of the method) or `type`.

EPashkin: `type` is better

EPashkin:
Renamed `owner_xxx` to `type_xxx`

EPashkin:
Ready for merge.
Generated files is compiling in gtk (after replacement `*mut` to `*const` in sys)
with 2 warning about unused `use glib::types;` and `use object::*;`,
so I plan change `codegen::general::uses` in next PR

EPashkin:
Removed TODO

gkoz:
Thanks!
@homu r+

PR #161, opened 2015-11-06
Rework uses

EPashkin:
Uses now look like this
```
use PageOrientation;
use PaperSize;
use Unit;
use ffi;
use glib::translate::*;
use glib::types;
use object::*;
```


gkoz:
Sorry for the delay.
@homu r+

PR #162, opened 2015-11-14
Aliases

EPashkin:
Generated WidgetPath needed gkoz/gtk-rs-glib#2 to build

### DIFF
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -83,7 +83,8 @@ fn analyze_function(env: &Env, func: &library::Function, type_tid: library::Type
     if unsupported_outs {
         warn!("Function {} has unsupported outs", func.c_identifier.as_ref().unwrap_or(&func.name));
         commented = true;
-    } else if !outs.is_empty() {
+    } else if !outs.is_empty() && !commented {
+        //TODO: move to out_parameters::analyze
         imports.add("std::mem".into(), func.version);
     }
 
--- a/Gir_Gtk.toml
+++ b/Gir_Gtk.toml
@@ -103,6 +103,8 @@ manual = [
     "Gtk.Dialog",
     "Gtk.Widget",
     "Gtk.Window",
+    "Gdk.Rectangle",
+    "cairo.RectangleInt",
 ]
 
 ignore = [
--- a/src/analysis/conversion_type.rs
+++ b/src/analysis/conversion_type.rs
@@ -48,6 +48,7 @@ impl ConversionType {
                 None => ConversionType::Unknown,
                 Unsupported => ConversionType::Unknown,
             },
+            Alias(ref alias) => ConversionType::of(library, alias.typ),
             Bitfield(_) => ConversionType::Direct,
             Record(_) => ConversionType::Pointer,
             Enumeration(_) => ConversionType::Direct,
--- a/src/analysis/out_parameters.rs
+++ b/src/analysis/out_parameters.rs
@@ -4,6 +4,7 @@ use std::vec::Vec;
 use env::Env;
 use library::*;
 use super::conversion_type::ConversionType;
+use super::rust_type::parameter_rust_type;
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub enum Mode {
@@ -72,7 +73,7 @@ fn can_as_return(env: &Env, par: &Parameter) -> bool {
     match ConversionType::of(&env.library, par.typ) {
         Direct => true,
         Scalar => true,
-        Pointer => false,
+        Pointer => parameter_rust_type(env, par.typ, ParameterDirection::Out, Nullable(false)).is_ok(),
         Unknown => false,
     }
 }
--- a/src/analysis/rust_type.rs
+++ b/src/analysis/rust_type.rs
@@ -71,7 +71,8 @@ fn rust_type_full(env: &Env, type_id: library::TypeId, nullable: Nullable, by_re
         //TODO: check usage library::Type::get_name() when no _ in this
     };
 
-    if type_id.ns_id != library::MAIN_NAMESPACE && type_id.ns_id != library::INTERNAL_NAMESPACE {
+    if type_id.ns_id != library::MAIN_NAMESPACE && type_id.ns_id != library::INTERNAL_NAMESPACE
+        && !implemented_in_main_namespace(&env.library, type_id) {
         let rust_type_with_prefix = rust_type.map(|s| format!("{}::{}",
             crate_name(&env.library.namespace(type_id.ns_id).name), s));
         if env.type_status(&type_id.full_name(&env.library)).ignored() {
@@ -194,6 +195,16 @@ fn use_by_ref(library: &library::Library, type_: &library::Type, direction: libr
             let type_ = library.type_(alias.typ);
             use_by_ref(library, type_, direction)
         }
+        _ => false,
+    }
+}
+
+fn implemented_in_main_namespace(library: &library::Library, type_id: library::TypeId) -> bool {
+    if library.namespace(library::MAIN_NAMESPACE).name != "Gtk" {
+        return false;
+    }
+    match &*type_id.full_name(library) {
+        "Gdk.Rectangle" => true,
         _ => false,
     }
 }
--- a/src/analysis/rust_type.rs
+++ b/src/analysis/rust_type.rs
@@ -54,6 +54,8 @@ fn rust_type_full(env: &Env, type_id: library::TypeId, nullable: Nullable, by_re
                 _ => err(&format!("Fundamental: {:?}", fund)),
             }
         },
+        Alias(ref alias) => rust_type_full(env, alias.typ, nullable, by_ref)
+                .map_any(|_| alias.name.clone()),
 
         Enumeration(ref enum_) => Ok(enum_.name.clone()),
         Bitfield(ref bitfield) => Ok(bitfield.name.clone()),
@@ -95,6 +97,7 @@ pub fn used_rust_type(env: &Env, type_id: library::TypeId) -> Result {
     use library::Type::*;
     match *env.library.type_(type_id) {
         Fundamental(library::Fundamental::Type) |
+            Alias(..) |
             Bitfield(..) |
             Record(..) |
             Class(..) |
@@ -109,7 +112,7 @@ pub fn parameter_rust_type(env: &Env, type_id:library::TypeId,
         direction: library::ParameterDirection, nullable: Nullable) -> Result {
     use library::Type::*;
     let type_ = env.library.type_(type_id);
-    let by_ref = use_by_ref(type_, direction);
+    let by_ref = use_by_ref(&env.library, type_, direction);
     let rust_type = rust_type_full(env, type_id, nullable, by_ref);
     match *type_ {
         Fundamental(fund) => {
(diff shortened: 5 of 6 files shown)


EPashkin:
Build error because [generate_constants](https://github.com/gkoz/gir/blob/master/src/codegen/sys/lib_.rs#L120-L142) uses `parameter_rust_type` that returns not ffi type names.
Before code looked
```
//pub const PANGO_GLYPH_EMPTY: Unknown rust type: "Glyph" = 268435455;
//pub const PANGO_GLYPH_INVALID_INPUT: Unknown rust type: "Glyph" = 4294967295;
//pub const PANGO_GLYPH_UNKNOWN_FLAG: Unknown rust type: "Glyph" = 268435456;
```
now 
```
pub const PANGO_GLYPH_EMPTY: Glyph = 268435455;
pub const PANGO_GLYPH_INVALID_INPUT: Glyph = 4294967295;
pub const PANGO_GLYPH_UNKNOWN_FLAG: Glyph = 268435456;
```
and must use `PangoGlyph`.
I can add hardcoded fix to `generate_constants` same as for `"&str"`

EPashkin:
Tried to find alias for replacing `Quark` but don't find any good, so just change conversion for him to reflect changes in glib.

gkoz:
Hmm, I wonder how a PR, that touches neither parser nor sys codegen, managed to break the sys tests. I'll take a better overall look tomorrow.

EPashkin:
See https://github.com/gkoz/gir/pull/162#issuecomment-156713825.
PR just uncomment some constants.

gkoz:
Ah, okay. Hardcoded is fine, the ffi_ng branch is supposed to replace that anyway.

EPashkin:
Fixed build, ready for merge

src/analysis/functions.rs
```diff
@@ -89,6 +95,10 @@ fn analyze_function(env: &Env, func: &library::Function, type_tid: library::Type
                 imports.add(s, func.version);
             }
         }
+        //TODO: change to check upcasts count
```
EPashkin: Forgot about this TODO, doing it now

EPashkin:
Done

EPashkin:
Next attempt for alias, not base on `GdkRectangle` 
Still in progress.
```
    #[cfg(gtk_3_12)]
    fn set_pointing_to(&self, rect: &gdk::Rectangle) {
        unsafe {
            ffi::gtk_popover_set_pointing_to(self.upcast().to_glib_none().0, rect);
        }
    }
```

EPashkin:
Counting GdkRectangle as direct arise problem with
`gtk_image_new_from_surface`: it lose optional
```
    #[cfg(gtk_3_10)]
    pub fn new_from_surface(surface: &cairo::Surface) -> Image {
        unsafe {
            Widget::from_glib_none(ffi::gtk_image_new_from_surface(surface)).downcast_unchecked()
        }
    }
```
not sure what better to do:
a) change conversion back to pointer or scalar (for that needed write wrapper in cairo and reexport it in Gdk instead ffi)
b) write some fix for body and optional records.
c) create other conversion type `DirectPointer`, that same as direct but allow optional

gkoz:
Suppose we had a
```rust
#[repr(C)]
struct Rectangle {
    x1: i32,
    y1: i32,
    x2: i32,
    x2: i32,
}
```
and faked conversions by using transmutes
```rust
impl<'a> ToGlibPtr<'a, *const GdkRectange> for &'a Rectangle {
    type Storage = &'a Rectangle;
    fn to_glib_none(&self) -> Stash<'a, &'a Rectangle> {
        Stash(transmute(*self), *self)
    }
}
```

EPashkin:
So it act as ConversionType::Pointer?
This conversion will be added to cairo (as I know rust don't allow this in other crates)?

EPashkin:
Ignore last comment.

EPashkin:
So my reexport commit need changed?

EPashkin:
Also needed implementation for `Uninitialized` as it needed for out parameter in `gtk_range_get_range_rect`

gkoz:
BTW the hacky translations aren't strictly necessary, we could start with ones that actually copy the data but are less risky.

I'd leave `gdk` alone for now.

It could be, handling aliases is more of a distraction at this time. We can't make the generator perfect so some functions will need to be done manually. Perhaps it's time to start experimenting with how exactly that is going to work.

gkoz:
> I'd leave `gdk` alone for now

 -- i.e. this `struct Rectangle` could even be defined in `gtk` first.

EPashkin:
So functions will look like this? (It need some hardcoding in `rust_type` to remove `gdk::`)
```
    #[cfg(gtk_3_12)]
    fn set_pointing_to(&self, rect: &Rectangle) {
        unsafe {
            ffi::gtk_popover_set_pointing_to(self.upcast().to_glib_none().0, rect.to_glib_none().0);
        }
    }

    #[cfg(gtk_3_10)]
    pub fn new_from_surface(surface: Option<&cairo::Surface>) -> Image {
        unsafe {
Widget::from_glib_none(ffi::gtk_image_new_from_surface(surface.to_glib_none().0)).downcast_unchecked()
        }
    }
```

gkoz:
Yes, that looks good.

Eventually might be a hardcoding issue *if* we try to make gdk and cairo share the same type.

EPashkin:
As `ToGlibPtr` already present in nonobject glib we can just declare `RectangleInt` in `cairo` and reexport it (and not GdkRectangle) in Gdk.

EPashkin:
By way `cairo_rectangle_int_t` not (x1,y1)-(x2,y2) and (x1,x2,width, height)

gkoz:
Thanks for the correction, I was too lazy to look it up. The point is moving from `c_int` in the original struct to `i32`, notably a different approach from what gi-rust is aiming to do, so this choice might be questioned or turn out wrong later.

EPashkin:
How about conversion type `DirectPointer`, this don't needed any change in libs and work directly with ffi types?

gkoz:
I don't have good arguments against that right now. Just that having `i32` and `u32` almost everywhere but one or two structs with `c_int`s doesn't look very nice. Especially if some ABI later manages to make `c_int != i32`.

EPashkin:
I'm afraid of slightly more than one - two: GLib.DateTime, GdkPoint,  GtkTargetEntry (it boxed with fields and without getters, not sure that it really needed get\set for using as it has good constructor), GtkBorder(also boxed, but with empty constructor, important fields and no get\set), others in https://github.com/gkoz/gir/issues/103#issuecomment-153145754.
So this issue requires a lot of thinking.

gkoz:
`DateTime` should be converted into whatever Rust counterpart we choose (e.g. `chrono::DateTime`) and is opaque, `TargetEntry` carries a pointer so it's definitely not POD. `Point` yep. `Border` is a weird case but at least the fields unambiguously are `u16`.

EPashkin:
Tried update macro `glib_boxed_wrapper` for boxed records with fields with `TargetEntry` as example (https://github.com/EPashkin/rust-gnome-glib/tree/glib_boxed_wrapper_with_fields).
@gkoz , how it look for you?
```
glib_wrapper! {
    pub struct TargetEntry(Boxed<ffi::GtkTargetEntry>);

    fields {
        get pointer (get_target) target: String,
        get direct (get_flags) flags: u32,
        get direct (get_info) info: u32,
    }

    match fn {
        copy => |ptr| ffi::gtk_target_entry_copy(ptr as *mut ffi::GtkTargetEntry),
        free => |ptr| ffi::gtk_target_entry_free(ptr),
    }
}

impl TargetEntry {
    pub fn new(target: &str, flags: u32, info: u32) -> TargetEntry {
        unsafe {
            from_glib_full(ffi::gtk_target_entry_new(target.to_glib_none().0, flags, info))
        }
    }

    pub fn test(&self){
        let target: String = self.get_target();
        let flags:u32 = self.get_flags();
        let info:u32 = self.get_info();
    }
}
```

gkoz:
The most complicated yet essential goal would be supporting [GdkEvent](https://developer.gnome.org/gdk3/stable/gdk3-Event-Structures.html) records.

`Deref` might be a better match than `AsRef`, you won't even have to call `as_ref` then.

The macros don't seem to buy us much because of all the variations. For instance, getting a string out might work better (assuming there is an appropriate implementation) as

```rust
    fn target(&self) -> Cow<str> {
        from_glib_borrow(self.0.target)
    }
```
but encoding this in the macro syntax complicates things. Even the integer case looks simple enough thanks to `Deref` (unless I'm missing something)
```rust
    fn info(&self) -> u32 {
        self.0.info
    }
```

EPashkin:
I don't even want think about `GdkEvent` (all unions will be problems)
`Deref` is much better.
Updated branch, not sure if setter done right way.
About custom functions, we can make macro that accept functions like this
```
glib_wrapper! {
    pub struct TargetEntry(Boxed<ffi::GtkTargetEntry>);

    impl {
        fn target(&self) -> Cow<str> {
            from_glib_borrow(self.0.target)
        }
        fn info(&self) -> u32 {
            self.0.info
        }
    }

    match fn {
        copy => |ptr| ffi::gtk_target_entry_copy(ptr as *mut ffi::GtkTargetEntry),
        free => |ptr| ffi::gtk_target_entry_free(ptr),
    }
}
```
May be it be better as it more customizable.

EPashkin:
As for `GdkEvent` can we do this?
Declare rust enum `Event_` with supported cases (possibly with not full field list) and `Unsupported` case and `struct Event(Event_, gdk_ffi::GdkEvent, MemoryManager<gdk_ffi::GdkEvent>)`
On `from_glib_full` create `Event_` by copying supported fields or `Event_::Unsupported` if case not supported.
On `to_glib_none` copy supported fields back to `gdk_ffi::GdkEvent` and return it.

EPashkin:
Not sure if any fields of received event need be changed by rust code.
If no thus fields then `to_glib_none` will be simply returning stored `gdk_ffi::GdkEvent`.

Also now `gdk_ffi::GdkEvent = c_void` it needed to change to `[u8;N]` where `N` is size of union.

EPashkin:
* If no thus fields => If no such fields

EPashkin:
About `https://github.com/gkoz/gir/pull/162#issuecomment-159004477`
I was wrong: for custom accessors we don't need macro as we can just define `impl TargetEntry` with these functions.

gkoz:
An enum doesn't allow to put a specific event type in a signature.

One possible strategy is to make `struct Event(Boxed<GdkEvent>)`, that converts to both `*mut GdkEvent` and `*mut GdkEventAny`, and make the rest wrap `Event`, e.g. `struct EventButton(Event)` with `Deref`.

EPashkin:
I talked about internal type that can be filled from `GdkEvent` and if needed copy changes back.
```
enum Event_{
   Key{
       is_release: bool,
       send_event: i8,
       time: u32,
       keyval: u32,
   },
   Button{
       type: BtnEventType,
       send_event: i8,
       time: u32,
       x: double,
       y: double,
       state: u32,
   },
...
   Unsupported,
}
```

EPashkin:
Oh, sorry, you saying about callback already, then you right, I missed that callback need accept single type.

EPashkin:
Rebased. Last commit only needed if `Rectangle` will placed in `gtk`

gkoz:
Looks like all aliases are tricky so this is difficult to evaluate. `Rectangle` needs `caller-allocates` support we don't have yet.

EPashkin:
I done caller support already, add PR after this.

EPashkin:
Or can add it to this PR

gkoz:
Fine, I'll merge this and we'll fix any problems as they come.

@homu r+

PR #163, opened 2015-11-18
Ignored functions

EPashkin:


gkoz:
While okay for now, we'll need one or more generalizations of these quirks:
 * there already are ignored and non-nullable, additionally unix-specific functions will need to be configured (the gio-unix problem);
 * the same applies to objects and perhaps constants, etc;
 * it could be useful to allow regexes (so everything `.*_unix_.*` could be matched at once).

EPashkin:
While global `ignored`, that applies not only functions but objects and notably constant is good,
it can greatly slow `gir` due large number of variants and checks so need use with caution and better separated by type.
As for local `ignored_function`, making it regex-based is good idea I try do it in this PR.

EPashkin:
Add regex support

Gir_Gtk.toml
```diff
@@ -357,3 +357,10 @@ status = "generate"
 [[object]]
 name = "Gtk.PaperSize"
 status = "generate"
+
+[[object]]
+name = "Gtk.WidgetPath"
+status = "generate"
+ignored_functions = [
+    '''iter_has_q\w+''',
```
gkoz: Does this need triple quotes? It seems they're necessary if there's a single quote inside the string.

EPashkin: if use single qupte `\` count as start escape string

gkoz: So this quote from [here](https://github.com/toml-lang/toml#string) is not correct?

> Literal strings are surrounded by single quotes. Like basic strings, they must appear on a single line:

>```
# What you see is what you get.
winpath  = 'C:\Users\nodejs\templates'
winpath2 = '\\ServerX\admin$\system32\'
quoted   = 'Tom "Dubs" Preston-Werner'
regex    = '<\i\c*\s*>'
```

EPashkin: Oh, not tried really single quote (always use double for strings). Its work.

gkoz:
Does this PR contain everything from #162, is it okay to just merge this?

EPashkin:
Removed triple quotes in config.
This PR ready for merge, and contains some cherry-picked commits that fixes building generated records, main functionality (aliases) still not copied.

gkoz:
@homu r+

PR #164, opened 2015-11-19
Comment out functions with ignored records and classes among parameters

gkoz:


EPashkin:
Only find strange changes in `gtk_search_bar_connect_entry` it now commented without any visible 'Ignored' for `Entry`.
Not sure that needed fix.

EPashkin:
```
    //#[cfg(gtk_3_10)]
    //pub fn connect_entry<T: Upcast<Entry>>(&self, entry: &T) {
    //    unsafe { TODO: call ffi::gtk_search_bar_connect_entry() }
    //}
```

gkoz:
Looks like `Entry` is not in the toml and so ignored by default. Adding it restores this method.

EPashkin:
Yes, but error that /*Ignored*/ not shown before `Entry` as printed by other methods.
IMHO it minor and can be ignored.

gkoz:
I'll see if it can be fixed easily

EPashkin:
Good fix.

gkoz:
@homu r+

PR #167, opened 2015-11-26
Add interfaces support to object analysis

gkoz:


EPashkin:
Not needed `Interface` case in `rust_type::use_by_ref`?
Same in `needed_upcast::needed_upcast`? Not sure about last.

EPashkin:
Better add some examples in `Gir_Gtk.toml` (as minimum generate `Gtk.Orientable`).

gkoz:
> Not needed Interface case in rust_type::use_by_ref?
Same in needed_upcast::needed_upcast?

You're right about both, updated.

EPashkin:
Thanks.

src/analysis/parents.rs
```diff
@@ -29,3 +29,26 @@ pub fn analyze(env: &Env, type_: &Class, imports: &mut Imports)
     parents.reverse();
     parents
 }
+
+pub fn analyze_interface(env: &Env, type_: &Interface, imports: &mut Imports)
+        -> Vec<StatusedTypeId> {
+    let mut parents = Vec::new();
+
+    for &parent_tid in &type_.prereq_parents {
+        let parent_type = env.type_(parent_tid).to_ref_as::<Class>();
```
EPashkin: `Interface`?

gkoz:
@homu r+

gkoz:
@homu r-

gkoz:
I'll look into that one

src/analysis/parents.rs
gkoz: The prerequisites are usually classes. Not sure if I ever saw an interface depend on other interfaces.

gkoz: But yeah, there's a `TreeModel` prerequisite to `TreeSortable` so this needs to be generalized.

EPashkin:
May be both class(`Widget`) and interface(`TreeModel`) prerequisites

PR #168, opened 2015-11-26
Interface prerequisites aren't limited to classes

gkoz:


gkoz:
This doesn't achieve `pub struct TreeSortable(Object<ffi::GtkTreeSortable>): Treemodel;` yet.

gkoz:
Apparently the same incorrect assumption was made in the `library` module

Gir_Gtk.toml
```diff
@@ -331,6 +331,14 @@ name = "Gtk.ToolPalette"
 status = "generate"
 
 [[object]]
+name = "Gtk.TreeModel"
```
EPashkin: Can you add it to `generate` array?

EPashkin:
Thanks.

EPashkin:
Thanks

EPashkin:
oops. Two `Gtk.Revealer'

gkoz:
Hmm, this means I might have lost something.

gkoz:
Removed the duplicate.

EPashkin:
From old version no missed files

EPashkin:
By way generated `Gtk.Container` contains all functions from https://github.com/gkoz/gtk-rs-gtk/blob/object_reform_gir/src/widgets/container.rs with only `get_children` differs.

gkoz:
Yeah, I'm going to switch more classes to `generated` in that branch.

@homu r+

PR #169, opened 2015-11-28
Caller allocates

EPashkin:


EPashkin:
Added example for outs with `caller-allocates=0`

EPashkin:
Still not solved problem with sys mentioned in https://github.com/gkoz/gir/issues/103#issuecomment-160305272. I though that https://github.com/gkoz/gir/tree/ffi_ng will solve it.

gkoz:
> Still not solved problem with sys mentioned in #103 (comment). I though that https://github.com/gkoz/gir/tree/ffi_ng will solve it.

So far I wasn't intending to override constness in `ffi_ng`, and the normal mode would still have to do something too.

EPashkin:
Current boxed macro don't define `impl<'a> $crate::translate::ToGlibPtr<'a, *mut $ffi_name> for $name` (was defined for objects)

gkoz:
Having such an impl would be going too far. I've been overriding that with a [cast to `*mut`](https://github.com/gtk-rs/gtk/blob/b400a92e5449cd26c56b8cc57f53fa8eb3789a58/src/widgets/tree_store.rs#L111-L114).

gkoz:
We could have something like this to avoid scary casts
```
unsafe fn force_mut<T>(ptr: *const T) -> *mut T {
    ptr as *mut T
}
```

EPashkin:
As all `instance-parameter` converted to `*mut` then ` as *mut _` will be everywhere.
We can add new conversion type for that but IMHO it not good solution.

gkoz:
By default a `*mut` should require `&mut` for owned `Boxed` types, only when we tell the generator `&` is enough, should there be an override. (also seems that fn doesn't have to be unsafe)

EPashkin:
Or new conversion type for records that always use ".to_glib_none_mut()`?

gkoz:
In the generator? Yes it probably makes sense to treat owned and shared pointer types differently.

EPashkin:
Seems I don't understand something. Can you give example for `owned Boxed` type and for `shared Boxed` type?

gkoz:
This is largely to avoid ambiguities: I'm not positive what exactly is considered boxed in gnome terminology. [Here](https://developer.gnome.org/gtk3/stable/ch02.html) some refcounted types are put under `Boxed`. [This](https://developer.gnome.org/gobject/unstable/gobject-Boxed-Types.html) suggests the refcounted types shouldn't be called that.

So I call owned `Boxed` types those which use `Boxed` wrapper. Shared ones use `Refcounted` but someone might still consider them boxed.

EPashkin:
So for all records with `Boxed<ffi::*>` (that all current records) we need use `&mut self` or `&mut param` in function descriptions and `to_glib_none_mut` for calling ffi?

EPashkin:
*all current records => currently generated records

gkoz:
> So for all records with `Boxed<ffi::*>` (that all current records) we need use `&mut self` or `&mut param` in function descriptions and `to_glib_none_mut` for calling ffi?

Yes if the pointer is `*mut` in the FFI signature, and most of the time it is.

If it's const, e.g.
```C
gint gtk_text_iter_get_line (const GtkTextIter *iter);
```
or we override the constness somehow (e.g. pretend `gtk_tree_store_iter_is_valid` takes `*const TreeIter`), it's `&self` and `param: &Struct`.

EPashkin:
So if parameter mutable in ffi and it overridden then on call ffi it will be converted with `as *mut _` or `.force_mut()`.
I think about it.
IMHO this PR ready and can be merged. Or it can be postponed until boxed record generation fixed.
Only found now that you added `auto\paper_size.rs` to `gtk` but not included it in `mod.rs`, though that it normally compiles with current sys.

gkoz:
`force_mut` might need a better name, which would express the promise that it will actually be treated as a const pointer.

`paper_size` had me confused for a moment because of the "Class ... not found" message, I'll put it back into `Gir.toml`.

I'm merging this and adding a `Rectangle` wrapper to make the compiler happy.

@homu r+

gkoz:
I intend to alter `glib::translate::if_some` to take a closure and then it can be used in place of `if ret { Some(...) } else { None }` with additional benefit that it accepts `gboolean` directly.

gkoz:
Um, actually `PaperSize` doesn't seem ready to be included yet.

PR #170, opened 2015-11-29
Mutable reference for non refcounted records

EPashkin:


EPashkin:
Found possible problem with `Boxed` macro:
`gtk_widget_path_iter_get_siblings` return `*const GtkWidgetPath` and your `from_glib_full` don't accept it (allow only `*mut`)

EPashkin:
Current stage allow to compile `gtk` with records.
Override the constness will go to other PR as I not sure how it better do.
Other possible problem is unneeded adding `use std::mem` for records (also not sure how do to fix it)

gkoz:
> `gtk_widget_path_iter_get_siblings` return `*const GtkWidgetPath` and your `from_glib_full` don't accept it (allow only `*mut`)

It might make sense to add a generic implementation for `*const T` bounded on existing `*mut T` implementation. This implementation would have to leave `from_glib_borrow` and `from_glib_full` `unimplemented!()` though. (`from_glib_borrow` currently assumes mutable borrow and `transfer: full` would not be placed on a const pointer).

---

> Override the constness will go to other PR as I not sure how it better do.

I've been thinking: there already is overriding return types nullability. A generalization of both needs could allow assigning quirks to any parameter or return value of a function, something like
```toml
[[object]]
name = "Gtk.TreeModel"
[[object.function]]
name = "get_path"
[object.function.parameter]
name = "iter"
const = true

[[object]]
name = "Gtk.EntryBuffer"
[[object.function]]
name = "get_text"
[object.function.return]
nullable = false
```

Some clumsiness is a consequence of the chosen config style. An alternative could look like
```toml
[Gtk.TreeModel.fn.get_path.parameter.iter]
const = true

[Gtk.EntryBuffer.fn.get_text.return]
nullable = false
```

gkoz:
Now that this `from_glib_borrow` deficiency has become apparent I might remove it and take a different approach: make `Borrow<T>` and `BorrowMut<T>` structs, that would deref to `T` and forget it when dropped. This would achieve the same goal of not touching `ref/unref` or `free` when borrowing and handing out references to `T` in such cases.

[This](https://github.com/gtk-rs/gtk/blob/master/src/signal.rs#L1349-L1350)
```rust
            f(FFIWidget::wrap_widget(this as *mut _), &mut from_glib_borrow(iter),
                TreePath::wrap_pointer(path));
```
would then change to
```rust
            f(FFIWidget::wrap_widget(this as *mut _), &mut BorrowMut::from(iter),
                TreePath::wrap_pointer(path));
```

EPashkin:
Are you sure that last comment belong this PR?
I can not estimate your decision, as not studied the current realization of signals.

gkoz:
Just had to put this somewhere so it wouldn't be a complete surprise and this is relevant to `FromGlibPtr` impls discussion. :) Callbacks (in particular, signals) are the main use case of the borrow mode, where we get a pointer that outlives the scope.

EPashkin:
Ping

gkoz:
Apologies, I haven't had much time lately but still hope to finish the reform this year.

The `Borrow` idea from above doesn't feel quite right, so I'm not implementing it for now.

@homu r+

EPashkin:
No problem, I also had no free time, so I don't even tried do my next PRs (about constness overriding and use rework)

PR #171, opened 2015-12-19
Function config with parameters immutability

EPashkin:
WIP


EPashkin:
All done

gkoz:
Very good work, thanks.
I guess it's too late to pick any nits, I'll open follow-up PRs to address things that get in the way.
@homu r+

Happy New Year!

PR #172, opened 2016-01-02
Small fixes

gkoz:
The second commit relies on the recently pushed updates to `glib#object_reform`


EPashkin:
Look good.

gkoz:
Added another one, fixes some unused and missing 'std::mem` and `std::ptr` imports.

EPashkin:
May be better extract `codegen::function_body_chunk::OutMemMode` to analyze and add method for imports?

EPashkin:
Current code good too.

gkoz:
I'm not very happy with the current patch will consider your suggestion.

gkoz:
While managing imports in analysis seems too early, communicating the needed imports back from chunks seems problematic as well. But it might make sense to move the imports analysis into a separate function.

gkoz:
The imports code is more isolated now and currently I doubt it can be moved to codegen.

gkoz:
@homu r+

PR #173, opened 2016-01-05
Add `module_name` config setting

gkoz:


src/codegen/objects.rs
```diff
@@ -23,13 +23,17 @@ pub fn generate(env: &Env, root_path: &Path, mod_rs: &mut Vec<String>, traits: &
             }
         };
 
-        let path = root_path.join(file_name(&class_analysis.full_name));
+        let mod_name = obj.module_name.clone().unwrap_or_else(|| {
```
EPashkin: May be better update `nameutils::file_name` to accept `&class_analysis.full_name` as `full_name` and `obj.module_name` as `configured_name` ?

gkoz: `file_name` just doesn't seem that useful. It derives the module name (this should not be its concern and has caused code duplication in practice) and puts an extension on it (which is trivial).

gkoz: We could move the first part into a function perhaps.

gkoz:
@homu r+

Issue #174, opened 2016-01-05
Generate safety assertions


gkoz:
All* non-method functions functions need to have safety assertions introduced in https://github.com/gtk-rs/gtk/pull/189, e.g.

```rust
impl Box {
    pub fn new(orientation: Orientation, spacing: i32) -> Box {
        assert_initialized_main_thread!();
        ...
```

\* if we take an object that has an assertion in its constructor, it can be skipped, but this optimization is not strictly necessary:
```rust
 impl Layout {
    pub fn new(hadjustment: &Adjustment, vadjustment: &Adjustment) -> Layout {
        skip_assert_initialized!();
        ...
```

The whole feature should probably be optional.

EPashkin:
Assertion only for objects "static" functions or some records need it too?
Can be term `non-method functions` replaced with `constructors`?

Is safe to say that all objects has assertion in constructor or it depends on other asserted objects?
Then second part equal check constructors for object parameters.

gkoz:
Whether an object is involved or not, we have to start with the assumption that all functions need the assertions but methods obviously don't. It should be workable to rely on all constructors, including the manually written ones, having the assertion.

PR #175, opened 2016-01-06
Generate safety assertion in non-member functions

EPashkin:
Part of #174

### DIFF
--- a/src/analysis/safety_assertion_mode.rs
+++ b/src/analysis/safety_assertion_mode.rs
@@ -1,4 +1,5 @@
 use analysis::parameter::Parameter;
+use env::Env;
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub enum SafetyAssertionMode {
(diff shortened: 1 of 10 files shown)


EPashkin:
Still thinking of other way to check when use `skip_assert_initialized`

src/analysis/safety_assertion_mode.rs
```diff
@@ -0,0 +1,36 @@
...
+        use library::Type::*;
+        if !env.config.generate_safety_asserts {
+            return None;
+        }
+        if params.len() > 0 && params[0].instance_parameter {
+            return None;
+        }
+        for par in params {
+            match *env.library.type_(par.typ) {
+                Class(..) => return Skip,
```
gkoz: Also `Interface`.

This gets into somewhat difficult area though: `Option<T>` or `&[T]` might allow calling the function without a previously created object (e.g. `Layout::new`).
It might be better to merge with just the `instance_parameter` case handled and then add improvements if any.

EPashkin:
Removed assert_skip for now, as it possible depend on nullables and #176

gkoz:
Seems to pass the `check_init_asserts` checks so this should be good enough for a while. Thanks!
@homu r+

Issue #176, opened 2016-01-06
Parameter `nullable = true` override doesn't seem to work


gkoz:
```toml
[[object]]
name = "Gtk.Entry"
status = "generate"
    [[object.function]]
    name = "set_placeholder_text"
        [[object.function.parameter]]
        name = "text"
        nullable = true
```
Expected outcome:
```rust
fn set_placeholder_text(&self, text: Option<&str>);
```

EPashkin:
Tried `[[object.function.parameter]]` ?

EPashkin:
And yes it not realized.

gkoz:
Oops, updating the issue with correct syntax.

gkoz:
Hmm, I thought that the opposite worked but looks like I confused it with `const = true`. Please let me know if you're looking into this so we don't both do it ;)

EPashkin:
Currently nullable for return work only for `nullable = false` as it default value `true` https://github.com/gkoz/gir/blob/master/src/analysis/return_value.rs#L28
For parameters we may be needed both cases as for enabling and for disabling so need add `nullable: Option<bool>` to https://github.com/gkoz/gir/blob/master/src/config/functions.rs#L40-L46

Seems I don't have time do it today.

gkoz:
OKay, thanks for the details.

PR #177, opened 2016-01-06
Add `gunichar` support with scalar conversion

gkoz:


EPashkin:
`gtk_entry_set_invisible_char` now work, but IMHO needed nullable override from #176.
And `gtk_entry_get_invisible_char` also needed unrealized return nullable override.

gkoz:
I hadn't thought of applying nullability to ints. This could work. You're right that it's more consistent to use `Option` here, but then it's far easier to just implement these two functions manually (because `char` can't just be compared to `0`).

We might want to work out a generator solution for harmonization of ints (as in `EntryBuffer`, `Notebook` and `invisible_char`) but feasibility is unclear.

gkoz:
Okay, this isn't that difficult, it's the conversion functions that take care of comparisons etc. I'll try the generator way.

gkoz:
This should settle the `get/set_invisible_char` case.

src/config/functions.rs
```diff
@@ -257,27 +256,34 @@ name = "par1"
...
 const = false
+nullable = false
 [[parameter]]
 name = "par3"
 const = true
+nullable = true
 [[parameter]]
 pattern = "par4"
 const = true
+nullable = true
```
EPashkin: Can you change this to None?

gkoz: I don't follow. `None` corresponds to absence of this line. Remove it?

EPashkin: Oh, yes, remove.

gkoz:
Updated

EPashkin:
Tnx

gkoz:
@homu r+

PR #179, opened 2016-01-07
Add skip assertion if use objects or interfaces

### LINKED ISSUE #174: Generate safety assertions

EPashkin:
Closes #174
Cover both `Option<Widget>` and `&[]` (GLib.List) exclusions cases.

### DIFF
--- a/src/analysis/safety_assertion_mode.rs
+++ b/src/analysis/safety_assertion_mode.rs
@@ -1,5 +1,6 @@
 use analysis::parameter::Parameter;
 use env::Env;
+use library;
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub enum SafetyAssertionMode {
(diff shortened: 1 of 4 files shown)


gkoz:
Another thing I forgot is only objects from the current library are eligible, e.g. `PixbufAnimation` constructors don't have the assertions: 
```patch
     pub fn new_from_animation<T: Upcast<gdk_pixbuf::PixbufAnimation>>(animation: &T) -> Image {
-        assert_initialized_main_thread!();
+        skip_assert_initialized!();
         unsafe {
```

EPashkin:
You right.
With limiting by types from current crate checks in `gtk_image_new_from_animation`, `gtk_status_icon_new_from_pixbuf`, `gtk_style_context_reset_widgets` will be not skipped.

gkoz:
Looks good!

gkoz:
@homu r+

PR #184, opened 2016-01-08
Add out_parameters Mode::Combined for functions with nontrivial return types

gkoz:


EPashkin:
Seems it cover #144 with minor difference. Thanks.

gkoz:
I wasn't completely sure if `ret` should be the first or last. It probably makes more sense to put it first like suggested in that issue.

gkoz:
And the return value is in the first position now.

EPashkin:
Thanks.

gkoz:
@homu r+

PR #187, opened 2016-01-10
Throwable functions

### LINKED ISSUE #130: Return Result for errors

EPashkin:
Closes #130

WIP

### DIFF
--- a/Gir_Gtk.toml
+++ b/Gir_Gtk.toml
@@ -22,15 +22,18 @@ generate = [
     "Gtk.AspectFrame",
     "Gtk.Bin",
     "Gtk.Box",
+    "Gtk.Builder",
     "Gtk.Button",
     "Gtk.ButtonBox",
     "Gtk.Calendar",
     "Gtk.CheckButton",
     "Gtk.ColorButton",
     "Gtk.ColorChooserWidget",
+    "Gtk.CssProvider",
     "Gtk.DrawingArea",
     "Gtk.EventBox",
     "Gtk.Expander",
+    "Gtk.FileChooser",
     "Gtk.FileChooserWidget",
     "Gtk.Fixed",
     "Gtk.FlowBox",
@@ -41,6 +44,7 @@ generate = [
     "Gtk.Frame",
     "Gtk.Grid",
     "Gtk.HeaderBar",
+    "Gtk.IconInfo",
     "Gtk.Image",
     "Gtk.InfoBar",
     "Gtk.Label",
@@ -66,6 +70,7 @@ generate = [
     "Gtk.RadioButton",
     "Gtk.Range",
     "Gtk.RecentChooserWidget",
+    "Gtk.RecentManager",
     "Gtk.Revealer",
     "Gtk.Scale",
     "Gtk.ScaleButton",
@@ -97,6 +102,7 @@ generate = [
 ]
 
 manual = [
+    "GLib.Error",
     "Gdk.Screen",
     "Gdk.Window",
     "GdkPixbuf.Pixbuf",
@@ -105,9 +111,12 @@ manual = [
     "Gtk.Buildable",
     "Gtk.Container",
     "Gtk.Dialog",
+    "Gtk.RecentData",
+    "Gtk.RecentInfo",
     "Gtk.Widget",
     "Gtk.Window",
     "Gdk.Rectangle",
+#    "Gdk.RGBA",
     "cairo.RectangleInt",
 ]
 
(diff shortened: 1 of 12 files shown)


EPashkin:
Working but needed some optimization on error detection

EPashkin:
.. now it detects by comparing it name with "error"

gkoz:
In most cases it's a [synthetical parameter](https://github.com/gkoz/gir/blob/master/src/parser.rs#L672-L683) though.

EPashkin:
IMHO In all cases.

gkoz:
In that case it seems straightforward to add a flag to `Parameter`.

EPashkin:
You right, something like 'is_error: bool'

EPashkin:
Other minor problem is double '(' in `Ok((ret))` but it accepted by compiler normally.
```
    pub fn new_from_file(file_name: &str) -> Result<PageSetup, glib::Error> {
        unsafe {
            let ret = from_glib_full(ffi::gtk_page_setup_new_from_file(file_name.to_glib_none().0, &mut error));
            if error.is_null() { Ok((ret)) } else { Err(from_glib_full(error)) }
        }
    }
```

EPashkin:
Added flag in parameters

gkoz:
Since `Error` is the universal error type, I'm inclined to reexport it in `gtk` making the return type more like `Result<PageSetup, Error>` (with appropriate `use Error;`).

EPashkin:
IMHO better add `glib::Result<T>=Result<T, glib::Error>`

EPashkin:
Not sure about need of reexporting `Error` or `Result`

gkoz:
The `Result` alias make sense but I'm wary of cases where we might want to return `Result` from functions that don't throw. The error type there would be other than this `Error`, probably just `()`.

EPashkin:
Is one reason why I prefer `glib::Result` in generated code.

gkoz:
My considerations are these:
 * Popular types not particularly tied to `glib`, like `Error` and `Result<T> = Result<T, Error>` should be available in other libraries so the user wouldn't have to `extern crate glib` just to mention them.
 * When reexporting this `glib::Result` I'd hate to get into a situation where some module might need both `Result<T, Error>` and  `Result<T, ()>` (or perhaps `Result<T, SomeObscureError>`). The only non-`Error` `Result` so far is `gtk::init`. This might not turn out to be a practical problem.
 * If I'm not mistaken, adding the `Result<T, Error>` alias later won't be a breaking change because the type remains the same.

EPashkin:
3. IMHO also

gkoz:
For now let's change `glib::Error` to `Error` if it's not much trouble.

It's nice to see half of "unsupported outs" warnings gone.

gkoz:
There's a slight problem with return (and outparam?) values conversions:
```rust
let ret = from_glib_full(ffi::gtk_page_setup_new_from_file(file_name.to_glib_none().0, &mut error));
if error.is_null() { Ok((ret)) } else { Err(from_glib_full(error)) }
```
We can't expect `ret` to be valid if there was an error. So this needs to be
```rust
let ret = ffi::gtk_page_setup_new_from_file(file_name.to_glib_none().0, &mut error);
if error.is_null() { Ok(from_glib_full(ret)) } else { Err(from_glib_full(error)) }
```

EPashkin:
Fixed `Ok((ret))`.
Next be `Error` and then `Ok(from_glib_full(ret))`

EPashkin:
Done `Error`

EPashkin:
Done `Ok(from_glib_full(ret))`

gkoz:
Perfect, thanks!
@homu r+

PR #189, opened 2016-01-13
 	Remove unneeded generated end extra line and add extra empty lines between trait function declarations

### LINKED ISSUE #188: Extra empty line at the end of struct
When generating a `struct`, it ends like this:

```Rust
struct Foo {
    fn a() {}

}
```

I'm working on fixing this.

GuillaumeGomez:
Fixes #188.

### DIFF
--- a/src/codegen/function.rs
+++ b/src/codegen/function.rs
@@ -14,10 +14,10 @@ use writer::ToCode;
 
 pub fn generate(w: &mut Write, env: &Env, analysis: &analysis::functions::Info,
     in_trait: bool, only_declaration: bool, indent: usize) -> Result<()> {
-
     let mut commented = false;
     let mut comment_prefix = "";
     let mut pub_prefix = if in_trait { "" } else { "pub " };
+    try!(writeln!(w, ""));
     match analysis.visibility {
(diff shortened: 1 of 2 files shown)


src/codegen/function.rs
```diff
@@ -14,10 +14,10 @@ use writer::ToCode;
 
 pub fn generate(w: &mut Write, env: &Env, analysis: &analysis::functions::Info,
     in_trait: bool, only_declaration: bool, indent: usize) -> Result<()> {
-
     let mut commented = false;
     let mut comment_prefix = "";
     let mut pub_prefix = if in_trait { "" } else { "pub " };
+    try!(writeln!(w, ""));
```
gkoz: Without an `if` this has an additional effect in trait definitions. Is that inteded? It doesn't match the commit/PR note.

GuillaumeGomez:
Commit message has been updated.

gkoz:
Okay, considering we'll be putting the docs there, adding blank lines between trait function items is a good thing.
Thanks!
@homu r+

gkoz:
Something's wrong with homu I guess.

PR #193, opened 2016-01-18
`to_glib_container` returns a `Stash`

gkoz:
See https://github.com/gtk-rs/glib/pull/92


EPashkin:
Thanks.
Still no use `to_glib_container` in `gtk` or internal `Gir_Gtk.toml` so can't check if it works :smile:

gkoz:
I've uncovered a use when generating `Window::set_default_icon_list`.

PR #194, opened 2016-01-18
Add analysis::InfoBase with common fields

EPashkin:


gkoz:
Makes sense. I'm a little apprehensive of implementing inheritance with `Deref` but it seems harmless for now.

On the unification note, when looking at error enums I've noticed that not only records and objects but even enums and likely bitfields can have associated functions, that are currently ignored.

Issue #195, opened 2016-01-19
Use idiomatic types for `filename` parameters


gkoz:
Upgrade the "upcasts" feature to support taking `Fundamental::Filename` as
```rust
fn select_filename<T: AsRef<Path>>(&self, filename: T) -> bool
```

EPashkin:
We can change https://github.com/gtk-rs/gir/blob/master/src/analysis/rust_type.rs#L52
to `std::path::Path` and `std::path::PathBuf`,
then change result of https://github.com/gtk-rs/gir/blob/master/src/analysis/needed_upcast.rs#L3 to enum (for AsRef).
But I have no idea what do with long double-dotted names and imports.

gkoz:
`AsRef<std::path::Path>` might be okay at first.

One more difficulty is the need to call to `as_ref`, e.g.

```rust
filename.as_ref().to_glib_none().0
```

gkoz:
A similar issue is `Gio.File` ([used here](https://developer.gnome.org/gtk3/stable/GtkAppChooserDialog.html#gtk-app-chooser-dialog-new) for instance) which is similar enough to `PathBuf` that it could be nice to have

```rust
pub fn new<T: IsA<Window>, U: Into<gio::File>>(parent: Option<&T>, flags: DialogFlags, file: U) -> AppChooserDialog
```

This conflicts with taking objects by reference though...

gkoz:
Implemented in #196

PR #196, opened 2016-01-20
Trait bounds

EPashkin:
Part of #195 
WIP


EPashkin:
All done.
Depends on merging https://github.com/gkoz/gtk-rs-glib/tree/path.
Don't like `.as_ref()` adding method but don't found better.

gkoz:
Thanks!

PR #198, opened 2016-01-28
Add doc generation

GuillaumeGomez:
cc @gkoz

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,14 +3,15 @@ name = "gir"
 version = "0.0.1"
 dependencies = [
  "case 0.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "docopt 0.6.73 (registry+https://github.com/rust-lang/crates.io-index)",
- "env_logger 0.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "git2 0.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "lazy_static 0.1.14 (registry+https://github.com/rust-lang/crates.io-index)",
- "log 0.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "regex 0.1.41 (registry+https://github.com/rust-lang/crates.io-index)",
- "toml 0.1.22 (registry+https://github.com/rust-lang/crates.io-index)",
- "xml-rs 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "docopt 0.6.78 (registry+https://github.com/rust-lang/crates.io-index)",
+ "env_logger 0.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "git2 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "lazy_static 0.1.15 (registry+https://github.com/rust-lang/crates.io-index)",
+ "log 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "regex 0.1.48 (registry+https://github.com/rust-lang/crates.io-index)",
+ "stripper_interface 0.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "toml 0.1.25 (registry+https://github.com/rust-lang/crates.io-index)",
+ "xml-rs 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -18,16 +19,16 @@ name = "advapi32-sys"
 version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "winapi 0.2.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winapi 0.2.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "winapi-build 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
 name = "aho-corasick"
-version = "0.3.2"
+version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "memchr 0.1.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -37,7 +38,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "bitflags"
-version = "0.3.2"
+version = "0.3.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
@@ -47,72 +48,80 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "cmake"
-version = "0.1.5"
+version = "0.1.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "gcc 0.3.17 (registry+https://github.com/rust-lang/crates.io-index)",
+ "gcc 0.3.21 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
 name = "docopt"
-version = "0.6.73"
+version = "0.6.78"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "regex 0.1.41 (registry+https://github.com/rust-lang/crates.io-index)",
+ "regex 0.1.48 (registry+https://github.com/rust-lang/crates.io-index)",
  "rustc-serialize 0.3.16 (registry+https://github.com/rust-lang/crates.io-index)",
  "strsim 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
 name = "env_logger"
-version = "0.3.1"
+version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "log 0.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "regex 0.1.41 (registry+https://github.com/rust-lang/crates.io-index)",
+ "log 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "regex 0.1.48 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
 name = "gcc"
-version = "0.3.17"
+version = "0.3.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "advapi32-sys 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "winapi 0.2.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winapi 0.2.5 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "gdi32-sys"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "winapi 0.2.5 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
 name = "git2"
-version = "0.3.1"
+version = "0.3.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "bitflags 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
- "libgit2-sys 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
- "url 0.2.37 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libgit2-sys 0.3.9 (registry+https://github.com/rust-lang/crates.io-index)",
+ "url 0.2.38 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
 name = "lazy_static"
-version = "0.1.14"
+version = "0.1.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "libc"
-version = "0.1.10"
+version = "0.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "libgit2-sys"
-version = "0.3.4"
+version = "0.3.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "cmake 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
- "libssh2-sys 0.1.30 (registry+https://github.com/rust-lang/crates.io-index)",
- "libz-sys 0.1.8 (registry+https://github.com/rust-lang/crates.io-index)",
- "openssl-sys 0.6.5 (registry+https://github.com/rust-lang/crates.io-index)",
- "pkg-config 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cmake 0.1.12 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libssh2-sys 0.1.34 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libz-sys 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "openssl-sys 0.7.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "pkg-config 0.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -125,32 +134,34 @@ dependencies = [
 
 [[package]]
 name = "libssh2-sys"
-version = "0.1.30"
+version = "0.1.34"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "cmake 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
- "libz-sys 0.1.8 (registry+https://github.com/rust-lang/crates.io-index)",
- "openssl-sys 0.6.5 (registry+https://github.com/rust-lang/crates.io-index)",
- "pkg-config 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cmake 0.1.12 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libz-sys 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "openssl-sys 0.7.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "pkg-config 0.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winapi 0.2.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "ws2_32-sys 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
 name = "libz-sys"
-version = "0.1.8"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "gcc 0.3.17 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
- "pkg-config 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "gcc 0.3.21 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "pkg-config 0.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
 name = "log"
-version = "0.3.2"
+version = "0.3.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "libc 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.5 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -160,26 +171,27 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "memchr"
-version = "0.1.6"
+version = "0.1.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "libc 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.5 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
 name = "openssl-sys"
-version = "0.6.5"
+version = "0.7.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "gcc 0.3.17 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
+ "gdi32-sys 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "libressl-pnacl-sys 2.1.6 (registry+https://github.com/rust-lang/crates.io-index)",
- "pkg-config 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "pkg-config 0.3.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "user32-sys 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
 name = "pkg-config"
-version = "0.3.5"
+version = "0.3.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
@@ -192,21 +204,21 @@ dependencies = [
 
 [[package]]
 name = "rand"
-version = "0.3.11"
+version = "0.3.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "advapi32-sys 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "libc 0.1.10 (registry+https://github.com/rust-lang/crates.io-index)",
- "winapi 0.2.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winapi 0.2.5 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
 name = "regex"
-version = "0.1.41"
+version = "0.1.48"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "aho-corasick 0.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "memchr 0.1.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "aho-corasick 0.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr 0.1.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex-syntax 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -221,6 +233,11 @@ version = "0.3.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "stripper_interface"
+version = "0.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "strsim"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -230,12 +247,12 @@ name = "tempdir"
 version = "0.3.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "rand 0.3.11 (registry+https://github.com/rust-lang/crates.io-index)",
+ "rand 0.3.13 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
 name = "toml"
-version = "0.1.22"
+version = "0.1.25"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "rustc-serialize 0.3.16 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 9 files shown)


src/parser.rs
```diff
@@ -884,6 +942,15 @@ impl Get for Attributes {
...
 
+fn read_text(parser: &mut Reader) -> Result<Option<String>, Error> {
+    let ret = match try!(parser.next()) {
+        Characters(text) => Ok(Some(text)),
+        _ => Ok(None),
+    };
+    try!(ignore_element(parser));
+    ret
+}
+
```
gkoz: Like I said previously I would like this function to resemble other `read_*` functions i.e. have a loop which handles `EndElement` explicitly, concatenates multiple `Characters` events if necessary and returns an error on `StartElement`.

src/codegen/doc/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+
+pub fn generate(env: &Env) {
+    let path =  env.config.target_path.join("doc.cmts");
+    println!("Generating documentation {:?}", path);
+    save_to_file(&path, env.config.make_backup,
+        |w| generate_doc(w, &env.library));
+}
+
+fn generate_doc(w: &mut Write, lib: &Library) -> Result<()> {
+    for i in 0..lib.namespaces.len() {
```
gkoz: Right now we only work on the "main" namespace in one generator run so just `let namespace = lib.namespaces.main();` will do.

src/codegen/doc/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+        let mut output = String::new();
+        let namespace = &lib.namespaces[i];
+
+        if let Some(ref namespace_doc) = namespace.doc {
+            for line in namespace_doc.split("\n") {
+                output.push_str(&format!("{}{}\n", FILE_COMMENT, line));
+            }
+        }
+        try!(create_sub_docs(&mut output, &namespace.constants, None));
+        for i in 0..namespace.types.len() {
```
gkoz: Don't use `for i in 0..len` in Rust unless there's a good reason. This one can be handled with

```rust
for ty in ns.types.iter().filter_map(|t| t.as_ref())
```

GuillaumeGomez: Oh the `.as_ref`, I totally forgot it... And yes, it's because of the moved errors that came back to this syntax.

src/codegen/doc/mod.rs
GuillaumeGomez: Ok, I'll change the code is consequence. I guess that each struct is in its own file and there isn't a namespace for one file (like I thought before).

src/codegen/doc/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+        w.push_str(&format!("{}{}\n{}\n",
+                            MOD_COMMENT,
+                            ty,
+                            format!("{}/// {}\n",
+                                    tabs,
+                                    class_doc.split("\n").collect::<Vec<&str>>().join(&format!("\n{}/// ", tabs)))));
+    }
+    for child_id in class.children.iter() {
+        if let &LType::Class(ref child) = lib.type_(child_id.clone()) {
+            try!(create_class_doc(w, &child, Some(Box::new(ty.clone())), lib));
```
gkoz: This feels like it will generate a lot of duplicate entries. Won't `handle_type` call `create_class_doc` on all children anyway?

No `create_fn_doc`?

EPashkin:
Can you rebase PR to master? First 2 commits was already merged.

src/codegen/doc/mod.rs
```diff
@@ -0,0 +1,183 @@
...
+    if let Some(ref class_doc) = class.doc {
+        w.push_str(&format!("{}{}\n{}\n",
+                            MOD_COMMENT,
+                            ty,
+                            format!("{}/// {}\n",
+                                    tabs,
+                                    class_doc.split("\n").collect::<Vec<&str>>().join(&format!("\n{}/// ", tabs)))));
+    }
+    for child_id in class.children.iter() {
+        if let &LType::Class(ref child) = lib.type_(child_id.clone()) {
```
gkoz: `TypeId` is `Copy`. It's confusing to see it `clone`d when you can `for &child_id in class.children.iter()` instead.

src/codegen/doc/mod.rs
GuillaumeGomez: It's wrong anyway, this code is based on the principle that a class can be a child of another one, which isn't the case here. `create_class_doc` function will be changed to take this in account.

src/codegen/doc/mod.rs
gkoz: This `Namespace` thing corresponds to a single `gir` file (a single `<namespace>` element). So `gtk` is one "namespace". When we run the generator on `gtk` it's unnecessary to generate stuff for `gdk` or `glib` etc.

GuillaumeGomez:
@EPashkin: Done. Also, this PR isn't intended to be merged as is. It's to make the code comparison easier in order to make @gkoz task to help me easier as well. :)

EPashkin:
Thanks.
While PR still in progress, I believe that it will be merged, so better fix sooner.

GuillaumeGomez:
@EPashkin: Already done anyway. :p

src/library.rs
```diff
@@ -192,20 +215,25 @@ pub struct Alias {
     pub c_identifier: String,
     pub typ: TypeId,
     pub target_c_type: String,
+    pub doc: Option<String>,
```
gkoz: Would you mind doing the same with `doc-deprecated`? It is relevant for doc generation and parser-wise is exactly like `doc`.

GuillaumeGomez: I'm still thinking about how I could generate it (I mean, I need to check if I can use `#[deprecated]`, otherwise I'll just make something like: "!!!!!!!!!!!! DEPRECATED STUFF !!!!!!!!!!!!!!!!!!!!!!!!".

gkoz: Appending this to docs with a `# Deprecated` subheading should be workable. Not in the parser though. `#[deprecated]` is better left for some other day. ;)

GuillaumeGomez: For now I have set deprecated doc to "========> DEPRECATED <=========". Should be enough. :p

src/codegen/doc/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+    ty.parent = parent;
+
+    try!(writeln!(w, "{}src/auto/{}.rs", FILE, class.name.to_lowercase()));
+    let found = if let Some(ref class_doc) = class.doc {
+        try!(writeln!(w, "{}{}\n{}",
+                      MOD_COMMENT,
+                      ty,
+                      format!("{}/// {}",
+                              tabs,
+                              class_doc.split("\n").collect::<Vec<&str>>().join(&format!("\n{}/// ", tabs)))));
```
gkoz: There doesn't seem to be a reason to `collect` rather than `write!` each line in a `for` loop. Additionally the `found` dance will become unnecessary.

GuillaumeGomez: I avoid calling `write!` as much as possible in these cases. The script is already slow engouh.

gkoz: I don't get it. Writes are buffered.

GuillaumeGomez: They do? Erf. Bad habit from C...

gkoz: https://github.com/gtk-rs/gir/blob/master/src/file_saver.rs#L18

This reminds me: notice the `Untabber` filter. You can avoid messing with tab size and `"    "` because a '\t' will get expanded properly.

GuillaumeGomez: Hum... I'm gonna try.

GuillaumeGomez: I just thought that a tab won't affect the fact that:
1. it will have to be translated internally into spaces, which I do directly, so no perfomance gain
2. I still have to generate as many tabs as needed (I think one tab = 4 spaces)

There isn't much to win in here.

gkoz: That's just for convenience: sometimes it's nice to not have to count spaces.

See also https://github.com/gtk-rs/gir/blob/master/src/writer/mod.rs#L6, you can import those consts instead of redefining them.

GuillaumeGomez: This, however, I'll use it!

src/codegen/doc/mod.rs
```diff
@@ -0,0 +1,236 @@
...
+    for child_id in class.children.iter() {
+        if let &LType::Class(ref child) = lib.type_(*child_id) {
+            try!(create_class_doc(w, &child, None, lib));
+        }
+    }
+    Ok(())
+}
+
+fn create_trait_doc(w: &mut Write, trait_: &library::Interface, parent: Option<Box<TypeStruct>>) -> Result<()> {
+    let tabs : String = iter::repeat(" ").take(compute_ident(&parent) * TAB_SIZE).collect::<Vec<&str>>().concat();
```
EPashkin: Don't `writer::primitives::tabs(num)` doing same?

src/library.rs
```diff
@@ -192,20 +220,28 @@ pub struct Alias {
...
+    pub doc_deprecated: Option<String>,
 }
 
+
```
EPashkin: Can you remove this?

src/library.rs
```diff
@@ -192,20 +220,28 @@ pub struct Alias {
...
     pub name: String,
     pub c_identifier: String,
     pub typ: TypeId,
     pub c_type: String,
     pub value: String,
+    pub doc: Option<String>,
+    pub doc_deprecated: Option<String>,
 }
 
+
```
EPashkin: And this.

src/config/work_mode.rs
```diff
@@ -2,8 +2,9 @@ use std::str::FromStr;
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub enum WorkMode {
-    Normal,     //generate widgets etc.
-    Sys,        //generate -sys with ffi
+    Normal,     // generate widgets etc.
+    Sys,        // generate -sys with ffi
+    Doc,        // generate documentation file
```
EPashkin: Can you add `doc` to https://github.com/gtk-rs/gir/blob/master/src/config/config.rs#L22 ?

src/codegen/doc/mod.rs
GuillaumeGomez: @gkoz: Said the same things a few comments above. And yes, I'll use it.

gkoz: I'd say this is an improvement on what I pointed at because I was just talking about constants and this functions gives you the whole string at once.

GuillaumeGomez:
I think I have adressed all comments. Anything else?

Cargo.lock
```diff
@@ -3,14 +3,15 @@ name = "gir"
...
- "xml-rs 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "docopt 0.6.78 (registry+https://github.com/rust-lang/crates.io-index)",
+ "env_logger 0.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "git2 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "lazy_static 0.1.15 (registry+https://github.com/rust-lang/crates.io-index)",
+ "log 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "regex 0.1.48 (registry+https://github.com/rust-lang/crates.io-index)",
+ "stripper_interface 0.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "toml 0.1.25 (registry+https://github.com/rust-lang/crates.io-index)",
+ "xml-rs 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
```
gkoz: Can the `Cargo.toml`/`Cargo.lock` update be put in a separate commit?

src/codegen/doc/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+}
+
+fn create_class_doc(w: &mut Write, class: &library::Class, parent: Option<Box<TypeStruct>>, lib: &Library) -> Result<()> {
+    let tabs : String = primitives::tabs(compute_ident(&parent));
+    let mut ty = class.convert();
+    ty.parent = parent;
+
+    try!(writeln!(w, "{}src/auto/{}.rs", FILE, class.name.to_lowercase()));
+    let found = if let Some(ref class_doc) = class.doc {
+        try!(writeln!(w, "{}{}", MOD_COMMENT, ty));
```
gkoz: The `found` business is unnecessary. Essentially you want to execute this line if `class_doc.is_some() && doc_deprecated.is_some()`. This can be expressed much more concisely.

src/codegen/doc/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+        if found == false {
+            try!(writeln!(w, "{}{}",
+                          MOD_COMMENT,
+                          ty));
+        }
+        try!(writeln!(w, "{}{}\n{}/// {}",
+                      tabs,
+                      DEPRECATED,
+                      tabs,
+                      class_doc.split("\n").collect::<Vec<&str>>().join(&format!("\n{}/// ", tabs))));
```
gkoz: Redundant `collect/join` has been discussed already I believe.

src/codegen/doc/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+    }
+    for i in 0..class.functions.len() {
```
gkoz: No C for loops please.

src/library.rs
```diff
@@ -6,6 +6,34 @@ use nameutil::split_namespace_name;
 use traits::*;
 use version::Version;
 
+use stripper_interface as stripper;
+use stripper_interface::Type as SType;
+use stripper_interface::TypeStruct;
+
+pub trait ToStripperType {
```
gkoz: This trait and its implementation for lack of better place should go to the `codegen::doc` module.

src/parser.rs
```diff
@@ -884,6 +946,22 @@ impl Get for Attributes {
     }
 }
 
+fn read_text(parser: &mut Reader) -> Result<Option<String>, Error> {
+    match try!(parser.next()) {
```
gkoz: I realize you had been too tired when fixing this place. ;) Note how every `read_*` function contains a loop not a single read.

src/parser.rs
```diff
@@ -884,6 +946,22 @@ impl Get for Attributes {
...
+        Characters(mut text) => {
+            Ok(if let Some(ret) = try!(read_text(parser)) {
+                text.push_str(&ret);
+                Some(text)
+            } else {
+                Some(text)
+            })
+        }
+        EndElement { .. } => Ok(None),
+        StartElement { .. } => Err(mk_error!("Unexpected StartElement", parser)),
```
gkoz: This would preferably be consistent with other "unexpected element" errors (e.g. featuring the element's name).

src/parser.rs
```diff
@@ -884,6 +946,22 @@ impl Get for Attributes {
...
+        _ => read_text(parser),
```
gkoz: This seems inconsistent with other `read_*` functions too.

src/codegen/doc/mod.rs
```diff
@@ -0,0 +1,232 @@
...
+use analysis::namespaces::MAIN;
+use env::Env;
+use file_saver::save_to_file;
+use library::{self, Library, ToStripperType};
+use library::Type as LType;
+use writer::primitives;
+
+use stripper_interface::{FILE, /*FILE_COMMENT, */MOD_COMMENT, TypeStruct};
+
+const DEPRECATED : &'static str = "/// ====>> DEPRECATED <<====";
```
gkoz: I feel this will look more rusty as just `"/// # Deprecated"`. ;)

gkoz: A '#' heading might need to be surrounded by newlines though.

GuillaumeGomez: Yes, but so much easier to see it like this. :D

EPashkin:
Not sure if it error but now in generated docs some comments uncommented:
```
=| trait Buildable
-/// GtkBuildable allows objects to extend and customize their deserialization
+GtkBuildable allows objects to extend and customize their deserialization
```
same with `=| struct Widget` and other tabless comments

GuillaumeGomez:
@EPashkin: No, it's abnormal.

src/codegen/doc/mod.rs
gkoz: And by `&&` I meant `||`... :blush:

GuillaumeGomez:
Updated.

EPashkin:
Now seems no "uncommented" comments.
Still not sure why output named `doc.cmts` while `rustdoc-stripper` accept only `comments.cmts`, but IMHO it not problem even if `rustdoc-stripper` can't regenerate comments now as it can be fixed later.

GuillaumeGomez:
@EPashkin: The name of the file doesn't really matter (you can set the input file inside the rustdoc-stripper options if you want). I thought it might be better for understanding.

src/codegen/doc/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+impl_to_stripper_type!(Bitfield, Type);
+impl_to_stripper_type!(Record, Type);
+impl_to_stripper_type!(Field, Variant);
+impl_to_stripper_type!(Union, Struct);
+impl_to_stripper_type!(Function, Fn);
+impl_to_stripper_type!(Interface, Trait);
+impl_to_stripper_type!(Class, Struct);
+impl_to_stripper_type!(Namespace, Mod);
+
+const DEPRECATED : &'static str = "/// ====>> DEPRECATED <<====";
```
gkoz: I don't know why the previous comment is gone, it will be more robust to add a `write_deprecated` function, which would emit
```
///
/// # Deprecated
///
```
This can't be done with just a const because of indents.

src/codegen/doc/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    for function in trait_.functions.iter() {
+        try!(create_fn_doc(w, &function, Some(Box::new(ty.clone()))));
+    }
+    Ok(())
+}
+
+fn create_enum_doc(w: &mut Write, enum_: &Enumeration, parent: Option<Box<TypeStruct>>) -> Result<()> {
+    let ident = compute_ident(&parent);
```
gkoz: Looks like `s/ident/indent/g` throughout the module is in order.

src/codegen/doc/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+        }
+    }
+    Ok(())
+}
+
+fn create_fn_doc(w: &mut Write, fn_: &Function, parent: Option<Box<TypeStruct>>) -> Result<()> {
+    let tabs : String = primitives::tabs(compute_ident(&parent) + 1);
+
+    let mut docs = match fn_.doc {
+        Some(ref d) => format!("{}/// {}\n", tabs, d.split("\n").collect::<Vec<&str>>().join(&format!("\n{}/// ", tabs))),
```
gkoz: I'd prefer this too was done straightforwardly without collecting into `docs` (not to mention that this function bypasses `write_lines`).
At worst you'll have empty `MOD_COMMENT` line for a couple of functions, which should not have a noticeable effect on stripper.

GuillaumeGomez:
Updated.

src/codegen/doc/mod.rs
```diff
@@ -0,0 +1,271 @@
...
+    for ty in namespace.types.iter().filter_map(|t| t.as_ref()) {
+        try!(handle_type(&mut w, &ty, None, lib));
+    }
+    Ok(())
+}
+
+fn handle_type(w: &mut Write, ty: &LType, parent: Option<Box<TypeStruct>>, lib: &Library) -> Result<()> {
+    match *ty {
+        LType::Alias(ref a) => {
+            try!(writeln!(w, "{}src/auto/{}.rs", FILE, a.name.to_lowercase()));
```
EPashkin: `to_lowercase()` not right function, better use `nameutil::module_name`

EPashkin: It convert `ActionBar` to `action_bar`

gkoz: This is going to be fun because `GLArea` and GLContext` need module name overrides from config. We could leave this off for another PR though.

EPashkin:
May be problem that files in `doc.cmts` placed in random order, but only if this file saved.
Still don't understand reason for randomness as `generate_doc` iterates `types: Vec<Option<Type>>`

src/codegen/doc/mod.rs
```diff
@@ -0,0 +1,271 @@
...
+            try!(write_lines(w, &m_doc, &tabs));
+        }
+    }
+    Ok(())
+}
+
+fn create_fn_doc(w: &mut Write, fn_: &Function, parent: Option<Box<TypeStruct>>) -> Result<()> {
+    let tabs : String = primitives::tabs(compute_indent(&parent) + 1);
+
+    if fn_.doc().is_none() && fn_.doc_deprecated().is_none() && fn_.ret.doc().is_none() && fn_.ret.doc_deprecated().is_none() {
```
gkoz: This doesn't fit on the github page it probably is longer than 100 chars.

src/codegen/doc/mod.rs
GuillaumeGomez: Yes, there are some cases where the file name doesn't correspond to what is inside.

src/codegen/doc/mod.rs
```diff
@@ -0,0 +1,271 @@
...
+        let mut leave = true;
+
+        for parameter in fn_.parameters.iter() {
```
gkoz: `parameters.iter().any()`?

src/codegen/doc/mod.rs
```diff
@@ -0,0 +1,271 @@
...
+        try!(write_deprecated(w, &tabs));
+        try!(write_lines(w, &class_doc, &tabs));
+    }
+
+    for function in class.functions.iter() {
+        try!(create_fn_doc(w, &function, Some(Box::new(ty.clone()))));
+    }
+    for child_id in class.children.iter() {
+        if let &LType::Class(ref child) = lib.type_(*child_id) {
+            try!(create_class_doc(w, &child, None, lib));
```
gkoz: Why is `create_class_doc` recursive? I've now confirmed my suspicion that it generates redundant copies of documentation.

gkoz:
It looks like you can't get away with not relying on analysis. But the good news is that a good bit of classes handling can be copied from https://github.com/gtk-rs/gir/blob/master/src/codegen/objects.rs#L10-L30

Here instead of taking the classes and interfaces from the library you iterate over "objects" from the config and take ones, that are not "ignore"d. This computes the correct module name and `Info::has_children` tells you if there is a trait or just a struct. Then functions, which have `instance_parameter = true` parameters go into traits.

EPashkin:
IMHO better don't modify iteration and check in `create_class_doc` with `env::type_status` and next do analysis

EPashkin:
Was wrong: `env::type_status` is not enough as also needed configures `module_name`, so
instead use something like `config.objects.get(name).unwrap_or(Default::default())`

GuillaumeGomez:
@EPashkin: How could this work? `name` can have conflicts, how does it differentiate between two indentical names?

gkoz:
That `name` would be a "full name" e.g. `Gtk.Button` but I don't think you'll have it handy, so you'll need to format it but then this approach becomes somewhat cumbersome.

GuillaumeGomez:
If I have to do any transformation on it, then I don't think it's a good idea.

EPashkin:
`name` is something like 'Gtk.GLArea' can be obtained with `type_id.full_name(&env.library)`.
You right about conflicting names, contained in `library::Class`, so you need pass to `handle_type` `namespace.name' or enumerate iterator and combine TypeId with MAIN.

EPashkin:
IMHO skipping ignored can be added later. I just commented @gkoz idea about iterating config.

EPashkin:
On second though iterating config is better that iterating library.

GuillaumeGomez:
Ok, should be it.

src/codegen/doc/mod.rs
```diff
@@ -0,0 +1,296 @@
...
+           x.doc().is_some() || x.doc_deprecated().is_some()
+       }) {
+        try!(writeln!(w, "{}///", tabs));
+        try!(writeln!(w, "{}/// # Parameters", tabs));
+        try!(writeln!(w, "{}///", tabs));
+    }
+    let mut wrote = false;
+    for parameter in fn_.parameters.iter() {
+        if let Some(ref parameter_doc) = parameter.doc() {
+            try!(writeln!(w, "{}/// ## {}:", tabs, parameter.name));
```
gkoz: This is where `instance_parameter` is relevant too. Most likely it should be skipped because in Rust it's passed implicitly.

gkoz: Anyway there are endless issues with parameters so refining that can be left for later.

EPashkin: Yes it really endless issues, at minimum we have out parameters to return value conversion.

GuillaumeGomez: I skip `instance_parameter` now as well.

EPashkin: Strange but on `Gir_Gtk.toml` only `# Parameters` found with `## error:` with out real parameter docs.

GuillaumeGomez: Yes, that's what I have too.

src/codegen/doc/mod.rs
```diff
@@ -0,0 +1,296 @@
...
+        |w| generate_doc(w, &env));
+}
+
+fn generate_doc(mut w: &mut Write, env: &Env) -> Result<()> {
+    //let namespace = lib.namespace(MAIN);
+
+    //try!(create_sub_docs(&mut output, &namespace.constants, None));
+    /*for ty in namespace.types.iter().filter_map(|t| t.as_ref()) {
+        try!(handle_type(&mut w, &ty, None));
+    }*/
```
gkoz: Iterating over `config.objects` is only suitable for objects but the rest (enums, records, etc.) still has to be done by iterating the library, so this loop is still needed, just remove `Class` and `Interface` from `handle_type`.

src/codegen/doc/mod.rs
```diff
@@ -0,0 +1,296 @@
...
+        LType::Function(ref f) => create_fn_doc(w, &f, parent),
+        LType::Interface(ref i) => create_class_doc(w, i, parent, is_trait),
+        LType::Class(ref c) => create_class_doc(w, c, parent, is_trait),
+        _ => Ok(()),
+    }
+}
+
+fn create_class_doc<T: FunctionTraitType + ToStripperType>(w: &mut Write, class: &T,
+                                                           parent: Option<Box<TypeStruct>>,
+                                                           is_trait: bool)
```
gkoz: `is_trait` is misleading because its meaning is `has_trait`

src/codegen/doc/mod.rs
EPashkin: All that have own file count as object in config.

GuillaumeGomez: So what do I do?

gkoz: I exepct the enums and bitfields to be missing from configs for a while. But that could be discussed I guess.
In the meantime we'd still like to document them even if it's right in the sys crates where they're now.

EPashkin: I suggest merge the PR and then finish the remaining issues.

src/parser.rs
```diff
@@ -884,6 +946,27 @@ impl Get for Attributes {
...
+        match try!(parser.next()) {
+            Characters(text) => {
+                ret_text = match ret_text {
+                    Some(t) => Some(format!("{}{}", t, text)),
+                    None => Some(text),
+                }
+            }
+            EndElement { .. } => break,
+            StartElement { name, .. } => return Err(mk_error!(&format!("Unexpected element: {}",
+                                                                       name),
```
gkoz: This is materially different from how other unexpected element errors are generated. `name` is not the same as `name.local_name`.

GuillaumeGomez: I don't see anything like this :p

src/parser.rs
```diff
@@ -884,6 +946,27 @@ impl Get for Attributes {
...
+                ret_text = match ret_text {
+                    Some(t) => Some(format!("{}{}", t, text)),
+                    None => Some(text),
+                }
+            }
+            EndElement { .. } => break,
+            StartElement { name, .. } => return Err(mk_error!(&format!("Unexpected element: {}",
+                                                                       name.local_name),
+                                                              parser)),
+            _ => continue,
```
gkoz: This is not how other `read_*` functions handle the catch-all case.

GuillaumeGomez: Sorry, fixed.

gkoz:
Yeah looks good. I think it's time to sqash most in-progress commits (not the `Cargo.*` one).

GuillaumeGomez:
Squashed enough?

gkoz:
Yeah.

My testing doesn't show any parameter documentation but I don't see an obvious problem in the code.

GuillaumeGomez:
That's rassuring and scary at the same time. I wasn't able to determine what I have missed.

EPashkin:
@GuillaumeGomez parameters docs was just not parsed and error get functions docs.
Fix https://github.com/EPashkin/gir/commit/7ef8557c3bad696b8c48bb8af4a270245128860c
IMHO still need some empty line before `# returns`.
And output parameters also need some change but it better do later.

EPashkin:
IMHO also better delete `Parameter::doc_deprecated` and `impl_to_stripper_type!(Parameter, Variant);` or at minimum delete `if let Some(ref parameter_doc) = parameter.doc_deprecated() {` that I missed.

GuillaumeGomez:
@EPashkin: Why that?

EPashkin:
@GuillaumeGomez sorry don't understand. No 'doc_deprecated' in 'parameters' and 'return-value`

GuillaumeGomez:
@EPashkin: Oh ok! I update my code then.

GuillaumeGomez:
@EPashkin: I rebased my code on your fix but I still don't have parameters. I also removed doc deprecated generation for parameters and return value.

EPashkin:
3 changes you don't included https://gist.github.com/EPashkin/be4a162bb75dab540ef8

EPashkin:
May be rebase way was wrong, as my comment was on your previous.

GuillaumeGomez:
My PR is 3 commits above gir/master. So are you sure you did include it correctly?

EPashkin:
Change I gisted is between your last 3c3a81f3c3803e45d5c5c8afa35fcfd36e02afd9 and "good" code.
For some reason your rebase don't get needed changes, so try do it manually.

GuillaumeGomez:
@EPashkin: That's an interesting bug. I added the changes. Weird that the rebase failed...

EPashkin:
Now present unnamed parameter from varargs addition in line parser.rs:882.
IMHO better set it to `None` as no varargs in Rust.
```
=| struct InfoBar§fn new_with_buttons
    /// Creates a new #GtkInfoBar with buttons. Button text/response ID
    /// pairs should be listed, with a %NULL pointer ending the list.
    /// Button text can be either a stock ID such as %GTK_STOCK_OK, or
    /// some arbitrary text. A response ID can be any positive number,
    /// or one of the values in the #GtkResponseType enumeration. If the
    /// user clicks one of these dialog buttons, GtkInfoBar will emit
    /// the “response” signal with the corresponding response ID.
    ///
    /// # Parameters
    ///
    /// ## first_button_text:
    /// stock ID or text to go in first button, or %NULL
    /// ## :
    /// response ID for first button, then additional buttons, ending
    ///    with %NULL
```

GuillaumeGomez:
@EPashkin: If name is empty, I don't display this parameter. Does it seem fine to you?

EPashkin:
Now docs equal my previous.

GuillaumeGomez:
@EPashkin: What do you mean? I still have parameters.

EPashkin:
I mean is good, change successfully merged and result equal my "proposed"

GuillaumeGomez:
Perfect then. :)

src/codegen/doc/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+    if let Some(ref docs) = fn_.doc() {
+        try!(write_lines(w, &docs, &tabs));
+    };
+    if let Some(ref docs) = fn_.doc_deprecated() {
+        try!(write_header(w, &tabs, "Deprecated"));
+        try!(write_lines(w, &docs, &tabs));
+    };
+
+    if fn_.parameters.iter().any(|x| {
+           x.instance_parameter == false && x.name.len() > 0 && x.doc().is_some()
```
EPashkin: Not better use `String::is_empty` instead `len`

src/codegen/doc/mod.rs
```diff
@@ -0,0 +1,283 @@
...
+       }) {
+        try!(write_header(w, &tabs, "Parameters"));
+    }
+    for parameter in fn_.parameters.iter() {
+        if parameter.instance_parameter || parameter.name.len() < 1 {
```
EPashkin: Same

gkoz:
Thanks for contributing to making nice documentation possible!

GuillaumeGomez:
@gkoz: Thanks for your great help! @EPashkin: And thanks a lot to you as well! :)

PR #200, opened 2016-01-28
Generate functions associated with bitfields, enums and unions

gkoz:
Composing `GdkEvent*` wrappers is difficult without any `gdk_event_*` functions, which are associated with the `GdkEvent` union


EPashkin:
Thanks

Issue #203, opened 2016-02-01
Formatting documentation


GuillaumeGomez:
For now, here is the list of things we need to update inside the parsed documentation:

- [x] Transform code blocks "|[&lt;!-- language="C" --&gt; // some code ]|" into "```C // some code ```".
- [x] C-types should be replaced by their rust equivalent (example "GtkWindow" into "Window").
- [x] Replace "@%s" values by "`%s`" (example "@accel_mods" into "`accel_mods`").

The following one will be done later:
- [x] Replace objects' properties by their rust equivalent (enums certainly) (example "#GtkAboutDialog:license-type"). -> it has been partially done

gkoz:
I wonder if it's feasible to parse the docs as markdown. And then output rustddoc-friendly markdown again.

GuillaumeGomez:
We won't know until we try unfortunately. :-/

gkoz:
This page seems to describe the syntax best: https://wiki.gnome.org/Projects/GTK%2B/DocumentationSyntax/Markdown
There's also https://developer.gnome.org/gtk-doc-manual/stable/documenting_syntax.html.en

gkoz:
Alright parsing markdown is probably pointless. Rustdoc understands it. It's the "extensions" that need to be handled. So the list can just be updated with this new information about the extensions syntax.

gkoz:
Okay so the custom syntax list is:

>    Use `function()` to refer to functions or macros which take arguments.
    Use `@param` to refer to parameters. Also use this when referring to parameters of other functions, related to the one being described.
    Use `%constant` to refer to a constant, e.g. `%G_TRAVERSE_LEAFS`.
    Use `#symbol` to refer to other types of symbol, e.g. structs and enums and macros which don't take arguments.
    Use `#Object::signal` to refer to a GObject signal.
    Use `#Object:property` to refer to a GObject property.
    Use `#Struct.field` to refer to a field inside a structure and `#GObjectClass.foo_bar()` to refer to a vmethod.

PR #204, opened 2016-02-03
Add doc formatting

GuillaumeGomez:
I'll need to improve the regex part because for now, only "#Gtk" objects are replaced in their rust equivalent (the problem is with code blocks, I don't want them to change the C-name to rust-name).

#203

### DIFF
--- a/src/codegen/doc/format.rs
+++ b/src/codegen/doc/format.rs
@@ -1,59 +1,65 @@
 use regex::{Captures, Regex};
 
+const LANGUAGE_SEP_BEGIN : &'static str = "<!-- language=\"";
 const LANGUAGE_SEP_END : &'static str = "\" -->";
-const LANGUAGE_BLOCK_BEGIN : &'static str = "|[<!-- language=\"";
+const LANGUAGE_BLOCK_BEGIN : &'static str = "|[";
 const LANGUAGE_BLOCK_END : &'static str = "\n]|";
 
-pub fn reformat_doc(input: &str) -> String {
-    let out = code_blocks_transformation(input);
-    replace_c_types(&out)
+lazy_static! {
+    static ref REG : Regex = Regex::new(r"#?(G[dt]k)([\w]*)").unwrap();
+    static ref REG2 : Regex = Regex::new(r"@(\w*)").unwrap();
 }
 
-fn code_blocks_transformation(input: &str) -> String {
+pub fn reformat_doc(input: &str) -> String {
+    code_blocks_transformation(input)
+}
+
+fn try_split<'a>(src: &'a str, needle: &str) -> (&'a str, Option<&'a str>) {
+    match src.find(needle) {
+        Some(pos) => (&src[..pos], Some(&src[pos + needle.len()..])),
+        None => (src, None),
+    }
+}
+
+fn code_blocks_transformation(mut input: &str) -> String {
     let mut out = String::new();
 
-    for entry in input.split(LANGUAGE_BLOCK_BEGIN) {
-        if out.is_empty() {
-            out.push_str(entry);
-            continue;
-        }
-        out.push_str("```");
-        let entry = get_language(entry, &mut out);
-        out.push_str(&format!("{}", entry.replace(LANGUAGE_BLOCK_END, "\n```")))
+    loop {
+        input = match try_split(input, LANGUAGE_BLOCK_BEGIN) {
+            (before, Some(after)) => {
+                out.push_str(&replace_c_types(before));
+                if let (before, Some(after)) = try_split(get_language(after, &mut out),
+                                                         LANGUAGE_BLOCK_END) {
+                    out.push_str(before);
+                    out.push_str("\n```");
+                    after
+                } else {
+                    after
+                }
+            }
+            (before, None) => {
+                out.push_str(&replace_c_types(before));
+                return out
+            }
+        };
     }
-    out
 }
 
 fn get_language<'a>(entry: &'a str, out: &mut String) -> &'a str {
-    if let Some(pos) = entry.find(LANGUAGE_SEP_END) {
-        out.push_str(&entry[0..pos]);
-        &entry[(pos + LANGUAGE_SEP_END.len())..]
-    } else {
-        out.push_str("\n");
-        entry
+    if let Some(pos) = entry.find(LANGUAGE_SEP_BEGIN) {
+        let entry = &entry[pos + LANGUAGE_SEP_BEGIN.len()..];
+        if let Some(pos) = entry.find(LANGUAGE_SEP_END) {
+            out.push_str(&format!("```{}", &entry[0..pos]));
+            return &entry[(pos + LANGUAGE_SEP_END.len())..]
+        }
     }
+    out.push_str("```");
+    entry
 }
 
 fn replace_c_types(entry: &str) -> String {
-    if let Ok(reg) = Regex::new(r"Gtk[\w]*") {
-        let mut out = vec!();
-
-        for (num, part) in entry.split("```").into_iter().enumerate() {
-            out.push(
-                if num & 1 == 0 {
-                    reg.replace_all(part, |caps: &Captures| {
-                        match caps.at(0) {
-                            Some(s) => format!("`{}`", &s[4..]),
-                            None => String::new(),
-                        }
-                    })
-                } else {
-                    part.to_owned()
-                }
-            );
-        }
-        out.join("```")
-    } else {
-        entry.to_owned()
-    }
+    let out = &REG.replace_all(entry, |caps: &Captures| {
+        format!("`{}::{}`", &caps[1].to_lowercase(), &caps[2])
+    });
+    REG2.replace_all(out, "`$1`")
 }
--- a/src/codegen/doc/format.rs
+++ b/src/codegen/doc/format.rs
@@ -1,0 +1,59 @@
+use regex::{Captures, Regex};
+
+const LANGUAGE_SEP_END : &'static str = "\" -->";
+const LANGUAGE_BLOCK_BEGIN : &'static str = "|[<!-- language=\"";
+const LANGUAGE_BLOCK_END : &'static str = "\n]|";
+
+pub fn reformat_doc(input: &str) -> String {
+    let out = code_blocks_transformation(input);
+    replace_c_types(&out)
+}
+
+fn code_blocks_transformation(input: &str) -> String {
+    let mut out = String::new();
+
+    for entry in input.split(LANGUAGE_BLOCK_BEGIN) {
+        if out.is_empty() {
+            out.push_str(entry);
+            continue;
+        }
+        out.push_str("```");
+        let entry = get_language(entry, &mut out);
+        out.push_str(&format!("{}", entry.replace(LANGUAGE_BLOCK_END, "\n```")))
+    }
+    out
+}
+
+fn get_language<'a>(entry: &'a str, out: &mut String) -> &'a str {
+    if let Some(pos) = entry.find(LANGUAGE_SEP_END) {
+        out.push_str(&entry[0..pos]);
+        &entry[(pos + LANGUAGE_SEP_END.len())..]
+    } else {
+        out.push_str("\n");
+        entry
+    }
+}
+
+fn replace_c_types(entry: &str) -> String {
+    if let Ok(reg) = Regex::new(r"Gtk[\w]*") {
+        let mut out = vec!();
+
+        for (num, part) in entry.split("```").into_iter().enumerate() {
+            out.push(
+                if num & 1 == 0 {
+                    reg.replace_all(part, |caps: &Captures| {
+                        match caps.at(0) {
+                            Some(s) => format!("`{}`", &s[4..]),
+                            None => String::new(),
+                        }
+                    })
+                } else {
+                    part.to_owned()
+                }
+            );
+        }
+        out.join("```")
+    } else {
+        entry.to_owned()
+    }
+}
--- a/src/codegen/doc/format.rs
+++ b/src/codegen/doc/format.rs
@@ -1,13 +1,14 @@
-use std::ascii::AsciiExt;
 use regex::{Captures, Regex};
+use analysis::namespaces::MAIN;
+use env::Env;
 
 const LANGUAGE_SEP_BEGIN : &'static str = "<!-- language=\"";
 const LANGUAGE_SEP_END : &'static str = "\" -->";
 const LANGUAGE_BLOCK_BEGIN : &'static str = "|[";
 const LANGUAGE_BLOCK_END : &'static str = "\n]|";
 
-pub fn reformat_doc(input: &str, namespace_name: &str) -> String {
-    code_blocks_transformation(input, namespace_name)
+pub fn reformat_doc(input: &str, env: &Env) -> String {
+    code_blocks_transformation(input, env)
 }
 
 fn try_split<'a>(src: &'a str, needle: &str) -> (&'a str, Option<&'a str>) {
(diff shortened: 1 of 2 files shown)


gkoz:
I'm pretty sure regex can handle the `|[...]|` blocks as well (might need to enable multi-line matching).

src/codegen/doc/format.rs
```diff
@@ -0,0 +1,48 @@
+use regex::{Captures, Regex};
+
+const LANGUAGE_SEP_END : &'static str = "\" -->";
+const LANGUAGE_BLOCK_BEGIN : &'static str = "|[<!-- language=\"";
+const LANGUAGE_BLOCK_END : &'static str = "\n]|";
+
+pub fn reformat_doc(input: &str) -> String {
+    let out = code_blocks_transformation(input);
+    replace_c_types(&out)
```
gkoz: You'll probably want to avoid transforming the insides of `|[..]|` blocks.

GuillaumeGomez:
Yes, absolutely. It's just that I'm a bit rusted in regex. I'll train a bit and then it'll be all good!

GuillaumeGomez:
I actually found an easier solution by not running regex on "```" blocks. Unless there is something which needs to be fixed, I think this PR is good to go for now.

src/codegen/doc/format.rs
```diff
@@ -0,0 +1,59 @@
...
+}
+
+fn code_blocks_transformation(input: &str) -> String {
+    let mut out = String::new();
+
+    for entry in input.split(LANGUAGE_BLOCK_BEGIN) {
+        if out.is_empty() {
+            out.push_str(entry);
```
gkoz: Can't you just call `replace_c_types` here?

gkoz: Although I'm not quite sure what's going to happen if there are two code blocks.

GuillaumeGomez: Oh indeed.

GuillaumeGomez: Actually, I'll update this loop a bit to avoid reparsing block codes.

src/codegen/doc/format.rs
```diff
@@ -0,0 +1,59 @@
...
+        out.push_str(&entry[0..pos]);
+        &entry[(pos + LANGUAGE_SEP_END.len())..]
+    } else {
+        out.push_str("\n");
+        entry
+    }
+}
+
+fn replace_c_types(entry: &str) -> String {
+    if let Ok(reg) = Regex::new(r"Gtk[\w]*") {
```
gkoz: This was probably supposed to match both `GtkFoo` and `#GtkFoo`? While here `Gdk` and `G` seem easy enough too. And you'll probably want to use a capture group for the word you want to preserve:
```
(?:#)?(?:G|Gdk|Gtk)([A-Z]\w*)
```

GuillaumeGomez:
Updated and added '@' in the same time.

gkoz:
Is there a way to not abuse `split`? Even regex would probably be clearer.

src/codegen/doc/format.rs
```diff
@@ -0,0 +1,73 @@
...
+        out.push_str(&entry[0..pos]);
+        &entry[(pos + LANGUAGE_SEP_END.len())..]
+    } else {
+        out.push_str("\n");
+        entry
+    }
+}
+
+fn replace_c_types(entry: &str) -> String {
+    if let (Ok(reg), Ok(reg2)) = (Regex::new(r"(?:#)?(?:G[A-Z][a-z]|Gdk|Gtk)[\w]*"), Regex::new(r"@\w*(\s|\n)")) {
```
gkoz: I don't think `G[A-Z][a-z]` is going to match anything.

GuillaumeGomez:
These two split make our task way easier. So why don't using them?

src/codegen/doc/format.rs
```diff
@@ -0,0 +1,73 @@
...
+        let out = &reg.replace_all(entry, |caps: &Captures| {
+            match caps.at(0) {
+                Some(mut s) => {
+                    if s.starts_with("#") {
```
gkoz: Regex can do all of this for you. Use a capture group for the `\w*` part.

gkoz:
> These two split make our task way easier. So why don't using them?

Because it's not easy to reason about.

src/codegen/doc/format.rs
GuillaumeGomez: It's problematic because of the GTypes. This G is already super annoying to handle. To be sure to match, i have to take two more letters, so I can't just take the last capture group.

src/codegen/doc/format.rs
gkoz: Actually I don't know why I suggested `(?:#)?` when `#?` is enough. Oops :)

src/codegen/doc/format.rs
gkoz: The key idea is that you don't handle anything. Regex does that.

GuillaumeGomez: No, for example, for the "GtkAccelGroup", the result will be "`Accel`roup``". I have to precise the rule.

src/codegen/doc/format.rs
GuillaumeGomez: No, for example, for the "GtkAccelGroup", the result will be "Accelroup``". I have to precise the rule.

GuillaumeGomez: And GUI will be matched as well when we don't want to.

src/codegen/doc/format.rs
gkoz: That's why I suggested `(?:G|Gdk|Gtk)([A-Z]\w*)`. Here the prefix must be followed by a capital. But it will be more reliable like this
```
(?:#|\b)(?:G|Gdk|Gtk)([A-Z]\w*)
```
No matching in the middle of a word now.

gkoz:
https://github.com/gtk-rs/gir/pull/204#discussion_r51950736 is invisible now so I'm copying it here.

>That's why I suggested `(?:G|Gdk|Gtk)([A-Z]\w*)`. Here the prefix must be followed by a capital. But it will be more reliable like this

>```
(?:#|\b)(?:G|Gdk|Gtk)([A-Z]\w*)
```

>No matching in the middle of a word now.

**Comment by GuillaumeGomez** [comment]
But GUI does match when we don't want it to.

**Comment by gkoz** [comment]
Yeah. Perhaps require that `G` is always after a `#`, maybe split this in two:
```
[@#]([\w:]+)
\b(?:Gdk|Gtk)([A-Z]\w*)
```

**Comment by gkoz** [comment]
Also you'll want to put the regexes into `lazy_static!`s because creating them many times is slow.

**Comment by gkoz** [comment]
> `\b(?:Gdk|Gtk)([A-Z]\w*)`
This becomes simply `\bG[dt]k([A-Z]\w*)`

**Comment by gkoz** [comment]
Emmm, and I'm leading you somewhat down the wrong path here. I mean in the `gtk` crate we don't want to just strip `Gdk`, we want to change it to `gdk::`. And vice versa.

**Comment by GuillaumeGomez** [comment]
Ok, I added the `gdk::` prefix when needed. What do you mean by "vice versa"? Adding it as well for `gtk::`?

For the regex, we'll to talk about it again. Wanted to do it all in once in regex isn't a good idea from my point of view.

**Comment by gkoz** [comment]
> Ok, I added the gdk:: prefix when needed. What do you mean by "vice versa"? Adding it as well for gtk::?

When processing `gdk` we'll want to change `Gtk` to `gtk::`. And there's `G`. So yeah, taking into account the main namespace is going to take some effort.

File: src/codegen/doc/format.rs [hunk 6]
```diff
@@ -0,0 +1,74 @@
...
+    }
+}
+
+fn replace_c_types(entry: &str) -> String {
+    let out = &REG.replace_all(entry, |caps: &Captures| {
+        match caps.at(0) {
+            Some(mut s) => {
+                if s.starts_with("#") {
+                    s = &s[1..];
+                }
```
gkoz: You need to stop counting characters manually. Capture the important part -- `\w*` into a capture group and you'll have it in `caps[1]`.

src/codegen/doc/format.rs
```diff
@@ -0,0 +1,73 @@
...
+        &entry[(pos + LANGUAGE_SEP_END.len())..]
+    } else {
+        out.push_str("\n");
+        entry
+    }
+}
+
+fn replace_c_types(entry: &str) -> String {
+    let out = &REG.replace_all(entry, |caps: &Captures| {
+        match caps.at(0) {
```
gkoz: `caps[0]` (the whole match) and any non-optional capture groups will always be defined.

src/codegen/doc/format.rs
```diff
@@ -0,0 +1,73 @@
...
+}
+
+pub fn reformat_doc(input: &str) -> String {
+    code_blocks_transformation(input)
+}
+
+fn code_blocks_transformation(input: &str) -> String {
+    let mut out = String::new();
+
+    for (num, entry) in input.split(LANGUAGE_BLOCK_BEGIN).into_iter().enumerate() {
```
gkoz: If you insist in doing this the hard way, I think the primitive you need here is a custom `split` like this:
```
fn try_split<'a>(src: &'a str, needle: &src) -> (&'a str, Option<&'a str>) {
    match src.find(needle) {
        Some(pos) => (src[..pos], Some(src[pos + needle.len())),
        None => (src, None),
    }
}
```
Then you'll only need one loop, likely without any `break`/`continue`.

Regex is going to be more robust still and easily handle optional language and make sure `|[` starts at newline if you ask it to.

src/codegen/doc/format.rs
GuillaumeGomez: Yes but for G[A-Z][a-z], a part of the word will be missing.

gkoz: And this is why `G[A-Z][a-z]` is not the best approach. It might be possible to get more creative in the "substance" part, e.g. `(G|G[dt]k)([A-Z]\w*[a-z]\w*)` so the prefix must be followed by an uppercase letter but there also must be at least one lowercase one there somewhere.

gkoz: I guess this still breaks on "GUIs". I just can't win, eh? :)

GuillaumeGomez: Yup, sorry. ^^'

gkoz: Oh but both `G[A-Z][a-z]` and my above suggestion will eat "GLib". If we don't perform library lookups let's just leave naked `G.*` symbols alone?

GuillaumeGomez: Yes, no other choice for now...

src/codegen/doc/format.rs
```diff
@@ -0,0 +1,71 @@
...
+        &entry[(pos + LANGUAGE_SEP_END.len())..]
+    } else {
+        out.push_str("```\n");
+        entry
+    }
+}
+
+fn replace_c_types(entry: &str) -> String {
+    let out = &REG.replace_all(entry, |caps: &Captures| {
+        let mut s = caps.at(0).unwrap();
```
gkoz: `at(0).unwrap()` is the same as `[0]`.

src/codegen/doc/format.rs
```diff
@@ -0,0 +1,71 @@
...
+        if s.starts_with("#") {
+            s = &s[1..];
+        }
```
gkoz: This piece is 100% unnecessary. Use capture groups.

src/codegen/doc/format.rs
```diff
@@ -0,0 +1,74 @@
...
+    entry
+}
+
+fn replace_c_types(entry: &str) -> String {
+    let out = &REG.replace_all(entry, |caps: &Captures| {
+        let mut s = &caps[0];
+        if s.starts_with("#") {
+            s = &s[1..];
+        }
+        format!("`{}::{}`", &s[..3].to_lowercase(), &s[3..])
```
gkoz: So with capture groups (`#?(G[dt]k)([\w]*)`) this would be as simple as `"{}::{}", caps[1], caps[2]` without any manual slicing.

src/codegen/doc/format.rs
```diff
@@ -0,0 +1,69 @@
...
+
+const LANGUAGE_SEP_BEGIN : &'static str = "<!-- language=\"";
+const LANGUAGE_SEP_END : &'static str = "\" -->";
+const LANGUAGE_BLOCK_BEGIN : &'static str = "|[";
+const LANGUAGE_BLOCK_END : &'static str = "\n]|";
+
+lazy_static! {
+    static ref REG : Regex = Regex::new(r"#?(G[dt]k)([\w]*)").unwrap();
+    static ref REG2 : Regex = Regex::new(r"@(\w*)").unwrap();
+}
```
gkoz: It's probably better to define them close to where they're used.

src/codegen/doc/format.rs
```diff
@@ -0,0 +1,69 @@
...
+            }
+        };
+    }
+}
+
+fn get_language<'a>(entry: &'a str, out: &mut String) -> &'a str {
+    if entry.starts_with(LANGUAGE_SEP_BEGIN) {
+        if let Some(pos) = entry.find(LANGUAGE_SEP_END) {
+            out.push_str(&format!("```{}", &entry[LANGUAGE_SEP_BEGIN.len()..pos]));
+            return &entry[(pos + LANGUAGE_SEP_END.len())..]
```
gkoz: This bit looks very similar to `try_split` I bet it could use that.

src/codegen/doc/format.rs
```diff
@@ -0,0 +1,69 @@
...
+        }
+    }
+    out.push_str("```");
```
gkoz:
```
```text
```
probably?

GuillaumeGomez:
@gkoz: I added the other types as well (the `%` is just out of our league for now). I think we're good on features. Review please?

src/codegen/doc/format.rs
```diff
@@ -0,0 +1,69 @@
+use std::ascii::AsciiExt;
+use regex::{Captures, Regex};
+
+const LANGUAGE_SEP_BEGIN : &'static str = "<!-- language=\"";
+const LANGUAGE_SEP_END : &'static str = "\" -->";
+const LANGUAGE_BLOCK_BEGIN : &'static str = "|[";
+const LANGUAGE_BLOCK_END : &'static str = "\n]|";
+
+pub fn reformat_doc(input: &str, namespace_name: &str) -> String {
```
gkoz: It seems more future-proof to pass `&analysis::namespaces::Info` instead of `namespace_name` here.

src/codegen/doc/format.rs
```diff
@@ -0,0 +1,69 @@
...
+    static ref REG : Regex = Regex::new(r"#?(G[dt]k)([\w]+:?:?\.?[\w-]+)").unwrap();
+    static ref REG2 : Regex = Regex::new(r"@(\w*)").unwrap();
+}
+
+fn replace_c_types(entry: &str, namespace_name: &str) -> String {
+    let out = &REG.replace_all(entry, |caps: &Captures| {
+        if caps[1].eq_ignore_ascii_case(namespace_name) {
+            format!("`{}`", &caps[2])
+        } else {
+            format!("`{}::{}`", &caps[1].to_lowercase(), &caps[2])
```
gkoz: Then you can take the `crate_name` of the namespace here.

src/codegen/doc/mod.rs
```diff
@@ -137,16 +143,16 @@ fn create_class_doc<T: FunctionTraitType + ToStripperType>(w: &mut Write, class:
     }
 
     if let Some(ref class_doc) = class.doc() {
-        try!(write_lines(w, &class_doc, &tabs));
+        try!(write_lines(w, &class_doc, &tabs, &namespace_name));
```
gkoz: Redundant referencing is confusing -- `namespace_name` already is a reference.

src/codegen/doc/format.rs
GuillaumeGomez: You mean `library::Namespace`, no?

gkoz: No but I got the struct name wrong: `analysis::namespaces::Namespace` is the one which has `crate_name`

GuillaumeGomez:
Updated, however I used `library::Namespace` before so I just used it once again. Please say it if it doesn't fit what you have in mind.

gkoz:
It doesn't have `crate_name` that was supposed to supply the crate name for regex to use.

GuillaumeGomez:
Updated.

gkoz:
Thanks!

PR #205, opened 2016-02-04
Make "ignored" classes, interfaces or records ignored rust types as well

### LINKED ISSUE #201: Failure to ignore a class when passed in a container
```toml
[options]
girs_dir = "gir-files"
library = "Gio"
version = "2.0"
min_cfg_version = "2.32"
target_path = "."
work_mode = "normal"
external_libraries = [
    "GLib",
    "GObject",
]

generate = [
    "Gio.AppInfo",
    "Gio.AppLaunchContext",
]
```

```rust
    pub fn get_display<T: IsA<AppInfo>>(&self, info: &T, files: &[File]) -> Option<String> {
        unsafe {
            from_glib_full(ffi::g_app_launch_context_get_display(self.to_glib_none().0, info.to_glib_none().0, files.to_glib_none().0))
        }
    }
```

gkoz:
Fix #201

### DIFF
--- a/src/analysis/rust_type.rs
+++ b/src/analysis/rust_type.rs
@@ -94,7 +94,7 @@ fn rust_type_full(env: &Env, type_id: library::TypeId, nullable: Nullable, ref_m
         let rust_type_with_prefix = rust_type.map(|s| format!("{}::{}",
             crate_name(&env.library.namespace(type_id.ns_id).name), s));
         if env.type_status(&type_id.full_name(&env.library)).ignored() {
-            rust_type = Err(rust_type_with_prefix.as_str().into());
+            rust_type = Err(format!("/*Ignored*/{}", rust_type_with_prefix.as_str()));
         } else {
             rust_type = rust_type_with_prefix;
         }
@@ -165,9 +165,7 @@ pub fn parameter_rust_type(env: &Env, type_id:library::TypeId,
             Bitfield(..) => format_parameter(rust_type, direction),
 
         Record(..) => {
-            if env.type_status(&type_id.full_name(&env.library)).ignored() {
-                Err(format!("/*Ignored*/{}", rust_type.as_str()))
-            } else if direction == library::ParameterDirection::InOut {
+            if direction == library::ParameterDirection::InOut {
                 Err(format!("/*Unimplemented*/{}", rust_type.as_str()))
             } else {
                 rust_type
@@ -177,9 +175,6 @@ pub fn parameter_rust_type(env: &Env, type_id:library::TypeId,
         Class(..) |
             Interface (..) => {
             match direction {
-                _ if env.type_status(&type_id.full_name(&env.library)).ignored() => {
-                    Err(format!("/*Ignored*/{}", rust_type.as_str()))
-                }
                 library::ParameterDirection::In |
                     library::ParameterDirection::Out |
                     library::ParameterDirection::Return => rust_type,
--- a/src/analysis/rust_type.rs
+++ b/src/analysis/rust_type.rs
@@ -7,12 +7,46 @@ use nameutil::crate_name;
 use super::conversion_type::ConversionType;
 use traits::*;
 
-pub type Result = result::Result<String, String>;
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum TypeError {
+    Ignored(String),
+    Mismatch(String),
+    Unimplemented(String),
+}
 
-impl AsStr for Result {
-    #[inline]
-    fn as_str(&self) -> &str {
-        self.as_ref().unwrap_or_else(|s| s)
+pub type Result = result::Result<String, TypeError>;
+
+fn into_inner(res: Result) -> String {
+    use self::TypeError::*;
+    match res {
+        Ok(s) |
+            Err(Ignored(s)) |
+            Err(Mismatch(s)) |
+            Err(Unimplemented(s)) => s,
+    }
+}
+
+impl IntoString for Result {
+    fn into_string(self) -> String {
+        use self::TypeError::*;
+        match self {
+            Ok(s) => s,
+            Err(Ignored(s)) => format!("/*Ignored*/{}", s),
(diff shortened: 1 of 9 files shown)


src/analysis/rust_type.rs
```diff
@@ -91,14 +112,22 @@ fn rust_type_full(env: &Env, type_id: library::TypeId, nullable: Nullable, ref_m
...
-            rust_type = Err(rust_type_with_prefix.as_str().into());
-        } else {
-            rust_type = rust_type_with_prefix;
+            rust_type = Err(rust_type.as_str().to_owned());
+        }
+        rust_type = rust_type.map_any(|s| format!("{}::{}",
+            crate_name(&env.library.namespace(type_id.ns_id).name), s));
+    }
+
+    if let Err(s) = rust_type {
```
EPashkin: This condition is true on other errors, like "Unsupported"

src/analysis/rust_type.rs
```diff
@@ -65,9 +65,30 @@ fn rust_type_full(env: &Env, type_id: library::TypeId, nullable: Nullable, ref_m
 
         Enumeration(ref enum_) => Ok(enum_.name.clone()),
         Bitfield(ref bitfield) => Ok(bitfield.name.clone()),
-        Record(ref record) => Ok(record.name.clone()),
-        Interface(ref interface) => Ok(interface.name.clone()),
-        Class(ref klass) => Ok(klass.name.clone()),
+        Record(ref record) => {
+            if env.type_status(&type_id.full_name(&env.library)).ignored() {
```
EPashkin: Repeated 3 times -> better make function, may include adding /*Ignored*/

src/analysis/rust_type.rs
gkoz: This is getting out of hand, I'll think about adding an enum that would also handle the "Unsupported" and "Ignored" comment.

EPashkin: Not better just add `check_for_ignored: bool` and check it here?
Then also functions will not needed.

src/analysis/rust_type.rs
gkoz: They can probably be collapsed into a single match arm because using `Type::get_name`.

EPashkin: Yes, with enumeration and bitfield also

EPashkin: If you will collapse matches, remove two TODOs about checking `library::Type::get_name()`

src/analysis/rust_type.rs
```diff
@@ -146,29 +193,24 @@ pub fn parameter_rust_type(env: &Env, type_id:library::TypeId,
...
             }
         }
         Alias(ref alias) => {
-            let res = format_parameter(rust_type, direction);
-            if parameter_rust_type(env, alias.typ, direction, nullable, ref_mode).is_ok() {
-                res
-            } else {
-                res.and_then(|s| Err(s))
-            }
+            let rust_type = rust_type.map_any(|s| format_parameter(s, direction));
```
EPashkin: Can this be moved into `and_then` ?

EPashkin: Now `Atom` shows as `/*Ignored*/gdk::Atom_` (Ex. in `gtk_text_buffer_register_serialize_tagset`)
Seems this not equal change.

gkoz:
I think I got `Alias` right now.

src/analysis/rust_type.rs
```diff
@@ -146,29 +193,28 @@ pub fn parameter_rust_type(env: &Env, type_id:library::TypeId,
...
         }
         Alias(ref alias) => {
-            let res = format_parameter(rust_type, direction);
-            if parameter_rust_type(env, alias.typ, direction, nullable, ref_mode).is_ok() {
-                res
-            } else {
-                res.and_then(|s| Err(s))
-            }
+            rust_type.and_then(|s| {
+                match parameter_rust_type(env, alias.typ, direction, nullable, ref_mode) {
```
gkoz: Eh, this should just have been `map_any`

src/analysis/rust_type.rs
```diff
@@ -53,21 +87,36 @@ fn rust_type_full(env: &Env, type_id: library::TypeId, nullable: Nullable, ref_m
...
-                .map_any(|_| alias.name.clone()),
-
-        Enumeration(ref enum_) => Ok(enum_.name.clone()),
-        Bitfield(ref bitfield) => Ok(bitfield.name.clone()),
-        Record(ref record) => Ok(record.name.clone()),
-        Interface(ref interface) => Ok(interface.name.clone()),
-        Class(ref klass) => Ok(klass.name.clone()),
+        Alias(ref alias) => {
+            rust_type_full(env, alias.typ, nullable, ref_mode)
+                .map_any(|_| alias.name.clone())
```
EPashkin: Now it always aliased name.
Why not `type_.get_name().to_owned()` ?

gkoz: Just `get_name` without checking if the actual type is not unimplemented?

EPashkin: Not. In map any instead `alias.name.clone()`

gkoz: We have the `alias` binding already. I don't see why go through `get_name`. Doesn't it return the same? For consistency?

EPashkin: For something like `Rectangle` alias for `RectangleInt`. (Just theoretical example)

EPashkin: Ops. Was wrong. Forget this ;)

EPashkin:
Thanks.

src/analysis/rust_type.rs
gkoz: `type_.get_name` [literally is](`https://github.com/gtk-rs/gir/blob/master/src/library.rs#L384`) the same as `alias.name.clone()`. Whether there's going to be a need to get the name of `alias.typ` is unclear at this point.

EPashkin: Yes, i mistaken `alias.name` for aliased name.

Issue #199, opened 2016-01-28
Generate signals


EPashkin:
Generate `pub trait xxxSignals` for classes with "IsA" implementation for classes with children
or direct implimentation for classes without children (same as for "xxxExt" traits).
For both cases generate trampoline function.

Example for ComboBox and TreeSelection: https://github.com/EPashkin/rust-gnome-gtk/tree/signal_sample

gkoz:
I'm still undecided about where exactly the signals belong. If they are generated alongside other functions, perhaps `xxxSignals` traits are unnecessary, just put them into `xxxExt`?

Also we should probably fix `callback_guard!` to not require importing the struct.

EPashkin:
Currently signals trait implemented with bound `IsA<ComboBox> + IsA<Object>`.

gkoz:
Yeah the `IsA<Object>` is required to get `ToGlibPtr<*mut GObject>` but it's harmless as a bound because all objects implement it.

EPashkin:
Other that bound don't see any problem with implementing in `xxxExt` or in same impl for childless class.

gkoz:
Generating correct trampoline signatures might require finishing the `ffi_ng` branch. Eventually I'd like the trampoline signatures to match C exactly (no `f: &Box<Fn(...)>`, but rather cast `user_data: gpointer` to `&Box<Fn(...)>`).

gkoz:
It's worse than that: signal definitions seem routinely omit `c:type` attributes:

```xml
      <glib:signal name="accel-activate" detailed="1">
...
        <parameters>
          <parameter name="acceleratable" transfer-ownership="none">
            <doc xml:space="preserve">the object on which the accelerator was activated</doc>
            <type name="GObject.Object"/>
          </parameter>
          <parameter name="keyval" transfer-ownership="none">
            <doc xml:space="preserve">the accelerator keyval</doc>
            <type name="guint" c:type="guint"/>
          </parameter>
          <parameter name="modifier" transfer-ownership="none">
            <doc xml:space="preserve">the modifier combination of the accelerator</doc>
            <type name="Gdk.ModifierType"/>
          </parameter>
        </parameters>
      </glib:signal>
```

EPashkin:
If it omit only for glib-related types then its not big problem

gkoz:
[...]

PR #206, opened 2016-02-05
Signals

EPashkin:
Part of #199
WIP

I don't planing read these signal attributes as don't see its use now: when, detailed, action, no-recurse, introspectable, no-hooks


gkoz:
So there is `no-recurse`! This is nice. Such signals could take `FnMut` instead of `Fn`.

src/parser.rs
```diff
@@ -13,6 +13,8 @@ use library::*;
 type Reader = EventReader<BufReader<File>>;
 type Attributes = Vec<OwnedAttribute>;
 
+const EMPTY_CTYPE: &'static str = "/*EMPTY*/";
```
gkoz: This is obviously suboptimal but I'm not sure if it's better to have two `Parameter` types or put up with optional `c_type`. Perhaps the composition&`Deref` trick could help here too.

EPashkin: Seems for signals separate classes not needed.
Used 111 types can has "object" c_type, and can filled with it (like `GdkWindow*`) , others 5 is array types and can be processed later.
Full list https://gist.github.com/EPashkin/b13bf3e8e4a6087a8078.
Seems no primitive types without c:type in signals.

EPashkin:
Add signals empty ctypes post detecting

EPashkin:
Made `enumerate` but can't do single pass update due needed `Library::type_` in `c_type_by_type_id`

EPashkin:
Rebased

EPashkin:
@gkoz There really needed type bounds in trampolines?
Its seems unused, and if I remove it in https://github.com/gtk-rs/gtk/blob/master/src/signal.rs#L1041-L1060 `gtk` still builds fine.

gkoz:
Uh-oh. This looks like a giant bug: `T` is not used at all! I think it was supposed to be more like this
```rust
    unsafe extern "C" fn void_trampoline<T>(this: *mut GtkComboBox, f: &Box<Fn(&T) + 'static>)
    where T: IsA<ComboBox> {
        callback_guard!();
        f(&ComboBox::from_glib_none(this).downcast_unchecked());
    }
```

gkoz:
We could make signals first-class types kinda like they used to be in rgtk.

```
unsafe trait Signal<Receiver> {
    type Closure;
    /// ...
}

enum ButtonClicked { }

unsafe impl<O: IsA<Button>> Signal<O> for ButtonClicked {
    type Closure = Fn(&O);
    // ...
}

trait ObjectExt {
    fn connect<S: Signal<Self>>(&self, f: S::Closure) -> u64;
    fn connect_after<S: Signal<Self>>(&self, f: S::Closure) -> u64;
}

button.connect::<ButtonClicked>(|btn| do_something());
```

EPashkin:
IMHO worth to try.

gkoz:
My example wouldn't compile though because `S::Closure` is unsized :( It has to be more complex.

GuillaumeGomez:
Any news on this @EPashkin?

EPashkin:
Sorry, no free time.
Anyway IMHO it don't need in next crate release as will subject of changes.

EPashkin:
Ready for merging.
Some part is not done, but I prefer do it in separate PRs:
1) Versions
2) Configuration (ignore)
3) Remove doubleboxing

GuillaumeGomez:
I think it's better to stop here, it's already a very big PR! Thanks a lot for your work @EPashkin!

EPashkin:
Strange, my generator don't want generate `MenuItem::connect_toggle_size_request`, I don't understand why.

EPashkin:
Now understand, it accept `gpointer` in Gtk-3.0.gir, so we don't know real type.
In https://github.com/gtk-rs/gtk/pull/300 states that it also `int`, not sure that it right.

EPashkin:
https://people.gnome.org/~shaunm/girdoc/C/Gtk.MenuItem.toggle_size_request.html
states that it `int*` not `int`

EPashkin:
@GuillaumeGomez  C-source states also, so it need be fixed (or disabled) in your PR.

EPashkin:
Minor changes in `codegen::signal::bounds` (removed if in match)

GuillaumeGomez:
I'll rmeove them from my PR.

gkoz:
When trying this on `gio` looks like some imports are missing. Funny thing about `gio`: I've only just noticed that we'll need an override to force generation of traits: `Gio-2.0.gir` can't know that `Gtk-3.0.gir` subclasses `Application`.

Do you think we should keep adding the `connect_` methods for now rather than design a generic API?

EPashkin:
Oh, I fully forgot about imports, so it may has problem even in gtk :(
Planned fixing imports after add version and ignore, when updating gtk.

`connect_` has good use with closures. Do not see how it can be replaced.

gkoz:
I'm gravitating toward something like
```rust
widget.connect::<WidgetDeleteEvent>(|this, event| { ... });
```
(eventually `widget.connect::<Widget::DELETE_EVENT>(...);`)

which could take several closure kinds:
```rust
Fn(&T, &Event) -> Inhibit + 'static
Fn(&T, &Event) + 'static
// possibly
Fn(&Event) -> Inhibit + 'static
Fn(&Event) + 'static
// maybe even?
Fn() -> Inhibit + 'static
Fn() + 'static // this one would allow `window.connect::<WidgetDeleteEvent>(gtk::main_quit)`
```

But this might not get fleshed out in a while.

EPashkin:
IMHO approach with `::<` will bad for IDE's (racer) autocomplete function.
Secondly we get some "God" function that also not good.

EPashkin:
Imports is sure problem so this PR still WIP

EPashkin:
Type `GtkBox` conflict with signals in almost all files, can I rename its generated type to 'Box_'?
Or better import `std::boxed::Box` as `Box_` and generate signal with it?

gkoz:
> Type GtkBox conflict with signals in almost all files, can I rename its generated type to Box_?
Or better import std::boxed::Box as Box_ and generate signal with it?

Tough one :( Is it feasible to avoid importing just this one type and always spell it as `::Box`?

EPashkin:
`::Box` work same as just `Box` from gtk. `std::boxed::Box` IMHO too long and working only after `use std;`

gkoz:
If there's no `use Box;`, I believe `Box` should resolve to `::std::boxed::Box` and `::Box` to `::auto::box_::Box`. But this special case it could be too complex to implement.

EPashkin:
You right about both parts: it work but unnecessary complex

EPashkin:
Currently I stay with `use std::boxed::Box as Box_`

gkoz:
Alright, let's look into renaming our `Box` to `Box_`. Reexporting it as `Box` manually should avoid breaking any code.

gkoz:
For reference, I've outlined the way to avoid double-boxing [here](https://github.com/gkoz/gtk-rs-glib/commit/628c541d02e09a37a3282998e3a05319c6060b19) but the significance of the monomorphization cost (how bad it is to generate a copy of each trampoline per closure) is not clear to me and in the best case this feels like premature optimization.

EPashkin:
If we reexporing gtk's `Box` as `Box` to users, then I don't see any reason to rename it internally.

EPashkin:
I currently don't see any problem with many copies for trampolines, this too small anyway.
Also I decide not combine trampolines and generate one trampoline for each signal.

We can do `double-boxing` fix after finishing applying signal generation to Gtk.

gkoz:
> If we reexporing gtk's Box as Box to users, then I don't see any reason to rename it internally.

I was thinking of trying to keep the code idiomatic. Signal bindings are the most fragile it seems and putting speed bumps like `Box_` or `::std::boxed::Box` into that code could make reviewing harder. There's value in everyone knowing that `Box` is what they expect.
At the same time users would have a choice of using `Box_` or `Box` whichever suits them best. But having two names for the struct could be confusing in its own way I suppose.

EPashkin:
Two Boxes mixed almost only in generated code, so IMHO review or Box_\Box dilemma its not problem.

gkoz:
I don't feel strongly about this so okay, let's go with your solution.

src/analysis/signals.rs
```diff
@@ -0,0 +1,76 @@
...
+    let trampoline_name = trampolines::analyze(env, signal, type_tid, in_trait, trampolines,
+                                               &mut used_types, version);
+
+    if trampoline_name.is_some() {
+        for s in used_types {
+            if let Some(i) = s.find("::") {
+                imports.add(&s[..i], version);
+            } else {
+                imports.add(&s, version);
+            }
```
gkoz: There should probably be an `Imports` method, that does this.

EPashkin: You right, this logic better be in one place, even if only 2 uses.

src/codegen/trampoline.rs
```diff
@@ -0,0 +1,147 @@
...
+    } else {
+        ("", " {")
+    };
+
+    let params_str = trampoline_parameters(env, analysis);
+    let func_str = func_string(env, analysis, None::<(&str, &str)>);
+    let ret_str = trampoline_returns(env, analysis);
+
+    try!(version_condition(w, env, analysis.version, false, 0));
+    try!(writeln!(w, "unsafe extern \"C\" fn {}{}({}, f: &Box_<{}>){}{}",
```
gkoz: I'd rather we continued to clean up the extern fns' signatures, put a `gpointer` here and cast it in the body.

EPashkin: Safer do it after gtk has current variant.
Also I plan use `Chunks` here too, so this will be reworked anyway.

EPashkin:
Now we got working version.
Gtk update coming soon.
There still some errors in gtk, I solve it ignoring some signals,
also add issues latter (or update #199)

EPashkin:
Rebased on skip_assert fix

gkoz:
I've only realized this now, almost too late: the generator doesn't know when to translate `bool` to `Inhibit` does it?

EPashkin:
Yes, I don't think about it.

EPashkin:
You mean `Inhibit` in closure return to `gboolean` in trampoline return?

gkoz:
Yes, e.g. generating `RangeExt::connect_change_value` would be a breaking change.

EPashkin:
Added issue, remove generation for Range

EPashkin:
AppVeyor build succeeded

EPashkin:
By way anyone know how do "Hooray" emoji?

mitaa:
http://www.emoji-cheat-sheet.com/

`:tada:`

:tada:

EPashkin:
Thanks. Just don't see it in github help.

PR #208, opened 2016-02-10
Drop version detection in favor of features, add version support for enum variants

gkoz:
This should ensure better API stability control as described in https://github.com/gtk-rs/gtk/issues/243 (new upstream library versions support should not introduce breaking changes in code that sticks to earlier versions).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ rust:
 addons:
   apt:
     packages:
-    - libgtk-3-dev
     - libssh2-1-dev
 script:
   - cargo build --release
--- a/src/codegen/function.rs
+++ b/src/codegen/function.rs
@@ -38,8 +38,7 @@ pub fn generate(w: &mut Write, env: &Env, analysis: &analysis::functions::Info,
 
     try!(writeln!(w, ""));
     try!(cfg_condition(w, &analysis.cfg_condition, commented, indent));
-    try!(version_condition(w, &env.config.library_name,
-        env.config.min_cfg_version, analysis.version, commented, indent));
+    try!(version_condition(w, env.config.min_cfg_version, analysis.version, commented, indent));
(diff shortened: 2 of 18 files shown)


EPashkin:
No problem with code.
2 possible problems with sys generation:
1. Features in 'atk-sys\Cargo.toml` sorted as strings.
2. sys_build build only basic version (3.4)?

gkoz:
> Features in 'atk-sys\Cargo.toml` sorted as strings.

The toml library is the limit factor here. Hopefully a ["round-trippable" fork](https://github.com/vosen/toml_document) will be ready soon.

> sys_build build only basic version (3.4)?

Yeah, I should upgrade the test script.

EPashkin:
Configuring "Gdk.EventType" better moved to `tests\sys\gir-gdk.toml`

EPashkin:
Don't see `touchpad_swipe` and `touchpad_pinch` in `Gdk.EventType` in sys. Seems `gir-files` need update.

gkoz:
That's right. I keep putting off the update to 3.18 because it could break some assumptions in the generator but we should probably try it anyway. And also because the last time I built GTK+ 3.18 it seemed to require an unreleased cairo version.

EPashkin:
Wrong file for `EventType config`? `tests/sys/gir-gdk-pixbuf.toml`

EPashkin:
Thanks.

EPashkin:
We forgot commit `tests/sys/*/Cargo.toml` after rebuild

gkoz:
Hmm yes, although the generator does the right thing to them now, stale manifests could cause problems in the future.

gkoz:
Apparently I haven't researched and tested the linking changes properly so this might have to be reverted, I'll be more sure tomorrow.

PR #209, opened 2016-02-11
Revert "Drop version detection in favor of features, add version support for enum variants"

gkoz:
Backing out the mistaken #208. I'll salvage good commits from it.


EPashkin:
Not better just repush master if it really that bad?

EPashkin:
Or github will be bad with it?

gkoz:
Okay, I suppose it won't affect a lot of people.

gkoz:
Rolled back `master`.

EPashkin:
Will you recreate `version_features` PR with WIP status? Or it unfixable?

gkoz:
Basically removing the build scripts was nonsense, they should at least emit `cargo:rustc-link-lib={package_name}` but perhaps even try calling `pkg-config` (but not fail if it's absent). So that is the part, which needs to be rewritten. The other commits will remain but I might have to resolve some merge conflicts.

PR #211, opened 2016-02-13
Extract equal functionality for config items to macros

EPashkin:


gkoz:
This looks like it could just be a trait, partly because newtypes don't seem to be called for.

src/config/identables.rs
```diff
@@ -0,0 +1,37 @@
...
+                        if let Some(par) = $item::parse(par, object_name) {
+                            v.push(par);
+                        }
+                    }
+                }
+
+                $items(v)
+            }
+
+            pub fn matched(&self, name: &str) -> Vec<&$item> {
```
gkoz: We might try experimenting with trait objects, e.g. would `Box<Iterator<Item = &$item>>` here?

EPashkin:
For trait case needed to many declaration:
only `matches` needed 2 function, for `parse` also 2 functions: 'parse` for item and `from` for items.
```
impl Identable for Function {
    fn ident(&self) -> &Ident { &self.ident }
}

impl Identables for Functions {
    type Item = Function;

    fn iter(&self) -> Iter<Self::Item> { self.0.iter() }
}
```

src/config/identables.rs
EPashkin: Currently it used in `Function::matched_parameters` with `extend_from_slice`

gkoz:
Some of boilerplate would come from the existence of newtypes but is there any point to hiding that `Vec<Function>` is a vector? A blanket implementation for `Vec<T: Parse + AsRef<Ident>>` should do the job.

EPashkin:
You right about unneeded hiding. I add other PR with traits to compare result.

EPashkin:
Closing in favor of #212

PR #212, opened 2016-02-13
Extract equal functionality for config items to traits

EPashkin:
Trait case for #211

### DIFF
--- a/src/config/identables.rs
+++ b/src/config/identables.rs
@@ -1,0 +1,34 @@
+use super::ident::Ident;
+use toml::Value;
+
+pub trait Parse : Sized {
+    fn parse(toml: &Value, name: &str) -> Option<Self>;
+}
+
+pub trait Identables {
+    type Item;
+
+    fn parse(toml: Option<&Value>, object_name: &str) -> Vec<Self::Item>;
+    fn matched(&self, name: &str) -> Vec<&Self::Item>;
+}
+
+impl<T: Parse + AsRef<Ident>> Identables for Vec<T> {
+    type Item = T;
+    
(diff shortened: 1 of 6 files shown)


src/analysis/functions.rs
```diff
@@ -9,6 +9,7 @@ use analysis::return_value;
 use analysis::rust_type::*;
 use analysis::safety_assertion_mode::SafetyAssertionMode;
 use config;
+use config::identables::Identables;
```
gkoz: It might make sense to reexport `Identables` in `mod traits`;

src/config/identables.rs
```diff
@@ -0,0 +1,34 @@
...
+    fn matched(&self, name: &str) -> Vec<&Self::Item>;
+}
+
+impl<T: Parse + AsRef<Ident>> Identables for Vec<T> {
+    type Item = T;
+    
+    fn parse(toml: Option<&Value>, object_name: &str) -> Vec<Self::Item> {
+        let mut v = Vec::new();
+        if let Some(fns) = toml.and_then(|val| val.as_slice()) {
+            for f in fns {
```
gkoz: More generic names for `f` and `fns` would be better.

gkoz:
I like this better than macros.

EPashkin:
Fixed

gkoz:
> It might make sense to reexport `Identables` in mod `traits`

Not sure if you disagree or misunderstood. What I meant is we already import `traits::*` so the `Identables` trait could get imported for free.

EPashkin:
I missed about `mod traits`.

EPashkin:
Reexported

EPashkin:
I also prefer this variant

gkoz:
Thanks!

PR #213, opened 2016-02-18
Small doc generator fixes and improvements

gkoz:

PR #214, opened 2016-02-19
More intelligent symbol name handling and other docs generator improvements

gkoz:
Implement more of #203

- Small fixes (#213)
- Add symbol dictionary to translate C names to Rust correctly
- Avoid mangling inline code in backquotes
- Mangle parameter names and change instance parameter to `self`
- Shield XML tags

### DIFF
--- a/src/analysis/symbols.rs
+++ b/src/analysis/symbols.rs
@@ -1,0 +1,144 @@
+use analysis::namespaces::{self, NsId};
+use case::CaseExt;
+use library::*;
+use std::collections::HashMap;
+
+#[derive(Clone, Debug, Default)]
+pub struct Symbol {
+    crate_name: Option<String>,
+    owner_name: Option<String>,
+    name: String,
+}
+
+impl Symbol {
+    pub fn full_rust_name(&self) -> String {
+        let mut ret = String::new();
+        if let Some(ref s) = self.crate_name {
+            ret.push_str(s);
+            ret.push_str("::");
+        }
+        if let Some(ref s) = self.owner_name {
+            ret.push_str(s);
+            ret.push_str("::");
+        }
+        ret.push_str(&self.name);
+        ret
+    }
+
+    pub fn make_trait_method(&mut self) {
+        let name = self.owner_name.take();
+        self.owner_name = name.map(|s| format!("{}Ext", s));
+    }
+}
+
+#[derive(Debug)]
+pub struct Info {
+    symbols: Vec<Symbol>,
+    c_name_index: HashMap<String, u32>,
+}
+
+pub fn run(library: &Library, namespaces: &namespaces::Info) -> Info {
+    let mut info = Info {
+        symbols: Vec::new(),
+        c_name_index: HashMap::new(),
+    };
+
+    info.insert("NULL", Symbol { name: "None".into(), ..Default::default() });
+    info.insert("FALSE", Symbol { name: "false".into(), ..Default::default() });
+    info.insert("TRUE", Symbol { name: "true".into(), ..Default::default() });
+
+    for (ns_id, ref ns) in library.namespaces.iter().enumerate() {
+        let ns_id = ns_id as NsId;
+        if ns_id == namespaces::INTERNAL {
+            continue
+        }
+
+        let crate_name = if ns_id == namespaces::MAIN {
+            None
+        } else {
+            Some(&namespaces[ns_id].crate_name)
+        };
+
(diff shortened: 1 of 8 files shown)


src/analysis/symbols.rs
```diff
@@ -0,0 +1,148 @@
...
+    info.insert("FALSE", Symbol { name: "false".into(), ..Default::default() });
+    info.insert("TRUE", Symbol { name: "true".into(), ..Default::default() });
+
+    for (ns_id, ref ns) in library.namespaces.iter().enumerate() {
+        let ns_id = ns_id as NsId;
+        if ns_id == namespaces::INTERNAL {
+            continue
+        }
+
+        let crate_name = || -> Option<String> {
```
EPashkin: This really need be function and not `Option<&str>`?

gkoz: Oops it used to be more involved but now the closure doesn't make sense.

src/analysis/symbols.rs
```diff
@@ -0,0 +1,147 @@
...
+            continue
+        }
+
+        let crate_name;
+        if ns_id == namespaces::MAIN {
+            crate_name = None;
+        }
+        else {
+            crate_name = Some(&namespaces[ns_id].crate_name);
+        }
```
EPashkin: `let crate_name = if ...` ?

src/analysis/symbols.rs
```diff
@@ -0,0 +1,147 @@
...
+impl Info {
+    pub fn by_c_name(&self, name: &str) -> Option<&Symbol> {
+        self.c_name_index.get(name).map(|&id| &self.symbols[id as usize])
+    }
+
+    pub fn by_c_name_mut(&mut self, name: &str) -> Option<&mut Symbol> {
+        if let Some(&id) = self.c_name_index.get(name) {
+            Some(&mut self.symbols[id as usize])
+        }
+        else {
```
EPashkin: One line `} else {`?

src/analysis/symbols.rs
gkoz: Trying different ways of formatting this kind of assignments. Do you find the following preferable?

```rust
        let crate_name =
            if ns_id == namespaces::MAIN {
                None
            }
            else {
                Some(&namespaces[ns_id].crate_name)
            };
```

EPashkin: No. I like variable and its "value" in one line, but if it impossible, IMHO `if` or `match` better be on first line.

src/analysis/symbols.rs
gkoz: A large part of gir still uses the above style. I guess I never noticed the discrepancy. I might evaluate rustfmt again to avoid this kind of issues :)

src/analysis/symbols.rs
gkoz: Okay looks like this style

```rust
        let crate_name = if ns_id == namespaces::MAIN {
            None
        } else {
            Some(&namespaces[ns_id].crate_name)
        };
```
is all over rustc and libstd code so I should get used to it.

EPashkin:
Now docs look more rusty.
Only last signals like `AppChooserWidget::application-selected`. Not sure if it can be normally converted.

EPashkin:
Strange warning in https://ci.appveyor.com/project/GuillaumeGomez/gir/build/1.0.14/job/uvtcd89k19u1eh51#L298

gkoz:
> Only last signals like AppChooserWidget::application-selected. Not sure if it can be normally converted.

Indeed that's not clear yet.

> Strange warning

I imagine powershell reacts to `WARN`s this way but I don't care enough to dig into that.

EPashkin:
Thanks.

src/analysis/symbols.rs
```diff
@@ -0,0 +1,144 @@
...
+    c_name_index: HashMap<String, u32>,
+}
+
+pub fn run(library: &Library, namespaces: &namespaces::Info) -> Info {
+    let mut info = Info {
+        symbols: Vec::new(),
+        c_name_index: HashMap::new(),
+    };
+
+    info.insert("NULL", Symbol { name: "None".into(), ..Default::default() });
```
GuillaumeGomez: I didn't want to make this change because in a lot of cases, this conversion is invalid. I hope it won't create too much troubles. :-/

gkoz: It's a problem either way, but unlike `NULL`, `None` will be correct in a number of cases.

GuillaumeGomez: NULL was correct in all cases in C doc. :p

PR #215, opened 2016-02-19
Build script update

gkoz:
- Update pkg-config to 0.3.7.
- Let pkg-config compare version numbers.
- If pkg-config can't be run emit hardcoded library names.
- On Windows always emit hardcoded library names to avoid the `-lmm32` issue.


EPashkin:
Thanks.

PR #216, opened 2016-02-19
sys build script: avoid duplicate metadata when it's not suppressed

gkoz:
The `!hardcode_shared_libs` branch should not print anything because `pkg_config` does that.


EPashkin:
Thanks

PR #218, opened 2016-02-22
Switch to new rustdoc-stripper interface

gkoz:


gkoz:
Closing to update stripper a bit more.

PR #224, opened 2016-03-03
Docs: include Since and deprecated versions, `[Deprecated]` tags

gkoz:


src/analysis/info_base.rs
```diff
@@ -35,3 +40,29 @@ impl InfoBase {
...
+        .map(|f| f.deprecated_version)
+        .max()
+        .unwrap_or(None);
+    let fn_deprecated_min = functions.iter()
+        .filter(|f| f.visibility == Visibility::Public)
+        .map(|f| f.deprecated_version)
+        .min()
+        .unwrap_or(None);
+    let deprecated_version = deprecated_version.or(fn_deprecated_min.and(fn_deprecated_max));
+    trace!("{}: {:?}", obj.name, deprecated_version);
```
EPashkin: Still needed? If yes add some text inside

gkoz: Sorry, I'll remove this.

EPashkin:
Other looks fine. Thanks.
On results notable changes is adding version condition in mod.rs to `Popover`. 
Apparently before it was not due to lack of version `gtk_popover_get_pointing_to`.

gkoz:
Yeah, that's why I added the object version override, Popover and Gesture need that.

I wonder if I should put `Since:` and `# Deprecated` sections in functions before parameters.

EPashkin:
I don't understand why `Popover` need version override.
Also don't see any changes to `Gir_Gtk.toml`, so can you add version for `Popover` for reference?

gkoz:
You could say overriding `Popover` or `Popover::get_pointing_to` versions has the same effect. It seems simpler and somewhat more consistent with gtkdoc to override the former. But perhaps the proper way would be to overriding each offending function.

EPashkin:
I was missed that version condition was not added but removed.
No question then.
Thanks.

PR #228, opened 2016-03-21
Use more expressive object macro syntax for cross-crate inheritance

gkoz:
This implementation is hacky but should serve at least as a proof of concept.
https://github.com/gtk-rs/glib/pull/118


Gir_Gio.toml
```diff
@@ -0,0 +1,30 @@
+[options]
+girs_dir = "gir-files"
+library = "Gio"
+version = "2.0"
+min_cfg_version = "2.32"
+target_path = "."
```
EPashkin: #target_path = "./gio" plz

Gir_Gio.toml
```diff
@@ -0,0 +1,30 @@
+[options]
+girs_dir = "gir-files"
```
EPashkin: girs_dir = "../gir-files"

gkoz:
Looks like `git2` trips an LLVM assertion on current nightly, we can ignore that.

Gir_Gio.toml
```diff
@@ -0,0 +1,30 @@
...
+    "Gio.MenuAttributeIter",
+    "Gio.MenuItem",
+    "Gio.MenuLinkIter",
+    "Gio.MenuModel",
+    "Gio.Permission",
+    "Gio.Settings",
+    "Gio.SimpleAction",
+    "Gio.SimpleActionGroup",
+    "Gio.SimplePermission",
+]
```
EPashkin: Not sure if it works on Windows but
    "Gio.Drive",
    "Gio.Mount",
    "Gio.Volume",
    "Gio.VolumeMonitor",

Gir_Gio.toml
EPashkin: And `/gio` in `.gitignore`

Gir_Gio.toml
gkoz: At least they don't seem to require `gio-unix` so we could try them too. There are some thread safety issues though:

> GVolumeMonitor is not [thread-default-context aware](https://developer.gnome.org/glib/stable/glib-The-Main-Event-Loop.html#g-main-context-push-thread-default), and so should not be used other than from the main thread, with no thread-default-context active.

We don't have a strategy for dealing with that outside `gdk`/`gtk`. Could consider lifting the main_thread stuff to `glib` but here the limitations are documented in a cryptic way :(

gkoz:
Updated.

gkoz:
Now we don't add the `[]` if the simpler parents syntax is sufficient, and format the extended syntax better.

EPashkin:
That much better, thanks.

EPashkin:
Sorry, not seen that it ready for merging.

Issue #124, opened 2015-09-25
Generate bitfields, enums and consts reexports


gkoz:
The best modules organization is not entirely clear now. Possibly add modules `bitfileds`, `enums`, `consts` under `auto`.

Initially this is just a bunch of reexports like this:

``` rust
pub use ffi::GtkAccelFlags as AccelFlags;
```

If #123 is decided in favor of converting enums, the `enums` module will contain the enums definitions and conversion impls.

EPashkin:
This old issue that already done long ago

Issue #108, opened 2015-08-05
Enum idiosyncrasies


gkoz:
I'm going to collect fun enum-related stories here.
- Icon sizes of the supposed [GtkIconSize](https://developer.gnome.org/gtk3/stable/gtk3-Themeable-Stock-Images.html#GtkIconSize) type are passed around as `int`s because that enum isn't exhaustive. https://github.com/GNOME/gtk/commit/0b4af241b60cb63cccc010bc7c94ab2c97134730. The native (i.e. not the FFI one) Rust enum has to account for that. Unfortunately the annotations are attached to function parameters of that type not to the type itself.

EPashkin:
Can we just add to this enum next values: Other(int) and DontScale=-1?

gkoz:
Yes but in the automation context the more interesting question is if it's possible to detect such enums and divine the required additional variants. The "don't scale" case might be better reflected with `Option::<IconSize>::None`.

EPashkin:
After better looking I found that we need find solution only for 3 functions `gtk_icon_set_render_icon`, `gtk_style_real_set_background`, `gtk_widget_create_pango_layout` that accepts -1.
All other function accepts only registered values, so user can just store value returned by `get_icon_register` and use it in function calls.
Only possible problem with IconSize matching but IMHO it mostly unneeded or can be solved direct comparing with stored value.
For 3 problem function we really can use optional size parameter, but it can't be detected automatically and need some config.

EPashkin:
Updated previous comment.

EPashkin:
When checking for `<type name="gint" c:type="GtkIconSize"/>` found many
`<type name="gint" c:type="int"/>`
and only 2 other ginted ctypes: `cairo_format_t` and `pid_t`.
GUinted ctypes `uid_t` with some `unsigned` or `unsigned int`.

gkoz:
So this issue is extremely rare? Good. We could implement the conversions manually and let the generator assume they exist.

It's not clear why `cairo_format_t` is replaced with an `int`. Manually adding conversions to/from `int` would take care of that I expect.

`<type name="gint" c:type="int"/>` is weird but probably harmless.

gkoz:
So, the current assumption in the code that enums are passed directly is problematic in light of these exceptions.

What do we do with `GType`? There's `enum Type` in `glib` so the generator should ideally treat it as a converted enum as well.

GuillaumeGomez:
I think we can close this issue now.

PR #229, opened 2016-03-24
Generate bitflags and enums in high-level crates

gkoz:
Before generating any more crates I'd like to start generating enums.

Some things to decide there are:

* Do we want to list all generated enums in the config? We probably should either change "object" to something more generic or add `enumeration`, `record` etc.

* The `__Nonexhaustive` variant protects from breakage caused by adding variants. This implementation always adds it but we may want a different policy. Adding `__Nonexhaustive` everywhere could potentially be a serious breaking change for gtk.

* The working assumption here is that in the future the sys crates won't have any enums or bitflags, just `c_int` and `c_uint` newtypes or type aliases and constants. But perhaps not?

Relevant issues: #108, #124.


EPashkin:
IMHO we can safe use `object` in config as `object to process` not as `decendant of GObject` etc (as we do with interfaces and records), so separating unneeded.
Only if "subtype" need some unique parameters may be reason to separating.

src/codegen/enums.rs
```diff
@@ -0,0 +1,154 @@
...
+use ffi;
+use glib_ffi;
+use glib::error::ErrorDomain;
+use glib::translate::*;
+"#));
+
+        for (_, type_) in env.library.namespace_types(namespaces::MAIN) {
+            if let Type::Enumeration(ref enum_) = *type_ {
+                try!(generate_enum(env, w, enum_));
+                if let Some (cfg) = version_condition_string(env.config.min_cfg_version,
```
EPashkin: Too many calls `version_condition_string` with these parameters, maybe its time to add new function with parameters `(env: &Env, version: Option<Version>)`?

src/codegen/enums.rs
```diff
@@ -0,0 +1,154 @@
...
+use std::path::Path;
+use traits::*;
+use version::Version;
+
+pub fn generate(env: &Env, root_path: &Path, mod_rs: &mut Vec<String>) {
+    let path = root_path.join("enums.rs");
+    mod_rs.push("\nmod enums;".into());
+    file_saver::save_to_file(path, env.config.make_backup, |w| {
+        try!(general::start_comments(w, &env.config));
+        try!(writeln!(w, "{}", r#"
```
EPashkin: This really needed be raw string? For enums.rs changing all raw to normal seems has no effect on result file.
Also mixing multiline strings starting from beginning of line and not slightly confusing.

EPashkin:
Don't see problem in `__Nonexhaustive`.

gkoz:
I'll clean those things up. And will add status checks against config.

gkoz:
Updated. Not sure if it's possible to make multiline literals prettier.

Bitfields and enums are ignored by default now and need to be listed in the config.

src/codegen/enums.rs
```diff
@@ -0,0 +1,158 @@
...
+                    c.type_id.map_or(false, |tid| tid.ns_id == namespaces::MAIN)
+            });
+        let mut first = true;
+        for config in configs {
+            if let Type::Enumeration(ref enum_) = *env.library.type_(config.type_id.unwrap()) {
+                if first {
+                    mod_rs.push("\nmod enums;".into());
+                    first = false;
+                }
+                if let Some (cfg) = version_condition_string(env, enum_.version, false, 0) {
```
EPashkin: I thinking about `version_condition_string_default(env, enum_.version)` here, as too many `(... ,false, 0)` in project.

src/codegen/general.rs
```diff
@@ -86,19 +85,17 @@ pub fn define_boxed_type(w: &mut Write, type_name: &str, glib_name: &str,
...
+                         indent: usize) -> Result<()> {
+    if let Some(s) = version_condition_string(env, version, commented, indent) {
         try!(writeln!(w, "{}", s));
     }
     Ok(())
 }
 
-pub fn version_condition_string(min_cfg_version: Version,
-        version: Option<Version>, commented: bool, indent: usize) -> Option<String> {
+pub fn version_condition_string(env: &Env, version: Option<Version>, commented: bool, indent: usize)        -> Option<String> {
```
EPashkin: Too long

src/codegen/objects.rs
```diff
@@ -18,7 +18,7 @@ pub fn generate(env: &Env, root_path: &Path, mod_rs: &mut Vec<String>, traits: &
         let class_analysis = match info {
             Some(info) => info,
             None => {
-                warn!("Class or interface {} not found.", obj.name);
+                info!("Class or interface {} not found.", obj.name);
```
EPashkin: IMHO if config states that type must generated and no such type, then it sure warning and maybe ever error,
and better seen in run log and not when checking file list later.

src/codegen/enums.rs
```diff
@@ -0,0 +1,158 @@
...
+                enum_.name, member.name));
+        }
+        if has_failed_member {
+            try!(writeln!(w, "\t\t\t_ => Some({}::Failed),", enum_.name));
+        } else {
+            try!(writeln!(w, "\t\t\t_ => Some({}::__Nonexhaustive),", enum_.name));
+        }
+
+        try!(writeln!(w, "{}",
+"        }
```
EPashkin: `try!(writeln!(w, "\t\t}}\n\t}}\n}}\n\n"));` ?
Not sure if it really better.

src/codegen/enums.rs
gkoz: Overall it seems the `tabs` parameter is used a lot and other ways of adding tabs are going to be a bit ugly too. Factoring out `commented` feels like diminishing returns or I'm just not seeing a clean way.

src/codegen/enums.rs
gkoz: Too dense :)

EPashkin:
@gkoz Merge when ready

EPashkin:
When tried regenerate gtk after this `GesturePan` lose its `#[cfg(feature = "v3_14")]`

EPashkin:
Ignore this, I just don't generated `Gtk.Orientation`

PR #232, opened 2016-03-28
Use rust_type when generating `Foo::uninitialized()` chunks

gkoz:


EPashkin:
Nightbuild error just travis bug?
Thanks, please, merge yourself (not sure if it better do before 279)

gkoz:
Yeah, travis trips up on a bug with codegen-units > 1

PR #235, opened 2016-04-08
Add Shared (refcounted) records support

gkoz:


src/codegen/record.rs
```diff
@@ -11,9 +11,15 @@ pub fn generate(w: &mut Write, env: &Env, analysis: &analysis::record::Info) ->
     try!(general::start_comments(w, &env.config));
     try!(general::uses(w, env, &analysis.imports));
 
-    let copy_fn = analysis.specials.get(&Type::Copy).expect("No copy function for record");
-    let free_fn = analysis.specials.get(&Type::Free).expect("No free function for record");
-    try!(general::define_boxed_type(w, &analysis.name, &type_.c_type, copy_fn, free_fn));
+    if let (Some(ref_fn), Some(unref_fn)) = (analysis.specials.get(&Type::Ref),
+                                             analysis.specials.get(&Type::Unref)) {
+        try!(general::define_shared_type(w, &analysis.name, &type_.c_type, ref_fn, unref_fn));
+    } else if let (Some(copy_fn), Some(free_fn)) = (analysis.specials.get(&Type::Copy),
```
EPashkin: As I remember some records has all four function (can't find any now), may be move this condition first?

gkoz: If a shared struct has `copy` we unhide it (`record.rs:65`) to allow obtaining an independent copy. `free` should not make any difference.

gkoz: There is one struct, which could be used as both refcounted and owned, but I can't remember its name.

EPashkin: Ok

EPashkin:
Not adding generation `glib::Bytes` and `ResourceLookupFlags`?

gkoz:
Updated `Gir_gio.toml`

EPashkin:
Thanks

PR #245, opened 2016-04-21
Fix `+IsA<Object>` without active signals in gesture.rs

EPashkin:


EPashkin:
IMHO simplest way to apply this, add it to https://github.com/gtk-rs/gtk/pull/306

gkoz:
Thanks!

PR #247, opened 2016-04-23
Cows

EPashkin:
WIP.

Now `gtk` generated `2min on my old PC, so I start trying to faster it by remove unneeded copying.
Seems I started with wrong part as changes give no effect at all :(

### DIFF
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::collections::vec_deque::VecDeque;
 use std::slice::Iter;
 use std::vec::Vec;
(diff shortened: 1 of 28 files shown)


gkoz:
Yeah guessing what causes slowness isn't productive. You'd need to use a profiler. Not sure if there are tools that can profile binaries built for the `-gnu` target or if it's possible to use the `-msvc` target. 

There's a chance that most of the time is spent the XML parser. It could be an interesting experiment to serialize `Library` into a faster to parse format and cache that representation.

EPashkin:
Yes, only effect of this PR is increasing build time, so I postpone it and start adding statistics gatherer.

gkoz:
Running under callgrind shows this instruction number distribution suggesting that parsing does indeed dwarf everything else.
```
57,308,350  => config::Config::new
12,133,476,457  => library::Library::read_file
1,165,692  => library::Library::postprocessing
5,866,579  => analysis::class_hierarchy::run
486,227,927  => codegen::generate
51,075,239  => env::Env::drop
```
(everything else is under 1 million).

EPashkin:
So it just reading Library?

EPashkin:
By https://github.com/gtk-rs/gir/pull/248 it doing something other: loading only 5sec from 26 or 121

EPashkin:
This needed no more, closing

PR #248, opened 2016-04-23
Add collecting statistics

EPashkin:
WIP.

Only collecting and printing times of loading and generating.

```
Gir:
Total  PT26.231993243S   26231
Loading  PT5.116987983S   5116
Generating  PT21.114965659S   21114

GTK:
Total  PT121.552481879S   121552
Loading  PT5.209217861S   5209
Generating  PT116.343246907S   116343
```


gkoz:
Interesting results. I wonder if disk I/O (possibly swapping) contributes to the slowdown. here's what I get with gtk:
```
Total  PT3.058763024S   3058
Loading  PT2.820055179S   2820
Generating  PT0.238701993S   238
```
At least this is consistent with my cachegrind results.

The API might look nicer with a RAII approach, e.g.
```rust
{
    let _watch = statistics.start(SWType::Type)
    //...
}

//or

let watch = statistics.start(SWType::Type)
//...
drop(watch);
```

EPashkin:
Agree about RAII

EPashkin:
May be you right about swapping:
```
D:\eap\rust\0\gir>time cargo run --release -- -c D:\eap\rust\0\gir\Gir_Gtk.toml -o D:\eap\rust\0\gir\gtk
     Running `D:\eap\rust\0\./target\release\gir.exe -c D:\eap\rust\0\gir\Gir_Gtk.toml -o D:\eap\rust\0\gir\gtk`
Total PT25.941996539S                25941
Loading PT5.137561408S                 5137
Generating PT20.804421931S                20804
0.00user 0.00system 0:26.77elapsed 0%CPU (0avgtext+0avgdata 208128maxresident)k
0inputs+0outputs (847major+0minor)pagefaults 0swaps
```

EPashkin:
I wondered about huge difference in generating times (116 vs 0.2) with normal on loading (5 vs 3) so I run generation on Ubuntu on same computer and get:
```
Total                      PT5.588866285S                 5588
Loading                    PT4.741487931S                 4741
Generating                 PT0.847373953S                  847
```
Not sure is problem in rust or your code or just windows (with antivirus).

gkoz:
I confirm that 'Generating' seems to take 100x more time on windows, though in my case some of it could be blamed on SSD vs HDD.

EPashkin:
On virtual boxed ubuntu on same folder (shared):
```
Total                     PT15.911299623S                15911
Loading                    PT6.076296568S                 6076
Generating                 PT9.834947744S                 9834
```

EPashkin:
So it rust on windows problem.
I check with `msvc` version to be sure.

gkoz:
https://github.com/gtk-rs/gir/blob/master/src/codegen/general.rs#L16
Its repeated calls into git that slow the generator down.

EPashkin:
Thanks, I starting to fix it

EPashkin:
.. or you already almost done fix, @gkoz ?

EPashkin:
My results after fix in #249 :tada:
```
Total                      PT6.481996734S                 6481
Loading                    PT5.963716757S                 5963
Generating                 PT0.518267755S                  518
```
 @gkoz Big thanks for fast finding source

EPashkin:
Ready for merge, I currently don't plan collect detailed statistic as problem solved.

src/main.rs
```diff
@@ -75,7 +84,17 @@ fn do_main() -> Result<(), Box<Error>> {
...
+    drop(watcher_loading);
+    {
+        let _watcher_generating = statistics.start(SWType::Generating);
+
+        codegen::generate(&env);
+    }
+
+    drop(watcher_total);
+
+    statistics.print();
```
gkoz: I'd rather we made this optional either via an command-line flag or logging via `info!`.

src/statistics.rs
```diff
@@ -0,0 +1,72 @@
...
+iterated_enum! {
+    SWType ; 3 => [
+        Total,
+        Loading,
+        Generating
+    ]
+}
+
+pub struct Watcher {
+    stopwatch: Rc<RefCell<Stopwatch>>,
```
gkoz: Since `Stopwatch` is `Copy` it can be put in a `Cell`.

src/main.rs
EPashkin: I change to `info!` then

src/statistics.rs
EPashkin: k

gkoz: Or rather it would make sense if the stopwatch crate had marked its methods as `inline`. As it is `RefCell` could even be better. Overall this doesn't affect performance anyway...

src/main.rs
EPashkin: Seems `info!` is too spammed with `cargo` I better change to command-line flag.

src/statistics.rs
EPashkin: With `Cell` I need get(), stop(), set(). IMHO `RefCell` better.

gkoz: Okay let's keep it as it is.

EPashkin:
Added command-line flag

gkoz:
> @gkoz Big thanks for fast finding source 

I haven't discovered a way to run a profiler though :( But tracing with `procmon` was enough to notice that it keeps rereading git-related files. There might be a bug in `libgit2`, that makes it so much slower on Windows.

EPashkin:
And I have `core.preloadindex = true, core.fscache = true` one of them IMHO run many git copies when checks index.

EPashkin:
AppVeyor build succeeded

gkoz:
Relevant: https://crates.io/crates/hprof

PR #250, opened 2016-04-25
Add used types from signal return

EPashkin:
Only affect generation `FileChooser::confirm-override`, all other returning types was lucky and used somewhere else in this module.


EPashkin:
AppVeyor build succeeded

EPashkin:
I add it to https://github.com/gtk-rs/gtk/pull/319

PR #251, opened 2016-04-25
Make cast generated enums to i32 non-scalar

EPashkin:


EPashkin:
Its second part of `enum as i32`. See https://github.com/gtk-rs/examples/pull/88

EPashkin:
AppVeyor build succeeded, @gkoz, please, merge this

EPashkin:
Seems I missread matches, need more `(_)`

EPashkin:
Or not?

EPashkin:
I tried micro-optimisation on match.

gkoz:
```
__Nonexhaustive(_) => panic!(),
```
This got me thinking: if we somehow get an unexpected variant from the library, should it be possible to pass it back unchanged? This would require complicated workarounds and making enums much fatter.

EPashkin:
We can add flag for some enums, that convert it to `__Nonexhaustive(i32)` on `from_glib` and use it in `to_glib`

EPashkin:
These enums still defended again direct converting, so I don't see any problems.

gkoz:
Problem is we can't allow putting arbitrary `i32`s there and that's where complicated workarounds come in.

EPashkin:
Sorry, don't understand :(

gkoz:
Type safety and (likely memory safety too) require that a user can't just put any random value in that variant and also can't copy it from one enum into another, so it'd be along the lines of

```rust
pub struct Discriminant<T> {
    value: i32,
    _dummy: PhantomData<*const T>,
};

impl<T> struct Discriminant<T> {
    // there shouldn't be a safe way to put arbitrary discriminants into enums
    pub unsafe fn new(value: i32) -> Self {
        Discriminant {
            value: value,
            _dummy: PhantomData::new(),
        }
    }
}

pub enum Foo {
    Member1,
    Member2,
    Member3,
    __Nonexhaustive(Discriminant<Foo>),
}
```

This looks like it's not worth the trouble at this point.

EPashkin:
IMHO it not much related to this PR ;)
Currently we safe from direct converting `enum <-> int` (not checked `int -> enum` but `FromPrimitive` trait seems discarded),
way make defence against `enum1 -> int -> enum2` you show IMHO not needed until some example arises.

PR #253, opened 2016-04-30
Deprecate functions and signals that deprecated by minimal version

EPashkin:


EPashkin:
In https://github.com/gtk-rs/gtk/pull/326 I found new "bug" with pattern: some functions was wrongly ignored in "auto/widget.rs".
While I strongly don't like it, may be we need add "^" and "$" to pattern in code.
@gkoz, @GuillaumeGomez  what you think

gkoz:
I agree that the lack of anchors is a hazard so it's safer to force-add them.

GuillaumeGomez:
I agree with @gkoz.

EPashkin:
Added anchors

gkoz:
Thanks!

Issue #241, opened 2016-04-21
Signals with optional parameters


EPashkin:
Part of #199 
ListBox: row-selected
Widget: hierarchy-changed, parent-set

For closure `Fn(&ListBox, Option<&ListBoxRow>) + 'static>` generated generated bad code.
Not sure if it need be fixed in `gir` or in `glib\gtk`.
```
unsafe extern "C" fn row_selected_trampoline(this: *mut GtkListBox, row: *mut GtkListBoxRow, f: gpointer) {
    callback_guard!();
    let f: &Box_<Fn(&ListBox, Option<&ListBoxRow>) + 'static> = transmute(f);
    f(&from_glib_none(this), Option<ListBoxRow>::from_glib_none(row))
}
```

EPashkin:
Added partial fix

EPashkin:
Seems gir generated working code, I close this.

PR #255, opened 2016-05-03
Fix signal parameters with Options

EPashkin:
Not final fix of #241.
Changes `Option<&ListBoxRow>` to `&Option<ListBoxRow>` in signal closure


gkoz:
:+1:

Issue #244, opened 2016-04-21
Signals returning Inhibit as gboolean


EPashkin:
# Closures for these must return `Inhibit` not `bool`.

RangeExt::change-value

gkoz: It appears a significant number of signals need Inhibit, with varying degrees of certainty:
- `EntryCompletion`
- `Label::activate-link` (?)
- `LinkButton::activate-link`
- `MenuShell::move-selected`
- `Notebook::change-current-page` and others (??)
- `Paned` (?)
- `SpinButton::output` (?)
- `StatusIcon::size-changed`
- `Switch::state-set`
- `ToolItem::create-menu-proxy`
- `Toolbar`

Signals with `Inhibit` detected by `_gtk_boolean_handled_accumulator` in `gtk` sources
~~`AboutDialog`: activate-link~~
`AccelGroup`: accel-activate (not meant to be used by applications)
~~`EntryCompletion`
insert-prefix
match-selected
cursor-on-match
`GLArea`: render (already manual with Inhibit)~~
`IMContext`
retrieve-surrounding
delete-surrounding
~~`Label`: activate-link
`LinkButton`: activate-link
`MenuShell`: move-selected~~
`Overlay`: get-child-position (output)
`PrintOperation`
paginate (?)
preview
~~`Range`: change-value (already manual with Inhibit)~~
`Socket`: plug-removed
`SpinButton`: output (already manual with Inhibit)
~~`Switch`: state-set
`TextTag`: event
`TextView`: extend-selection
`Toolbar`: popup-context-menu
`ToolItem`: create-menu-proxy
`TreeView` all already manual but not Inhibit
test-expand-row
test-collapse-row~~
`Widget` (+ already manual with Inhibit, - not added, some added but return `bool`)
~~draw +
mnemonic-activate +
focus +
keynav-failed +
event +
button-press-event +
button-release-event +
touch-event + (?)
scroll-event +
motion-notify-event +
delete-event +
destroy-event +
key-press-event +
key-release-event +
enter-notify-event +
leave-notify-event +
configure-event +
focus-in-event +
focus-out-event +
map-event +
unmap-event +
property-notify-event +
selection-clear-event - (+)
selection-request-event - (+)
selection-notify-event - (+)
proximity-in-event +
proximity-out-event +
drag-failed - (+)
drag-motion - (-)
drag-drop - (-)
visibility-notify-event - (+)
window-state-event +
damage-event +
grab-broken-event +
query-tooltip currently not Inhibit
[...]

PR #256, opened 2016-05-11
Inhibit

EPashkin:
Part of #244

### DIFF
--- a/Gir_Gtk.toml
+++ b/Gir_Gtk.toml
@@ -50,7 +50,6 @@ generate = [
     "Gtk.IconInfo",
     "Gtk.Image",
     "Gtk.InfoBar",
-    "Gtk.Label",
     "Gtk.Layout",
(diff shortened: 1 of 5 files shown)


EPashkin:
Applying this to `gtk` - https://github.com/gtk-rs/gtk/pull/338
@gkoz can you check issue for mistakes. IMHO we can add fixing of `Widget::show-help` (https://github.com/gtk-rs/gir/issues/244#issuecomment-217665868) to that PR.

gkoz:
Thanks!
It wasn't obvious to me before but clearly this feature should be temporary, to be subsumed by general types rewriting (`i32` -> `u32`, `bool` -> `Inhibit`, `char*` -> `TreePath` etc.)

EPashkin:
There is nothing more permanent than a temporary patch :cry:
But I hope that we will come up with how to handle it properly.

PR #257, opened 2016-05-12
Allow union parameters

EPashkin:
Currently only enabled union in `Gtk` is `Event`


EPashkin:
Part of https://github.com/gtk-rs/gtk/pull/339 and #244

PR #258, opened 2016-05-14
Detect union ctype

EPashkin:
For https://github.com/gtk-rs/gtk/pull/340

### DIFF
--- a/src/library_postprocessing.rs
+++ b/src/library_postprocessing.rs
@@ -124,18 +124,25 @@ impl Library {
     }
 
     fn c_type_by_type_id(&self, tid: TypeId) -> Option<String> {
-        use library::Type::*;
         let type_ = self.type_(tid);
-        let glib_name = type_.get_glib_name();
-        if glib_name.is_none() { return None; }
-        let glib_name = glib_name.unwrap();
-        let detected_c_type = match *type_ {
+        type_.get_glib_name().map(|glib_name| {
+            if self.is_referenced_type(type_) {
+                format!("{}*", glib_name)
+            } else {
+                glib_name.to_string()
+            }
+        })
+    }
+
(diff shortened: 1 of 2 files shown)


src/library_postprocessing.rs
```diff
@@ -130,6 +130,17 @@ impl Library {
         if glib_name.is_none() { return None; }
         let glib_name = glib_name.unwrap();
         let detected_c_type = match *type_ {
+            Alias(ref alias) => {
+                if let Some(aliased) = self.c_type_by_type_id(alias.typ) {
+                    if aliased.find("*").is_some() {
+                        format!("{}*", glib_name)
```
gkoz: What's going to happen if it's something different from `TypeName*` (perhaps a const pointer)?

EPashkin: It called recursively and I don't see any other cases as other data used.

EPashkin: * as no other data used

EPashkin: I can extract from this function `fn is_pointer(...)->bool` to make logic clearer

EPashkin:
Simplified logic

src/library_postprocessing.rs
```diff
@@ -124,18 +124,26 @@ impl Library {
     }
 
     fn c_type_by_type_id(&self, tid: TypeId) -> Option<String> {
-        use library::Type::*;
         let type_ = self.type_(tid);
         let glib_name = type_.get_glib_name();
```
gkoz: Now I understand. Also after splitting out `is_referenced_type` this bit can become simpler:
```rust
type_.get_glib_name().map(|glib_name| {
...
```

EPashkin: You right
Fixed

gkoz:
Thanks.

Issue #259, opened 2016-05-15
Cannot regenerate gtk files


GuillaumeGomez:
I updated gir to last master branch update. When I run:

```shell
cargo run --release -- -c ../gtk/Gir.toml -d ../gir-files -o gtk
```

I get the following error:

```shell
"../gtk/Gir.toml": 452:6-452:6 error: expected a newline after table definition
```

Any idea?

PS: I also had to change the line 63 in `src/statistics.rs` because Duration doesn't implement `Display` trait (so I added `.as_secs()`).

EPashkin:
Can you push your gtk master?

GuillaumeGomez:
It's the same than gtk-rs/gtk. That's why I'm surprised.

EPashkin:
Confirmed that after `cargo update` `std::time::Duration: std::fmt::Display` don't have display

GuillaumeGomez:
Do you want me to open a PR for this change?

EPashkin:
If you can.
Seems toml also changed parser, so comment need moved
```
 [[object]]
 name = "Gtk.Editable"
 status = "generate"
-    [[object.signal]] #out #manual
-    pattern = ".+"
-    ignore = true
+[[object.signal]]
+pattern = ".+"
+ignore = true #out #manual
```

GuillaumeGomez:
Ok, I'll open a PR for gtk as well.

EPashkin:
Whitespace change in my diff was mistake.

PR #260, opened 2016-05-15
Fix Display error for Duration

GuillaumeGomez:


GuillaumeGomez:
Fixes #259

src/statistics.rs
```diff
@@ -60,7 +60,7 @@ impl Statistics {
         for typ in SWType::iter() {
             if let Some(sw) = self.stopwatches.get(typ) {
                 let sw = sw.borrow();
-                let elapsed = sw.elapsed();
+                let elapsed = sw.elapsed().as_secs();
```
EPashkin: IMHO better completely remove this column, only mssec is enough

GuillaumeGomez:
Updated.

EPashkin:
:+1:

GuillaumeGomez:
I let you handle the merge then. :wink:

gkoz:
So you've found one more reason to not run `cargo update` randomly? ;)
I'd hold off modifying any `Gir.toml`s for now, because I'm not convinced `[table] # comment` actually is invalid.

GuillaumeGomez:
> So you've found one more reason to not run cargo update randomly? ;)

:-°

GuillaumeGomez:
I actually did because of the compiler error I got. Not a wisely choice... I'm removing the Cargo.lock change.

GuillaumeGomez:
Updated.

EPashkin:
IMHO without `cargo.lock` this PR has no meaning

GuillaumeGomez:
:cry:

GuillaumeGomez:
So what do I do?

EPashkin:
Removing `Display` was introduced yesterday in https://github.com/ellisonch/rust-stopwatch/commit/1ffee80fc41a70836e85a8f4f614510c831d78be we just was "lucky" and find it immediately.

EPashkin:
Toml changed month ago https://github.com/alexcrichton/toml-rs/commit/848227c6053019918b726fa7356f5dfb87861420

gkoz:
The toml change was merged only 3 days ago. As to statistics, we ever need to touch that code again, I'd be in favor of just using the `hprof` crate.

EPashkin:
About toml, IMHO in https://github.com/alexcrichton/toml-rs/commit/2cd1116e062f9af389f97044360a9cfbdb127c82 is good reason to restriction

EPashkin:
As we already merged toml-related fix we better forget about it, adding Issue or PR to toml don't change that fact.
So currently we has only these options:
1) Postpone solution,
2) Do minor fix to work with stopwatch 0.0.7 
3) Replace stopwatch with hprof (gkoz suggested)
@GuillaumeGomez If you will do 3th in this PR it be best, but 1) and 2) also good for me

EPashkin:
*both 2 and 3 options must include `Cargo.lock` changes.

GuillaumeGomez:
Please choose and I'll will update in this way.

EPashkin:
As I understand @gkoz stated that him don't like 2 and prefer 3.
I also stated my option.
So final choice up to you, @GuillaumeGomez, if you have time to apply `hprof` then go 3, else close this PR (1).

GuillaumeGomez:
I'll take option 3 then. Sorry @EPashkin :wink:

GuillaumeGomez:
Updated.

src/main.rs
```diff
@@ -63,10 +62,11 @@ fn do_main() -> Result<(), Box<Error>> {
         Err(err) => return Err(Box::new(err)),
     };
 
-    let mut statistics = Statistics::new();
+    let statistics = Profiler::new("gir profiler");
```
EPashkin: Just "Gir"?

src/config/members.rs
```diff
@@ -1,5 +1,5 @@
 use super::ident::Ident;
-use super::identables::{Identables, Parse};
+use super::identables::Parse;
```
EPashkin: Thank for notice that.

src/main.rs
GuillaumeGomez: Sure? It isn't very beautiful. I agree for another but just "Gir"? :-/

EPashkin:
Maybe need call `statistics.disable();` after loading config if it disabled.

EPashkin:
* or just don't do `start_frame`

src/main.rs
EPashkin: "Timing information for gir profiler:" IMHO no good,
prefer "Timing information for Gir:"

GuillaumeGomez: Fine by me.

EPashkin:
May be better remove `watcher_total`

GuillaumeGomez:
> May be better remove watcher_total

It's "convenient": you don't have to make the addition yourself. :D

src/main.rs
EPashkin: IMHO it `tautology`

GuillaumeGomez:
> Maybe need call `statistics.disable();` after loading config if it disabled.

It doesn't impact performance so why bother?

GuillaumeGomez:
So I changed the message. Anything else?

EPashkin:
Thanks, others was just possibilities.

EPashkin:
I leave final decision to @gkoz

EPashkin:
Strange `cargo update` still changes `Cargo.lock`.

EPashkin:
Sorry, I though that you update it.

GuillaumeGomez:
I thought I did as well. Sorry, updating.

GuillaumeGomez:
Ok, now it has been updated.

EPashkin:
:+1:

gkoz:
Looks good. I guess we need to bump rust version in appveyor.

GuillaumeGomez:
Done as well.

gkoz:
Thanks

PR #261, opened 2016-05-16
Generate urls (when possible)

GuillaumeGomez:


EPashkin:
@GuillaumeGomez, if diffs line this is really what you want, then all good. But IMHO link in markdown looks like `[I'm an inline-style link](https://www.google.com)`
```diff
<!-- struct AboutDialog -->
-The `AboutDialog` offers a simple way to display information about
+The `./struct.AboutDialog.html` offers a simple way to display information about
 a program like its logo, name, copyright, website and license. It is
 also possible to give credits to the authors, documenters, translators
 ```

**Comment by GuillaumeGomez** [comment]
I forgot to commit/squash/push. Sorry about that. Updated.

**Comment by EPashkin** [comment]
Current version also not right. Ticks go inside brackets. I make temporary branch for illustrate: https://github.com/gtk-rs/gir/tree/tmp

**Comment by GuillaumeGomez** [comment]
Hum, right. I didn't pay attention to backticks.

**Comment by GuillaumeGomez** [comment]
Fixed.

File: src/codegen/doc/format.rs [hunk 1]
```diff
@@ -82,20 +84,45 @@ fn format(mut input: &str, symbols: &symbols::Info) -> String {
     }
 }
 
-fn replace_c_types(entry: &str, symbols: &symbols::Info) -> String {
+fn to_url(type_name: &str, refs: &TypeReferences) -> Option<String> {
+    let ty = type_name.split("::").collect::<Vec<&str>>();
+
+    if let Some(t) = refs.get_type(&ty.last().unwrap()) {
+        Some(format!("./{}.{}.html", t.ty, t.name))
```
gkoz: Is there a reason for having the `./` bit?

GuillaumeGomez: It makes the fact that this is a path more obvious. It's in prevision for future inter-crate update.

gkoz:
I'm getting redundant `Ext`s like this with gtk:
```patch
-Use `BoxExt::set_child_packing` to reset the expand,
+Use `BoxExtExt::set_child_packing` to reset the expand,
```

src/codegen/doc/format.rs
gkoz: What else if not a relative path could start with letters?

GuillaumeGomez: Fine, I remove it.

gkoz:
```patch
- `Box:spacing` property. If `child` is a widget at one of the
+ [`Box:spacing`](./struct.Box.html) property. If `child` is a widget at one of the
```
I'm not sure it's a good idea to link property and signal references to just the struct's page. There'd be expectation that they lead somewhere useful.

src/codegen/doc/format.rs
```diff
@@ -82,20 +84,45 @@ fn format(mut input: &str, symbols: &symbols::Info) -> String {
     }
 }
 
-fn replace_c_types(entry: &str, symbols: &symbols::Info) -> String {
+fn to_url(type_name: &str, refs: &TypeReferences) -> Option<String> {
+    let ty = type_name.split("::").collect::<Vec<&str>>();
```
gkoz: If you just want the last item there's no need to collect.

src/codegen/doc/mod.rs
```diff
@@ -44,46 +44,115 @@ pub fn generate(env: &Env) {
...
+
+    pub fn get_type(&self, type_name: &str) -> Option<TypeStruct> {
+        for ty in self.refs.iter() {
+            if &ty.name == type_name {
+                return Some(ty.clone())
+            }
+        }
+        None
+    }
+}
```
gkoz: `HashMap`?

GuillaumeGomez: That's what I did at first. Not so sure what the better solution is. A `HashMap` would duplicate data.

src/codegen/doc/mod.rs
```diff
@@ -44,46 +44,115 @@ pub fn generate(env: &Env) {
...
+
+fn create_references(env: &Env) -> TypeReferences {
+    let mut refs = TypeReferences::new();
+
+    for (tid, type_) in env.library.namespace_types(MAIN) {
+        if let Some(obj) = env.config.objects.get(&tid.full_name(&env.library))
+            .and_then(|obj| if obj.status.ignored() { None } else { Some(obj) }) {
+            for ty in match *type_ {
+                LType::Class(..) => {
+                    let ty = analysis::object::class(env, obj).unwrap();
```
gkoz: It'd be nice to find a way to not run analysis many times.

GuillaumeGomez: Agreed. It's a big time loss (and it doubles the output).

src/codegen/doc/mod.rs
```diff
@@ -44,46 +44,115 @@ pub fn generate(env: &Env) {
...
+                    vec!(TypeStruct::new(SType::Struct, &ty.name),
+                         TypeStruct::new(SType::Trait, &format!("{}Ext", &ty.name)))
+                }
+                LType::Interface(..) => {
+                    let ty = analysis::object::interface(env, obj).unwrap();
+                    vec!(TypeStruct::new(SType::Struct, &ty.name),
+                         TypeStruct::new(SType::Trait, &format!("{}Ext", &ty.name)))
+                }
+                LType::Record(..) => {
+                    let type_id = analysis::record::new(env, obj).unwrap().type_id;
```
gkoz: This analysis run doesn't seem to do anything useful.

src/codegen/doc/mod.rs
GuillaumeGomez: Actually we could just loop over my generated vec and it'd be all good I guess.

src/codegen/doc/mod.rs
gkoz: Performing a linear search of the same growing vector over and over again, i.e. N^2 over the total number of eligible types in the library gives me a very bad vibe.

gkoz:
Generally, are you sure it'd not be better to expand `analysis::symbols` instead? There seem to be duplication of concerns in the making (I assume you'll want to process methods too).

src/codegen/doc/mod.rs
GuillaumeGomez: So let's throw memory in fire and handle all this with `HasMap`! :D

GuillaumeGomez:
This is a first shot. I have a very bad understanding of all the underlying mechanisms, which lead to the code you can see now.

gkoz:
The symbols module only tracks names now but it could also store types. It could be difficult to make it play nice with the struct/trait dichotomy. But in return you'd get the rest function handling for free.

GuillaumeGomez:
I think the "ExtExt" issue. I guess running the analysis twice created it. You'll certainly find out better than I.

EPashkin:
IMHO this link need hashtag (when signals added)
```
[`AboutDialog::activate-link`](struct.AboutDialog.html)
```
Also don't see reason to having link to `struct.AboutDialog.html` in same file.

EPashkin:
Possible link to wrong page
```
[`gio::Application::startup`](struct.Application.html) signal
````
Only found other possible hashtagable is properties (very few, ex. Label) and `Widget::draw` that can be signal or function.

Other wrong link: must link to `TreeViewColumnSizing` (if we have separate page for enum)
```
[`TreeViewColumnSizing::Fixed`](struct.Fixed.html)
```

gkoz:
I've taken another look at `analysis::symbols`, it should be possible to make it track additional type information. There already is a [provision for dealing with trait methods](https://github.com/gtk-rs/gir/blob/master/src/analysis/object.rs#L110).

@EPashkin I think we've come to a point when it's necessary to make a separate analysis pass, so the codegen could just look up analysis results.

EPashkin:
@gkoz Agree

gkoz:
After #267 classes and interfaces will get analyzed by the time codegen is run ensuring the trait methods get fixed up. So moving forward with this would be to make `analysis::symbols` aware of item types and teach it to emit relative paths for links.

PR #265, opened 2016-08-20
Fix imports from glib when generate glib

EPashkin:
Fix use `glib::` errors in https://github.com/gtk-rs/glib/issues/135 and https://github.com/gtk-rs/glib/pull/136


GuillaumeGomez:
Sounds good to me! :+1:

gkoz:
Nice one, thanks!

PR #266, opened 2016-08-21
Proper ffi namespace use in glib

hfiguiere:
Fix the ffi namespace use in glib. See PR gtk-rs/glib#136 for some of the details.


EPashkin:
:+1:

hfiguiere:
thanks a lot !

Issue #136, opened 2015-10-03
Change analysis into proper passes and rearrange data


gkoz:
Calling only the analyzers you need from codegen seems limiting and requires workarounds to avoid duplicate work.

A more robust scheme would be to do the analysis first and then let the codegen use that data. Expected benefits:
 * parser cleanup, by moving the patching logic to a separate stage,
 * library cleanup, by removing the bits of analysis from it,
 * analysis stages can rely on the data from earlier ones and agree on viral blacklisting (commenting out) effects,
 * possibly, weaker coupling between stages.

This is just a first approximation, I've not thought about the ways to organize the data. `TypeId` would probably turn into `DefId`. We could probably learn a thing or two from `rustc`.

EPashkin:
Good long-term objective, but I do not have any ideas how to do it.

gkoz:
In the (still in progress) `ffi_ng` branch I'm translating the "gir" representation created by the parser into the "foreign" one. The "foreign" analysis is going to be responsible for applying quirks, creating helper types for unions, tracking visibility and other contagious properties, generating rust FFI type strings etc. The generator just has to iterate through the definitions and do some `format!`s.

gkoz:
I think now's a good time to try implementing some changes to this effect. I intend to continue work on the `ffi_ng` branch soon.

PR #267, opened 2016-08-22
Add a separate analysis pass

### LINKED ISSUE #252: Incorrect trait name in doc symbol dictionary
```patch
-To set the value of a child property, use `ContainerExt::child_set_property`,
+To set the value of a child property, use `Container::child_set_property`,
 `ContainerExt::child_set` or `ContainerExt::child_set_valist`.
 To obtain the value of a child property, use
-`ContainerExt::child_get_property`, `ContainerExt::child_get` or
+`Container::child_get_property`, `ContainerExt::child_get` or
```

gkoz:
Objects (classes and interfaces) and records ('generated' and 'manual' ones from all namespaces) get analyzed prior to generation. The `Info`s are collected into vectors with no indexing by name yet.
This partially implements #136. Fixes #252.

### DIFF
--- a/src/codegen/doc/mod.rs
+++ b/src/codegen/doc/mod.rs
@@ -46,41 +46,44 @@ pub fn generate(env: &Env) {
 
 fn generate_doc(mut w: &mut Write, env: &Env) -> Result<()> {
     try!(write_file_name(w, None));
+    let mut generators: Vec<(&str, Box<Fn(&mut Write, &Env) -> Result<()>>)> = Vec::new();
 
-    for (tid, type_) in env.library.namespace_types(MAIN) {
-        let obj = env.config.objects.get(&tid.full_name(&env.library))
-            .and_then(|obj| if obj.status.ignored() { None } else { Some(obj) });
-        match *type_ {
-            LType::Class(..) => {
-                if let Some(obj) = obj {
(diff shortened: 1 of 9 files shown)


EPashkin:
:+1:  Thanks

src/codegen/doc/mod.rs
```diff
@@ -46,41 +47,49 @@ pub fn generate(env: &Env) {
 
 fn generate_doc(mut w: &mut Write, env: &Env) -> Result<()> {
     try!(write_file_name(w, None));
+    let mut generators: Vec<(&str, Box<Fn(&mut Write, &Env) -> Result<()>>)> = Vec::new();
+
+    let ignored = |info: &InfoBase| {
+        info.type_id.ns_id != MAIN ||
+            env.config.objects.get(&info.full_name).map_or(true, |obj| obj.status.ignored())
```
EPashkin: only needed `!= MAIN`

gkoz: I'm assuming we don't want to emit docs for items which are ignored and thus absent from the crate.

EPashkin: This items ignored on analysis phase and not get to `objects` or `records`

gkoz: Yeah, good point :)

PR #270, opened 2016-08-31
Add single digit version variant

### LINKED ISSUE #269: version::Version fail with single digit version
If you have this in the Gir.toml
````
min_cfg_version = "230"
````

The build.rs generated contain:
````
let version = {
        "230.0"
};
````

Which in the end fails with pkg-config.
````
$ pkg-config --modversion gudev-1.0
230
````

The build output is:
````
error: failed to run custom build command for `gudev-sys v0.1.0 (file:///home/hub/source/gudev-rs/gudev-sys)`
process didn't exit successfully: `/home/hub/source/gudev-rs/gudev-sys/target/debug/build/gudev-sys-ed1f2ccef31fd523/build-script-build` (exit code: 1)
--- stderr
`"pkg-config" "--libs" "--cflags" "gudev-1.0 >= 230.0"` did not exit successfully: exit code: 1
--- stdout
Requested 'gudev-1.0 >= 230.0' but version of gudev-1.0 is 230
````

I tracked it all down to ````src/version.rs````.

This is the sample repository:
https://github.com/hfiguiere/gudev-rs

### PR DESCRIPTION
**Pull request by EPashkin**
Possible fix #269

### CODE REVIEW

**Comment by gkoz** [comment]
Would it be cleaner to make `Version` an `enum { Full(u16, u16, u16), Short(u16) }`?

**Comment by EPashkin** [comment]
Maybe, but how "transfer" info about preferred type?
By type of  parsed `min_cfg_version` value?

**Comment by gkoz** [comment]
We could just choose the format a the time of parsing a version. No dots => `Short`.

**Comment by EPashkin** [comment]
I mean, fill `VersionMode` mode by result of parsing `min_cfg_version` or
remove `VersionMode` and pass `min_cfg_version` to `to_cfg` as format sample?

**Comment by gkoz** [comment]
I think `VersionMode` can be avoided, essentially this line is the direct cause of the issue and it would just format differently

```rust
try!(writeln!(w, "{{\n\t\t\"{}\"\n\t}};", env.config.min_cfg_version));
```

gkoz:
Edited ^

EPashkin:
So don't change feature and configs on short version?

gkoz:
I believe it will work out with just updated formatting and ordering in `mod version`. This library's gir has `version="165"` attributes. If you change `230` to something lower I guess it should be apparent whether that works.

EPashkin:
Reworked. Result seems good, but I don't have `gudev` package and can't test building.

hfiguiere:
it works great here. Thanks a lot !

gkoz:
We probably ought to handle ordering more carefully. How are `Full(230, 0, 0)` and `Short(230)` supposed to compare?
`pkg-config` doesn't appear to have a sensible interpretation of a single-number version:  `229.999 <= 230` and `230 < 230.`.

EPashkin:
IMHO ordering between `Full` and `Short` doesn't matter as in one package seems used only one type.

gkoz:
Okay, we can look into keeping just partial ordering later. Thanks!

PR #274, opened 2016-09-15
Don't return from used_ffi_type basic types

### LINKED ISSUE #272: Trying to import u64 when generating signals
If a signal signature has a `u64` gir may try to `use u64;` and break the build.
https://github.com/gtk-rs/gtk/issues/381#issuecomment-247316925

EPashkin:
Closes #272 
It get no changes to current gtk, but not sure that it good.
Maybe better add comparsion to list ["i8", "u8", ..."] from `ffi_inner`,
or change same string from `ffi_inner` to something with ':'


gkoz:
Seems reasonable, thanks.

PR #275, opened 2016-09-24
Enums primary detects error domain by function "quark"

EPashkin:
For https://github.com/gtk-rs/gtk/issues/381#issuecomment-247492968

### DIFF
--- a/src/codegen/enums.rs
+++ b/src/codegen/enums.rs
@@ -131,8 +131,13 @@ impl FromGlib<ffi::{ffi_name}> for {name} {{
     }
 }
 "));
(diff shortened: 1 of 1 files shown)


gkoz:
I was hoping we could automate it: add `quark` to "special" functions then if there's a `quark` function and `glib:error-domain` is present set its value to that function's name.

EPashkin:
I try do it

EPashkin:
@gkoz thanks for idea, it works fine

gkoz:
Huh, it turned out more elegant than I expected. Thanks!

PR #276, opened 2016-10-11
Parse fields in classes and generate correct class structs

eugene2k:
This pull request makes it possible for the generator to generate proper class structs, which is very useful for creating custom widgets for example. I haven't tested it much, mostly because I'm not very familiar with gtk-rs (or gtk+, for that matter), so if you see any problems with it, please point them out to me.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -737,6 +737,7 @@ impl Library {
                         }
                         "doc" => doc = try!(read_text(parser)),
                         "doc-deprecated" => doc_deprecated = try!(read_text(parser)),
+                        "doc-version" => try!(ignore_element(parser)),
                         x => return Err(mk_error!(format!("Unexpected element <{}>", x), parser)),
                     }
                 }


src/codegen/sys/lib_.rs
```diff
@@ -72,6 +72,7 @@ fn generate_lib(w: &mut Write, env: &Env) -> Result<()>{
 
     try!(writeln!(w, "\n}}"));
 
+
```
GuillaumeGomez: Why this empty line?

src/codegen/sys/lib_.rs
```diff
@@ -232,10 +233,23 @@ fn prepare_enum_member_name(name: &str) -> String {
...
+        
+        let comment = if commented { "//" } else { "" };
+        if lines.is_empty() {
+            try!(writeln!(w, "{}#[repr(C)]\n{0}pub struct {}(c_void);\n", comment, klass.c_type));
+        }
+        else {
+            try!(writeln!(w, "{}#[repr(C)]\n{0}pub struct {} {{", comment, klass.c_type));
+
+            for line in lines {
+                try!(writeln!(w, "{}{}", comment, line));
```
GuillaumeGomez: Isn't a comma missing in here?

src/codegen/sys/lib_.rs
```diff
@@ -232,10 +233,23 @@ fn prepare_enum_member_name(name: &str) -> String {
...
-fn generate_classes_structs(w: &mut Write, classes: &[&Class]) -> Result<()> {
+fn generate_classes_structs(w: &mut Write, env: &Env, classes: &[&Class]) -> Result<()> {
     try!(writeln!(w, ""));
     for klass in classes {
-        try!(writeln!(w, "#[repr(C)]\npub struct {}(c_void);", klass.c_type));
+        let (lines, commented) = generate_fields(env, &klass.name, &klass.fields);
+        
+        let comment = if commented { "//" } else { "" };
+        if lines.is_empty() {
+            try!(writeln!(w, "{}#[repr(C)]\n{0}pub struct {}(c_void);\n", comment, klass.c_type));
```
GuillaumeGomez: Could you use named parameters instead of `{}` and `{0}` please?

src/codegen/sys/lib_.rs
```diff
@@ -232,10 +233,23 @@ fn prepare_enum_member_name(name: &str) -> String {
...
+        }
+        else {
+            try!(writeln!(w, "{}#[repr(C)]\n{0}pub struct {} {{", comment, klass.c_type));
```
GuillaumeGomez: Same comment as above.

GuillaumeGomez:
Sorry for the delay.

cc @EPashkin @gkoz

src/codegen/sys/lib_.rs
eugene2k: Actually, my whole file is different... wth...

eugene2k: Okay, this was stupid, but forgot to switch branches :)

src/codegen/sys/lib_.rs
eugene2k: err... where?

GuillaumeGomez: Aren't you writing struct's fields? (It's totally possible that I misunderstood what you were doing here.)

eugene2k: Not quite, the fields are in `generate_fields`, which generates indented lines that end with a comma. I could move the comma out of generate fields, though.

eugene2k: Also, the original code wasn't written by me (I think @EPashkin wrote it), I simply modified it to work with classes as well as records, as I didn't see any reason it shouldn't.

GuillaumeGomez: Ok, so it confirms what I thought. My bad. :p

GuillaumeGomez:
All good for me now. Waiting for @EPashkin and @gkoz.

EPashkin:
Changes mady by this PR in sys crate: https://github.com/EPashkin/rust-gnome-sys/commit/be6f7c530095f271fb4ce90ed736e85572223d87
Not sure that we want show gtk internals, but for custom widget it may be needed (don't know how do create custom widget).

EPashkin:
May be this need be separated as special mode.

eugene2k:
Hmm... seeing as how the same types of fields are randomly private and public, I can force them all to be private, until someone actually raises an issue...

gkoz:
I'm coming round to making all fields public as requested in https://github.com/gtk-rs/sys/issues/25. Given the advances in bindgen and apparent intention of the Rust project to integrate more C interoperability in 2017, it seems undesirable to diverge from C headers too much, and C doesn't have private fields.

eugene2k:
> C doesn't have private fields
Which doesn't seem to be much of a problem for the GTK developers creating them...
/rant

EPashkin:
@GuillaumeGomez, how you think, better make all fields public?
While `pub priv_: gpointer` look strange, it may be solve some problem with custom objects.

EPashkin:
On other side, if we find, that all pub is wrong and unneeded changing it back it's be `breaking change`

GuillaumeGomez:
I think that all FFI struct's fields should be public. Otherwise, it depends if it is an internal mechanism or not.

EPashkin:
@GuillaumeGomez, seems it works, as you say.
@eugene2k, sorry for long wait, can you squash commits to one?

EPashkin:
@GuillaumeGomez, about extra empty lines from https://github.com/gtk-rs/gir/blob/master/src/codegen/sys/lib_.rs#L245 etc: I think that better be replaced later with something like 
```
if(!interfaces.is_empty()) {
    try!(writeln!(w, "//Interfaces"));
}
```

GuillaumeGomez:
@EPashkin: You mean `// Interfaces`, right? :p
@eugene2k: Just like @EPashkin said: please squash your commits into one and then I merge.

eugene2k:
@EPashkin, @GuillaumeGomez 
I seem to have messed up my repo branches, so I'll resubmit the pull request in a moment

GuillaumeGomez:
Sure, no worries.

Issue #278, opened 2016-10-15
Gir should sometimes generate Option<Type> instead of Type


antoyo:
Hello.
For this [signal](https://webkitgtk.org/reference/webkit2gtk/stable/WebKitWebView.html#WebKitWebView-create), the return type generated by the binding should be `Option<Widget>` instead of only `Widget` because the return value can be null:
"a newly allocated WebKitWebView widget or NULL to propagate the event further."
[Here is the currently generated binding.](https://github.com/antoyo/webkit2gtk-rs/blob/master/src/auto/web_view.rs#L458)
Thanks to fix this issue.

EPashkin:
Try use config https://github.com/gtk-rs/gtk/blob/master/Gir.toml#L467-L468
If it no effect we try to add usage of `nullable` for signals.

antoyo:
It does not seem to work for signals.
I tried the following:
```toml
[[object]]
name = "WebKit2.WebView"
status = "generate"
    [[object.signal]]
    name = "create"
    version = "2.6"
        [object.signal.return]
        nullable = true
```
Is it the right syntax?

EPashkin:
Added PR for this (https://github.com/EPashkin/gir/tree/nullable_in_signals).
@antoyo, check it, please.

antoyo:
It works.
Thanks!

PR #280, opened 2016-10-15
Add object.signal.return.nullable config

EPashkin:
Fix issue #278

### DIFF
--- a/src/analysis/trampolines.rs
+++ b/src/analysis/trampolines.rs
@@ -44,6 +44,7 @@ pub fn analyze(env: &Env, signal: &library::Signal, type_tid: library::TypeId, i
     //Fake
     let configured_functions: Vec<&config::functions::Function> = Vec::new();
 
+    //TODO: move to object.signal.return config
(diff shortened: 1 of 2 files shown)


src/analysis/trampolines.rs
```diff
@@ -94,19 +95,33 @@ pub fn analyze(env: &Env, signal: &library::Signal, type_tid: library::TypeId, i
...
+
+        let nullable_override = configured_signals.iter()
+            .filter_map(|f| f.ret.nullable)
+            .next();
+        if let Some(val) = nullable_override {
+            ret_nullable = val;
+        }
     }
 
+    let ret = library::Parameter {
```
GuillaumeGomez: Won't these changes create troubles with #282?

GuillaumeGomez:
cc @gkoz

src/analysis/trampolines.rs
EPashkin: What kind of troubles? 282 build on top of this PR

Issue #281, opened 2016-10-15
Gir generates a Display impl that does not compile


antoyo:
Hello.
Gir generates this [Display impl](https://github.com/antoyo/webkit2gtk-webextension-rs/commit/9c9ffd875c880fee3b7f09dd53e054d54ee65b7b) that I needed to comment because it does not compile.
If you look at the code above, you wiil see that the `to_string()` method returns a `Result<String, Error>`, so the format string `"{}"` is wrong.
Could you please fix this issue?
Thanks.

antoyo:
Your fix works.
Thanks for quickly fixing the issues!

PR #282, opened 2016-10-15
Throwable to string

EPashkin:
Fix generating Display for records with throwable `to_string` for #281


GuillaumeGomez:
cc @gkoz

gkoz:
Thanks!

PR #292, opened 2016-11-04
Nullable signal parameters

### LINKED ISSUE #291: Some trampoline parameter can be null
Hello.
There are some parameters of trampoline that can be null (like the `redirected_response` parameter of the [`send-request` signal](https://webkitgtk.org/reference/webkit2gtk/stable/WebKitWebPage.html#WebKitWebPage-send-request)) so the use of `from_glib_none()` causes a panic when the parameter is null.
[See generated code here.](https://github.com/antoyo/webkit2gtk-webextension-rs/blob/master/src/auto/web_page.rs#L123)
Thanks to fix this issue.

EPashkin:
Fixes #291

cc @gkoz @GuillaumeGomez 
Contains config refactoring (splitting `Identables`)

### DIFF
--- a/src/analysis/parameter.rs
+++ b/src/analysis/parameter.rs
@@ -1,6 +1,7 @@
 use std::borrow::Cow;
 
 use config::functions::Function;
+use config::signals::Signal;
 use env::Env;
 use library;
 use nameutil;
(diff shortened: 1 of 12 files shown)


GuillaumeGomez:
Since I don't know much about this part in `gir`, I can't give approval on the change itself. However the code looks good for me. Waiting for @gkoz's approval.

EPashkin:
Removed commit "Combined analysis for functions and signals parameters" as it will be separated again later.

GuillaumeGomez:
Thanks!

PR #297, opened 2016-12-08
Parse fields in classes and generate correct class structs (resubmit)

eugene2k:
This is a resubmit of #276 with the commits squashed and some reordering of functions in codegen/sys/lib_.rs for diff readability.


EPashkin:
👍 
Thanks @eugene2k

Issue #139, opened 2015-10-04
Avoid generating a method if a parent already has it


gkoz:
For instance, `gtk_widget_activate` vs `gtk_menu_item_activate`.
Emit a warning -- in cases I've seen so far they're essentially duplicate, but who knows.
Bonus points for comparing the signatures.

EPashkin:
IMHO some function names will be good in description.

gkoz:
Yes, for an example see https://github.com/gtk-rs/gtk/issues/168, also `gtk_tree_view_get_n_columns` vs `gtk_tree_model_get_n_columns`.

gkoz:
This turns out not so easy. Currently analysis results aren't stored so you can't just pass over them again marking the duplicates. Looks like this will have to wait for #136

EPashkin:
Agree :cry:

EPashkin:
For current attempt I found that these methods is duplicated (by name)
Button (parent Widget has methods with "v3_20")
fn get_focus_on_click(&self) -> bool;
fn set_focus_on_click(&self, focus_on_click: bool);
CellAreaBox (parent CellLayout has different number of arguments):
pub fn pack_end<T: IsA<CellRenderer>>(&self, renderer: &T, expand: bool, align: bool, fixed: bool)
pub fn pack_start<T: IsA<CellRenderer>>(&self, renderer: &T, expand: bool, align: bool, fixed: bool)
ComboBox (same as Button)
fn get_focus_on_click(&self) -> bool;
fn set_focus_on_click(&self, focus_on_click: bool);
ComboBoxText (parent Container has `fn remove<T: IsA<Widget>>(&self, widget: &T)`)
pub fn remove(&self, position: i32)
Container (parent Widget has different number of arguments):
fn child_notify<T: IsA<Widget>>(&self, child: &T, child_property: &str);
FileChooserButton (same as Button)
pub fn get_focus_on_click(&self) -> bool
pub fn set_focus_on_click(&self, focus_on_click: bool)
Menu (parent Widget has different number of arguments):
pub fn set_accel_path(&self, accel_path: Option<&str>)
MenuButton (parent Widget has `fn get_direction(&self) -> TextDirection`
#[cfg(feature = "v3_6")]
pub fn get_direction(&self) -> ArrowType
#[cfg(feature = "v3_6")]
pub fn set_direction(&self, direction: ArrowType)
MenuItem (same as Menu):
fn set_accel_path(&self, accel_path: Option<&str>);
[...]

PR #305, opened 2017-01-12
Don't generate a method if a parent already has it

EPashkin:
Part of #139 

WIP

### DIFF
--- a/src/analysis/record.rs
+++ b/src/analysis/record.rs
@@ -53,7 +53,8 @@ pub fn new(env: &Env, obj: &GObject) -> Option<Info> {
     imports.add("glib::translate::*", None);
     imports.add("ffi", None);
 
-    let mut functions = functions::analyze(env, &record.functions, record_tid, &obj, &mut imports);
+    let mut functions = functions::analyze(env, &record.functions, record_tid, &obj,
+                                           &mut imports, None, None);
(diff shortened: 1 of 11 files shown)


EPashkin:
@GuillaumeGomez, seem this PR done.
I added PR in gtk with results and in example for build fix after change.

src/analysis/record.rs
```diff
@@ -84,7 +85,7 @@ pub fn new(env: &Env, obj: &GObject) -> Option<Info> {
         deprecated_version: deprecated_version,
         cfg_condition: obj.cfg_condition.clone(),
     };
-    
+
```
GuillaumeGomez: Oh nice one! ;)

src/codegen/general.rs
```diff
@@ -118,6 +118,16 @@ pub fn version_condition_string(env: &Env, version: Option<Version>, commented:
     }
 }
 
+pub fn not_version_condition(w: &mut Write, version: Option<Version>, commented: bool,
+                         indent: usize) -> Result<()> {
```
GuillaumeGomez: The indent is broken here.

GuillaumeGomez:
Seems good to me. As always, when you feel ready, just merge it. Thanks for your work!

PR #314, opened 2017-02-08
Add Into<&str> for optional parameters

GuillaumeGomez:
Remains to call `.into()` on the arguments (didn't look yet where I can do it in the code).

cc @EPashkin

### DIFF
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -114,23 +114,23 @@ mod tests {
     fn get_new_all() {
         let mut bounds: Bounds = Default::default();
         let typ = BoundType::IsA;
-        assert_eq!(bounds.add_parameter("a", "", typ), true);
-        assert_eq!(bounds.add_parameter("a", "", typ), false);  //Don't add second time
-        assert_eq!(bounds.add_parameter("b", "", typ), true);
-        assert_eq!(bounds.add_parameter("c", "", typ), true);
-        assert_eq!(bounds.add_parameter("d", "", typ), true);
-        assert_eq!(bounds.add_parameter("e", "", typ), true);
-        assert_eq!(bounds.add_parameter("f", "", typ), true);
-        assert_eq!(bounds.add_parameter("g", "", typ), true);
-        assert_eq!(bounds.add_parameter("h", "", typ), false);
+        assert_eq!(bounds.add_parameter("a", "", typ, false), true);
+        assert_eq!(bounds.add_parameter("a", "", typ, false), false);  //Don't add second time
+        assert_eq!(bounds.add_parameter("b", "", typ, false), true);
+        assert_eq!(bounds.add_parameter("c", "", typ, false), true);
+        assert_eq!(bounds.add_parameter("d", "", typ, false), true);
+        assert_eq!(bounds.add_parameter("e", "", typ, false), true);
+        assert_eq!(bounds.add_parameter("f", "", typ, false), true);
+        assert_eq!(bounds.add_parameter("g", "", typ, false), true);
+        assert_eq!(bounds.add_parameter("h", "", typ, false), false);
     }
 
     #[test]
     fn get_parameter_alias_info() {
         let mut bounds: Bounds = Default::default();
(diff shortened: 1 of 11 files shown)


GuillaumeGomez:
Actually, I could just implement the implementation for `Into<Option<&'a str>>` instead. Much easier this way!

src/analysis/bounds.rs
```diff
@@ -50,8 +55,16 @@ impl Bounds {
             _ => String::new(),
         }
     }
-    pub fn add_parameter(&mut self, name: &str, type_str: &str, bound_type: BoundType) -> bool {
-        if self.used.iter().any(|ref n| n.0 == name)  { return false; }
+    pub fn add_parameter(&mut self, name: &str, type_str: &str, bound_type: BoundType,
+                         is_nullable: bool) -> bool {
+        if self.used.iter().any(|ref n| n.0 == name) { return false; }
+        if bound_type == BoundType::Into {
+            if is_nullable == false { return true; }
```
EPashkin: If `is_nullable` used only for this check, may be better move it to `type_for` or add separate check function that must called on `BoundType::Into`

GuillaumeGomez: It seemed more logical this way. I still think we should keep in here to avoid having bounds logic splitted.

EPashkin:
@GuillaumeGomez, cast `comments.into() as Option<&'a str>` really needed?
about_dialog.rs normally builds without it.

GuillaumeGomez:
In a lot of cases it's needed. Don't know why unfortunately... Rust automatic coercion is a bit stupid. But at least, with this it works everytime.

EPashkin:
I rebuild all modules without `as`. Affected `gio`, `glib`, `gtk`
And examples builds fine (after fix notebook.rs)
IMHO as `T: Into<Option<&'a str>>>(` function `T.info()` can only one result type.

GuillaumeGomez:
In my case, it's gtk which didn't build without the explicit cast (I didn't try the other libs). I have the following rustc verison:

```
rustc 1.17.0-nightly (ea7a6486a 2017-02-04)
```

EPashkin:
Strange my 2 versions builds fine. I add PR to gtk for CI test

rustc 1.17.0-nightly (0648517fa 2017-02-03)
binary: rustc
commit-hash: 0648517faf1e2cf37c8b6770cbd0180a816ed9a0
commit-date: 2017-02-03
host: x86_64-pc-windows-gnu
release: 1.17.0-nightly
LLVM version: 3.9

rustc 1.17.0-nightly (29dece1c8 2017-02-08)
binary: rustc
commit-hash: 29dece1c8bbebf7ae8034ef0826b119281730937
commit-date: 2017-02-08
host: x86_64-pc-windows-gnu
release: 1.17.0-nightly
LLVM version: 3.9

GuillaumeGomez:
Ok, I'll try without the cast after updating the version locally. So strange...

GuillaumeGomez:
After a little rustup update it works without the cast. Really strange. I remove the cast then!

GuillaumeGomez:
Updated.

src/analysis/bounds.rs
```diff
@@ -44,14 +49,23 @@ impl Bounds {
             _ => None,
         }
     }
-    pub fn to_glib_extra(library: &Library, type_id: TypeId) -> String {
+    pub fn to_glib_extra(library: &Library, type_id: TypeId, is_nullable: bool) -> String {
```
EPashkin: @GuillaumeGomez  While I prefer `library::Nullable` over `bool` I can merge as is.
You don't want squash or add Into for other options?

GuillaumeGomez: > @GuillaumeGomez While I prefer library::Nullable over bool I can merge as is.

Then I'll update in this way.

> You don't want squash

I can, just tell me which commits you want me to keep.

> add Into for other options?

Not for now. :P

EPashkin:
Thanks, I merge this.

GuillaumeGomez:
Oh ok. I'll make a small cleanup then.

PR #316, opened 2017-02-10
Moved bounds logic out parameter

EPashkin:
@GuillaumeGomez I leave merge to you, don't merge if it conflict with your plans.

### DIFF
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -3,7 +3,7 @@ use std::slice::Iter;
 use std::vec::Vec;
 
 use env::Env;
-use library::{Fundamental, Library, Nullable, Type, TypeId};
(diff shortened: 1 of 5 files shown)


GuillaumeGomez:
As long as it produces a correct result, it's ok for me. At least, the check is now centralized. I should try to implement the `Into` trait for more things when I have time... Anyway, thanks for the update!

EPashkin:
I missed that all code in function.rs starting `if !par.instance_parameter && par.direction != ParameterDirection::Out {` can be moved to `Bounds` with making many it functions private.
@GuillaumeGomez please do that on your next attempt.

GuillaumeGomez:
Won't be soon so I think you'll do it before me. (Lot of works on `sysinfo` and `rustc`)

PR #319, opened 2017-02-26
Remove unused uses in generated enums.rs

EPashkin:
Removes lines in `enums.rs` if no `impl ErrorDomain` generated.
In `pango` in allow remove dependency to `glib-sys` added in https://github.com/gtk-rs/pango/pull/54
```
use glib_ffi;
use glib::error::ErrorDomain;
```


EPashkin:
Also removed many unused warnings imposed by properties.

PR #320, opened 2017-03-02
Update toml, regex and xml-rs version

EPashkin:
Updating to new last toml as `rustc-serialize` planed for deprecation https://internals.rust-lang.org/t/rust-release-milestone-predictions/4591
Only last is `docopt` that also planing use serde https://github.com/docopt/docopt.rs/issues/128

PS. There was problem with sys crates https://github.com/alexcrichton/toml-rs/pull/151
PPS. When upgrading found that `regex 0.2.0` don't implement `Eq` so this code fails and upgrade will need other PR later (if any).
```
#[derive(Clone, Debug, Eq, PartialEq)]
pub enum Ident {
    Name(String),
    Pattern(Regex),
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,18 +10,18 @@ build = "build.rs"
 exclude = ["Gir*.toml", "tests/**/*", "*.md"]
 
 [dependencies]
-docopt = "*"
-xml-rs = "*"
-toml = "*"
-case = "*"
-env_logger = "*"
-error-chain = { version = "*", default-features = false }
-git2 = "0.4" #locked due 0.6 dependency curl-sys won't builds on appveyor
-lazy_static = "*"
-log = "*"
-regex = "*"
-time = "*"
-hprof = "*"
+docopt = "0.7"
+xml-rs = "0.4"
+toml = "0.3"
+case = "0.1"
+env_logger = "0.4"
+error-chain = { version = "0.10", default-features = false }
+git2 = "0.4" #due 0.6 dependency curl-sys won't builds on appveyor
+lazy_static = "0.2"
+log = "0.3"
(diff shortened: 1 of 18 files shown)


EPashkin:
Fixed regex update

EPashkin:
Upgraded xml-rs

EPashkin:
@GuillaumeGomez I merge PR tomorrow if you do not see the problems

Cargo.toml
```diff
@@ -11,7 +11,7 @@ exclude = ["Gir*.toml", "tests/**/*", "*.md"]
 
 [dependencies]
 docopt = "*"
-xml-rs = "0.2.1"
+xml-rs = "*"
```
GuillaumeGomez: I'm clearly not a big fan of this. Also, we won't be able to publish we use '*' (even if we don't care).

GuillaumeGomez:
Except for my comment, I'm 100% 👍

Cargo.toml
EPashkin: Change to "0.2" and add version for others *?

GuillaumeGomez: Yes please. At least we'll have a bit more control over dependencies.

EPashkin:
Done.
Also upgraded git2 with removing unneeded features

EPashkin:
Not sure that we really needed same `build-dependencies`

EPashkin:
Was wrong about `build-dependencies` forgot that it

GuillaumeGomez:
Ok. So like I said, it's good for me. Merge when you want to.

EPashkin:
Removed unneeded `into_owned`
Will merge tomorrow

PR #321, opened 2017-03-04
Add into conversion for every nullable type

GuillaumeGomez:
However, I can't make it work because (small subset of errors):

```
error[E0191]: the value of the associated type `Storage` (from the trait `glib::translate::ToGlibPtr`) must be specified
   --> src/auto/window.rs:993:49
    |
993 |     fn set_transient_for<'a, T: Into<Option<&'a IsA<Window>>>>(&self, parent: T) {
    |                                                 ^^^^^^^^^^^ missing associated type `Storage` value

error[E0191]: the value of the associated type `GlibType` (from the trait `glib::wrapper::Wrapper`) must be specified
   --> src/auto/window.rs:993:49
    |
993 |     fn set_transient_for<'a, T: Into<Option<&'a IsA<Window>>>>(&self, parent: T) {
    |                                                 ^^^^^^^^^^^ missing associated type `GlibType` value
```

Apparently, it cannot determine the type if there is too much indirection. Quite annoying. :-/

Edit: Rust issue about problem that caused need of local variables https://github.com/rust-lang/rust/issues/40613

### DIFF
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -9,12 +9,12 @@ use analysis::rust_type::bounds_rust_type;
 use library::{Function, Fundamental, Nullable, Type, TypeId, ParameterDirection};
 use traits::IntoString;
 
-#[derive(Copy, Clone, Eq, Debug, PartialEq)]
+#[derive(Clone, Eq, Debug, PartialEq)]
 pub enum BoundType {
     IsA,
     AsRef,
-    //lifetime
-    Into(char),
+    // lifetime
+    Into(char, Option<Box<BoundType>>),
 }
 
 #[derive(Clone, Eq, Debug, PartialEq)]
@@ -66,7 +66,7 @@ impl Bounds {
     pub fn add_for_parameter(&mut self, env: &Env, func: &Function, par: &mut Parameter) {
         if !par.instance_parameter && par.direction != ParameterDirection::Out {
             if let Some(bound_type) = Bounds::type_for(env, par.typ, par.nullable) {
-                let to_glib_extra = Bounds::to_glib_extra(bound_type);
+                let to_glib_extra = Bounds::to_glib_extra(bound_type.clone());
                 par.to_glib_extra = to_glib_extra;
                 let type_name = bounds_rust_type(env, par.typ);
                 if !self.add_parameter(&par.name, &type_name.into_string(), bound_type) {
@@ -80,23 +80,33 @@ impl Bounds {
         use self::BoundType::*;
         match *env.library.type_(type_id) {
             Type::Fundamental(Fundamental::Filename) => Some(AsRef),
-            Type::Fundamental(Fundamental::Utf8) if *nullable => Some(Into('_')),
-            Type::Class(..) => {
+            Type::Fundamental(Fundamental::Utf8) if *nullable => Some(Into('_', None)),
+            Type::Class(..) if !*nullable => {
                 if env.class_hierarchy.subtypes(type_id).next().is_some() {
                     Some(IsA)
                 } else {
                     None
                 }
             }
-            Type::Interface(..) => Some(IsA),
-            _ => None,
+            Type::Class(..) => {
+                if env.class_hierarchy.subtypes(type_id).next().is_some() {
+                    Some(Into('_', Some(Box::new(IsA))))
+                } else {
+                    Some(Into('_', None))
+                }
+            }
+            Type::Interface(..) if !*nullable => Some(IsA),
+            Type::Interface(..) => Some(Into('_', Some(Box::new(IsA)))),
+            _ if !*nullable => None,
+            _ => Some(Into('_', None)),
         }
     }
     fn to_glib_extra(bound_type: BoundType) -> String {
         use self::BoundType::*;
         match bound_type {
             AsRef => ".as_ref()".to_owned(),
-            Into(_) => ".into()".to_owned(),
+            Into(_, None) => ".into()".to_owned(),
+            Into(_, Some(x)) => format!(".into(){}", Bounds::to_glib_extra(*x)),
             _ => String::new(),
         }
     }
@@ -105,7 +115,7 @@ impl Bounds {
         if let BoundType::Into(_) = bound_type {
             if let Some(lifetime) = self.unused_lifetimes.pop_front() {
                 self.lifetimes.push(lifetime);
-                bound_type = BoundType::Into(lifetime);
+                bound_type = BoundType::Into(lifetime, x);
             } else {
                 return false;
             }
@@ -133,7 +143,15 @@ impl Bounds {
             match used.bound_type {
                 IsA => imports.add("glib::object::IsA", None),
                 AsRef => imports.add_used_type(&used.type_str, None),
-                Into(_) => {}
+                Into(_, Some(ref x)) => {
+                    match **x {
+                        IsA => imports.add("glib::object::IsA", None),
+                        // This case shouldn't be possible normally.
+                        AsRef => imports.add_used_type(&used.2, None),
+                        _ => {}
+                    }
+                }
+                Into(_, None) => {}
             }
         }
     }
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -21,7 +21,7 @@ pub enum BoundType {
 }
 
 impl BoundType {
-    fn set_lifetime(ty_: BoundType, lifetime: char) -> BoundType {
+    fn with_lifetime(ty_: BoundType, lifetime: char) -> BoundType {
         match ty_ {
             BoundType::IsA(_) => BoundType::IsA(Some(lifetime)),
             BoundType::AsRef(_) => BoundType::AsRef(Some(lifetime)),
@@ -64,13 +64,14 @@ impl Bound {
     pub fn get_for_property_setter(env: &Env, var_name:&str, type_id: TypeId,
                                    nullable: Nullable) -> Option<Bound> {
         match Bounds::type_for(env, type_id, nullable) {
-            Some(BoundType::IsA) => {
+            Some(BoundType::IsA(l)) => {
                 let type_str = bounds_rust_type(env, type_id);
-                Some(Bound{
-                    bound_type: BoundType::IsA,
+                Some(Bound {
+                    bound_type: BoundType::IsA(l),
                     parameter_name: var_name.to_owned(),
                     alias: 'T',
                     type_str: type_str.into_string(),
+                    info_for_next_type: false,
                 })
             }
             _ => None,
@@ -125,8 +126,7 @@ impl Bounds {
         use self::BoundType::*;
         match bound_type {
             AsRef(_) => ".as_ref()".to_owned(),
-            Into(_, None) => ".into()".to_owned(),
-            Into(_, Some(x)) => format!(".into(){}", Bounds::to_glib_extra(*x)),
+            Into(_, Some(x)) => Bounds::to_glib_extra(*x),
             _ => String::new(),
         }
     }
@@ -145,9 +145,14 @@ impl Bounds {
         };
         let type_str = if let Some((Some(sub), lifetime)) = sub {
             if let Some(alias) = self.unused.pop_front() {
-                self.used.push((name.into(), alias, type_str.into(),
-                                BoundType::set_lifetime(*sub, lifetime), true));
-                format!("{}", alias)
+                self.used.push(Bound {
+                    bound_type: BoundType::with_lifetime(*sub, lifetime),
+                    parameter_name: name.to_owned(),
+                    alias: alias,
+                    type_str: type_str.to_owned(),
+                    info_for_next_type: true,
+                });
+                alias.to_string()
             } else {
                 return false;
             }
@@ -160,6 +165,7 @@ impl Bounds {
                 parameter_name: name.to_owned(),
                 alias: alias,
                 type_str: type_str.to_owned(),
+                info_for_next_type: false,
             });
             true
         } else {
@@ -167,22 +173,14 @@ impl Bounds {
         }
     }
     pub fn get_parameter_alias_info(&self, name: &str) -> Option<(char, BoundType)> {
-        let mut stop = false;
         self.used.iter()
                  .find(move |ref n| {
-            if stop == true {
-                true
-            } else if n.parameter_name == name {
-                if n.info_for_next_type == false {
-                    true
-                } else {
-                    stop = true;
-                    false
-                }
+            if n.parameter_name == name {
+                !n.info_for_next_type
             } else {
                 false
             }})
-            .map(|t| (t.alias, t.bound_type))
+            .map(|t| (t.alias, t.bound_type.clone()))
     }
     pub fn update_imports(&self, imports: &mut Imports) {
         //TODO: import with versions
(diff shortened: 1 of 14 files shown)


EPashkin:
It compiles with `fn set_focus_child<'a, T: IsA<Widget>, U: Into<Option<&'a U>>>(&self, child: U);
`

EPashkin:
Was wrong (used U: U)

EPashkin:
It compiled after fix

EPashkin:
Sorry was wrong again, working version:
```rust
fn set_focus_child<'a, T: 'a + IsA<Widget>, U: Into<Option<&'a T>>>(&self, child: U) {
    unsafe {
        let child = child.into();
        ffi::gtk_container_set_focus_child(self.to_glib_none().0, child.to_glib_none().0);
    }
}
```

GuillaumeGomez:
So like I thought, we have to add a second type parameter. Not cool. :-/ Won't we have issues with Ext traits? I guess we'll have to generate them again...

EPashkin:
Seems you need separate `IsWindow` from other references:
`fn set_help_overlay<'a, T: Into<Option<&'a ShortcutsWindow>>>(&self, help_overlay: T)`
don't need `let child = child.into();`

GuillaumeGomez:
Uh? That's quite strange...

GuillaumeGomez:
Right, this code breaks the case we have (`str` instead of `&str`):

```rust
pub fn new<'a, P: Into<Option<&'a [str]>>>(size: i32, min: f64, max: f64, step: f64, icons: P) -> ScaleButton {
```

It should be:

```rust
pub fn new<'a, P: Into<Option<&'a [&'a str]>>>(size: i32, min: f64, max: f64, step: f64, icons: P) -> ScaleButton {
```

EPashkin:
Slices can't be optional (`[]` used instead it), so `gir` must additional check for all types of arrays

EPashkin:
`gtk_entry_set_invisible_char` affected in wrong way, it don't needed reference to `char`
Maybe instead `TYPE_PARAMETERS_START` better just expand "TUVWXYZ" to "TUVWXYZSRQPO.."

EPashkin:
Other idea about `Bounds`: instead `unused: VecDeque` use `next_type_alias_index: u8` and add function to convert it to `char` (maybe with storing `u8` in `used` but IMHO it unneeded)

GuillaumeGomez:
> Slices can't be optional (`[]` used instead it), so `gir` must additional check for all types of arrays

Good to know, it'll make my life easier!

> `gtk_entry_set_invisible_char` affected in wrong way, it don't needed reference to `char`

Yes, I saw it as well. But I think it's a minor side-effect for now, it'll certainly be easy to change.

> Maybe instead `TYPE_PARAMETERS_START` better just expand "TUVWXYZ" to "TUVWXYZSRQPO.."

I wonder... I like keeping things in the "good" order. Also, with this way of doing it, we can now easily expand the type parameter names.

> Other idea about `Bounds`: instead unused: `VecDeque` use `next_type_alias_index: u8` and add function to convert it to `char` (maybe with storing `u8` in `used` but IMHO it unneeded)

I like manipulating `char`s directly. Maybe some day (who knows?) we'll need to manipulate type parameters which will look like `AA`. I hope not but I prefer to be cautious. So for now, I think I'll keep using `char`, except if you have a very good reason to change this.

EPashkin:
> I wonder... I like keeping things in the "good" order. Also, with this way of doing it, we can now easily expand the type parameter names.

Minor problem with `TYPE_PARAMETERS_START` that if we found that it need changes we need change (regenerate) all crates to reflect that.
`"TUVWXYZSRQPO.."` don't mean remove `TYPE_PARAMETERS_START` it idea about use aliases T, U, V, .. Z for functions that need few type parameters and S, R, Q, ..., A (may be starting with AA next) for rare others.

> I like manipulating chars directly. Maybe some day (who knows?) we'll need to manipulate type parameters which will look like AA. I hope not but I prefer to be cautious. So for now, I think I'll keep using char, except if you have a very good reason to change this.

Sorry, I don't understand: you against store index in `Bound.used` (same as I) or don't like change `VecDeque` to index and converter index to char?

GuillaumeGomez:
> Minor problem with TYPE_PARAMETERS_START that if we found that it need changes we need change (regenerate) all crates to reflect that.
"TUVWXYZSRQPO.." don't mean remove TYPE_PARAMETERS_START it idea about use aliases T, U, V, .. Z for functions that need few type parameters and S, R, Q, ..., A (may be starting with AA next) for rare others.

I prefer to start at a common point (by using `TYPE_PARAMETERS_START`), even for functions with small number of type parameters. I don't mind regenerating all crates in order to reflect it.

> Sorry, I don't understand: you against store index in `Bound.used` (same as I) or don't like change `VecDeque` to index and converter index to char?

What I wanted to say is that I didn't feel like using an `u8` indirection to then convert it to `char` was a good idea. However, I don't care much about where we store the index.

GuillaumeGomez:
Ok, just one last type of error:

```
error: borrowed value does not live long enough
   --> src/auto/window.rs:995:101
    |
995 |             ffi::gtk_window_set_transient_for(self.to_glib_none().0, parent.into().to_glib_none().0);
    |                                                                      -------------                  ^ temporary value dropped here while still borrowed
    |                                                                      |
    |                                                                      temporary value created here
    |
    = note: values in a scope are dropped in the opposite order they are created
    = note: consider using a `let` binding to increase its lifetime
```

Seems pretty obvious and difficult to do at the same time. How can add something like:

```rust
let into_parent = parent.into();
```

?

EPashkin:
This can be added to `codegen\function_body_chunk.rs`
After calling `self.write_out_variables(&mut body);` add function with adding for all variables something like
```
        let value = Chunk::Custom(format!("{}.into();", name));
        chunks.push(Chunk::Let{
            name: name.clone(),
            is_mut: false,
            value: Box::new(value),
            type_: None
        });

```

EPashkin:
And better don't name variable "into_parent"

GuillaumeGomez:
> And better don't name variable "into_parent"

What I had in mind was:

```
into_{variable_name}
```

So `into_parent` was just random. But if it doesn't satisfy you, what would you prefer?

EPashkin:
If you define variable with different name then you need also change it in function call, IMHO simpler just "redefine" without change name (as I show in code)

GuillaumeGomez:
Ah right. Feel stupid for not thinking about shadowing...

GuillaumeGomez:
We now have a last set of error:

```
> cargo build
   Compiling gtk v0.1.1 (file:///Users/imperio/rust/gtk)
error: `parent` does not live long enough
  --> src/auto/app_chooser_dialog.rs:35:5
   |
33 |             Widget::from_glib_none(ffi::gtk_app_chooser_dialog_new_for_content_type(parent.to_glib_none().0, flags.to_glib(), content_type.to_glib_none().0)).downcast_unchecked()
   |                                                                                     ------ borrow occurs here
34 |         }
35 |     }
   |     ^ `parent` dropped here while still borrowed
   |
   = note: values in a scope are dropped in the opposite order they are created

error: `window` does not live long enough
   --> src/auto/application.rs:104:5
    |
102 |             ffi::gtk_application_inhibit(self.to_glib_none().0, window.to_glib_none().0, flags.to_glib(), reason.to_glib_none().0)
    |                                                                 ------ borrow occurs here
103 |         }
104 |     }
    |     ^ `window` dropped here while still borrowed
    |
    = note: values in a scope are dropped in the opposite order they are created

error: `widget` does not live long enough
   --> src/auto/cell_renderer.rs:238:5
    |
236 |             from_glib(ffi::gtk_cell_renderer_get_state(self.to_glib_none().0, widget.to_glib_none().0, cell_state.to_glib()))
    |                                                                               ------ borrow occurs here
237 |         }
238 |     }
    |     ^ `widget` dropped here while still borrowed
    |
    = note: values in a scope are dropped in the opposite order they are created

error: `parent` does not live long enough
  --> src/auto/color_chooser_dialog.rs:33:5
   |
31 |             Widget::from_glib_none(ffi::gtk_color_chooser_dialog_new(title.to_glib_none().0, parent.to_glib_none().0)).downcast_unchecked()
   |                                                                                              ------ borrow occurs here
32 |         }
33 |     }
   |     ^ `parent` dropped here while still borrowed
   |
   = note: values in a scope are dropped in the opposite order they are created

error: `parent` does not live long enough
  --> src/auto/font_chooser_dialog.rs:31:5
   |
29 |             Widget::from_glib_none(ffi::gtk_font_chooser_dialog_new(title.to_glib_none().0, parent.to_glib_none().0)).downcast_unchecked()
   |                                                                                             ------ borrow occurs here
30 |         }
31 |     }
   |     ^ `parent` dropped here while still borrowed
   |
   = note: values in a scope are dropped in the opposite order they are created

error: `icon_widget` does not live long enough
  --> src/auto/menu_tool_button.rs:35:5
   |
33 |             ToolItem::from_glib_none(ffi::gtk_menu_tool_button_new(icon_widget.to_glib_none().0, label.to_glib_none().0)).downcast_unchecked()
   |                                                                    ----------- borrow occurs here
34 |         }
35 |     }
   |     ^ `icon_widget` dropped here while still borrowed
   |
   = note: values in a scope are dropped in the opposite order they are created

error: `icon_widget` does not live long enough
  --> src/auto/tool_button.rs:35:5
   |
33 |             ToolItem::from_glib_none(ffi::gtk_tool_button_new(icon_widget.to_glib_none().0, label.to_glib_none().0)).downcast_unchecked()
   |                                                               ----------- borrow occurs here
34 |         }
35 |     }
   |     ^ `icon_widget` dropped here while still borrowed
   |
   = note: values in a scope are dropped in the opposite order they are created

error: aborting due to 7 previous errors

error: Could not compile `gtk`.
```

Seems like the `let` binding isn't enough... :-/

EPashkin:
Sorry, I missed comments about error.
Not fully understand why it's error but can fixed with 2 let per variable
```
    pub fn inhibit<'a, 'b, P: IsA<Window> + 'a, Q: Into<Option<&'a P>>, R: Into<Option<&'b str>>>(&self, window: Q, flags: ApplicationInhibitFlags, reason: R) -> u32 {
        let window = window.into();
        let window = window.to_glib_none();
        let reason = reason.into();
        unsafe {
            ffi::gtk_application_inhibit(self.to_glib_none().0, window.0, flags.to_glib(), reason.to_glib_none().0)
        }
    }
```

GuillaumeGomez:
Urg. Not very beautiful. :-/

EPashkin:
May be it better fixed on `rustc` side: temporary object, created by `to_glib_none()` IMHO must be dropped at end of unsafe block, while error says that borrow ends at end of function

GuillaumeGomez:
@EPashkin: I think we'll need to make an intermediate `let` binding. I tried to make it but I break everything every time. Do you know where should I make this change (creating the Chunk like I did for `.into()`) please?

EPashkin:
IMHO `function_body_chunk::add_into_conversion` need work on list of bounds without recalculating conditions and add `let x = x.to_glib_xxx()` from info in `BoundType::Into`.
As for ffi-function parameter like `x.0` maybe simpler add `Chunk::FfiCallStashParameter(name)` and switch to it in `generate_func_parameters` if parameter name found in `Into` bounds

GuillaumeGomez:
Yes, my `function_body_chunk::add_into_conversion` really need to be reworked. I wondered if it'd be better to directly add a variable into the `ParameterIn` struct instead.

> As for ffi-function parameter like x.0 maybe simpler add Chunk::FfiCallStashParameter(name) and switch to it in generate_func_parameters if parameter name found in Into bounds

I didn't understand this part.

EPashkin:
You can add optional variable to `ParameterIn` (it not memory wise but not needed searching).
For initialization I recommend add function `pub fn bounds(&mut self, bounds: &AnalysisBounds)`
that only add bounds for problem `Into`.
For last part: you need change `window.to_glib_none().0` to `window.0` in ffi function call.
So you need change `generate_func_parameters`.
I propose add new chunk type for this task and use it for branch `In { ref parameter, ref bound } if bound.is_some() =>`
while branch `In { ref parameter, None } =>` stays `Chunk::FfiCallParameter...`

GuillaumeGomez:
Ok, updated to last version of gir. Now let's fix this issue!

GuillaumeGomez:
Ok, the generated is now working ( 🎉 ). However, the code isn't very clean. Any way I could improve it a bit @EPashkin ?

src/analysis/bounds.rs
```diff
@@ -79,61 +96,117 @@ impl Bounds {
...
     pub fn get_parameter_alias_info(&self, name: &str) -> Option<(char, BoundType)> {
-        self.used.iter().find(|n| n.parameter_name == name)
-            .map(|t| (t.alias, t.bound_type))
+        let mut stop = false;
+        self.used.iter()
+                 .find(move |ref n| {
+            if stop == true {
+                true
+            } else if n.parameter_name == name {
+                if n.info_for_next_type == false {
```
EPashkin: can changed just to `!n.info_for_next_type` with same result, `stop` unneeded then

src/analysis/bounds.rs
```diff
@@ -79,61 +96,117 @@ impl Bounds {
...
+            _ => Some(Into(Some('_'), None)),
         }
     }
     fn to_glib_extra(bound_type: BoundType) -> String {
         use self::BoundType::*;
         match bound_type {
-            AsRef => ".as_ref()".to_owned(),
-            Into(_) => ".into()".to_owned(),
+            AsRef(_) => ".as_ref()".to_owned(),
+            //Into(_, None) => ".into()".to_owned(),
```
EPashkin: Unneeded?

src/analysis/bounds.rs
```diff
@@ -79,61 +96,117 @@ impl Bounds {
...
+        let type_str = if let Some((Some(sub), lifetime)) = sub {
+            if let Some(alias) = self.unused.pop_front() {
+                self.used.push(Bound {
+                    bound_type: BoundType::set_lifetime(*sub, lifetime),
+                    parameter_name: name.to_owned(),
+                    alias: alias,
+                    type_str: type_str.to_owned(),
+                    info_for_next_type: true,
+                });
+                format!("{}", alias)
```
EPashkin: `char` have `to_string` method

src/codegen/function_body_chunk.rs
```diff
@@ -117,6 +119,32 @@ impl Builder {
                 chunks.push(Chunk::AssertInitializedAndInMainThread),
         }
     }
+    fn add_into_conversion(&self, chunks: &mut Vec<Chunk>) {
+        for param in self.parameters.iter().filter(|x| match **x {
```
EPashkin: `filter_map`?

src/codegen/parameter.rs
```diff
@@ -20,12 +20,16 @@ impl ToParameter for Parameter {
...
-                            type_str = format!("Option<&{}{}>", mut_str, t)
+                        BoundType::IsA(_) => if *self.nullable {
+                            // should not happen!
+                            type_str = String::new()
                         } else {
                             type_str = format!("&{}{}", mut_str, t)
                         },
-                        BoundType::AsRef | BoundType::Into(_) => type_str = t.to_string(),
+                        BoundType::Into(_, Some(_)) => {
+                            type_str = format!("{}", t)
```
EPashkin: char to string?

src/codegen/translate_to_glib.rs
```diff
@@ -12,9 +12,10 @@ impl TranslateToGlib for chunk::parameter_ffi_call_in::Parameter {
         use analysis::conversion_type::ConversionType::*;
         match ConversionType::of(library, self.typ) {
             Direct => self.name.clone(),
-            Scalar => format!("{}{}", self.name, ".to_glib()"),
+            Scalar => format!("{}{}{}",
+                              self.name, if !*self.nullable { "" } else { ".into()" }, ".to_glib()"),
             Pointer => {
-                let (left, right) = to_glib_xxx(self.transfer, self.ref_mode);
+                let (left, right) = to_glib_xxx(self.transfer, self.ref_mode, self.is_into);
```
EPashkin: IMHO switch by `self.is_into` better be here, not in to_glib_xxx

EPashkin:
After fast look it great, will look more latter

src/analysis/bounds.rs
```diff
@@ -2,19 +2,32 @@ use std::collections::vec_deque::VecDeque;
...
+    // lifetime
+    IsA(Option<char>),
+    // lifetime <- shouldn't be used but just in case...
+    AsRef(Option<char>),
+    // lifetime (if none, not a reference, like for primitive types), wrapper type
+    Into(Option<char>, Option<Box<BoundType>>),
+}
+
+impl BoundType {
+    fn set_lifetime(ty_: BoundType, lifetime: char) -> BoundType {
```
EPashkin: with_lifetime?

src/analysis/bounds.rs
```diff
@@ -79,61 +96,117 @@ impl Bounds {
...
         }
     }
     fn to_glib_extra(bound_type: BoundType) -> String {
         use self::BoundType::*;
         match bound_type {
-            AsRef => ".as_ref()".to_owned(),
-            Into(_) => ".into()".to_owned(),
+            AsRef(_) => ".as_ref()".to_owned(),
+            //Into(_, None) => ".into()".to_owned(),
+            Into(_, Some(x)) => format!("{}", Bounds::to_glib_extra(*x)),
```
EPashkin: to_glib_extra return string

GuillaumeGomez:
Thanks for your review! I'll try to update it as soon as possible.

src/codegen/translate_to_glib.rs
GuillaumeGomez: Hum... We can't do it on every parameter so we still need to check the enum variant. I think it's good as is (but maybe I misunderstood you, if so, can you explain me what you have in mind more precisely please?).

EPashkin: You right, I somewhat missed difference when I checked last time.

src/analysis/parameter.rs
```diff
@@ -67,5 +69,16 @@ pub fn analyze(env: &Env, par: &library::Parameter, configured_functions: &[&Fun
         ref_mode: ref_mode,
         is_error: par.is_error,
         to_glib_extra: String::new(),
+        is_into: *nullable && !par.instance_parameter && match *env.library.type_(par.typ) {
```
GuillaumeGomez: Filename is already generated with `Into`, so trying to put it again is a bad idea (the generated code doesn't compile actually).

GuillaumeGomez:
Updated.

@EPashkin: When you're okay with the changes, I'll squash my commits. :)

EPashkin:
👍 Thanks

GuillaumeGomez:
Squashed. Once merged, I'll update regenerate the other crates as well.

EPashkin:
Before regenerate IMHO better solve #328 as well.
Seems author gone away, so maybe I add PR without problem lines?

GuillaumeGomez:
I can ask him directly if you want.

EPashkin:
Do it please, no answers for half month :(

GuillaumeGomez:
Done, we'll see.

GuillaumeGomez:
@EPashkin: CIs passed. I let you handle the merge. :)

EPashkin:
Thanks.

GuillaumeGomez:
So should I regen gtk-rs crates with this since the other PR won't get merged soon?

EPashkin:
After you merge https://github.com/gtk-rs/gir/pull/358

GuillaumeGomez:
Well, it's a gir PR and you're the maintainer so I guess it's your call. 😉

EPashkin:
Ok, I merged it.
Only possible delay is https://github.com/gtk-rs/gtk/issues/488 but as it change only one file it can be added later.
So seems all ready for mass-regen

GuillaumeGomez:
Let's go!

PR #328, opened 2017-03-31
Add default values for fundamental types and Records

aruiz:
Lack of defaults for a bunch of types made binding auto generation more difficult

### DIFF
--- a/src/analysis/properties.rs
+++ b/src/analysis/properties.rs
@@ -174,7 +174,6 @@ pub fn get_type_default_value(env: &Env, type_tid: library::TypeId, type_: &libr
                 Fundamental::Utf8 => some("None::<&str>"),
                 Fundamental::Float => some("&0f32"),
                 Fundamental::Double => some("&0f64"),
-                Fundamental::None => some("&0"),
                 Fundamental::Int8 => some("&0i8"),
                 Fundamental::UInt8 => some("&0u8"),
                 Fundamental::Int16 => some("&0i16"),
--- a/src/analysis/properties.rs
+++ b/src/analysis/properties.rs
@@ -174,12 +174,33 @@ pub fn get_type_default_value(env: &Env, type_tid: library::TypeId, type_: &libr
                 Fundamental::Utf8 => some("None::<&str>"),
                 Fundamental::Float => some("&0f32"),
                 Fundamental::Double => some("&0f64"),
+                Fundamental::None => some("&0"),
+                Fundamental::Int8 => some("&0i8"),
+                Fundamental::UInt8 => some("&0u8"),
+                Fundamental::Int16 => some("&0i16"),
+                Fundamental::UInt16 => some("&0u16"),
+                Fundamental::Int32 => some("&0i32"),
+                Fundamental::UInt32 => some("&0u32"),
+                Fundamental::Int64 => some("&0i64"),
+                Fundamental::UInt64 => some("&0u64"),
+                Fundamental::Char => some("&0i8"),
+                Fundamental::UChar => some("&0u8"),
+                Fundamental::Size => some("&0isize"),
+                Fundamental::SSize => some("&0usize"),
+                Fundamental::Pointer => some("::std::ptr::null_mut()"),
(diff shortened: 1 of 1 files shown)


src/analysis/properties.rs
```diff
@@ -174,12 +174,33 @@ pub fn get_type_default_value(env: &Env, type_tid: library::TypeId, type_: &libr
                 Fundamental::Utf8 => some("None::<&str>"),
                 Fundamental::Float => some("&0f32"),
                 Fundamental::Double => some("&0f64"),
+                Fundamental::None => some("&0"),
```
EPashkin: You expect property of type: None ?

src/analysis/properties.rs
```diff
@@ -174,12 +174,33 @@ pub fn get_type_default_value(env: &Env, type_tid: library::TypeId, type_: &libr
...
+                Fundamental::UInt16 => some("&0u16"),
+                Fundamental::Int32 => some("&0i32"),
+                Fundamental::UInt32 => some("&0u32"),
+                Fundamental::Int64 => some("&0i64"),
+                Fundamental::UInt64 => some("&0u64"),
+                Fundamental::Char => some("&0i8"),
+                Fundamental::UChar => some("&0u8"),
+                Fundamental::Size => some("&0isize"),
+                Fundamental::SSize => some("&0usize"),
+                Fundamental::Pointer => some("::std::ptr::null_mut()"),
```
EPashkin: Gir currently don't support pointers so this will be unusable

aruiz: I precisely added this to support a pointer paramspec property in gdbus (it actually is a function, but there is g_param_spec_function)

Pointers should always default to NULL anyway

src/analysis/properties.rs
aruiz: I just copy pasted all the fundamentals, if you think there are some we should skip I'm fine with removing them.

src/analysis/properties.rs
```diff
@@ -174,12 +174,33 @@ pub fn get_type_default_value(env: &Env, type_tid: library::TypeId, type_: &libr
...
+                /* We are assuming that these will just work by setting them to 0 */
+                Fundamental::Short => some("&0"),
```
EPashkin: IMHO this won't work as `g_object_get_property` normally check of type before assigning,
so this need some conversion to`libc::c_short` and `libc::c_long`

src/analysis/properties.rs
```diff
@@ -174,12 +174,33 @@ pub fn get_type_default_value(env: &Env, type_tid: library::TypeId, type_: &libr
...
+                Fundamental::Long => some("&0"),
+                Fundamental::ULong => some("&0"),
+                Fundamental::Type => some("&gobject_sys::G_TYPE_NONE"),
                 _ => None,
             }
         }
         Type::Bitfield(_) => some("&0u32"),
         Type::Enumeration(_) => some("&0"),
         Type::Class(..) |
+        Type::Record(..) |
```
EPashkin: Strange I though we already have it for `RGBA`  ex. `TextTag::get_property_background_rgba`, was wrong, we only generated `set_property_background_rgba`
I check if it work at evening.

EPashkin:
Checked that adding `Record` getters generate fixable changes on gtk\gdk, mostly `RGBA` and `Rectangle`
@aruiz Only last to remove Fundamental::None if you don't have suitable property getter :wink:.
Also if you don't have property getters to runtime test Fundamental::Short etc. better remove it

EPashkin:
@aruiz Any progress with `Fundamental::Short`?

aruiz:
What do you want me to do with Short? remove it? Ushort as well?

EPashkin:
> Also if you don't have property getters to runtime test Fundamental::Short etc. better remove it
I mean Short, UShort. Long, ULong.

EPashkin:
@aruiz Seems you don't have time :(

GuillaumeGomez:
So @aruiz confirmed me that he was busy. So if you want to takeover this PR @EPashkin, just do so. :)

EPashkin:
Replaced by #358

PR #331, opened 2017-04-01
Fixed processing of properties "construct_only" flag

EPashkin:
Affected modules gio, gdk, gtk

### DIFF
--- a/src/analysis/properties.rs
+++ b/src/analysis/properties.rs
@@ -32,7 +32,6 @@ pub fn analyze(env: &Env, props: &[library::Property], type_tid: library::TypeId
     let mut properties = Vec::new();
 
     for prop in props {
-        if prop.construct_only { continue; }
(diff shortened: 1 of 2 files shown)


EPashkin:
By misprint gir generating many not working setters for properties like https://developer.gnome.org/gtk3/stable/GtkAssistant.html#GtkAssistant--use-header-bar

EPashkin:
@GuillaumeGomez I right about "construct-only" really not writable?

GuillaumeGomez:
I think yes. It's not intended to be modified after being initialized. Thanks!

EPashkin:
I will update gio, gdk, gtk after @aruiz fixed #328 and it merged.
May be in this time I also fix #332

GuillaumeGomez:
Ok! I'll try to go back on my PR as well in the next week.

PR #333, opened 2017-04-03
Add support for parsing unions in classes

sdroege:
See commit message. While this does not add proper support for unions yet, it at least allows the parser to not fail on Gst-1.0.gir.

Proper union support is blocked by https://github.com/rust-lang/rust/issues/32836

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -203,6 +203,11 @@ impl Library {
                             fields.push(try!(self.read_field(parser, ns_id, &attributes))),
                         "virtual-method" => try!(ignore_element(parser)),
                         "doc" => doc = try!(read_text(parser)),
+                        "union" => {
+                            let (union_fields, _, doc) = try!(self.read_union(parser, ns_id));
+                            let typ = Type::union(self, union_fields);
+                            fields.push(Field { typ: typ, doc: doc, ..Field::default() });
+                        }
                         x => bail!(mk_error!(format!("Unexpected element <{}>", x), parser)),
                     }
                 }


EPashkin:
When applied to v1.10 it add to end of `GstPadProbeInfo`, `GstControlBinding`, `GstPad`
```rust
    //union,
```
As it NOP I don't see any problem with merging this.
@GuillaumeGomez What you think?

sdroege:
The problem I see here is if such a struct is embedded in another one. The offsets of all following fields will be wrong, e.g.

```rust
#[repr(C)]
pub struct GstPad {
    pub object: GstObject,
    [...]
    //union,
}
#[repr(C)]
pub struct GstProxyPad {
    pub pad: GstPad,
    priv_: *mut GstProxyPadPrivate,
}
```

However it allows at least for something to be generated.

EPashkin:
@sdroege, you right.
`_truncated_record_marker` mean that structure truncated and it size smaller that it need to be.
In gtk we don't need direct access to class struct so we ignored this problem.

GuillaumeGomez:
@EPashkin: I didn't see any problem but seems like it could create problems in the future. So your call.

EPashkin:
@sdroege Thanks

sdroege:
@EPashkin In GStreamer it's needed, but as any of this would have to happen in unsafe code anyway whoever writes that code should be aware of that anyway :) And having it parsed is better than nothing, at least it allows me to switch away from manually written FFI in https://github.com/sdroege/rsplugin at some point.

PR #335, opened 2017-04-04
Use FFI instead of Rust types for constants too

sdroege:
Otherwise we will make them to possibly incorrect Rust types, which do
not exist, and by this we map them to the same types as used inside
structs, function parameters, etc.

Case in question was GStreamer's GST_CLOCK_TIME_NONE which is of type
GstClockTime in C. By using the Rust type this would map to "ClockTime",
which does not exist in the FFI bindings.

### DIFF
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -157,10 +157,11 @@ fn generate_constants(w: &mut Write, env: &Env, constants: &[Constant]) -> Resul
             type_ = "&'static str".into();
             value = format!("r##\"{}\"##", value)
         } else if type_ == "gboolean" {
+            let prefix = if env.config.library_name == "GLib" { "" } else { "glib::" };
             if value == "true" {
-                value = "glib::GTRUE".into();
+                value = format!("{}GTRUE", prefix);
             } else {
-                value = "glib::GFALSE".into();
+                value = format!("{}GFALSE", prefix);
             }
         }
         try!(writeln!(w, "{}pub const {}: {} = {};", comment,
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -155,10 +155,13 @@ fn generate_constants(w: &mut Write, env: &Env, constants: &[Constant]) -> Resul
         let mut value = constant.value.clone();
         if type_ == "*mut c_char" {
             type_ = "&'static str".into();
-            value = format!("r##\"{}\"##", value);
-        } else if type_ == "Glyph" && env.config.library_name == "Pango"  {
-            //Fix single constant alias
-            type_ = "PangoGlyph".into();
+            value = format!("r##\"{}\"##", value)
+        } else if type_ == "gboolean" {
+            if value == "true" {
+                value = "glib::GTRUE".into();
+            } else {
+                value = "glib::GFALSE".into();
+            }
         }
         try!(writeln!(w, "{}pub const {}: {} = {};", comment,
             constant.c_identifier, type_, value));
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -2,8 +2,6 @@ use std::collections::HashMap;
 use std::io::{Result, Write};
 use case::CaseExt;
 
-use analysis::ref_mode::RefMode;
-use analysis::rust_type::parameter_rust_type;
 use codegen::general::{self, version_condition};
 use env::Env;
 use file_saver::*;
@@ -145,11 +143,8 @@ fn generate_constants(w: &mut Write, env: &Env, constants: &[Constant]) -> Resul
         try!(writeln!(w, "// Constants"));
     }
     for constant in constants {
-        let direction = ParameterDirection::In;
-        let ref_mode = RefMode::of(&env.library, constant.typ, direction);
         let (mut comment, mut type_) =
-            match parameter_rust_type(env, constant.typ, direction,
-                                      Nullable(false), ref_mode) {
+            match ffi_type(env, constant.typ, &constant.c_type) {
                 Ok(x) => ("", x),
                 x @ Err(..) => ("//", x.into_string()),
             };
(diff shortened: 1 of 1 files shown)


src/codegen/sys/lib_.rs
```diff
@@ -158,7 +153,7 @@ fn generate_constants(w: &mut Write, env: &Env, constants: &[Constant]) -> Resul
             comment = "//";
         }
         let mut value = constant.value.clone();
-        if type_ == "&str" {
+        if type_ == "*mut c_char" {
             type_ = "&'static str".into();
             value = format!("r##\"{}\"##", value);
         } else if type_ == "Glyph" && env.config.library_name == "Pango"  {
```
sdroege: This was probably also the reason for the Pango special case here. Does someone remember/know?

sdroege:
The problem with this here now is that this changes types. What was a i32 constant before will be now a c_int. Arguably that's more correct though.

sdroege:
It also breaks the build for various things, but e.g.
```rust
pub const G_SOURCE_REMOVE: bool = false;
```
seems suboptimal anyway. It's a gboolean (which is a c_int), and which is not necessarily the same as a Rust bool.

src/codegen/sys/lib_.rs
EPashkin: Yes, it was due Rust types used. Can be removed in this PR.

EPashkin:
@sdroege Seems for `gboolean` constants you need use `GTRUE` in case of glib or `glib::GTRUE` for all others, may be it need changed just to 0/1, but I prefer named. Not fully sure about 1, but `GTRUE` defined that way.
All other changes in https://github.com/gtk-rs/sys look good for me, thanks

sdroege:
Yes it would have to be GTRUE, and that's better/more descriptive than 0/1 :) I'll change that if you generally agree with this change and are not worried about the API changes in all generated -sys bindings.

sdroege:
And one thing that is also not ideal yet IMHO is that C string literal constants are converted to Rust string literals. They're unfortunately not equivalent, the latter missing the terminating '\0'. So for directly passing the string constants to C functions you still need to go through CString.

They could probably be changed to ASCII string literals in Rust: b"blablabla\0"

EPashkin:
Int constant IMHO stays same.
`bool` constants seems don't used in RUST code as all gtk builds after `GTRUE` fix.
So I agree with this change.

About non raw literals: currently I think that `"text".to_glib_none().0` is normal in fii call and adding '\0' not needed.

cc @GuillaumeGomez

GuillaumeGomez:
So I'm not sure to follow completely in here. As I understood, it means that it'll expose c types like `c_int`? If that's the case I'm against it. Otherwise it's fine for me.

sdroege:
> About non raw literals: currently I think that "text".to_glib_none().0 is normal in fii call and adding '\0' not needed.

Yes but that involves dynamic memory allocations, or not? Which seems unnecessary if these constants are just being passed to the C functions anyway.

> So I'm not sure to follow completely in here. As I understood, it means that it'll expose c types like c_int?

It does expose c_int for constants that were i32 before, yes. Just like it already uses c_int for function arguments and struct fields.

If I'm not mistaken this is also more correct as c_int is guaranteed to be the same as a C "int", but a C "int" might not be the same as a "i32". (It however is for all platforms currently supported by GLib).

sdroege:
>>    About non raw literals: currently I think that "text".to_glib_none().0 is normal in fii call and adding '\0' not needed.

> Yes but that involves dynamic memory allocations, or not? Which seems unnecessary if these constants are just being passed to the C functions anyway.

One solution for this would be to append an explicit \0 to the string constants. Then you could directly use str::as_bytes() and pass that to the C functions without any dynamic memory allocations.

GuillaumeGomez:
Actually I never paid attention to our string constants, I'm suprised to see how we handle them: why not just writing for example:

```rust
b"text\0".as_ptr() as *const c_char
```

? (just as @sdroege proposed)

sdroege:
I'm working on a pull request for that one btw, a bit tricky in the details.

EPashkin:
I not against use binary string, just don't forget escape it in case. This was removed in https://github.com/gtk-rs/gir/pull/84/commits/9defa38e7dd18c0818e2560b7534ebb9ea87e068

sdroege:
With my pending 3 pull requests I have my "gstreamer-sys" building without errors and warnings btw.

So the remaining question here is whether using c_int instead of i32 is acceptable for constants. I'd say yes, as a) the same is done in structs/functions already and b) it's more correct (a C int is a c_int, not necessarily a i32). @GuillaumeGomez what is your reasoning for preferring i32 here?

GuillaumeGomez:
My only problem is for functions/methods arguments. I have no issue with constants.

sdroege:
For functions it is already using c_int before my changes

GuillaumeGomez:
I know... So if your change is only for constant, it's ok for me. I thought it was more global, my bad.

sdroege:
Great, thanks for looking! I'll fix the remaining unit test failures then

src/codegen/sys/lib_.rs
```diff
@@ -158,12 +153,15 @@ fn generate_constants(w: &mut Write, env: &Env, constants: &[Constant]) -> Resul
...
+        if type_ == "*mut c_char" {
             type_ = "&'static str".into();
-            value = format!("r##\"{}\"##", value);
-        } else if type_ == "Glyph" && env.config.library_name == "Pango"  {
-            //Fix single constant alias
-            type_ = "PangoGlyph".into();
+            value = format!("r##\"{}\"##", value)
+        } else if type_ == "gboolean" {
+            if value == "true" {
+                value = "glib::GTRUE".into();
```
EPashkin: IMHO here need switch as glib don't need `glib::` (with `env.config.library_name` or https://github.com/gtk-rs/gir/blob/074a1ca9ddd13ae82fccdfd74070696734f72fb5/src/analysis/namespaces.rs#L85 + library::MAIN_NAMESPACE)

EPashkin:
@sdroege Thanks

PR #338, opened 2017-04-05
Add support for correctly generating bitflag constants

sdroege:
> This depends (and includes https://github.com/gtk-rs/gir/pull/335 )

In the GIR they are given with an integer value but we have to convert
them to the corresponding bitflags type.

### DIFF
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -164,8 +164,14 @@ fn generate_constants(w: &mut Write, env: &Env, constants: &[Constant]) -> Resul
                 value = format!("{}GFALSE", prefix);
             }
         }
-        try!(writeln!(w, "{}pub const {}: {} = {};", comment,
-            constant.c_identifier, type_, value));
+
+        if let Some(_) = env.library.type_(constant.typ).maybe_ref_as::<Bitfield>() {
+            try!(writeln!(w, "{}pub const {}: {} = {2} {{ bits: {} }};", comment,
+                constant.c_identifier, type_, value));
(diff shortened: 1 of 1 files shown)


EPashkin:
@sdroege PR not affect your sys at all. Can you show diff it produced?

sdroege:
It affects mine after my fixes to GStreamer's GIR, see e.g. https://github.com/sdroege/gstreamer-sys/blob/master/gstreamer-sys/src/lib.rs#L866

This would otherwise be a plain integer, and that fails to compile obviously

src/codegen/sys/lib_.rs
```diff
@@ -158,15 +153,25 @@ fn generate_constants(w: &mut Write, env: &Env, constants: &[Constant]) -> Resul
...
+        } else if type_ == "gboolean" {
+            let prefix = if env.config.library_name == "GLib" { "" } else { "glib::" };
+            if value == "true" {
+                value = format!("{}GTRUE", prefix);
+            } else {
+                value = format!("{}GFALSE", prefix);
+            }
+        }
+
+        if env.library.type_(constant.typ).maybe_ref_as::<Bitfield>().is_some() {
```
EPashkin: Change to `if let` please.

EPashkin:
I think merge it after it rebasing (to clear diff).
cc @GuillaumeGomez

sdroege:
Done, rebased it and used if-let

GuillaumeGomez:
Good for me! Thanks!

EPashkin:
Thanks

PR #339, opened 2017-04-05
Create C string constants as *const c_char with trailing \0 instead of &'static str

sdroege:
> This depends on (and includes) https://github.com/gtk-rs/gir/pull/335

This reverts part of 9defa38e7dd18c0818e2560b7534ebb9ea87e068

Reason for this change is that it allows to use the constants directly
for calling C functions without dynamic memory allocations, and still
allows to convert them to a &str without any copying by using CStr.

### DIFF
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -157,10 +157,11 @@ fn generate_constants(w: &mut Write, env: &Env, constants: &[Constant]) -> Resul
             type_ = "&'static str".into();
             value = format!("r##\"{}\"##", value)
         } else if type_ == "gboolean" {
+            let prefix = if env.config.library_name == "GLib" { "" } else { "glib::" };
             if value == "true" {
-                value = "glib::GTRUE".into();
+                value = format!("{}GTRUE", prefix);
             } else {
-                value = "glib::GFALSE".into();
+                value = format!("{}GFALSE", prefix);
             }
         }
         try!(writeln!(w, "{}pub const {}: {} = {};", comment,
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -155,10 +155,13 @@ fn generate_constants(w: &mut Write, env: &Env, constants: &[Constant]) -> Resul
         let mut value = constant.value.clone();
         if type_ == "*mut c_char" {
             type_ = "&'static str".into();
-            value = format!("r##\"{}\"##", value);
-        } else if type_ == "Glyph" && env.config.library_name == "Pango"  {
-            //Fix single constant alias
-            type_ = "PangoGlyph".into();
+            value = format!("r##\"{}\"##", value)
+        } else if type_ == "gboolean" {
+            if value == "true" {
+                value = "glib::GTRUE".into();
+            } else {
+                value = "glib::GFALSE".into();
+            }
         }
         try!(writeln!(w, "{}pub const {}: {} = {};", comment,
             constant.c_identifier, type_, value));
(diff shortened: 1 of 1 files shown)


EPashkin:
We use this strings in rust part, good that in only in one place
https://github.com/gtk-rs/glib/blob/da33c35ac3e622c007c8d7720d8a6072d35aa21e/src/lib.rs#L132
so theoretically we can just add it manually.
cc @GuillaumeGomez

GuillaumeGomez:
As long as we're sure they'll never be used directly by users, it's good for me.

EPashkin:
On other side: end-user may will use constants like `gio_ffi::G_FILE_ATTRIBUTE_OWNER_USER` in functions of gio crate.

EPashkin:
Not sure that using ffi consts with ffi functions take priority on using normal consts and normal functions

sdroege:
> On other side: end-user may will use constants like gio_ffi::G_FILE_ATTRIBUTE_OWNER_USER in functions of gio crate.

Shouldn't the end-user always use some &'static str constant from gio, not gio_ffi? Which then gets translated from a &str to a C string as needed.

It seems to me like there should be FFI string constants (as generated here), and &str constants which would be used by the non-sys crates and exposed to the end-user.

EPashkin:
The we need add all string consts manually to gio and all other crates. IMHO it too much and reduce effect of generation.

sdroege:
Are you currently re-exporting (all?) constants from the -sys crate in the end-user bindings crate?

An alternative would be to create a &'static str with a trailing \0. Then you could still use it without dynamic memory allocations. It only looks ugly in the docs, etc.

EPashkin:
Currently we reexporting string in one crate.
Not only in docs though, it will be bad in all functions what accept size (not sure if there any).

sdroege:
What would you suggest then? Ideally the end-user bindings crate could just do something like the following

```rust
pub const SOME_CONSTANT: &'static str = CStr::from_ptr(whatever_ffi::SOME_CONSTANT);
```

That way both use cases would be covered. But it would require CStr::from_ptr() to be a const_fn and it requires to use nightly at this point.

EPashkin:
Don't see way do convert to `&'static str`.
My last attempt:
```rust
lazy_static! {
   pub static ref KEY_FILE_DESKTOP_GROUP_: CString = unsafe{CString::from_raw(ffi::G_KEY_FILE_DESKTOP_GROUP as *mut _)};
   pub static ref KEY_FILE_DESKTOP_GROUP: &'static str = &KEY_FILE_DESKTOP_GROUP_.to_string_lossy();
}

```

sdroege:
`CStr::from_raw(ptr).to_str().unwrap()` should work

EPashkin:
Thanks.
For definition
```
lazy_static! {
pub static ref KEY_FILE_DESKTOP_GROUP: &'static str = unsafe{CStr::from_ptr(ffi::G_KEY_FILE_DESKTOP_GROUP).to_str().unwrap()};
}
```
code
```
let s:&'static str = &glib::KEY_FILE_DESKTOP_GROUP;
println!("{}: '{}'", s.len(), s);
```
gives `13: 'Desktop Entry'`

EPashkin:
@GuillaumeGomez, seems this fixable and can be merged, only need extra PR for glib to fix sys difference.

GuillaumeGomez:
Like I said, as long as it doesn't impact users' API usage, it's fine for me.

EPashkin:
@sdroege thanks

sdroege:
@EPashkin Are you going to work on the changes needed in glib and others for exporting the constants correctly?

EPashkin:
Yes, I add PR's on glib and sys in hour or two.

sdroege:
Perfect, thanks! Can you CC me there?

EPashkin:
Sure

PR #341, opened 2017-04-08
Mark private struct fields in -sys bindings as public

sdroege:
This is needed for various structs in GStreamer because either
1) it's a private padding field, but the struct has to be declared
   as a constant to be useful (GstPluginDesc)
2) it's a semi-private field that in C would be accessed via a macro

This also makes the GValue type field public, and solves the following
FIXME comment: https://github.com/gtk-rs/glib/blob/master/src/value.rs#L144

### DIFF
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -366,8 +366,6 @@ fn generate_fields(env: &Env, struct_name: &str, fields: &[Field]) -> (Vec<Strin
             continue;
         }
 
-        let vis = if field.private { "" } else { "pub " };
-
         if let Some(ref c_type) = field.c_type {
             let name = mangle_keywords(&*field.name);
             let mut c_type = ffi_type(env, field.typ, c_type);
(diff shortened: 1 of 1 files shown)


src/codegen/sys/lib_.rs
```diff
@@ -394,11 +392,11 @@ fn generate_fields(env: &Env, struct_name: &str, fields: &[Field]) -> (Vec<Strin
                 if c_type.is_err() {
                     commented = true;
                 }
-                lines.push(format!("\t{}{}: {},", vis, name, c_type.into_string()));
+                lines.push(format!("\tpub{}: {},", name, c_type.into_string()));
```
EPashkin: Extra space needed

src/codegen/sys/lib_.rs
```diff
@@ -394,11 +392,11 @@ fn generate_fields(env: &Env, struct_name: &str, fields: &[Field]) -> (Vec<Strin
...
             }
             else {
-                lines.push(format!("\t{}{}: [{:?} {}],",
-                                   vis, name, field.typ, field.typ.full_name(&env.library)));
+                lines.push(format!("\tpub{}: [{:?} {}],",
```
EPashkin: Same, produced `pubstate: GdkModifierType`

EPashkin:
@GuillaumeGomez, you not against it? It can effect on sys docs.

sdroege:
It would cause all currently non-pub fields to be in the docs, or were you thinking of other side-effects for the docs?

EPashkin:
Only about appearance of non-public fields

GuillaumeGomez:
I don't see the interest to add `sys` at the end of `sys` structs. We can put the namespace before every sys struct.

EPashkin:
@GuillaumeGomez Sorry, but I don't understand how it related to this PR.

GuillaumeGomez:
@EPashkin: ah damn, I think I commented on the bad PR, my bad...

sdroege:
Also having these fields non-private in the -sys bindings shouldn't be much of a problem. We're in unsafe territory here anyway, and this just makes it easier (and safer!) to use those fields. You could otherwise always do raw pointer casting or transmute.

EPashkin:
@sdroege I merging this.
Please, Rebase other PRs as IMHO its will be broken

sdroege:
Thanks, will do

sdroege:
... and all rebased

PR #342, opened 2017-04-08
Only special-case structs named "Value" from the GObject library

sdroege:


EPashkin:
This not affect `gtk-rs\sys` so I merge this

sdroege:
Yes it only prevents that code to trigger wrongly on other libraries :)

PR #343, opened 2017-04-08
Handle pointers to unions and special-case GMutex and GWeakRef unions

sdroege:
This allows for more struct fields to be correctly generated.

PR #344, opened 2017-04-08
Add special handling for "padding unions" as used in GStreamer structs

sdroege:
> Depends on (and includes) https://github.com/gtk-rs/gir/pull/343

These look like

```C
    struct {
      [..]
    
      union {
        gpointer _gst_reserved[GST_PADDING];
        struct {
          gint some_new_field;
        } abi;
      } ABI;
    }
```

And can be directly replaced with the _gst_reserved variant of the enum. The *reserved* variant is always the biggest one, otherwise the C library broke its ABI.

### DIFF
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -365,7 +365,7 @@ fn generate_fields(env: &Env, struct_name: &str, fields: &[Field]) -> (Vec<Strin
     // TODO: GLib/GObject special cases until we have proper union support in Rust
     let is_gweakref = env.config.library_name == "GObject" && struct_name == "WeakRef";
 
-    for field in fields {
+    'fields: for field in fields {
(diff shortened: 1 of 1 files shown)


src/codegen/sys/lib_.rs
```diff
@@ -344,15 +362,46 @@ fn generate_fields(env: &Env, struct_name: &str, fields: &[Field]) -> (Vec<Strin
...
-        if !truncated && (is_union || is_bits) {
+        let is_ptr = {
+             if let Some(ref c_type) = field.c_type {
+                 !rustify_pointers(c_type).0.is_empty()
+             } else {
+                 false
+             }
+        };
+
+        // TODO: Special case for padding unions like used in GStreamer
```
EPashkin: Please, add class\record name for example

EPashkin:
I don't see any problem with this and #343, will merge it after #341 approved and merged.

EPashkin:
Thanks

PR #345, opened 2017-04-09
Add support for guintptr and gintptr

sdroege:


src/library.rs
```diff
@@ -132,7 +134,7 @@ pub enum Fundamental {
     Unsupported,
 }
 
-pub const FUNDAMENTAL: [(&'static str, Fundamental); 31] = [
+const FUNDAMENTAL: &'static [(&'static str, Fundamental)] = &[
```
sdroege: This is necessary because [T; 33] does not implement any traits anymore (and we need Iterator below), it only goes up to 32.

src/analysis/conversion_type.rs
```diff
@@ -38,6 +38,8 @@ impl ConversionType {
                 ULong => ConversionType::Direct,
                 Size => ConversionType::Direct,
                 SSize => ConversionType::Direct,
+                IntPtr => ConversionType::Direct,
+                UIntPtr => ConversionType::Direct,
```
EPashkin: Move it to end please

EPashkin: Wrong place, I mean in library

EPashkin:
Seems GValue can't hold of these
so change https://github.com/gtk-rs/glib/blob/master/src/value.rs#L535-L542 don't needed
aslo this not "conflict" with https://github.com/gtk-rs/gir/pull/328

sdroege:
I'll move it later. What do you mean with conflict?

EPashkin:
It added properties support for many types and if GValue has functions for gintptr then 328 or your need be changed to adding that.

sdroege:
Ah sure but that's irrelevant after I removed the GValue bits here

sdroege:
Like this?

EPashkin:
Sorry, I forgot about Unsupported case. This need be before it.

sdroege:
No problem :) Like now?

EPashkin:
Yes, Thanks

PR #346, opened 2017-04-09
Generate GMutex union with correct alignment constraints on 32 bit pl…

sdroege:
…atforms

While it is always 64 bits in size, on 32 bit platforms it only requires
32 bit alignment due to being expressed by 2x32 bit integers.

### DIFF
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -254,9 +254,11 @@ fn generate_unions(w: &mut Write, env: &Env, items: &[&Union])
         if let Some(ref c_type) = item.c_type {
             // TODO: GLib/GObject special cases until we have proper union support in Rust
             if env.config.library_name == "GLib" && c_type == "GMutex" {
-                // Two c_uint or a pointer => 64 bits on all
-                // platforms currently supported by GLib
-                try!(writeln!(w, "#[repr(C)]\npub struct {}(u64); // union", c_type));
+                // Two c_uint or a pointer => 64 bits on all platforms currently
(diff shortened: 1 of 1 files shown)


src/codegen/sys/lib_.rs
```diff
@@ -256,7 +256,8 @@ fn generate_unions(w: &mut Write, env: &Env, items: &[&Union])
             if env.config.library_name == "GLib" && c_type == "GMutex" {
                 // Two c_uint or a pointer => 64 bits on all
                 // platforms currently supported by GLib
-                try!(writeln!(w, "#[repr(C)]\npub struct {}(u64); // union", c_type));
+                try!(writeln!(w, "#[cfg(target_pointer_width = \"32\")]\n#[repr(C)]\npub struct {0}([u32; 2]);\n\
+                                  #[cfg(target_pointer_width = \"64\")]\n#[repr(C)]\npub struct {0}(*mut c_void);", c_type));
```
EPashkin: Why not leave 64bit version u64?

EPashkin: And IMHO need updated comment about alignment

sdroege: It's a pointer in C, so I thought it's cleaner to make it a pointer in Rust too

EPashkin: Ok, only comment last

sdroege: Ok, done :)

EPashkin:
Thanks

PR #347, opened 2017-04-09
Also handle "padding" in addition to "reserved" as a valid union-padd…

sdroege:
…ing field name

Same as the previous. "padding" is used by GStreamer elsewhere


EPashkin:
@sdroege Thanks, I merge it when CI done

Issue #123, opened 2015-09-24
Consider checking returned enums for validity


gkoz:
It might be a good idea to test if the enums we get from the other side have valid values. This could require using newtype `c_int` wrappers instead of enums in the sys layer and implementing conversions between that and Rust enums.

EPashkin:
Note: there was long discussion about enums in https://github.com/gtk-rs/gdk/pull/154

EPashkin:
Problem:
While enum copy with `FromGlib` implementation was intended to fix this issue, 
current [implementation](https://github.com/gtk-rs/gdk/blob/master/src/auto/enums.rs#L48-L66)
actually don't fix it as wrong ffi-side enum values actually [count as one of cases](https://github.com/gtk-rs/gdk/pull/154#issuecomment-296823161).
Also copying enum introduced enums with different layout from FFI-side enum and even different integer values for same name.

EPashkin:
Solution 1.
Just rexport enums from FFI-side.
Proposed by @johncf. He even added  PR for it #354.

EPashkin:
Solution 2.
Stay with copy but change implementation to match integers. This way default branch with panic can be added.

EPashkin:
Solution 3.
https://github.com/gtk-rs/gdk/pull/154#issuecomment-296994347
Proposed by @GuillaumeGomez. I don't understand it :sob:

GuillaumeGomez:
:rofl:

EPashkin:
@GuillaumeGomez, @sdroege as https://github.com/gtk-rs/gir/pull/354 appeared again, we need decide what to do.

GuillaumeGomez:
So to put it simply: we have the choice between safety and performance. So I propose we add both (the performance will be `_unchecked`).

sdroege:
What exactly is the proposal to do here, the API to be added?

johncf:
> So to put it simply: we have the choice between safety and performance.

If I remember correctly, the status quo right now is no more safer than what remains after the changes in #354.

GuillaumeGomez:
@johncf: I agree on this, that's why I propose to add both. The "safe" one will use `match` and the "performant" one will just cast.

johncf:
@GuillaumeGomez That change can come later after we land these changes.

Basically, [this](https://github.com/gtk-rs/gdk/pull/154#issuecomment-296826269) was what you had proposed to do while converting from integers to enums. That would be a pervasive change too with a more profound effect overall.

[...]

PR #354, opened 2017-04-24
Avoid redefining enums and flags in safe API side

johncf:
**WIP: Do not merge!**

### Crates to Update

- [ ] Glib
- [ ] Gio
- [x] [Gdk](https://github.com/gtk-rs/gdk/pull/154)
- [x] [Gtk](https://github.com/gtk-rs/gtk/pull/531)
- [x] [Pango](https://github.com/gtk-rs/pango/pull/78)
- [ ] Gdk-Pixbuf
- [ ] SourceView

### DIFF
--- a/src/analysis/conversion_type.rs
+++ b/src/analysis/conversion_type.rs
@@ -55,7 +55,7 @@ impl ConversionType {
             Bitfield(_) => ConversionType::Scalar,
             Record(_) => ConversionType::Pointer,
             Union(_) => ConversionType::Pointer,
-            Enumeration(_) => ConversionType::Scalar,
+            Enumeration(_) => ConversionType::Direct,
             Interface(_) => ConversionType::Pointer,
             Class(_) => ConversionType::Pointer,
             CArray(_) => ConversionType::Pointer,
--- a/src/analysis/conversion_type.rs
+++ b/src/analysis/conversion_type.rs
@@ -52,7 +52,7 @@ impl ConversionType {
                 Unsupported => ConversionType::Unknown,
             },
             Alias(ref alias) => ConversionType::of(library, alias.typ),
-            Bitfield(_) => ConversionType::Scalar,
+            Bitfield(_) => ConversionType::Direct,
             Record(_) => ConversionType::Pointer,
             Union(_) => ConversionType::Pointer,
             Enumeration(_) => ConversionType::Direct,
--- a/src/codegen/enums.rs
+++ b/src/codegen/enums.rs
@@ -1,192 +1,24 @@
 use analysis::namespaces;
-use case::CaseExt;
-use codegen::general::{self, version_condition, version_condition_string};
+use codegen::general::version_condition_string;
 use config::gobjects::GObject;
 use env::Env;
-use file_saver;
 use library::*;
-use std::collections::HashSet;
-use std::io::prelude::*;
-use std::io::Result;
 use std::path::Path;
-use traits::*;
-use version::Version;
 
-pub fn generate(env: &Env, root_path: &Path, mod_rs: &mut Vec<String>) {
+pub fn generate(env: &Env, _root_path: &Path, mod_rs: &mut Vec<String>) {
     let configs: Vec<&GObject> = env.config.objects.values()
         .filter(|c| {
             c.status.need_generate() &&
                 c.type_id.map_or(false, |tid| tid.ns_id == namespaces::MAIN)
         })
         .collect();
-    let mut has_get_quark = false;
-    let mut has_any = false;
+    mod_rs.push(String::from(""));
     for config in &configs {
         if let Type::Enumeration(ref enum_) = *env.library.type_(config.type_id.unwrap()) {
-            has_any = true;
-            if get_error_quark_name(enum_).is_some(){
-                has_get_quark = true;
-                break;
+            if let Some (cfg) = version_condition_string(env, enum_.version, false, 0) {
+                mod_rs.push(cfg);
             }
+            mod_rs.push(format!("pub use ffi::{} as {};", enum_.c_type, enum_.name));
         }
     }
-
-    let path = root_path.join("enums.rs");
-    file_saver::save_to_file(path, env.config.make_backup, |w| {
-        try!(general::start_comments(w, &env.config));
-        try!(writeln!(w, ""));
-        try!(writeln!(w, "use ffi;"));
-        if env.namespaces.glib_ns_id == namespaces::MAIN {
-            if has_get_quark {
-                try!(writeln!(w, "use ffi as glib_ffi;"));
-                try!(writeln!(w, "use error::ErrorDomain;"));
-            }
-            try!(writeln!(w, "use translate::*;"));
-        } else {
-            if has_get_quark {
-                try!(writeln!(w, "use glib_ffi;"));
-                try!(writeln!(w, "use glib::error::ErrorDomain;"));
-            }
-            try!(writeln!(w, "use glib::translate::*;"));
-        }
-        try!(writeln!(w, ""));
-
-        if has_any {
-            mod_rs.push("\nmod enums;".into());
-        }
-        for config in &configs {
-            if let Type::Enumeration(ref enum_) = *env.library.type_(config.type_id.unwrap()) {
-                if let Some (cfg) = version_condition_string(env, enum_.version, false, 0) {
-                    mod_rs.push(cfg);
-                }
-                mod_rs.push(format!("pub use self::enums::{};", enum_.name));
-                try!(generate_enum(env, w, enum_, config));
-            }
-        }
-
-        Ok(())
-    });
-}
-
-fn generate_enum(env: &Env, w: &mut Write, enum_: &Enumeration, config: &GObject) -> Result<()> {
-    struct Member {
-        name: String,
-        c_name: String,
-        version: Option<Version>,
-    }
-
-    let mut members: Vec<Member> = Vec::new();
-    let mut vals: HashSet<String> = HashSet::new();
-
-    for member in &enum_.members {
-        let member_config = config.members.matched(&member.name);
-        let is_alias = member_config.iter().any(|m| m.alias);
-        if is_alias || vals.contains(&member.value) {
-            continue
-        }
-        vals.insert(member.value.clone());
-        let version = member_config.iter().filter_map(|m| m.version).next();
-        members.push(Member {
-            name: member.name.to_camel(),
-            c_name: member.c_identifier.clone(),
-            version: version,
-        });
-    }
-
-    try!(version_condition(w, env, enum_.version, false, 0));
-    try!(writeln!(w, "#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]"));
-    try!(writeln!(w, "pub enum {} {{", enum_.name));
-    for member in &members {
-        try!(version_condition(w, env, member.version, false, 1));
-        try!(writeln!(w, "\t{},", member.name));
-    }
-    try!(writeln!(w, "{}", "    #[doc(hidden)]
-    __Nonexhaustive(()),
-}
-"));
-
-    try!(version_condition(w, env, enum_.version, false, 0));
-    try!(writeln!(w, "#[doc(hidden)]
-impl ToGlib for {name} {{
-    type GlibType = ffi::{ffi_name};
-
-    fn to_glib(&self) -> ffi::{ffi_name} {{
-        match *self {{", name = enum_.name, ffi_name = enum_.c_type));
-    for member in &members {
-        try!(version_condition(w, env, member.version, false, 3));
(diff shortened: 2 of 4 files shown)


sdroege:
You probably want to do the same thing for flags then. Otherwise seems like a good idea to me, but it should still be possible to manually override the enums/flags.

johncf:
@sdroege I'm not sure why/how, but I got the notification to your comment today!!

Anyways, I'll try to do the same for flags soon...

> but it should still be possible to manually override the enums/flags.

I don't understand this. Are you referring to an unsafe `transmute` being a possibility?

sdroege:
> I don't understand this. Are you referring to an unsafe transmute being a possibility?

No, I meant that via configuration it should be still possible to mark the enum/flag e.g. as "manual" or to not generate it at all, and then implement it manually (or not at all). Your patch probably covers that case already

johncf:
Yes, I believe those checks are dealt with [prior](https://github.com/gtk-rs/gir/search?utf8=%E2%9C%93&q=gstatus&type=) [to](https://github.com/gtk-rs/gir/search?utf8=%E2%9C%93&q=type_status&type=) [reaching](https://github.com/gtk-rs/gir/blob/1644ef1/src/analysis/supertypes.rs) the codegen module.

johncf:
Updated to remove redefinitions of bitflags.

johncf:
Also updated https://github.com/gtk-rs/gdk/pull/154 to witness the effect of this change.

johncf:
@GuillaumeGomez @sdroege @EPashkin As of now, Gdk, Gtk, and Pango have been updated to reflect these changes. Can anyone test these locally? I'll proceed with more libraries once you're able to verify these changes to be working correctly.

GuillaumeGomez:
I won't be available starting this evening until the 24th of june. But after this sure. :D

EPashkin:
I check how changes works on Windows at tomorrow.

sdroege:
So the main thing we're *losing* now is that previously unknown enum values would panic (which as I said in the other issue can also happen because of something that is not a bug!), now they would cause undefined behaviour.

It would make sense to add a
```Rust
#[doc(hidden)]
__Nonexhaustive
```
like in the previously generated enum.rs to the sys enums, to prevent users from not handling this case and having it propagate to *safe* code as undefined behaviour (if you do something that looks like an exhaustive match on the enum to compiler, but actually isn't due to unknown enum values, things can go very wrong).

johncf:
> So the main thing we're losing now is that previously unknown enum values would panic

That's incorrect. Because, as of now, we use `mem::transmute` to convert to a sys enum ([example](https://github.com/gtk-rs/gdk/blob/master/src/auto/device_tool.rs#L67)), whose behavior is already undefined if the enum that it is transmuted to does not have a corresponding value. So adding a `__Nonexhaustive(())` variant really serves little purpose. It _might_ be possible to come up with an example where this undefined behavior is caught by the `__Nonexhaustive` variant to cause panic instead, but the fact remains that it is undefined and not a "solution" to the problem.

> to prevent users from not handling this case and having it propagate to safe code as undefined behaviour

I'd prefer to pay the cost of checking for correct versions and API compatibility elsewhere (eg. during compilation, or during dynamic library load) instead of every time a method is called at runtime.

sdroege:
Ah you're right, it's all bad already independent of your change :) Let's discuss that in the other ticket then.

Nonetheless, adding the __Nonexhaustive variant to the sys enums would be useful (as that's how the non-sys enum duplicate is defined right now) to prevent users from missing the `_` match and then having their code fail later if new enum values are added.

johncf:
> Nonetheless, adding the __Nonexhaustive variant to the sys enums would be useful (as that's how the non-sys enum duplicate is defined right now) to prevent users from missing the _ match and then having their code fail later if new enum values are added.

Oh I see... But wouldn't this change complicate things a bit. Wouldn't a cleaner solution be to add `#[cfg(feature = "v3_22")]` tags on variants introduced later? This way, if the user chooses to compile with features of a later version, it would produce a compile error, but not compiling with that feature would still be fine.

sdroege:
The problem with the features is
1) It is user configuration, it doesn't happen automatically if you compile against e.g. GTK >= 3.22 that the v3_22 feature is enabled.
2) The Rust bindings might be generated for GTK 3.22 but GTK 3.24 adds a new enum variant. There is no reason why the Rust bindings should stop to work, after all the new GTK version is backwards compatible

Why do you think this change (adding `__Nonexhaustive`) would complicate things a bit? The enums (the duplicates) in the non-sys crates already have the `__Nonexhaustive` variant (before your change), so all application code should be fine already. Only the sys crates didn't, but adding it there shouldn't cause many problems and adding it there would allow the re-exporting of the enums to provide exactly the same API.

EPashkin:
There problem with flags: just reexporting ffi version don't reeexport its items :(

johncf:
> It is user configuration, it doesn't happen automatically if you compile against e.g. GTK >= 3.22 that the v3_22 feature is enabled.

I thought that was more convenient, since there wouldn't be any unexpected API incompatibility.

> The Rust bindings might be generated for GTK 3.22 but GTK 3.24 adds a new enum variant. There is no reason why the Rust bindings should stop to work, after all the new GTK version is backwards compatible

It wouldn't stop working, right? Gtk 3.24 adds a new enum variant, but since I'm still compiling with 3.22 feature, so 3.24 enum variant will not be included in the enum.

---
But there's another problem with features that I'm not sure how to handle, which is: how would a Gtk-dependent (e.g. sourceview, or my own custom widget) library check for the feature version enabled for the gtk crate? If there's not straightforward solution for this, then it might have to be abandoned anyway.

---
> Why do you think this change (adding __Nonexhaustive) would complicate things a bit?

Oh I was actually thinking about removing `transmute` too when I wrote that. So the minor complications were that:

1. How to choose the numeric value of `__Nonexhaustive`? (max + 1?)
2. If panicking should be avoided, how can it be passed to the ffi side later?
3. [meta] Adding a `transmute` replacement will nullify the reduction in lines due to the current PR.

---

@EPashkin Woops! Should I export `ffi::flags::*`, or `pub use ffi::flags`?

EPashkin:
IMHO it need be something like this (almost same as it was before)
```
pub use ffi::GtkTreeModelFlags as TreeModelFlags;
pub use ffi::TREE_MODEL_ITERS_PERSIST;
pub use ffi::TREE_MODEL_LIST_ONLY;
```

johncf:
@EPashkin Ah, yeah! I haven't checked it yet. I'll make the appropriate changes, once this discussion reaches a conclusion.

sdroege:
> It wouldn't stop working, right? Gtk 3.24 adds a new enum variant, but since I'm still compiling with 3.22 feature, so 3.24 enum variant will not be included in the enum.

The C library will potentially pass the new enum variant to the Rust code, which then ends up not being mappable to the Rust enum

> how would a Gtk-dependent (e.g. sourceview, or my own custom widget) library check for the feature version enabled for the gtk crate?

You can require/enable features in your dependencies in the Cargo.toml

> So the minor complications were that:
>    1) How to choose the numeric value of __Nonexhaustive? (max + 1?)
>    2) If panicking should be avoided, how can it be passed to the ffi side later?
>    3) [meta] Adding a transmute replacement will nullify the reduction in lines due to the current PR.

Indeed, we need to find a numeric value for that because it's a repr(C) enum. And there is none that is safe to choose because the C library might at any point in the future use exactly that one.

And we indeed can't pass that to FFI, unless we make it `__Nonexhaustive(u32)` or similar (but then the enum can't be repr(C) anymore).

johncf:
> The C library will potentially pass the new enum variant to the Rust code, which then ends up not being mappable to the Rust enum

This would mean that the C library function which returned that enum variant was not backwards-compatible.

> You can require/enable features in your dependencies in the Cargo.toml

I meant that wouldn't the gtk-dependent library (eg. sourceview) need to know the Gtk features the user-binary enabled in the binary's Cargo.toml. Is that possible? To expand further on that, say I (sourceview) want to match on an enum that has an added variant on newer version, how would I condition that variant, without the ability to see the Gtk features that my library's user would enable? I don't think there's an automatic solution here. I (sourceview) would also have to give the user with a feature set that the user need to enable manually along with Gtk features, and both would have to be compatible.

> Maybe duplicating the enum on the Rust side is not that bad after all (especially because autogenerated) compared to the loss of safety and possibilities to "work around" those.

And we are back to square one! :sob: Fitting a large C library into idiomatic Rust library is exhausting! (Even though this was just a minor detail!)

EPashkin:
@johncf You can define features to your library that required need version from gtk and limit function usage by this.

johncf:
@EPashkin Is it possible to provide something like the following: (`lib.rs`)
```rust
#[cfg(gtk feature = "v3_22")]
fn twenty_two() -> ... { ... }

fn default() -> ... { ... }
```
A user should be able to use this library even without enabling the `v3_22` feature _on the Gtk crate_ (`default` is always available). If `v3_22` is enabled on the Gtk crate, then this library would also provide the function `twenty_two`. I don't think this is possible. For this, this library should probably export another feature which the user can enable independently (which should then make the library depend on Gtk with `v3_22` crate. I'm not very sure whether this is possible either).

EPashkin:
It possible in this way:
in Cargo.toml
```toml
[features]
use_gtk_3_10 = ["gtk/v3_10"]
```
in lib:
```rust
#[cfg(feature = "use_gtk_3_10")]
fn twenty_two() -> ... { ... }

fn default() -> ... { ... }
```

sdroege:
>>    The C library will potentially pass the new enum variant to the Rust code, which then ends up not being mappable to the Rust enum

> This would mean that the C library function which returned that enum variant was not backwards-compatible.

It depends on the meaning of the enum. It can be backwards compatible in the same way that adding enum variants to a `__Nonexhaustive` enum in Rust can be backwards compatible. Consider e.g. an enum that defines a file type (JPEG, GIF, PNG) and a function that guesses what type of file you have. It would be ok to add a e.g. JPEG2000 here as long as the enum had a `__Nonexhaustive` variant (your users would match on `_` and handle those as "unsupported file type").

johncf:
@EPashkin Ah I see, thanks! I couldn't understand [the documentation](http://doc.crates.io/manifest.html#the-features-section) for this when I checked right before...

johncf:
@sdroege 

> It depends on the meaning of the enum. [..]

What I meant was that, if passing the same set of parameters that were passed in `v3_22` now passed to `v3_24` of the same function returned a different enum variant than before, that is fundamentally a backwards-incompatible change in that function.

sdroege:
> What I meant was that, if passing the same set of parameters that were passed in v3_22 now passed to v3_24 of the same function returned a different enum variant than before, that is fundamentally a backwards-incompatible change in that function.

Sure, but that would then also require all the processes for API/ABI changes in the C library and bring us into a completely different territory.

What do you think would be the problem in the "file type" example I gave above?

johncf:
> What do you think would be the problem in the "file type" example I gave above?

Let's say the function checks the type of a file given as the argument. In `v3_6`, it returned either `UNK`, `JPG` and `PNG`. Now in `v3_8`, it also supports `SVG`. Then I'd say it is a breaking change since passing `diag.svg` in `v3_6` is not compatible with that in `v3_8`.

sdroege:
> >  What do you think would be the problem in the "file type" example I gave above?

> Let's say the function checks the type of a file given as the argument. In v3_6, it returned either UNK, JPG or PNG. Now in v3_8, it also supports SVG. Then I'd say it is a breaking change since passing diag.svg in v3_6 is not compatible with that in v3_8.

On the C level it could be compatible, if documented that it might return new enum variants in the future. On the Rust level such a function would be compatible in the same way (and `__Nonexhaustive` so users have to handle that case!). What we now need to find is a way to make this work safely between both, which should be possible via conversion trait impls and having `__Nonexhaustive` in the Rust enum. And possibly having the conversion trait impl map "unknown" values `x` to `__Nonexhaustive(x)` or similar.

EPashkin:
@sdroege Seems you like https://github.com/gtk-rs/gir/issues/123#issuecomment-297091954 more that reexport. I right?

johncf:
@sdroege You're right, if it is documented that one of the variants (like `UNK`) might change in future, then it might be considered compatible ...in a sense. If you feel this is indeed the right way to go, please feel free to close these PRs.

sdroege:
> @sdroege Seems you like #123 (comment) more that reexport. I right?

Something like that, yes (but more than that). But the problem already exists now at the sys crates: we directly use the repr(C) enum there (in function declarations, etc), and that could lead to undefined behaviour if a value that is not inside the enum is returned from C.

> @sdroege You're right, if it is documented that one of the variants (like UNK) might change in future, then it might be considered compatible ...in a sense.

As long as you document that support for more can be added later, or just omit the UNK variant to begin with and instead use a GError to signal that, it seems fine.

> If you feel this is indeed the right way to go, please feel free to close these PRs.

Not sure about that, needs further discussion.

johncf:
I think closing this is the right thing to do now.

PR #358, opened 2017-04-29
Add default values for fundamental types and Records

EPashkin:
Replacement of #328

PR #359, opened 2017-04-30
Added doc_hidden option for functions, properties and child_properties

### LINKED ISSUE #356: Add config options to hide function doc
Gir should generate `#[doc(hidden)]` for functions marked with it.
- [ ] Child property ([main reason](https://github.com/gtk-rs/gtk/pull/496#issuecomment-298170503))
- [ ] Normal function
- [ ] Trampoline  ??

Proposed option example `hide_doc = true`

EPashkin:
Closes #356

### DIFF
--- a/src/codegen/child_properties.rs
+++ b/src/codegen/child_properties.rs
@@ -7,6 +7,7 @@ use env::Env;
 use library;
 use writer::primitives::tabs;
 use nameutil;
+use super::general::doc_hidden;
 use super::property_body;
 use traits::IntoString;
 use writer::ToCode;
@@ -32,6 +33,7 @@ fn generate_func(w: &mut Write, env: &Env, prop: &ChildProperty, in_trait: bool,
 
     try!(writeln!(w, ""));
 
+    try!(doc_hidden(w, prop.doc_hidden, comment_prefix, indent));
     let decl = declaration(env, prop, is_get);
     try!(writeln!(w, "{}{}{}{}{}", tabs(indent),
         comment_prefix, pub_prefix, decl, decl_suffix));
--- a/README.md
+++ b/README.md
@@ -72,6 +72,8 @@ cfg_condition = "mycond"
     version = "3.12"
     # prefixed function with #[cfg(mycond)]
     cfg_condition = "mycond"
+    # prefixed function with #[doc(hidden)]
+    doc_hidden = true
         # override for parameter
         [[object.function.parameter]]
         # filter by name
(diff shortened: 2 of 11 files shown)


EPashkin:
cc @GuillaumeGomez

src/codegen/child_properties.rs
```diff
@@ -32,6 +33,9 @@ fn generate_func(w: &mut Write, env: &Env, prop: &ChildProperty, in_trait: bool,
 
     try!(writeln!(w, ""));
 
+    if prop.doc_hidden {
```
GuillaumeGomez: Just a thought: why not putting this check into the `doc_hidden` function directly? Since you're doing this check every time, I think it'd make more sense.

GuillaumeGomez:
Just a little nit/concern/suggestion but otherwise 👍

src/codegen/child_properties.rs
EPashkin: Ok

EPashkin:
Updated, I merge when CI passed.

GuillaumeGomez:
Perfect, thanks! ;)

GuillaumeGomez:
Great! Now there is just the gtk issue remaining.

EPashkin:
You mean regen?
I off for hour - two.

GuillaumeGomez:
Yes, I'm talking about this error with the new gir version:

```
Compiling gtk v0.1.1 (file:///Users/imperio/rust/gtk)
error[E0277]: the trait bound `glib::Variant: glib::IsA<glib::Object>` is not satisfied
  --> src/auto/actionable.rs:75:25
   |
75 |         let mut value = Value::from(None::<&glib::Variant>);
   |                         ^^^^^^^^^^^ the trait `glib::IsA<glib::Object>` is not implemented for `glib::Variant`
   |
   = note: required because of the requirements on the impl of `glib::value::SetValueOptional` for `glib::Variant`
   = note: required because of the requirements on the impl of `std::convert::From<std::option::Option<&glib::Variant>>` for `glib::Value`
   = note: required by `std::convert::From::from`

error[E0277]: the trait bound `glib::Variant: glib::IsA<glib::Object>` is not satisfied
  --> src/auto/actionable.rs:79:15
   |
79 |         value.get()
   |               ^^^ the trait `glib::IsA<glib::Object>` is not implemented for `glib::Variant`
   |
   = note: required because of the requirements on the impl of `glib::value::FromValueOptional` for `glib::Variant`

error: aborting due to 2 previous errors

error: Could not compile `gtk`.
```

EPashkin:
Again https://github.com/gtk-rs/gir/pull/358#issuecomment-298190578 :wink:
And please add `doc_ignore` for all child props added in  https://github.com/gtk-rs/gtk/pull/481

GuillaumeGomez:
Oh damn, didn't see that comment. My bad!

PR #361, opened 2017-05-06
Fix into parameters lifetime

EPashkin:
Fix of https://github.com/gtk-rs/gtk/issues/501

cc @GuillaumeGomez


GuillaumeGomez:
Thanks!

PR #364, opened 2017-05-10
Generate *Ext traits by default and allow disabling them instead

sdroege:
WIP: The configuration should probably be renamed to allow_no_trait or
similar instead.

Reason for this is that the general case is that subclasses of any
GObject are allowed, and in your own crate you don't generally know if
other crates or any other code is going to create subclasses. Only in
few cases a class is "final" and does not allow subclasses.

@EPashkin Any opinions on this one?

### DIFF
--- a/src/analysis/object.rs
+++ b/src/analysis/object.rs
@@ -42,6 +42,20 @@ impl Deref for Info {
     }
 }
 
+pub fn has_known_subtypes(env: &Env, class_tid: library::TypeId) -> bool {
+    for child_tid in env.class_hierarchy.subtypes(class_tid) {
+        let child_name = child_tid.full_name(&env.library);
+        let status = env.config.objects.get(&child_name)
+            .map(|o| o.status)
+            .unwrap_or_default();
+        if status.normal() {
+            return true;
+        }
+    }
+
+    false
+}
+
 pub fn class(env: &Env, obj: &GObject, deps: &[library::TypeId]) -> Option<Info> {
     info!("Analyzing class {}", obj.name);
     let full_name = obj.name.clone();
@@ -70,17 +84,8 @@ pub fn class(env: &Env, obj: &GObject, deps: &[library::TypeId]) -> Option<Info>
 
     // Sanity check the user's configuration. It's unlikely that not generating
     // a trait is wanted if there are subtypes in this very crate
-    if !generate_trait {
-        for child_tid in env.class_hierarchy.subtypes(class_tid) {
-            let child_name = child_tid.full_name(&env.library);
-            let status = env.config.objects.get(&child_name)
-                .map(|o| o.status)
-                .unwrap_or_default();
-            if status.normal() {
-                error!("Not generating trait for {} although subtypes exist", full_name);
-                break;
-            }
-        }
+    if !generate_trait && has_known_subtypes(env, class_tid) {
+        error!("Not generating trait for {} although subtypes exist", full_name);
     }
 
     let mut trampolines = trampolines::Trampolines::with_capacity(klass.signals.len());
--- a/src/analysis/object.rs
+++ b/src/analysis/object.rs
@@ -66,7 +66,7 @@ pub fn class(env: &Env, obj: &GObject, deps: &[library::TypeId]) -> Option<Info>
 
     let supertypes = supertypes::analyze(env, class_tid, &mut imports);
 
-    let mut has_children = obj.force_trait;
+    let mut has_children = obj.generate_trait;
 
     for child_tid in env.class_hierarchy.subtypes(class_tid) {
         let child_name = child_tid.full_name(&env.library);
--- a/src/analysis/object.rs
+++ b/src/analysis/object.rs
@@ -68,14 +68,18 @@ pub fn class(env: &Env, obj: &GObject, deps: &[library::TypeId]) -> Option<Info>
 
     let mut generate_trait = obj.generate_trait;
 
-    for child_tid in env.class_hierarchy.subtypes(class_tid) {
-        let child_name = child_tid.full_name(&env.library);
-        let status = env.config.objects.get(&child_name)
-            .map(|o| o.status)
-            .unwrap_or_default();
-        if status.normal() {
-            generate_trait = true;
-            break;
+    // Sanity check the user's configuration. It's unlikely that not generating
+    // a trait is wanted if there are subtypes in this very crate
+    if !generate_trait {
+        for child_tid in env.class_hierarchy.subtypes(class_tid) {
+            let child_name = child_tid.full_name(&env.library);
+            let status = env.config.objects.get(&child_name)
+                .map(|o| o.status)
+                .unwrap_or_default();
+            if status.normal() {
+                error!("Not generating trait for {} although subtypes exist", full_name);
+                break;
+            }
         }
     }
 
--- a/src/analysis/object.rs
+++ b/src/analysis/object.rs
@@ -17,7 +17,7 @@ pub struct Info {
     pub c_type: String,
     pub get_type: String,
     pub supertypes: Vec<general::StatusedTypeId>,
-    pub has_children: bool,
+    pub generate_trait: bool,
     pub has_constructors: bool,
     pub has_methods: bool,
     pub has_functions: bool,
@@ -66,7 +66,7 @@ pub fn class(env: &Env, obj: &GObject, deps: &[library::TypeId]) -> Option<Info>
 
     let supertypes = supertypes::analyze(env, class_tid, &mut imports);
 
-    let mut has_children = obj.generate_trait;
+    let mut generate_trait = obj.generate_trait;
 
     for child_tid in env.class_hierarchy.subtypes(class_tid) {
         let child_name = child_tid.full_name(&env.library);
@@ -74,12 +74,12 @@ pub fn class(env: &Env, obj: &GObject, deps: &[library::TypeId]) -> Option<Info>
             .map(|o| o.status)
             .unwrap_or_default();
         if status.normal() {
-            has_children = true;
+            generate_trait = true;
             break;
         }
     }
 
-    if has_children {
+    if generate_trait {
         imports.add("glib::object::IsA", None);
     }
 
@@ -93,7 +93,7 @@ pub fn class(env: &Env, obj: &GObject, deps: &[library::TypeId]) -> Option<Info>
     special_functions::unhide(&mut functions, &specials, special_functions::Type::Copy);
     special_functions::analyze_imports(&specials, &mut imports);
 
-    let signals = signals::analyze(env, &klass.signals, class_tid, has_children,
+    let signals = signals::analyze(env, &klass.signals, class_tid, generate_trait,
                                    &mut trampolines, obj, &mut imports);
     let properties = properties::analyze(env, &klass.properties, class_tid, obj, &mut imports,
                                          &signatures, deps);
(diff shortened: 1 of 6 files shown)


EPashkin:
@sdroege Reasonable, can you give examples of final classes?

EPashkin:
If this accepted `README.md` and object analyze  need changed too.

sdroege:
> @sdroege Reasonable, can you give examples of final classes?

Everything coming from G_DECLARE_FINAL_TYPE, and every type where in C the struct definitions are not inside the public headers. It's nothing that we can detect easly

> If this accepted README.md and object analyze need changed too.

Sure :) Would you prefer the configuration to be renamed to something more sensible though? allow_no_trait would seem like a good candidate

EPashkin:
Seems it really almost no final classes.
I prefer `trait=false` with renaming config variable `force_trait` to `generate_trait`

sdroege:
Ok, will update accordingly later

EPashkin:
This PR caused strange build error in gtk/auto/css_provider.rs 
```
   Compiling gtk v0.1.3 (file:///D:/eap/rust/0/gtk)
error[E0034]: multiple applicable items in scope
  --> D:/eap/rust/0/gtk\src\auto\css_provider.rs:55:30
   |
55 |         write!(f, "{}", self.to_string())
   |                              ^^^^^^^^^ multiple `to_string` found
   |
note: candidate #1 is defined in an impl of the trait `auto::css_provider::CssProviderExt` for the type `_`
  --> D:/eap/rust/0/gtk\src\auto\css_provider.rs:98:5
   |
98 | /     fn to_string(&self) -> String {
99 | |         unsafe {
100| |             from_glib_full(ffi::gtk_css_provider_to_string(self.to_glib_none().0))
101| |         }
102| |     }
   | |_____^
   = note: candidate #2 is defined in an impl of the trait `std::string::ToString` for the type `_`
```

EPashkin:
Also it generated trait when no trait members ex for `gtk::DrawingAreaExt`
And causes strange error in https://github.com/gtk-rs/examples/blob/master/src/bin/menu_bar.rs despite trait must be included with `use gtk::prelude::*;`
```
   Compiling gtk-rs-examples v0.0.1 (file:///D:/eap/rust/0/examples)
error: no method named `set_submenu` found for type `gtk::MenuItem` in the current scope
  --> src\bin\menu_bar.rs:58:10
   |
58 |     file.set_submenu(Some(&menu));
   |          ^^^^^^^^^^^
   |
   = help: items from traits can only be used if the trait is in scope; the following trait is implemented but not in scope, perhaps add a `use` for it:
   = help: candidate #1: `use gtk::MenuItemExt;`
```

sdroege:
I'll look at the CSS thing.

For the menubar trait, maybe the traits should all be directly re-exported from the lib.rs? Not sure why the trait is needed there though, and empty traits should probably be just skipped. I'll take a look

sdroege:
The CSS trait confusion is solved now, empty traits are not generated anymore, and the example you linked above also compiles for me now. Please review again :)

sdroege:
One problem here seems to be that now almost everything goes into the Ext trait, but we can't manually implement any of those functions of the Ext trait. One would have manually write an ExtManual trait for those things.

I guess not really a problem in the end, just slightly ugly, and it's certainly not affecting existing code.

EPashkin:
Now has problem after generating gio app_launch_context.rs and gdk drag_context.rs, both has `use glib;` under feature and examples don't builds.

I still get 6 errors in menu_bar after fix `use glib` manually

sdroege:
I'll take a look at gio later. How can I reproduce the menu bar problem?

EPashkin:
I build pending branch of examples with using http://doc.crates.io/specifying-dependencies.html#overriding-with-local-dependencies to replace all gtk-rs crates.
When replacing only gtk it compiles.
Trying to find what crate cause problem.

EPashkin:
Problem solved: gio also have `MenuItemExt` and we rexport it in gtk::prelude (not sure why).

src/analysis/object.rs
```diff
@@ -104,7 +104,7 @@ pub fn class(env: &Env, obj: &GObject, deps: &[library::TypeId]) -> Option<Info>
     let child_properties = child_properties::analyze(env, obj.child_properties.as_ref(), class_tid,
                                                      &mut imports);
 
-    if has_children && !properties.is_empty() {
+    if generate_trait && !properties.is_empty() {
```
EPashkin: `||` here, seems we found old bug :wink:
It source bad gio and gdk generation.

EPashkin: Now unneeded using in gio simple_permision.rs

EPashkin: Seems this need changed to with removing lines 82-84
```
    let has_methods = functions.iter()
            .any(|f| f.kind == library::FunctionKind::Method);
    if (generate_trait && has_methods) || !properties.is_empty() {
        imports.add("glib", None);
        imports.add("glib::object::IsA", None);
    }
```

src/analysis/object.rs
```diff
@@ -66,20 +66,20 @@ pub fn class(env: &Env, obj: &GObject, deps: &[library::TypeId]) -> Option<Info>
...
         let child_name = child_tid.full_name(&env.library);
         let status = env.config.objects.get(&child_name)
             .map(|o| o.status)
             .unwrap_or_default();
         if status.normal() {
-            has_children = true;
+            generate_trait = true;
             break;
         }
     }
```
EPashkin: This for unneeded

src/codegen/doc/mod.rs
```diff
@@ -91,7 +91,7 @@ fn create_object_doc(w: &mut Write, env: &Env, info: &analysis::object::Info) ->
     let symbols = env.symbols.borrow();
     let ty = TypeStruct::new(SType::Struct, &info.name);
     let ty_ext = TypeStruct::new(SType::Trait, &format!("{}Ext", info.name));
-    let has_trait = info.has_children;
+    let has_trait = info.generate_trait;
```
EPashkin: Maybe something like `generate_trait(info)`

src/analysis/object.rs
EPashkin: Or better as you done in `generate_trait` function

EPashkin:
@GuillaumeGomez you don't remember reason for [this](https://github.com/gtk-rs/gtk/blob/master/src/prelude.rs#L8) reexport?

sdroege:
Thanks, I'll update the pull request tomorrow :) what other changes did you do elsewhere to make things work?

src/analysis/object.rs
EPashkin: Unoptimized version that give needed result:
```
    let has_methods = functions.iter()
            .any(|f| f.kind == library::FunctionKind::Method);
    let has_signals = signals.iter().any(|s| s.trampoline_name.is_ok());
    if generate_trait && (!properties.is_empty() || !child_properties.is_empty() || has_signals) {
        imports.add("glib", None);
    }
    if generate_trait && (has_methods || !properties.is_empty() || !child_properties.is_empty() || has_signals) {
        imports.add("glib::object::IsA", None);
    }
```

EPashkin:
In gir I only changed imports, on gtk removed gio reexport.
After this all generated code compiles without additional warnings and look normal at first glance.

src/analysis/object.rs
EPashkin: First if don't need `child_properties`

EPashkin:
Affected crates: gio, pango, gdk, gtk

GuillaumeGomez:
> @GuillaumeGomez you don't remember reason for [this](https://github.com/gtk-rs/gtk/blob/master/src/prelude.rs#L8) reexport?

Not at all. We should try to remove it and see what's affected.

EPashkin:
Examples builds fine.

EPashkin:
After some fix and last check ;)

src/analysis/object.rs
sdroege: Why? If obj.generate_trait=false (i.e. set false in the configuration), we would still want to generate a trait if it is actually needed. Or not?

sdroege:
Alright, thanks for the review and checking things :) I'll update the pull request with your comments and then let's check everything again.

For future reference, which crates are you all testing with any changes to the code generator? Only those under the gtk-rs project, or also others? And which testcases do you compile/run?

src/analysis/object.rs
EPashkin: I don't think that Gir must override human decision about generating trait,
at maximum in can produce error! to inform.

EPashkin:
I check meaningful changes for all crates under gtk-rs
and try build examples pending branch for default feature and gtk_3_22 (almost as CI will do later).

src/analysis/object.rs
sdroege: OK, error! it is then :)

EPashkin: And separate for to function please.

sdroege:
Please see attached. I hope I got all your review comments correctly. Note that I just calculate whether a trait is to be generated once in the analysis, instead of duplicating all these conditions in the code generator and docs.

src/analysis/object.rs
```diff
@@ -66,23 +66,23 @@ pub fn class(env: &Env, obj: &GObject, deps: &[library::TypeId]) -> Option<Info>
...
+    if !generate_trait {
+        for child_tid in env.class_hierarchy.subtypes(class_tid) {
+            let child_name = child_tid.full_name(&env.library);
+            let status = env.config.objects.get(&child_name)
+                .map(|o| o.status)
+                .unwrap_or_default();
+            if status.normal() {
+                error!("Not generating trait for {} although subtypes exist", full_name);
+                break;
+            }
```
EPashkin: IMHO this better separate function like `has_subtypes` here.
But I don't insist this

src/codegen/trait_impls.rs
```diff
@@ -28,17 +28,25 @@ fn lookup<'a>(functions: &'a [Info], name: &str) -> &'a Info {
         .unwrap()
 }
 
-fn generate_display(w: &mut Write, type_name: &str, func: &Info) -> Result<()> {
+fn generate_display(w: &mut Write, type_name: &str, func: &Info, in_trait: bool) -> Result<()> {
     use analysis::out_parameters::Mode;
+
```
EPashkin: Unneeded empty line

src/codegen/trait_impls.rs
```diff
@@ -49,43 +57,67 @@ impl fmt::Display for {type_name} {{
...
 
-fn generate_ord(w: &mut Write, type_name: &str, func: &Info) -> Result<()> {
+fn generate_ord(w: &mut Write, type_name: &str, func: &Info, in_trait: bool) -> Result<()> {
+    let call = if in_trait {
+        format!("{type_name}Ext::{func_name}(self, other)",
+                type_name = type_name, func_name = func.name)
+    } else {
+        format!("self.{func_name}(other)",
+                func_name = func.name)
+    };
```
EPashkin: This block repeated many times, maybe better move it to function?

sdroege:
Next try :)

src/codegen/trait_impls.rs
```diff
@@ -28,18 +28,49 @@ fn lookup<'a>(functions: &'a [Info], name: &str) -> &'a Info {
...
+        if in_trait {
+            args_string.push_str(", ");
+        }
+        for arg in args {
+            args_string.push_str(arg);
+            args_string.push_str(", ");
+        }
+        // Drop final ", "
+        let len = args_string.len();
+        args_string.truncate(len - 2);
```
EPashkin: args_string.push_str(&args.join(", ")); ?

sdroege: Done, good point

EPashkin:
@sdroege Thanks.
I don't fully understand usability of reviews, but IMHO you now have `Merge pull request` button.
Please do merge when CI passed.

sdroege:
There's no such button for me :) It says
>  Only those with write access to this repository can merge pull requests. 

Thanks for the review

EPashkin:
Bad, then I merge it

sdroege:
Thanks. I'll send pull requests for updating the other crates tomorrow.

Working on the free functions generation currently

PR #365, opened 2017-05-10
Add support for generating alias types

sdroege:
These come without any conversion traits implemented as they should be
directly compatible with the versions in the -sys crates.

### DIFF
--- a/src/codegen/alias.rs
+++ b/src/codegen/alias.rs
@@ -1,0 +1,57 @@
+use analysis::namespaces;
+use analysis::rust_type::rust_type;
+use codegen::general;
+use config::gobjects::GObject;
+use env::Env;
+use file_saver;
+use library::*;
+use std::io::prelude::*;
+use std::io::Result;
+use std::path::Path;
+use traits::*;
+
+pub fn generate(env: &Env, root_path: &Path, mod_rs: &mut Vec<String>) {
+    let configs: Vec<&GObject> = env.config.objects.values()
+        .filter(|c| {
+            c.status.need_generate() &&
+                c.type_id.map_or(false, |tid| tid.ns_id == namespaces::MAIN)
+        })
+        .collect();
+    let mut has_any = false;
+    for config in &configs {
+        if let Type::Alias(_) = *env.library.type_(config.type_id.unwrap()) {
+            has_any = true;
+            break;
+        }
+    }
+
+    if !has_any {
+        return;
+    }
+
+    let path = root_path.join("alias.rs");
+    file_saver::save_to_file(path, env.config.make_backup, |w| {
+        try!(general::start_comments(w, &env.config));
+        try!(writeln!(w, ""));
+
+        if has_any {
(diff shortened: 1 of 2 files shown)


EPashkin:
IMHO this not right.
Example "Gtk.Allocation" is alias of "Gdk.Rectangle" and `gdk::Rectangle` and `ffi::GdkRectangle` need conversion (though GtkAllocation don't used)

sdroege:
How should that be handled, do you have any suggestions?

EPashkin:
IMHO it just use conversion from original, was used when `Gdk.Rectangle` was alias for `cairo.RectangleInt`.
Don't remember how we declared it.
Currently don't see any usable aliases.

src/codegen/alias.rs
```diff
@@ -0,0 +1,52 @@
...
+    let configs: Vec<&GObject> = env.config.objects.values()
+        .filter(|c| {
+            c.status.need_generate() &&
+                c.type_id.map_or(false, |tid| tid.ns_id == namespaces::MAIN)
+        })
+        .collect();
+    let mut has_any = false;
+    for config in &configs {
+        if let Type::Alias(_) = *env.library.type_(config.type_id.unwrap()) {
+            has_any = true;
```
EPashkin: break?

sdroege:
So you would just let the crate define these manually, instead of autogenerating?

src/codegen/alias.rs
sdroege: Yes :)

src/codegen/alias.rs
```diff
@@ -0,0 +1,52 @@
...
+        }
+    }
+
```
EPashkin: not better to return here if !has_any?

sdroege: It's the same code as used in enum.rs

sdroege:
I see that Gdk.Rectangle is manually implemented. That also works for me, especially as my aliases are just u64s so far anyway.

src/codegen/alias.rs
EPashkin: I know that we don't exit early in flags and enums but here better exit

EPashkin:
After checking code I found that your word about `directly compatible with the versions in the -sys crates` don't used in code
and this PR don't affect your crates with early exit.
So I think it can be merged.

sdroege:
I don't understand what you mean.

It's probably safe to merge because existing crates won't have the aliases in the .toml, so no conflicts possible with ones that already define their own implementations (like Gdk.Rectangle). Is that what you mean?

EPashkin:
Gdk.Rectange now separate type, not alias. So currently we don't use any aliases in `gtk-rs` crates at all.
Generating aliases may cause problem if it aliased type from some crate (as was for Gdk.Rectangle), but we can fix this later.

sdroege:
OK, I'll update the pull request later then. Thanks

sdroege:
Updated accordingly

EPashkin:
Thanks, @sdroege

PR #366, opened 2017-05-10
Fully qualify the glib::object::Object for the IsA<_> constraints

sdroege:
In crates like gtk or gstreamer, which have their own GtkObject and
GstObject, this otherwise will result in wrong bounds. Specifically the
one for GObject is missing, which is required for things like properties
and signals.

### DIFF
--- a/src/analysis/object.rs
+++ b/src/analysis/object.rs
@@ -105,7 +105,7 @@ pub fn class(env: &Env, obj: &GObject, deps: &[library::TypeId]) -> Option<Info>
                                                      &mut imports);
 
     if has_children && !properties.is_empty() {
-        imports.add("Object", None);
+        imports.add("glib", None);
     }
     //don't `use` yourself
     imports.remove(&name);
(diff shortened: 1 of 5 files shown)


EPashkin:
@GuillaumeGomez I not against this. Affected crates: gio, gtk

sdroege:
I expect that you already have some kind of workaround for this in gio and gtk? Or was it just never a problem here? For GStreamer, stuff fails to compile because the glib::object::Object bound is missing in the trait default impls for being able to call g_object_get_property() for example.

EPashkin:
We don't have any xxxObject other that Glib.Object so we even don't even though about it.

sdroege:
There's Gtk.Object at least

GuillaumeGomez:
I don't have any argument against it so it's up to you @EPashkin.

EPashkin:
@sdroege no such type as "Gtk.Object" in `Gtk-3.0.gir`, we just rexports it with some usable traits to remove need add `glib` dependence

EPashkin:
@sdroege Thanks, I merge this.
You will add PR's for gtk and gio after other 2 PRs merged?

sdroege:
> You will add PR's for gtk and gio after other 2 PRs merged?

Sure, but I'll probably have a few more in the next days

Issue #295, opened 2016-11-28
Generate functions


antoyo:
Hello.
Is there a way to generate a wrapper for functions (not methods)?
For instance, in gnome-keyring, there are [many functions](https://developer.gnome.org/gnome-keyring/stable/gnome-keyring-Simple-Password-Storage.html).
Is there a way to tell `gir` to generate a wrapper for these functions?
Thanks.

EPashkin:
Currently `git` can generate only object methods.
I will think way to do it for functions

EPashkin:
But it won't help you much with `gnome-keyring`, as almost all functions in `Simple Password Storage` has variable parameters (`...`), `Keyrings` has callbacks, e.t.c.

EPashkin:
Also `gnome-keyring` it deprecated in favor of `libsecret`

sdroege:
How this is handled in the C# bindings code generator (and IIRC also in Vala), is that all the free functions are placed inside the main namespace or similar (i.e. a static class with static methods in C#, named after the namespace).

For here, maybe they can just be placed as free functions in src/auto/mod.rs? Then they would be available as $crate_name::$function_name by default.

If you agree with that, or have other suggestions, I could work on that. I know more or less what needs to be done code-wise here.

sdroege:
@EPashkin Functions for interfaces seems slightly different though. Shouldn't those be just mapped to impls of the same functions in the interface traits? Functions for interfaces have some kind of "owner" (the interface), while free functions don't really.

EPashkin:
@sdroege Interfaces functions handled already in #309.
As for free functions we IMHO need `src/auto/functions.rs` that will be reexpored in main module.
In gtk we already exported [gtk_drag_finish](https://github.com/gtk-rs/gtk/blob/master/src/drag_context.rs) as part of trait for `gdk::DragContext` may be others too.
Also we need way to control generation, maybe with fictive object like "Gtk.*"

sdroege:
Why src/auto/functions.rs instead of directly src/auto/mod.rs? I don't mind either way in any case

For overriding, "Gtk.drag_finish" would work. In the GIR it's organized the same way. It's under the Gtk namespace, and a XML node with the name drag_finish.

EPashkin:
I prefer `auto/mod.rs` contains only reexports.
As these function will need manual reexport on crate level it all same where to put them

sdroege:
OK, let's go with that then :)

EPashkin:
[...]

PR #371, opened 2017-05-17
Generate functions

sdroege:
https://github.com/gtk-rs/gir/issues/295

All functions are now unconditionally generated unless "ignore=true" for them. None are re-exported automatically and would have to be added to e.g. src/lib.rs for re-exporting from auto::functions::$function_name.
Not sure if this is ideal. Reason for not re-exporting them automatically is that you usually want to group them into useful modules, e.g. in the case of GStreamer all "gst_parse_*" functions would be placed into a gst::parse module.

Any opinions?

Last commit is just some random cleanup that I happened to notice while reading the code again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ generate = ["Gtk.Widget", "Gtk.Window"]
 manual = ["Gtk.Button"]
 ```
 
-So in here, both `GtkWidget` and `GtkWindow` will be fully generated and functions/methods using `GtkButton` will be uncommented.
+So in here, both `GtkWidget` and `GtkWindow` will be fully generated and functions/methods using `GtkButton` will be uncommented. To generate code for all global functions, add `Gtk.*` to the `generate` array.
 
 Sometimes Gir understands the object definition incorrectly or the `.gir` file contains incomplete or wrong definition, to fix it, you can use the full object configuration:
 
@@ -141,6 +141,21 @@ status = "generate"
     # define starting version when member added
     version = "3.18"
 ```
+
+For global functions, the members can be configured by configuring the `Gtk.*` object:
+
+```toml
+[[object]]
+name = "Gtk.*"
+status = "generate"
+    [[object.member]]
+    name = "stock_list_ids"
+    # allows to ignore global functions
+    ignore = true
+```
+
+Note that you must not place `Gtk.*` into the `generate` array and
+additionally configure its members.
 
 ### Generation
 
--- a/src/analysis/mod.rs
+++ b/src/analysis/mod.rs
@@ -1,6 +1,7 @@
 use std::collections::BTreeMap;
 
 use env::Env;
+use library;
 use library::{Type, TypeId};
 
 pub mod bounds;
@@ -36,6 +37,7 @@ pub mod trampolines;
 pub struct Analysis {
     pub objects: BTreeMap<String, object::Info>,
     pub records: BTreeMap<String, record::Info>,
+    pub global_functions: Option<info_base::InfoBase>,
 }
 
 pub fn run(env: &mut Env) {
@@ -69,7 +71,44 @@ pub fn run(env: &mut Env) {
     }
     if !to_analyze.is_empty() {
         error!("Not analyzed {} objects due unfinished dependencies", to_analyze.len());
+        return;
     }
+
+    // Analyze free functions as the last step once all types are analyzed
+    analyze_global_functions(env);
+}
+
+fn analyze_global_functions(env: &mut Env) {
+    let ns = env.library.namespace(library::MAIN_NAMESPACE);
+
+    let full_name = format!("{}.*", ns.name);
+
+    let obj = match env.config.objects.get(&*full_name) {
+        Some(obj) if obj.status.need_generate() => obj,
+        _ => return,
+    };
+
+    let functions: Vec<_> = ns.functions.iter().filter(|f| f.kind == library::FunctionKind::Global).collect();
+    if functions.is_empty() {
+        return;
+    }
+
+    let mut imports = imports::Imports::new();
+    imports.add("glib::translate::*", None);
+    imports.add("ffi", None);
+
+    let functions = functions::analyze(env, &functions, TypeId::tid_none(), &obj, &mut imports, None, None);
+
+    imports.clean_glib(env);
+
+    env.analysis.global_functions = Some(info_base::InfoBase {
+        full_name: full_name,
+        type_id: TypeId::tid_none(),
+        name: "*".into(),
+        functions: functions,
+        imports: imports,
+        ..Default::default()
+    });
 }
 
 fn analyze(env: &mut Env, tid: TypeId, deps: &[TypeId]) {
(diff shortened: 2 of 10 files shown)


src/analysis/mod.rs
```diff
@@ -69,7 +71,47 @@ pub fn run(env: &mut Env) {
...
+fn analyze_global_functions(env: &mut Env) {
+    let ns = env.library.namespace(library::MAIN_NAMESPACE);
+
+    let full_name = format!("{}.*", ns.name);
+
+    let obj = match env.config.objects.get(&*full_name) {
+        Some(obj) => obj,
+        None => return,
+    };
+
```
EPashkin: IMHO better `check obj.status.need_generate()` before processing.
This way we can disable generation without deleting object from Gir.toml

sdroege: Done, good point

src/analysis/mod.rs
```diff
@@ -69,7 +71,47 @@ pub fn run(env: &mut Env) {
...
+    let functions: Vec<_> = ns.functions.iter().filter(|f| f.kind == library::FunctionKind::Global).collect();
+    if functions.is_empty() {
+        return;
+    }
+
+    let mut imports = imports::Imports::new();
+    imports.add("glib::translate::*", None);
+    imports.add("ffi", None);
+
+    let functions = functions::analyze(env, &functions, Default::default(), &obj, &mut imports, None, None);
```
EPashkin: Please change `Default::default` to `TypeId::tid_none()`

EPashkin:
Readme need update

src/analysis/mod.rs
sdroege: Ok :)

sdroege:
> Readme need update

Yes, was going to do that once the way how the configuration works is considered good :)

src/analysis/mod.rs
```diff
@@ -69,7 +71,47 @@ pub fn run(env: &mut Env) {
...
+        full_name: full_name,
+        type_id: TypeId::tid_none(),
+        name: "*".into(),
+        functions: functions,
+        specials: Default::default(),
+        imports: imports,
+        version: None,
+        deprecated_version: None,
+        cfg_condition: None,
+    });
```
EPashkin: InfoBase implement Default so don't need many default values
`type_id: TypeId::tid_none()` better stays to be same as `analyze` parameter

sdroege: `type_id: TypeId::tid_none()` is there already now. How would you like to use Default here? Like this?
```Rust
let mut info = Default::default();
info.name = ...;
...
```

sdroege: Ah with ..Default::default() ?

sdroege: Done that now, seems indeed better

src/codegen/functions.rs
```diff
@@ -0,0 +1,31 @@
...
+use file_saver;
+use codegen::general;
+use codegen::function;
+
+pub fn generate(env: &Env, root_path: &Path, mod_rs: &mut Vec<String>) {
+    info!("Generate global functions");
+
+    let functions = match env.analysis.global_functions {
+        None => { return; }
+        Some(ref functions) => functions,
```
EPashkin: Swap cases, return don't need braces. Maybe this better look as `if let`?

sdroege: Done. "if let" looks more noisy in these cases IMHO

EPashkin:
Code looks good, Thanks @sdroege 
While we wait for syntax confirmation, I add functions for other gtk-rs crates.

sdroege:
> While we wait for syntax confirmation

From whom? @GuillaumeGomez ? :)

EPashkin:
Yes

EPashkin:
`Imports::clean_glib` need skip `glib` too

EPashkin:
In needed for glib functions

EPashkin:
.. it can just remove "glib" in second line

sdroege:
> Imports::clean_glib need skip glib too
Ok :)

sdroege:
It's definitely called for global functions, but it never goes into the correct case to add std::mem

EPashkin:
g_ucs4_to_utf16 has `caller-allocates=1` for long

EPashkin:
@sdroege !par.caller_allocates need go https://github.com/gtk-rs/gir/blob/master/src/analysis/out_parameters.rs#L102

GuillaumeGomez:
You now have the syntax confirmation. ;)

EPashkin:
@sdroege I try finish all gtk-rs crates before merging this. It can take all evening

sdroege:
Sure, I'll be off until tomorrow but then do the out_parameter part and the Readme. Thanks for taking care of the other crates.
For mine it all seems to work fine

EPashkin:
Add `Type if env.namespaces.glib_ns_id == library::MAIN_NAMESPACE => ok("types::Type"),`
before https://github.com/gtk-rs/gir/blob/master/src/analysis/rust_type.rs#L101

sdroege:
@GuillaumeGomez My main question was about whether all global functions should be generated by default, or only those that are explicitly set to "ignore=false". As they have to be re-exported one by one usually.

sdroege:
Readme changes come tomorrow

src/config/gobjects.rs
```diff
@@ -3,6 +3,7 @@ use std::collections::BTreeMap;
 use std::str::FromStr;
 use toml::Value;
 
+use library;
 use library::{Library, TypeId};
```
EPashkin: IMHO better import MAIN_NAMESPACE and use it directly

sdroege: Done, and README.md updated too

README.md
```diff
@@ -142,6 +142,21 @@ status = "generate"
...
+name = "Gtk.*"
+status = "generate"
+    [[object.member]]
+    name = "stock_list_ids"
+    # allows to ignore global functions
+    ignore = true
+```
+
+Note that you must not place `Gtk.*` into the `generate` array and
+additionally configure its members.
```
EPashkin: IMHO first paragraph of "TOML file" section already stated that.

sdroege: It does, but I think it's worth repeating here because "Gtk.*" is a bit special. Conceptually you would think of the functions as the things to generate, not "Gtk.*". I can remove it if you prefer

EPashkin: Ok.

EPashkin:
Thanks, @sdroege

sdroege:
@EPashkin You're going to work on updating all gtk-rs crates and exporting the global functions? Or just when needed at a later time?

EPashkin:
I already prepared changes for all your crates and add PR's at evening.
There may be question what functions need reexport (currently only replacing already exported).
Also not sure what to do with already implemented manual parts (currently mostly remove if it same).

EPashkin:
All 5 PRs added and linked to this PR.
@GuillaumeGomez please take look when you have time.

GuillaumeGomez:
I finish what I'm working on and then I take a look!

EPashkin:
@GuillaumeGomez I added last pr for reexport. Please check these 3 PRs.

Issue #368, opened 2017-05-15
Add configuration for marking bool-returning functions as error indication


sdroege:
There are many functions all over the place that return a boolean, and this being true/false indicates whether the function was successful or not. Not all of them are also returning an actual value via out-parameter, but instead have their side-effects as only result.

I'd suggest adding a configuration for the .toml file for these functions ("bool_return_is_error"?), and converting them into Result<(), E> to be able to use try! and the ? operator.

Question now is what to put into the E. We don't have anything meaningful to put there as the C library does not give any further information. But using () is also not very satisfying. As such I would suggest passing a type to "bool_return_is_error".

Just what to pass to the constructor? Nothing? Or we could pass file name, line number of where it happens. We could pass a &'static str for the (C or Rust) function name.

Any comments, ideas, ...?

EPashkin:
We can add `glib::BoolError(pub &'static str)` and create it with `"type::name"` or `"trait::name"`.
As alternative we can define error type in manual way and write it name in `bool_return_is_error`.
But IMHO it is not worth the effort.
Anyway not need bother with file name and line number as standard backtrace contains both.

EPashkin:
By backtrace I mean https://github.com/brson/error-chain

sdroege:
That would also work and seem simpler, yes. Let's go with glib::BoolError(pub &'static str) for the time being.

I'll work on this after the other pull requests and the free functions.

EPashkin:
Thank you in advance.

sdroege:
> By backtrace I mean https://github.com/brson/error-chain

You would like glib to depend on error-chain? I would just implement Debug, Display and Error for the new glib::BoolError type to keep dependencies small.

EPashkin:
User can use this crate for error processing and it collect stacktrace on initial error conversion.
We need implement these trait for `glib::BoolError` but IMHO currently don't need use error_chain in your crates

sdroege:
Right, but for error-chain to produce a backtrace that starts in our bool-returning function, glib::BoolError would have to use error-chain, or not?

sdroege:
Anyway, we seem to be on the same page here then. I'll implement that tomorrow or so, and then we can discuss about the details with actual code :)

EPashkin:
As generated function only has one error return IMHO user only need know place in his code.

EPashkin:
@sdroege Seems this done and can be closed?

PR #373, opened 2017-05-20
Add configuration for converting bool return values to Result<(), Boo…

sdroege:
…lError>

https://github.com/gtk-rs/gir/issues/368

Example of the output/configuration can be found here: https://github.com/sdroege/gstreamer-rs/commit/f727ef4e77f34cdc54c02d3e1d8feea4398184af

### DIFF
--- a/src/analysis/return_value.rs
+++ b/src/analysis/return_value.rs
@@ -44,17 +44,18 @@ pub fn analyze(env: &Env, func: &library::Function, type_tid: library::TypeId,
         parameter_rust_type(env, func.ret.typ, func.ret.direction, Nullable(false), RefMode::None).is_err()
     };
 
-    let mut bool_return_is_error = !configured_functions.is_empty() && configured_functions.iter().any(|f| f.ret.bool_return_is_error);
-    if bool_return_is_error && func.ret.typ != TypeId::tid_bool() {
-        error!("Ignoring bool_return_is_error configuration for non-bool returning function {}", func.name);
-        bool_return_is_error = false;
-    } else if bool_return_is_error {
-        if env.namespaces.glib_ns_id == namespaces::MAIN {
-            imports.add("error", None);
+    let bool_return_is_error = configured_functions.iter().filter_map(|f| f.ret.bool_return_is_error.as_ref()).next();
+    let bool_return_error_message = bool_return_is_error.and_then(|m| {
+        if func.ret.typ != TypeId::tid_bool() {
+            error!("Ignoring bool_return_is_error configuration for non-bool returning function {}", func.name);
+            None
         } else {
-            imports.add("glib", None);
+            let ns = if env.namespaces.glib_ns_id == namespaces::MAIN { "error" } else { "glib" };
+            imports.add(ns, None);
+
+            Some(m.clone())
         }
-    }
+    });
 
     let mut base_tid = None;
 
@@ -78,7 +79,7 @@ pub fn analyze(env: &Env, func: &library::Function, type_tid: library::TypeId,
         parameter: parameter,
         base_tid: base_tid,
         commented: commented,
-        bool_return_is_error: if bool_return_is_error { Some(func.c_identifier.as_ref().unwrap().clone()) } else { None },
+        bool_return_is_error: bool_return_error_message,
     }
 }
 
--- a/src/analysis/return_value.rs
+++ b/src/analysis/return_value.rs
@@ -1,19 +1,22 @@
 use analysis::ref_mode::RefMode;
 use analysis::rust_type::*;
+use analysis::imports::Imports;
+use analysis::namespaces;
 use config;
 use env::Env;
-use library::{self, Nullable};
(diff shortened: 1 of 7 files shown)


sdroege:
If that is all good, next step would be going through all the bool returning functions everywhere and checking if they use the bool for signalling error

EPashkin:
Strange, but it causes program termination on windows on gtk and gdk

sdroege:
I'll check in a bit. You mean running gir on glib/gtk?

EPashkin:
Yes, seems problem with rust incremental compilation.
It failed it build d000fa713bba after b010d34e19

sdroege:
You mean after cargo clean it works fine?

EPashkin:
Yes, even just removing target/release works.
It don't happened to you?
```
>rustc -Vv
rustc 1.19.0-nightly (5dfcd85fd 2017-05-19)
binary: rustc
commit-hash: 5dfcd85fd4bae49445383baadf472fbdb414a0e6
commit-date: 2017-05-19
host: x86_64-pc-windows-gnu
release: 1.19.0-nightly
LLVM version: 4.0
```

sdroege:
I'm usually not using nightly, unless I have to

src/analysis/return_value.rs
```diff
@@ -41,6 +44,18 @@ pub fn analyze(env: &Env, func: &library::Function, type_tid: library::TypeId,
         parameter_rust_type(env, func.ret.typ, func.ret.direction, Nullable(false), RefMode::None).is_err()
     };
 
+    let mut bool_return_is_error = !configured_functions.is_empty() && configured_functions.iter().any(|f| f.ret.bool_return_is_error);
+    if bool_return_is_error && func.ret.typ != TypeId::tid_bool() {
+        error!("Ignoring bool_return_is_error configuration for non-bool returning function {}", func.name);
+        bool_return_is_error = false;
+    } else if bool_return_is_error {
+        if env.namespaces.glib_ns_id == namespaces::MAIN {
```
GuillaumeGomez: I think this could be rewritten as follow:

```rust
imports.add(if env.namespaces.glib_ns_id == namespaces::MAIN { " error" } else { "glib" }, None);
```

However, not sure if this is better...

src/analysis/return_value.rs
```diff
@@ -63,6 +78,7 @@ pub fn analyze(env: &Env, func: &library::Function, type_tid: library::TypeId,
         parameter: parameter,
         base_tid: base_tid,
         commented: commented,
+        bool_return_is_error: if bool_return_is_error { Some(func.c_identifier.as_ref().unwrap().clone()) } else { None },
```
EPashkin: Please, move it to variable. And IMHO "g_setenv" more confising "glib::setenv".
May be just change config value to string?

EPashkin:
@sdroege On stable it don't repeated for me

src/analysis/return_value.rs
sdroege: g_setenv describes the function that actually failed (it's not the Rust code that failed), and glib::setenv might not actually exist (it might be exported as glib::some_module::setenv instead for example).

Allowing the config value as a string could make sense (it would allow to provide a more useful error string per function), and maybe allow a boolean as fallback which then does the above?

EPashkin: Yes, boolean fallback also will works.

sdroege: Ok, I'll change it like that then. Good idea in any case, thanks!

sdroege:
> @sdroege On stable it don't repeated for me

Probably makes sense to report that to the Rust people then if it's reproducible :)

For everything else, I added the changes as another commit for easier reviewing now.

EPashkin:
String error still contains ffi function name?

EPashkin:
IMHO better replace message completely

EPashkin:
On windows error 100% reproducible in both way, I will confirm that error repeated on linux too before reporting.

sdroege:
> String error still contains ffi function name?

Yeah, as an indication where the actual error happened. It's prepended to the user given message, if any, to always have it available. I'm using the ffi name because that's where the error happened (not in the Rust code) and we don't know with which name the function is going to be exported later.

I'm not using only the user given string to give some more context automatically.

> IMHO better replace message completely

What do you mean?

EPashkin:
I prefer show message from config without addition, it give author more possibilities.
And I don't see any meaningful text that can be add to "g_func_name: " in constant way as we don't know what really happened.

sdroege:
OK, so maybe just only the configuration string. And get rid of allowing a book, it's easy enough to come up with a meaningful message anyway. Let's keep it simple then

sdroege:
Done so now, only a string is allowed now and it's directly used for the error

EPashkin:
Sorry I forgot about it, please, update readme

sdroege:
Done

EPashkin:
Thanks

PR #374, opened 2017-05-21
Don't prepend a comma before the first non-out parameter if the first…

sdroege:
… one is an out parameter

We otherwise generate things like

```Rust
pub fn value_init_and_copy(, src: &glib::Value) -> glib::Value
```

Here the first parameter is an out parameter (that is converted to the return type).


EPashkin:
Thanks

PR #378, opened 2017-05-27
Add Into for child properties as well

GuillaumeGomez:


src/analysis/bounds.rs
```diff
@@ -21,6 +21,15 @@ pub enum BoundType {
...
 impl BoundType {
+    pub fn is_into(&self) -> bool {
+        match *self {
+            BoundType::Into(_, _) => true,
+            _ => false,
+        }
+    }
+}
+
+impl BoundType {
```
EPashkin: Why separate impl?

src/analysis/child_properties.rs
```diff
@@ -83,6 +92,22 @@ fn analyze_property(env: &Env, prop: &config::ChildProperty, child_name: &str,
...
+        let mut is_into = false;
+        let mut bounds_str = String::new();
+        let dir = library::ParameterDirection::In;
+        let set_params = if let Some(bound) = Bounds::type_for(env, typ, nullable) {
+            is_into = bound.is_into();
+            let r_type = bounds_rust_type(env, typ).into_string();
+            let mut bounds = Bounds::default();
+            bounds.add_parameter("P", &r_type, bound);
+            let s_bounds = function::bounds(&bounds);
+            bounds_str.push_str(&s_bounds[1..s_bounds.len() - 1]);
```
EPashkin: IMHO better add comment why this needed with example

EPashkin:
There possible problem that function declaration use both automatic and hardcoded trait constraint name, but IMHO it can be ignored as setters always has 3 parameters.
```diff
-    fn set_child_name<T: IsA<Widget>>(&self, item: &T, name: Option<&str>);
+    fn set_child_name<'a, P: Into<Option<&'a str>>, T: IsA<Widget>>(&self, item: &T, name: P);
```

src/analysis/bounds.rs
GuillaumeGomez: Oh damn, didn't pay attention.

src/analysis/child_properties.rs
GuillaumeGomez: Good point!

GuillaumeGomez:
> There possible problem that function declaration use both automatic and hardcoded trait constraint name, but IMHO it can be ignored as setters always has 3 parameters.

This is exactly why I didn't bother much. We know for sure that we won't have more arguments so no need to try to make the whole thing more complicated than necessary.

EPashkin:
Thanks

Issue #376, opened 2017-05-22
Not parsing/handling array specific GIR fields (length parameter position, zero-terminated)


sdroege:
See e.g. https://wiki.gnome.org/action/show/Projects/GObjectIntrospection/Annotations#Array_length for an example with parameter position (the GIR does not contain the parameter name but its position).

sdroege:
I'm not sure how to get this information from src/parser.rs to analysis/codegenerator. Would I just parse them and store them in the library::Parameter type as additional fields of types Option<u32> / Option<bool>, and then pass those through the different layers?

For the length parameter position it would then be needed in the code generator to handle two cases:
1) Normal parameter (e.g. `fn f(length: u32, arr: &[Something])`). The length parameter would have to be removed from the generated function, and instead pass arr.len() to the C function
2) Out parameters, which I would have to research how that is handled at all currently

For the zero-terminated parameter, it only affects the code generator inside the function body and it needs to make sure to
1) For normal parameters to add a ptr::null() at the end before passing it to the C function
2) For out parameters to get the length by looking for a terminating ptr::is_null() in the array

EPashkin:
IMHO normal to store `length` in library::Parameter as `len_position: Option<int>`.
You right about input parameters. Note: for direct arrays we always add last element in https://github.com/gtk-rs/glib/blob/master/src/translate.rs#L459-L474.
About out/return parameters:
in cases when size returned we use `from_glib_(none|container|full)_num` in manual code,
for NULL-ended we have https://github.com/gtk-rs/glib/blob/master/src/translate.rs#L856-L866

sdroege:
Thanks for the pointers. The problem with all these is that they always require copying, which is not always required. But I guess that's for a later time then. Let's first make it work automatically at all ;)

sdroege:
While implementing all that, should probably also add a mapping for C arrays of gchar/guchar/guint8 and similar.

EPashkin:
Yes, that would be great

sdroege:
Still working on this btw, just busy with other things at the same time and this requires some minor code reorganization so that function_body_chunk can get all the information where its needed :)

PR #385, opened 2017-06-11
Add support for gsize/gssize by using usize/isize

sdroege:


sdroege:
Note that this exposes some API in GTK (e.g. `gtk_recent_chooser_get_uris`) which previously was not exposed, but those need https://github.com/gtk-rs/gir/issues/376 properly fixed before being correct.

EPashkin:
Thanks.

Many function generated by GLib-2.0.gir is failing (Ex. `g_convert_with_fallback` don't have `out`) and need ignored.
Gio also affected.

sdroege:
Yes, those are also on my list. It's 2-3 separate issues :)

EPashkin:
@sdroege you planing apply this to `glib` in near time?
It currently "block" checking all other gir PR's
If in plan only ignoring not compiling function then I can do it today.

sdroege:
Probably won't get to this specific one today though, but it's high up in my list next week

PR #387, opened 2017-06-19
Add support for array-length for parameters and return values

sdroege:
https://github.com/gtk-rs/gir/issues/376

Only pointer arrays are currently supported, the translation trait is even called FromGlibPtrContainer. This should be changed really to also be able to handle all kinds of integer (incl. byte) arrays.

### DIFF
--- a/src/codegen/function.rs
+++ b/src/codegen/function.rs
@@ -131,7 +131,7 @@ pub fn body_chunk(env: &Env, analysis: &analysis::functions::Info) -> Chunk {
     let ret_array_length = if let Some(pos) = analysis.ret.parameter.as_ref().and_then(|p| p.array_length) {
         // The first parameter is &self in case of methods
         let pos_offset = if analysis.kind == library::FunctionKind::Method { 1 } else { 0 };
-        analysis.parameters.iter().nth((pos + pos_offset) as usize)
+        analysis.parameters.get((pos + pos_offset) as usize)
     } else {
         None
     };
@@ -148,7 +148,7 @@ pub fn body_chunk(env: &Env, analysis: &analysis::functions::Info) -> Chunk {
                 use analysis::out_parameters::Mode;
                 // The actual return value was inserted at position 0
                 let pos_offset = if analysis.outs.mode == Mode::Combined || analysis.outs.mode == Mode::Throws(true) { 1 } else { 0 };
-                analysis.parameters.iter().nth((pos + pos_offset) as usize)
+                analysis.parameters.get((pos + pos_offset) as usize)
             } else {
                 None
             };
@@ -158,7 +158,7 @@ pub fn body_chunk(env: &Env, analysis: &analysis::functions::Info) -> Chunk {
             let array_length = if let Some(pos) = par.array_length {
                 // The first parameter is &self in case of methods
                 let pos_offset = if analysis.kind == library::FunctionKind::Method { 1 } else { 0 };
-                analysis.parameters.iter().nth((pos + pos_offset) as usize)
+                analysis.parameters.get((pos + pos_offset) as usize)
             } else {
                 None
             };
--- a/src/codegen/function.rs
+++ b/src/codegen/function.rs
@@ -12,6 +12,7 @@ use super::parameter::ToParameter;
 use super::return_value::{out_parameters_as_return, ToReturnValue};
 use writer::primitives::tabs;
 use writer::ToCode;
+use library;
 
 pub fn generate(w: &mut Write, env: &Env, analysis: &analysis::functions::Info,
                 in_trait: bool, only_declaration: bool, indent: usize) -> Result<()> {
@@ -72,12 +73,24 @@ pub fn declaration(env: &Env, analysis: &analysis::functions::Info) -> String {
 
     let bounds = bounds(&analysis.bounds);
 
+    let array_lengths: Vec<_> = analysis.parameters.iter()
+                                                    .filter_map(|p| p.array_length)
+                                                    .collect();
+
     let mut skipped = 0;
     for (pos, par) in analysis.parameters.iter().enumerate() {
         if outs_as_return && analysis.outs.iter().any(|p| p.name==par.name) {
             skipped += 1;
             continue;
         }
+
+        // The first parameter is &self in case of methods
+        let pos_offset = if analysis.kind == library::FunctionKind::Method { 1 } else { 0 };
+        if pos >= pos_offset && array_lengths.contains(&((pos - pos_offset) as u32)) {
+            skipped += 1;
+            continue;
+        }
+
         if pos > skipped { param_str.push_str(", ") }
         let s = par.to_parameter(env, &analysis.bounds);
         param_str.push_str(&s);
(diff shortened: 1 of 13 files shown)


sdroege:
With https://github.com/gtk-rs/gir/pull/389 and this, Gtk.RecentInfo can be fully autogenerated btw.

src/codegen/function_body_chunk.rs
```diff
@@ -74,8 +81,9 @@ impl Builder {
             }
             _ => Uninitialized,
         };
+        let array_length = array_length.map(|p| (p.name.clone(), rust_type(env, p.typ).unwrap()));
         self.parameters.push(Parameter::Out {
-            parameter: parameter.into(),
+            parameter:  parameter_ffi_call_out::Parameter::new(parameter, array_length),
```
EPashkin: Extra space

src/codegen/return_value.rs
```diff
@@ -36,12 +36,12 @@ impl ToReturnValue for analysis::return_value::Info {
 pub fn out_parameter_as_return_parts(analysis: &analysis::functions::Info)
                                      -> (&'static str, &'static str) {
     use analysis::out_parameters::Mode::*;
-    let is_tuple = analysis.outs.len() > 1;
+    let num_outs = analysis.outs.iter().filter(|p| p.array_length.is_none()).count();
```
EPashkin: fn len() now unused

EPashkin:
Bad change in glib:key_file - missed tuple
```diff
-    pub fn get_string_list(&self, group_name: &str, key: &str) -> Result<(Vec<String>, usize), Error> {
+    pub fn get_string_list(&self, group_name: &str, key: &str) -> Result<Vec<String>, usize, Error> {
```

EPashkin:
..but still return usize

src/codegen/translate_from_glib.rs
```diff
@@ -64,18 +64,22 @@ impl TranslateFromGlib for analysis::return_value::Info {
...
+    match (transfer, array_length) {
+        (Transfer::None, &None) => ("from_glib_none(".into(), ")".into()),
+        (Transfer::Full, &None) => ("from_glib_full(".into(), ")".into()),
+        (Transfer::Container, &None) => ("from_glib_container(".into(), ")".into()),
+        (Transfer::None, &Some((ref array_length_name, _))) => ("from_glib_none_num(".into(), format!(", {} as usize)", array_length_name)),
+        (Transfer::Full, &Some((ref array_length_name, _))) => ("from_glib_full_num(".into(), format!(", {} as usize)", array_length_name)),
+        (Transfer::Container, &Some((ref array_length_name, _))) => ("from_glib_container_num(".into(), format!(", {} as usize)", array_length_name)),
     }
 }
+
```
EPashkin: Extra line

EPashkin:
Too many long lines but code too.
Maybe code will be simpler is analysis parameter has `is_array_len` flag.

EPashkin:
or even SkipReason: None/InArrayLen/OutArrayLen

EPashkin:
*Too many long lines but it not big problem as code works.

sdroege:
Thanks, I'll clean this up tomorrow. For the code style, what do you think about my suggestion to just run things through rustfmt soon and then require pull requests to use the correct, automatically enforced format?

EPashkin:
also maybe better pass `array_length: Option<(String, String)>` directly to builder.

EPashkin:
About rustfmt, currently it change nothing in gir or panic with 'failed to emit error: operation not supported by the terminal'
Maybe it because some cargo bug: it build only if error stream relocated to file.
When run with 2>_ it also fails on build.rs when opening file `gir\\src/git.rs` (windows related problem).
Result code sometimes looks strange:
```rust
pub fn generate(
    w: &mut Write,
    env: &Env,
    prop: &ChildProperty,
    in_trait: bool,
    only_declaration: bool,
    indent: usize,
) -> Result<()> {
    try!(generate_func(
        w,
        env,
        prop,
        in_trait,
        only_declaration,
        indent,
        true,
    ));
    try!(generate_func(
        w,
        env,
        prop,
        in_trait,
        only_declaration,
        indent,
        false,
    ));

    Ok(())
}
```
So I don't against single apply in separate PR, but IMHO it don't ready to enforce.

sdroege:
The comments should all be covered now. Remaining problem is `g_key_file_set_string_list()`, which has the wrong type in the sys crate already. The list should be `*const *const c_char` but is not detected as such and also looks wrong in the .gir file. Once that is fixed, the `ToGlibPtr` instance for `[&str]` from glib/src/translate.rs has to be enabled again.

Unrelated, `g_strv_contains()` is also wrong but not in the sys crate. It should be `&[&str]` for the first parameter.

EPashkin:
Gio generation failed, other lib changes seems fine
```
D:\eap\rust\0\gir>cargo run --release -- -s -c D:\eap\rust\0\gio\Gir.toml
    Finished release [optimized] target(s) in 0.0 secs
     Running `D:\eap\rust\0\./target\release\gir.exe -s -c D:\eap\rust\0\gio\Gir.toml`
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src\libcore\option.rs:329
stack backtrace:
   0: std::sys_common::backtrace::_print
   1: std::panicking::default_hook::{{closure}}
   2: std::panicking::default_hook
   3: std::panicking::rust_panic_with_hook
   4: std::panicking::begin_panic
   5: std::panicking::begin_panic_fmt
   6: rust_begin_unwind
   7: core::panicking::panic_fmt
   8: core::panicking::panic
   9: gir::codegen::return_value::out_parameters_as_return
  10: gir::codegen::function::declaration
  11: gir::codegen::function::generate
  12: gir::codegen::object::generate
  13: gir::file_saver::save_to_file
  14: gir::codegen::objects::generate
  15: gir::codegen::normal_generate
  16: gir::do_main
  17: gir::main
  18: _rust_maybe_catch_panic
  19: std::rt::lang_start
  20: _tmainCRTStartup
  21: mainCRTStartup
  22: free_fail_stack_return
error: process didn't exit successfully: `D:\eap\rust\0\./target\release\gir.exe -s -c D:\eap\rust\0\gio\Gir.toml` (exit code: 101)
```

src/codegen/function.rs
```diff
@@ -115,17 +128,42 @@ pub fn body_chunk(env: &Env, analysis: &analysis::functions::Info) -> Chunk {
     }
 
     let outs_as_return = !analysis.outs.is_empty();
+    let ret_array_length = if let Some(pos) = analysis.ret.parameter.as_ref().and_then(|p| p.array_length) {
+        // The first parameter is &self in case of methods
+        let pos_offset = if analysis.kind == library::FunctionKind::Method { 1 } else { 0 };
+        analysis.parameters.iter().nth((pos + pos_offset) as usize)
```
EPashkin: clippy warning: called `.iter().nth()` on a Vec. Calling `.get()` is both faster and more readable

src/codegen/function.rs
```diff
@@ -115,17 +128,42 @@ pub fn body_chunk(env: &Env, analysis: &analysis::functions::Info) -> Chunk {
...
         .outs_mode(analysis.outs.mode);
 
     for par in &analysis.parameters {
         if outs_as_return && analysis.outs.iter().any(|p| p.name==par.name) {
-            builder.out_parameter(env, par);
+            let array_length = if let Some(pos) = par.array_length {
+                use analysis::out_parameters::Mode;
+                // The actual return value was inserted at position 0
+                let pos_offset = if analysis.outs.mode == Mode::Combined || analysis.outs.mode == Mode::Throws(true) { 1 } else { 0 };
+                analysis.parameters.iter().nth((pos + pos_offset) as usize)
```
EPashkin: same

sdroege:
https://bugzilla.gnome.org/show_bug.cgi?id=784020 (glib)
https://bugzilla.gnome.org/show_bug.cgi?id=784021 (pango)
https://bugzilla.gnome.org/show_bug.cgi?id=784022 (gtk)

For string parameters + array length, we currently still generate wrong code it seems. For some reason it is interpreted as a `&[&str]` then instead of a `&str`. E.g. in Pango after changing the .gir file for `set_markup` like this
```diff
-            <type name="utf8" c:type="const char*"/>
+            <array length="1" zero-terminated="0" c:type="char*">
+              <type name="utf8" c:type="char"/>
+            </array>
```

@EPashkin Any idea why, or should I research/debug?

sdroege:
https://bugzilla.gnome.org/show_bug.cgi?id=784035 (bug in gobject-introspection that causes it to drop `const`)

EPashkin:
`<type name="utf8" c:type="char"/>` gir think that type "utf8" is string and ignore "char" without * in normal mode.
In sys mode it maybe do different

EPashkin:
One symbol has other type name. IMHO pango_layout_set_markup accepts just text, and len need used from it.

sdroege:
> <type name="utf8" c:type="char"/> gir think that type "utf8" is string and ignore "char" without * in normal mode.

Ah good point. I guess we need to parse the c:type of this, or the surrounding <array> to know for sure. Otherwise this is ambiguous.

 > In sys mode it maybe do different

In sys mode it seems to count the asterisks (i.e. look at the c:type instead)

> One symbol has other type name. IMHO pango_layout_set_markup accepts just text, and len need used from it.

What do you mean? It also has to be changed the same way as `set_text`, yes.

EPashkin:
I mean that pango_layout_set_markup accepts not array but utf8 string and array-length attribute may be cant applied to it. Same as "set_text".

EPashkin:
I meant that string `uft8` != array [u8] (at minimum it "implies" that contains valid utf8 characters)
so I don't think that annotations in https://bugzilla.gnome.org/show_bug.cgi?id=784021 can be applied to strings.

sdroege:
All those functions are working with UTF8 strings. The markup ones just in addition require the strings to be valid Pango markup (or things will fail), just like GTK labels if you configure them to show Pango markup

EPashkin:
IMHO all array-related problems is solved, can I merge this, or you want add something?

sdroege:
It's doing the wrong things for strings, I'd like to fix that as part of this. Otherwise the array length annotations on strings (as exist in various libraries already) cause wrong code to be generated.

But you could merge it now already if you want, and I'll work on that in another pull request then.

EPashkin:
Ok, I wait then

sdroege:
Good to merge. I discussed the string+length stuff with gobject-introspection people on IRC, and there's no way to correctly annotate that currently unless your "string" is actually an array of bytes (and not a UTF8 string), in which case it would be (array length=len) (element-type guint8).

EPashkin:
Thanks, @sdroege.
After merge I will wait for regen PRs (glib and gtk most needed as it don't builds).

sdroege:
I do them, you do them, ...?

EPashkin:
> I wait 
😉

sdroege:
Alright, I'll try to get to that tonight then :)

sdroege:
All submitted

EPashkin:
@sdroege can you add link to all these PRs to this or better to #376 ?

Issue #350, opened 2017-04-15
Out String parameters


EPashkin:
Allow to generate functions that returning string as output parameters:
g_action_parse_detailed_name
g_menu_attribute_iter_get_next //types differ in mutability
g_menu_link_iter_get_next //types differ in mutability
gtk_image_get_icon_name //types differ in mutability
gtk_image_get_stock

GuillaumeGomez:
Is this issue still not fixed?

EPashkin:
Almost.
At it generated, but without deprecated `gtk_image_get_stock`
But `gtk_image_get_icon_name` causes panic in `from_glib_none` when called on Image, created from file.
So maybe function description need changed to `fn get_icon_name(&self) -> (Option<String>, i32);`

sdroege:
Why does it panic?

EPashkin:
Seems this image just don't have name, so ffi function return null.

sdroege:
Ah, nothing badly broken then.

EPashkin:
Yes, only need change return string to optional

PR #389, opened 2017-06-19
Add support for String/PathBuf out parameters

sdroege:
Part of #350

### DIFF
--- a/src/analysis/out_parameters.rs
+++ b/src/analysis/out_parameters.rs
@@ -96,9 +96,9 @@ pub fn analyze_imports(env: &Env, func: &Function, imports: &mut Imports) {
     for par in &func.parameters {
         if par.direction == ParameterDirection::Out {
             match *env.library.type_(par.typ) {
-                Type::Fundamental(..) |
-                    Type::Bitfield(..) |
+                Type::Bitfield(..) |
                     Type::Enumeration(..) => imports.add("std::mem", func.version),
+                Type::Fundamental(fund) if fund != Fundamental::Utf8 && fund != Fundamental::Filename => imports.add("std::mem", func.version),
                 _ if !par.caller_allocates => imports.add("std::ptr".into(), func.version),
                 _ => (),
             }
--- a/src/analysis/rust_type.rs
+++ b/src/analysis/rust_type.rs
@@ -202,16 +202,8 @@ pub fn parameter_rust_type(env: &Env, type_id: library::TypeId,
     let type_ = env.library.type_(type_id);
     let rust_type = rust_type_full(env, type_id, nullable, ref_mode);
     match *type_ {
-        Fundamental(fund) => {
-            if fund == library::Fundamental::Utf8 || fund == library::Fundamental::Filename {
-                match direction {
-                    library::ParameterDirection::In |
-                        library::ParameterDirection::Return => rust_type,
-                    _ => Err(TypeError::Unimplemented(into_inner(rust_type))),
-                }
-            } else {
-                rust_type.map_any(|s| format_parameter(s, direction))
-            }
+        Fundamental(..) => {
+            rust_type.map_any(|s| format_parameter(s, direction))
         }
 
         Alias(ref alias) => {
--- a/src/chunk/chunk.rs
+++ b/src/chunk/chunk.rs
@@ -19,6 +19,7 @@ pub enum Chunk {
     Let{name: String, is_mut: bool, value: Box<Chunk>, type_: Option<Box<Chunk>>},
     Uninitialized,
     UninitializedNamed{name: String},
+    NullPtr,
     NullMutPtr,
     Custom(String),
     Tuple(Vec<Chunk>, TupleMode),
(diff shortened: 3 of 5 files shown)


src/analysis/out_parameters.rs
```diff
@@ -96,9 +96,9 @@ pub fn analyze_imports(env: &Env, func: &Function, imports: &mut Imports) {
     for par in &func.parameters {
         if par.direction == ParameterDirection::Out {
             match *env.library.type_(par.typ) {
-                Type::Fundamental(..) |
-                    Type::Bitfield(..) |
+                Type::Bitfield(..) |
                     Type::Enumeration(..) => imports.add("std::mem", func.version),
+                Type::Fundamental(fund) if fund != Fundamental::Utf8 && fund != Fundamental::Filename => imports.add("std::mem", func.version),
```
EPashkin: Too long line

sdroege: Where would you like it to be split? :) Before the `if` or before the `=>`?

It might make sense to run everything through the new rustfmt (rustfmt-nightly) at some point to make everything consistent.

EPashkin:
In glib many:
```
error[E0308]: mismatched types
  --> D:/eap/rust/0/glib\src\auto\functions.rs:52:62
   |
52 |         let ret = ffi::g_ascii_strtod(nptr.to_glib_none().0, &mut endptr);
   |                                                              ^^^^^^^^^^^ types differ in mutability
   |
   = note: expected type `*mut *mut i8`
              found type `&mut *const _`
```
and
```
error: no associated item named `uninitialized` found for type `std::string::String` in the current scope
    --> D:/eap/rust/0/glib\src\auto\functions.rs:1767:26
     |
1767 |         let mut outbuf = String::uninitialized();
     |                          ^^^^^^^^^^^^^^^^^^^^^
     |
     = help: items from traits can only be used if the trait is implemented and in scope; the following trait defines an item `uninitialized`, perhaps you need to implement it:
     = help: candidate #1: `translate::Uninitialized`
```

sdroege:
Needs some fixes for things in GLib then, thanks!

EPashkin:
`pango_skip_space` has "inout" parameter and it now processed. IMHO we don't support "inout" parameters even for int.
1 Error in Gtk :
```
error[E0308]: mismatched types
   --> D:/eap/rust/0/gtk\src\auto\image.rs:234:61
    |
234 |             ffi::gtk_image_get_stock(self.to_glib_none().0, &mut stock_id, &mut size);
    |                                                             ^^^^^^^^^^^^^ types differ in mutability
    |
    = note: expected type `*mut *mut i8`
               found type `&mut *const _`
```
After removing bad functions examples compiled

src/analysis/out_parameters.rs
EPashkin: IMHO near && and near =>. Not sure before or after in both cases.

src/analysis/rust_type.rs
```diff
@@ -202,16 +202,8 @@ pub fn parameter_rust_type(env: &Env, type_id: library::TypeId,
     let type_ = env.library.type_(type_id);
     let rust_type = rust_type_full(env, type_id, nullable, ref_mode);
     match *type_ {
-        Fundamental(fund) => {
-            if fund == library::Fundamental::Utf8 || fund == library::Fundamental::Filename {
-                match direction {
-                    library::ParameterDirection::In |
-                        library::ParameterDirection::Return => rust_type,
```
EPashkin: Here lost "Unimplemented" for `InOut`.
Seems we actually support inout parameters for other types. I don't remember this :(

src/analysis/rust_type.rs
```diff
@@ -202,16 +202,8 @@ pub fn parameter_rust_type(env: &Env, type_id: library::TypeId,
...
-                    _ => Err(TypeError::Unimplemented(into_inner(rust_type))),
-                }
-            } else {
-                rust_type.map_any(|s| format_parameter(s, direction))
-            }
+        Fundamental(..) => {
+            rust_type.map_any(|s| format_parameter(s, direction))
         }
```
EPashkin: Unneeded brackets

sdroege:
For GLib, I think all the `g_ascii_*` and `g_unichar_*` and `g_str*` and `g_utf8_*` functions should be ignored as there are alternatives in `std` and they generally do *special* things. This covers all the failures in GLib.

Pango should be fine after ignoring inout again. Gtk is fine after ignoring `gtk_image_get_stock` (which was not enabled before, has wrong types in C and is deprecated anyway).

EPashkin:
I agree that all bad glib and pango functions unneeded and can be safety ignored.
`gtk_image_get_stock` while can be fixed with `ptr::null_mut` need some utility function to detect mutable array elements.

EPashkin:
.. and currently better ignored too.

EPashkin:
@sdroege, Thanks

sdroege:
> can be fixed with ptr::null_mut need some utility function to detect mutable array elements

We could do some better detection for that based on the actual C type (`const char **` vs. `char **`), instead of just looking at the transfer annotation. If ever needed for more than a couple of functions. Would need a single change in the place where currently between NullPtr and NullMutPtr is decided.

sdroege:
Also there needs to be a check added for forbidding mutable string parameters (or implementing them), see e.g. `pango_skip_space()`. This will now create a copy of the string, pass it to the function, have it modified and then destroyed.

PR #390, opened 2017-06-19
Run everything through rustfmt-nightly 0.1.5

sdroege:
Just a proposal for now. Might make sense to increase the line length size or otherwise customize, and put a rustfmt configuration into the repository then.


EPashkin:
IMHO better do it after ll your PR merged. Also latest produce slightly different result

sdroege:
> IMHO better do it after ll your PR merged

Sure, but what do you think about the changes? Anything that should be done different?

> Also latest produce slightly different result

Which latest do you mean?

EPashkin:
IMHO better stay with default config.
I missed that rustfmt has separate crate for nightly version, so used just normal rustfmt.

EPashkin:
I only has visual problem with multiline function declarations and calls, but I can live with it :wink:
So if you will rebase other PRs on top of this, I can merge it now.

sdroege:
The nightly version is using the parser from rustc directly, the other one is not developed any further. Otherwise the results should be more or less the same still

sdroege:
Better wait until the other pull requests are ready, especially the large array length one

sdroege:
Let's get this in after the array length change is merged? I'll rebase it then

sdroege:
Updated to latest master

EPashkin:
Thanks.

PR #391, opened 2017-06-21
Don't support String InOut parameters and &mut String Out parameters

sdroege:
These are not implemented yet.

See https://github.com/gtk-rs/gir/pull/389#issuecomment-309989621


EPashkin:
Thanks, @sdroege

Issue #379, opened 2017-05-28
Send/Sync impls / missing bounds


sdroege:
Currently the assumption seems to be that all GObjects and other data types are not Send and/or Sync, and every callback (e.g. for signals) is only ever called from the same thread where the connection was made to.

This is true in the context of Gtk/Gdk (neither Send nor Sync) or Pango (Send but not Sync), but not for many things in GIO (and even GLib, see GClosure or the GObject::notify signal, both being Send+Sync) or especially GStreamer. For the case of GStreamer the default would be to assume that everything is Send+Sync, and every signal or other callback's closure needs Sync+Send bounds (the number of exceptions can be counted on one hand).

As such it would be good to somehow be able to configure this (as GI does not have any annotations for this):

1. A global per-library configuration for a default. `default_concurrency` with values `none`, `send`, `sync`, `send+sync`. The default should probably stay `none`
2. Per object / signal configuration for the same, which overrides the above. `concurrency`

This would then impl the corresponding traits for the structs, and require the corresponding trait bounds for signals, etc.

The main open question is how to handle this in the context of sub-typing. While e.g. GObject::notify should be Send+Sync in the case of Send+Sync sub-types (e.g. GstElement), it is not required to be so otherwise (e.g. GtkWidget).

Any comments, ideas, ...? :)

EPashkin:
What problem can arise is base function don't have extra constraint and we add it only for function that use it (like signals connect)?

sdroege:
Can you give a concrete example for what you mean?

sdroege:
You mean for example connecting to e.g. the `notify` signal of GObject with a closure without `Send+Sync` for an object that actually is `Send+Sync`? In that case your closure could be called from multiple threads at once, while whatever data you pass into the closure might not be thread-safe at all (e.g. using Rc instead of Arc, not using Mutex but RefCell, ...).

EPashkin:
If glib::signal::connect don't have Send+Sync function constraint,
but we add it for gio::ActionGroupExt::connect_action_added (just for example, don't checked that it really needed), there will be problem?
If not then your proposal is very good.

EPashkin:
It really problem that we can't add limits on part glib::signal::connect call (for connecting "notify::activate" as need for GtkSwitch)

sdroege:
[...]

PR #393, opened 2017-06-25
Implement Send/Sync configuration for objects/records

sdroege:
This allows configuring crate-wide and per object/record if an object is
supposed to be Send, Sync or Send + Sync. The marker traits will
automatically be implemented and signals callbacks on those objects will
get the appropriate trait bounds.

Note that it is still required to manually ensure that any signal of
parent or child types has at least the same trait bounds.

https://github.com/gtk-rs/gir/issues/379

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,30 @@ status = "generate"
 Note that you must not place `Gtk.*` into the `generate` array and
 additionally configure its members.
 
+In various cases, GObjects or boxed types can be used from multiple threads
+and have certain concurrency guarantees. This can be configured with the
+`concurrency` setting at the top-level options or per object. It will
+automatically implement the `Send` and `Sync` traits for the resulting object
+and set appropriate trait bounds for signal callbacks. The default is `none`,
+and apart from that `send` and `send+sync` are supported.
+
+```toml
+[[object]]
+# object's fullname
+name = "Gtk.SomeClass"
+# can be also "manual" and "ignore" but it's simpler to just put the object in the same array
+status = "generate"
+# concurrency of the object, default is set in the top-level options or
+# otherwise "none". Valid values are "none", "send" and "send+sync"
+concurrency = "send+sync"
+```
+
+Note that `send` is only valid for types that are either not reference counted
+(i.e. `clone()` copies the object) or that are read-only (i.e. no API for
+mutating the object exists). `send+sync` is valid if the type can be sent to
+different threads and all API allows simultaneous calls from different threads
+due to internal locking via e.g. a mutex.
+
 ### Generation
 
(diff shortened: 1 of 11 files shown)


EPashkin:
`config::parse_status_shorthands` also need concurrency support
Ex. `gio::AppLaunchContext` don't changed on global concurrency change

sdroege:
Will check later. I also noticed that Send without Sync is unsafe for all types with reference counting. This makes it necessary to mark many GIO types as neither and applications would have to wrap them in something thread safe and unsafe impl both marker traits then.
Types that copy are fine though. And for GStreamer it's also all fine as basically everything is Send+Sync or Send+"Copy".

Sync without Send probably does not exist, it seems to make no sense. Or am I missing something?

Comments?

EPashkin:
You right. Just declaring type Sync don't make it sync, types need replace internal `Rc` with `Arc`, that possible will be problem with current glib's traits
Agree that Sync without Send make no sense.

sdroege:
That's not the problem. The problem is Send I'm combination with reference counting: this combination is only Send if the underlying object is Sync. So in our case, ref counted types are Send if the underlying type is Send and Sync, or Send and read-only by API.

The reference counting itself is Sync in GObject. That's not the problem. It is always like Arc.


So in summary: Send+Sync is always fine, Send is fine if underlying type is copying instead of reference counting (some boxed types) or read-only (only getter style APIs). Sync only makes no sense.
The user will have to configure that correctly, we can't reliably check that. Just like when normally implementing Sync and Send in Rust (it requires unsafe because the compiler can't know).

And this means that types in GIO and GLib that are Send in theory can't be marked as that because of the reference counting and the user has to wrap them in a safe construct that gets an unsafe Send implementation.



So. I propose to remove Sync-only, and apart from that we document the above. And I fix the problem you found. That should cover everything and allows me to continue with the GStreamer bindings :)

EPashkin:
Agree.


I completely forgot that `gtk-rs\glib` don't use `Rc` actually: it just used C pointers in `Rc<RefCell<T>>` way, so I make previous remarks. Sorry.

EPashkin:
@sdroege And, please, update `README.md` with new config options.

sdroege:
Except for Readme.md update this should be good now, please test and I'll write a little bit later :)

EPashkin:
Seems now all generated as planned.

sdroege:
README.md updated too now

EPashkin:
In code "send+sync" in readme "send-sync"

sdroege:
Thanks, fixed

EPashkin:
Thanks, @sdroege

PR #394, opened 2017-06-26
Add customization of crate names for external libraries

EPashkin:
Added extra config for external libraries
```
[options]
external_libraries = [
   "GLib",
   "Gdk",
   "GdkPixbuf",
]

[external_libraries]
crate-name="NameSpace"
```

Edit: switched parameters

### DIFF
--- a/src/config/external_libraries.rs
+++ b/src/config/external_libraries.rs
@@ -1,0 +1,115 @@
+use toml;
+
+use super::error::*;
+use nameutil::crate_name;
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct ExternalLibrary {
+    pub namespace: String,
+    pub crate_name: String,
+}
+
+pub fn read_external_libraries(toml: &toml::Value) -> Result<Vec<ExternalLibrary>> {
+    let mut external_libraries = match toml.lookup("options.external_libraries") {
+        Some(a) => {
+            try!(a.as_result_vec("options.external_libraries"))
+                .iter()
+                .filter_map(|v| v.as_str().map(String::from))
+                .map(|namespace| {
+                    ExternalLibrary {
+                        crate_name: crate_name(&namespace),
+                        namespace: namespace,
+                    }
+                })
+                .collect()
+        }
+        None => Vec::new(),
+    };
+    let custom_libs = toml.lookup("external_libraries").and_then(|v| v.as_table());
+    if let Some(custom_libs) = custom_libs {
+        for custom_lib in custom_libs {
+            let crate_name = custom_lib.0;
+            if let Some(namespace) = custom_lib.1.as_str() {
+                let lib = ExternalLibrary {
+                    namespace: namespace.to_owned(),
+                    crate_name: crate_name.clone(),
+                };
+                external_libraries.push(lib);
+            } else {
+                bail!("For external library \"{}\" namespace must be string", crate_name);
+            }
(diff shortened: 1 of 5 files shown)


EPashkin:
@antoyo This allow fix [problem](https://github.com/antoyo/webkit2gtk-rs/issues/4#issuecomment-310909823), check it please.
@sdroege, maybe it fixed your problem too.

antoyo:
It works fine.
Thanks!

EPashkin:
Seems I take wrong turn.
May be it better be this?
```
[external_libraries]
crate-name="NameSpace"
```

antoyo:
I don't mind: either way is fine.

EPashkin:
I better change

EPashkin:
Updated

src/config/external_libraries.rs
```diff
@@ -0,0 +1,116 @@
...
+   "GLib",
+   "Gdk",
+   "GdkPixbuf",
+]
+
+[external_libraries]
+coollib="CoolLib"
+other-lib="OtherLib"
+"#,
+        );
```
EPashkin: Very strange format for function call (with ending comma), but as it compiles seems it OK.

src/config/external_libraries.rs
```diff
@@ -0,0 +1,115 @@
...
+[options]
+external_libraries = [
+   "GLib",
+   "Gdk",
+   "GdkPixbuf",
+]
+
+[external_libraries]
+coollib="CoolLib"
+other-lib="OtherLib"
```
sdroege: Are all possible crate names valid toml keys? Or are there crate names that wouldn't be allowed here?

EPashkin: If it valid for `[dependencies]`

sdroege: Good point! 👍

EPashkin:
@sdroege, this enough for GStreamer-sys?

sdroege:
This doesn't allow renaming the sys crate itself, right? Only other sys crates it is depending on?

EPashkin:
Only dependencies, for current crate name better fix https://github.com/gtk-rs/gir/blob/master/src/nameutil.rs#L136. But IMHO it correctly works for `GStreamer`:
`assert_eq!(crate_name("GStreamer"), "gstreamer");`
Can you say more what you want?

sdroege:
For the -sys crate, the generated crate name is `gst` (and `gst-base` etc). I want it to be `gstreamer` (and `gstreamer-base` etc).

EPashkin:
It strange: Gir don't change `name` in `[lib]` and `[package]` in `Cargo.toml` on regen,
so change need only once. Seems I missed something.

EPashkin:
@sdroege, Tried regen for https://github.com/sdroege/gstreamer-sys,
other that now customized external libraries I found only problematic rewriting `package.links` with crate name,
so simply moved it to part, that fill empty Cargo.toml

sdroege:
Fine with me if that all works :) Thanks!

sdroege:
@EPashkin Can you send me the changes you did for gstreamer-sys so I can merge them?

EPashkin:
Sorry, I don't make any changes in gstreamer-sys. Just checked what changes on regen.

sdroege:
Ah, no problem then :) Just wanted to prevent doing your changes another time

PR #395, opened 2017-06-26
Generate correctly when mutable reference into Into

### LINKED ISSUE #382: The Into trait doesn't take into account if the variable is mutable
For example, when I generated for GtkSourceView, I got the following function:

```rust
fn create_context<'a, P: Into<Option<&'a gtk::TextIter>>>(&self, position: P) -> Option<CompletionContext> {
    let position = position.into();
    let position = position.to_glib_none_mut();
    unsafe {
        from_glib_none(ffi::gtk_source_completion_create_context(self.to_glib_none().0, position.0))
    }
}
```

However, the good header is:

```rust
fn create_context<'a, P: Into<Option<&'a mut gtk::TextIter>>>(&self, position: P) -> Option<CompletionContext> {
```

So `position` needs to be `&mut` instead of just `&`.

GuillaumeGomez:
Fixes #382.

### DIFF
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -16,14 +16,14 @@ pub enum BoundType {
     IsA(Option<char>),
     // lifetime <- shouldn't be used but just in case...
     AsRef(Option<char>),
-    // lifetime (if none, not a reference, like for primitive types), wrapper type
-    Into(Option<char>, Option<Box<BoundType>>),
+    // lifetime (if none, not a reference, like for primitive types), wrapper type, is mutable
+    Into(Option<char>, Option<Box<BoundType>>, bool),
 }
 
 impl BoundType {
     pub fn is_into(&self) -> bool {
         match *self {
-            BoundType::Into(_, _) => true,
+            BoundType::Into(_, _, _) => true,
             _ => false,
         }
     }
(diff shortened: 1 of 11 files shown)


GuillaumeGomez:
cc @EPashkin

EPashkin:
I already commented in https://github.com/gtk-rs/sourceview/pull/7.
Maybe I was wrong, but currently I don't see usage of this PR.

GuillaumeGomez:
For sourceview, the problem can be overlook but more globally, this addresses potential future issues: if the wrapped type should be a mutable reference, actually, it isn't generated as a mutable one.

EPashkin:
I remember only 2 possibilities for optional for mutable reference:
1) is "out" param, we handle this by always supplying value,
2) "inout" params, we currently handle it for simply types (Ex. gtk_editable_insert_text, as nonoptional)
In gtk optional inouts only gtk_init_*, we handle it manually or ignore.
In gdk: gdk_inits (same as gtk)
In glib: g_prefix_error (ignored)
In pango: pango_matrix_transform_pixel_rectangle (manual, as nonoptional) and pango_matrix_transform_rectangle (same)

GuillaumeGomez:
So possibilities exist! :D

EPashkin:
I don't see where it can be legally applied.
Only last possibilities in pango but from docs I don't see how it can be used without `rect`.

GuillaumeGomez:
Well, I'll let you decide. Just close it if you really don't see it useful. ;)

EPashkin:
@GuillaumeGomez, Sorry, but I don't see application of this PR in gtk-rs crates, so I close it.

PR #397, opened 2017-07-02
Destroy

GuillaumeGomez:
It fixes a problem I encountered when generated the binding for the `Attribute` class in pango.

In the gir file:

```xml
<method name="destroy" c:identifier="pango_attribute_destroy">
  <doc xml:space="preserve">Destroy a #PangoAttribute and free all associated memory.</doc>
  <return-value transfer-ownership="none">
    <type name="none" c:type="void"/>
  </return-value>
  <parameters>
    <instance-parameter name="attr" transfer-ownership="none">
      <doc xml:space="preserve">a #PangoAttribute.</doc>
      <type name="Attribute" c:type="PangoAttribute*"/>
    </instance-parameter>
  </parameters>
</method>
```

### DIFF
--- a/src/analysis/special_functions.rs
+++ b/src/analysis/special_functions.rs
@@ -25,7 +25,7 @@ impl FromStr for Type {
             "compare" => Ok(Compare),
             "copy" => Ok(Copy),
             "equal" => Ok(Equal),
-            "free" => Ok(Free),
+            "free" | "destroy" => Ok(Free),
             "is_equal" => Ok(Equal),
             "ref" | "ref_" => Ok(Ref),
             "to_string" => Ok(ToString),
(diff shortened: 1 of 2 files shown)


EPashkin:
Now GtkWindow and GdkWindow lost `destroy` function.
IMHO this problem, maybe we need show some special function for non boxed.
Can't test it now as my rust installation works strange even when I changed to stable.

```
error: no matching package named `atk-sys` found (required by `gtk-sys`)
location searched: file:///D:/eap/rust/0/sys/atk-sys
version required: ^0.3.4
```

EPashkin:
Error was from my test for https://github.com/gtk-rs/gir/pull/394, completly forgot about it.

Missing `destroy` in `Gtk` broke `treeview.rs`

GuillaumeGomez:
Strange error... Did you try `cargo update`? :p

> Missing `destroy` in `Gtk` broke `treeview.rs`

Not sure to understand. I'll try to generate `Gtk` with this gir version to see what's going on. :D

GuillaumeGomez:
Ok, I found the issue. That'll be fun to "fix". T_T

GuillaumeGomez:
Ok, should be fixed now.

src/analysis/special_functions.rs
```diff
@@ -35,29 +35,55 @@ impl FromStr for Type {
     }
 }
 
-pub type Infos = BTreeMap<Type, String>; //Type => glib_name
+pub type Infos = BTreeMap<Type, String>; // Type => glib_name
+
+fn update_func(func: &mut FuncInfo, type_: Type) -> bool {
```
EPashkin: Strange return value: return true if failed.
Can you invert it?

src/analysis/special_functions.rs
```diff
@@ -35,29 +35,55 @@ impl FromStr for Type {
...
+    let mut destroy = None;
 
-    for func in functions.iter_mut() {
+    for (pos, func) in functions.iter_mut().enumerate() {
         if let Ok(type_) = Type::from_str(&func.name) {
-            if func.visibility != Visibility::Comment {
-                func.visibility = visibility(type_, func.parameters.len());
+            if func.name == "destroy" {
+                destroy = Some((func.glib_name.clone(), pos));
+                continue
```
EPashkin: Please `continue;`

src/analysis/special_functions.rs
```diff
@@ -35,29 +35,55 @@ impl FromStr for Type {
...
+            }
+            if update_func(func, type_) {
+                continue
```
EPashkin: same

EPashkin:
Next comment can be ignoring and not fixed.
I don't like that names used in both `from_str` and `extract`,
maybe better add new case to `enum Type` with comment that it will be not used in `specials` ?

GuillaumeGomez:
The problem with this solution is that it'll spread out of this file and require more checks. Or maybe I misunderstood you?

EPashkin:
IMHO only `from_str`, ` extract` and `visibility` affected when adding new case to `enum Type` if "destroy" added as `Type::Free` to `special` in current `if destroy.is_some() && has_copy && !has_free {`

GuillaumeGomez:
I'm not sure this is really important... How much do you want it? :D

EPashkin:
As I say from beginning: its not important.
I can merge PR in current state if you don't want do it now.

GuillaumeGomez:
It's not that I don't want to: I don't see the point but I'm open to do it if you think it is important.

EPashkin:
I merge.
Thanks @GuillaumeGomez

GuillaumeGomez:
Don't hesitate to update the code as you have it in mind. :)

PR #398, opened 2017-07-04
Import everything from auto in case a type is an alias of an auto type

GuillaumeGomez:
(Another issue when generating pango haha)

cc @EPashkin


EPashkin:
@GuillaumeGomez thanks

EPashkin:
@sdroege I almost sure it don't cause problem in your crates, but to be sure, please, check this results too.

sdroege:
Should be fine

PR #402, opened 2017-07-06
Full enablement of Unions using nightly Rust

Luke-Nukem:
Unions are fully enabled behind a feature-gate. Both non-nightly and nightly Rust appear to be working well.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -152,7 +152,7 @@ impl Library {
                             try!(self.read_class(parser, ns_id, &attributes));
                         }
                         "record" => {
-                            try!(self.read_record(parser, ns_id, &attributes));
+                            try!(self.read_record_start(parser, ns_id, &attributes));
                         }
                         "union" => {
                             try!(self.read_named_union(parser, ns_id, &attributes));
@@ -265,13 +265,30 @@ impl Library {
                         "virtual-method" => try!(ignore_element(parser)),
                         "doc" => doc = try!(read_text(parser)),
                         "union" => {
-                            let (union_fields, _, doc) = try!(self.read_union(parser, ns_id));
-                            let typ = Type::union(self, union_fields);
-                            fields.push(Field {
-                                typ: typ,
-                                doc: doc,
-                                ..Field::default()
-                            });
+                            #[cfg(feature = "use_unions")]
+                            {
+                                use self::Type::*;
+                                if let Union(u) =
+                                        try!(self.read_union_unsafe(parser, ns_id, attrs)) {
+                                    let u_doc = u.doc.clone();
+                                    let type_id = Type::union(self, u, ns_id);
+                                    fields.push(Field {
+                                        typ: type_id,
+                                        doc: u_doc,
+                                        ..Field::default()
+                                    });
+                                };
+                            }
+                            #[cfg(not(feature = "use_unions"))]
+                            {
+                                let (union_fields, _, doc) = try!(self.read_union(parser, ns_id));
+                                let typ = Type::union(self, union_fields);
+                                fields.push(Field {
+                                    typ: typ,
+                                    doc: doc,
+                                    ..Field::default()
+                                });
+                            }
                         }
                         x => bail!(mk_error!(format!("Unexpected element <{}>", x), parser)),
                     }
@@ -302,18 +319,32 @@ impl Library {
         Ok(())
     }
 
-    fn read_record(&mut self, parser: &mut Reader, ns_id: u16, attrs: &Attributes) -> Result<()> {
+    fn read_record_start(&mut self, parser: &mut Reader,
+                         ns_id: u16, attrs: &Attributes) -> Result<()> {
+
+        if let Some(typ) = try!(self.read_record(parser, ns_id, attrs)) {
+            let name = typ.get_name().clone();
+            self.add_type(ns_id, &name, typ);
+        }
+        Ok(())
+    }
+
+    fn read_record(&mut self, parser: &mut Reader, ns_id: u16, attrs: &Attributes)
+        -> Result<Option<Type>> {
         let mut name = try!(
             attrs
                 .by_name("name")
-                .ok_or_else(|| mk_error!("Missing record name", parser))
+                .ok_or_else(|| mk_error!("Missing union name", parser))
         );
         let mut c_type = try!(
             attrs
                 .by_name("type")
                 .ok_or_else(|| mk_error!("Missing c:type attribute", parser))
         );
-        let get_type = attrs.by_name("get-type");
+        let get_type = match attrs.by_name("get-type") {
+            Some(s) => Some(s.to_string()),
+            None => None
+        };
         let version = try!(self.parse_version(parser, ns_id, attrs.by_name("version")));
         let deprecated_version = try!(self.parse_version(
             parser,
@@ -324,6 +355,8 @@ impl Library {
         let mut fns = Vec::new();
         let mut doc = None;
         let mut doc_deprecated = None;
+        #[cfg(feature = "use_unions")]
+        let mut union_count = 1;
         loop {
             let event = try!(parser.next());
             match event {
@@ -341,16 +374,60 @@ impl Library {
                             ))
                         }
                         "union" => {
-                            let (union_fields, _, doc) = try!(self.read_union(parser, ns_id));
-                            let typ = Type::union(self, union_fields);
-                            fields.push(Field {
-                                typ: typ,
-                                doc: doc,
-                                ..Field::default()
-                            });
+                            #[cfg(feature = "use_unions")]
+                            {
+                                use self::Type::*;
+                                if let Union(mut u) =
+                                        try!(self.read_union_unsafe(parser, ns_id, attrs)) {
+                                    // A nested union->struct->union typically has no c_type
+                                    // so we iterate over fields to find it. These fields are
+                                    // within the nested union->struct if found
+                                    let mut nested = true;
+                                    for f in &mut u.fields {
+                                        if f.c_type.is_none() ||
+                                                c_type == u.c_type.as_ref().unwrap() {
+                                            nested = true;
+                                            u.name = format!("{}_u{}",c_type,union_count);
+                                            u.c_type = Some(format!("{}_u{}",c_type,union_count));
+                                        }
+                                    }
+                                    let ctype = u.c_type.clone();
+                                    let u_doc = u.doc.clone();
+                                    let type_id = Type::union(self, u, ns_id);
+                                    if nested {
+                                        fields.push(Field {
+                                            name: format!("u{}", union_count),
+                                            typ: type_id,
+                                            doc: u_doc,
+                                            c_type: ctype,
+                                            ..Field::default()
+                                        });
+                                    } else {
+                                        fields.push(Field {
+                                            typ: type_id,
+                                            doc: u_doc,
+                                            c_type: ctype,
+                                            ..Field::default()
+                                        });
+                                    }
+                                    union_count += 1;
+                                };
+                            }
+                            #[cfg(not(feature = "use_unions"))]
+                            {
+                                let (union_fields, _, doc) = try!(self.read_union(parser, ns_id));
+                                let typ = Type::union(self, union_fields);
+                                fields.push(Field {
+                                    typ: typ,
+                                    doc: doc,
+                                    ..Field::default()
+                                });
+                            }
                         }
                         "field" => {
-                            fields.push(try!(self.read_field(parser, ns_id, &attributes)));
+                            if let Ok(f) = self.read_field(parser, ns_id, &attributes) {
+                                fields.push(f);
+                            }
                         }
                         "doc" => doc = try!(read_text(parser)),
                         "doc-deprecated" => doc_deprecated = try!(read_text(parser)),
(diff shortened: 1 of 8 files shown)


EPashkin:
@Luke-Nukem Can you rebase this to master with removing merges?
Why `GVariantBuilder_u1` etc. added to all generated libs in https://github.com/gtk-rs/sys when this feature enabled?
Bad changes in glib-sys with bitsized fields: GDate as example
```diff
#[repr(C)]
 pub struct GDate {
-    _truncated_record_marker: c_void,
-    //julian_days: guint: 32,
-    //julian: guint: 1,
-    //dmy: guint: 1,
-    //day: guint: 6,
-    //month: guint: 4,
-    //year: guint: 16,
+    pub julian_days: c_uint,
+    pub julian: c_uint,
+    pub dmy: c_uint,
+    pub day: c_uint,
+    pub month: c_uint,
+    pub year: c_uint,
 }
```

Strange that GScanner has `//union` comments before, but now it seems fixed, thanks

src/parser.rs
```diff
@@ -362,9 +438,10 @@ impl Library {
             }
         }
 
-        if attrs.by_name("is-gtype-struct").is_some() {
-            return Ok(());
-        }
+        // TODO: unsure of how to handle this
+        //if attrs.by_name("is-gtype-struct").is_some() {
+        //    return Ok(());
+        //}
```
EPashkin: Seems `read_record` return type need change to `Result<Option<Type>>`

Luke-Nukem:
Regarding the appending of `_u1` to various unions or `_s1` to nested union->structs, it's because without it we end up with each struct->union->struct having the same c_type, and thus the same name for each.

Regarding
```
#[repr(C)]
 pub struct GDate {
-    _truncated_record_marker: c_void,
-    //julian_days: guint: 32,
-    //julian: guint: 1,
-    //dmy: guint: 1,
-    //day: guint: 6,
-    //month: guint: 4,
-    //year: guint: 16,
+    pub julian_days: c_uint,
+    pub julian: c_uint,
+    pub dmy: c_uint,
+    pub day: c_uint,
+    pub month: c_uint,
+    pub year: c_uint,
 }
```
Well of course... We can't initialise a newly created struct like that. Have to declare it first, then initialise it. If it were an enum instead of a struct then it might work.

If you look at;
```
struct GDate {
  guint julian_days : 32; /* julian days representation - we use a
                           *  bitfield hoping that 64 bit platforms
                           *  will pack this whole struct in one big
                           *  int
                           */

  guint julian : 1;    /* julian is valid */
  guint dmy    : 1;    /* dmy is valid */

  /* DMY representation */
  guint day    : 6;
  guint month  : 4;
  guint year   : 16;
};
```
each field is a guint, which translates to a c_uint. And they are initialised (which we can't do in rust).

src/parser.rs
Luke-Nukem: Yes but it also needs an initialised Type to return. This is where I am unsure what to do... I could maybe return a `default()` initialised Type?

EPashkin:
As I remember size GDate in C will be 64bit or 8bytes with `julian`, `dmy` and `day` stored in byte 5)
Your version: 6*4=24 bytes;

src/parser.rs
EPashkin: Why it can't return `Ok(None)` in this case? And don't add type in `read_record_start`

Luke-Nukem: `Result<Option<Type>>`.... I don't know why my mind completely ignored that....
Done.

Luke-Nukem:
> As I remember size GDate in C will be 64bit or 8bytes with julian, dmy and day stored in byte 5)
> Your version: 6*4=24 bytes;

Looking at [guint](https://developer.gnome.org/glib/stable/glib-Basic-Types.html#guint) it seems it will be a particular size depending on a `#define G_MAXUINT UINT_MAX`.

 `UINT_MAX` is defined in Rust too, and c_uint size is defined by that. So essentially it looks like both `guint` and `c_uint` are defined the same way, and so should be the same size.

src/parser.rs
```diff
@@ -21,6 +21,7 @@ pub fn is_empty_c_type(c_type: &str) -> bool {
     c_type == EMPTY_CTYPE
 }
 
+
```
EPashkin: Unneeded extra line

src/parser.rs
```diff
@@ -322,24 +323,19 @@ impl Library {
...
-        let mut c_type = try!(
-            attrs
-                .by_name("type")
-                .ok_or_else(|| mk_error!("Missing c:type attribute", parser))
-        );
+        -> Result<Option<Type>> {
+        let mut name = try!(attrs.by_name("name")
+                .ok_or_else(|| mk_error!("Missing record name", parser)));
+        let mut c_type = try!(attrs.by_name("type")
+                .ok_or_else(|| mk_error!("Missing c:type attribute", parser)));
```
EPashkin: [rustfmt-nigthly](https://crates.io/crates/rustfmt-nightly) change this 2 variables back.

EPashkin:
Seems you ignore that
```
guint julian : 1;
guint dmy : 1; 
```
stored in one byte and all struct size can be lesser that guint size (not *2 as in your case).
Rust IMHO don't support this syntax,
this why we "ignore" stucts with these fields.
good that seems none of these structs used in unions,
so you can fallback to old code for this case

Also still not solved problem with including `GVariantBuilder_u1` and `GVariantBuilder_s1` without next usage to all your sys crates.

Luke-Nukem:
```
guint julian : 1;
guint dmy : 1; 
```
Hmm, I thought that was initialising those to the integer number 'n'...

>Also still not solved problem with including GVariantBuilder_u1 and GVariantBuilder_s1 without next usage to all your sys crates.

I don't understand what you mean... The result is;
```
#[repr(C)]
pub struct GVariantBuilder {
    pub u1: GVariantBuilder_u1,
}
 
#[repr(C)]
pub union GVariantBuilder_u1 {
    pub s1: GVariantBuilder_s1,
    pub x: [size_t; 16],
}
 
#[repr(C)]
pub struct GVariantBuilder_s1 {
    pub partial_magic: size_t,
    pub type_: *const GVariantType,
    pub y: [size_t; 14],
}
```
which to me seems fine. You would need to write the API on top of this to accommodate the nesting... 

I think the whole issue comes from in the gir;
```
   <record (trimmed info)>
      <union name="u" c:type="u">
        <record name="s" c:type="s">
```
I can't seem to get those names out, and if I did and the C_Type ended up being name 'u' or 's' then this would cause issues. I can make a small change to not add a number to those two fields, but then I'm not sure it that will cause issues later either.

EPashkin:
Result is fine. Problem that it added to all sys crates not only on glib (7 unused copies for each these union and inner struct).

EPashkin:
Same with `GDoubleIEEE754_s1`, `GFloatIEEE754_s1`.
Problem at seems that `Library::record` and `Library::union` add type to MAIN_NAMESPACE

Luke-Nukem:
Ah I see.. Yes I was thinking about adding special cases, but then I tried to make it general.

```
    #[cfg(feature = "use_unions")]
    pub fn union(library: &mut Library, u: Union) -> TypeId {
        let fields = u.fields;
        let field_tids: Vec<TypeId> = fields.iter().map(|f| f.typ).collect();
        let typ = Type::Union(Union {
            name            : u.name,
            c_type          : u.c_type,
            fields          : fields,
            functions       : u.functions,
            doc             : u.doc,
        });
        library.add_type(MAIN_NAMESPACE, &format!("#{:?}", field_tids), typ)
    }
```
Is it that do you think? 

Regarding the bit field structs, I'll see if I can use [bitflags](https://github.com/rust-lang-nursery/bitflags) crate.

EPashkin:
Yes, it one of functions, that cause problem (I thought that it in Library). It need `ns_id: u16` parameter.
Bitflags IMHO don't help with this and better just use "old" code with commenting.

Luke-Nukem:
@EPashkin are you sure about the bitflags? Mozilla is using it with bindgen for their servo stuff, and the results seem to wok well.

EPashkin:
You can try ;)

EPashkin:
Strange, but duplicates still present :(

src/parser.rs
```diff
@@ -302,18 +319,32 @@ impl Library {
...
+        if let Some(typ) = try!(self.read_record(parser, ns_id, attrs)) {
+            let name = typ.get_name().clone();
+            self.add_type(ns_id, &name, typ);
+        }
+        Ok(())
+    }
+
+    fn read_record(&mut self, parser: &mut Reader, ns_id: u16, attrs: &Attributes)
+        -> Result<Option<Type>> {
+        let name = try!(
```
EPashkin: let mut name

Luke-Nukem: Doh!

Luke-Nukem:
I'm not seeing any duplicates here 😕

EPashkin:
I meant that (for example) both atk-sys/src/lib.rs and glib-sys/src/lib.rs
have `GVariantBuilder_u1`, and it don't used in `atk-sys` at all.

EPashkin:
Sorry, something bugged on my machine: duplicate actually removed.
Only last problem with bitsized structs

EPashkin:
Minor bug in result formating:
https://github.com/gtk-rs/gir/pull/402/files#diff-0c37b0c87c3c91119e579feee853b5d0R368 only needed if feature disabled.

Luke-Nukem:
Okay, that should be the final issue. Except the result formatting? I can't see what you are referring too.

EPashkin:
Seems github generate broken links.
I meant last write in `codegen/sys/lib_::generate_unions`
```rust
    if !items.is_empty() {
        try!(writeln!(w, ""));
    }

```

EPashkin:
Not produce strange results for GIOChannel, GScannerConfig,  gobject-sys::GClosure. gtk-sys::GtkTextAttributes.
Before all after bit-sized fields was commented too.
```rust
#[repr(C)]
pub struct GIOChannel {
    pub ref_count: c_int,
    pub funcs: *mut GIOFuncs,
    pub encoding: *mut c_char,
    pub read_cd: GIConv,
    pub write_cd: GIConv,
    pub line_term: *mut c_char,
    pub line_term_len: c_uint,
    pub buf_size: size_t,
    pub read_buf: *mut GString,
    pub encoded_read_buf: *mut GString,
    pub write_buf: *mut GString,
    pub partial_write_buf: [c_char; 6],
    //use_buffer: guint: 1,
    //do_encode: guint: 1,
    //close_on_unref: guint: 1,
    //is_readable: guint: 1,
    //is_writeable: guint: 1,
    //is_seekable: guint: 1,
    pub reserved1: gpointer,
    pub reserved2: gpointer,
}
```

EPashkin:
Hm, Seems https://github.com/rust-lang/rust/pull/42068 already stabilized part of unions and it in already beta.
Glib has only problem with bit-sized `GDoubleIEEE754_s1` and `GFloatIEEE754_s1` as it non-Copy.

EPashkin:
Good thing that I don't see usage `GFloatIEEE754` and `GDoubleIEEE754` in any functions and we can try ignore its after next release.

Luke-Nukem:
We're almost there 😄 

I see why the proper truncation is required for fields after bitflags. Just needed to allow either is_bits or !truncated.

EPashkin:
Thanks, @Luke-Nukem, Now I don't see any potential problems in generated code.
Can you remove empty line after `generate_unions` declaration?
And also please squash this PR to one commit as there too many merges in it.

src/parser.rs
Luke-Nukem: Ah, my rustfmt is outdated. Are you running everything through rustfmt?

EPashkin: Not on regular basic, but sometimes.

EPashkin: Mostly because latest tend to split function declaration to multiple lines.

Luke-Nukem:
I think I did the squash correctly? It's commit #f56da73

EPashkin:
Yes, #f56da73 is good, but next is not (it merge again).
Please, reset your master branch to this commit, and force push it to your repo.

Luke-Nukem:
Got it. Figured if there was an issue it would be an easy solve. Never done a squash before, very handy.

Luke-Nukem:
Hold fire a few minutes, I'll clean up the commit log

EPashkin:
Actually you added to squashed commit some changes from current master (like adding external_libraries.rs and using it) but git is smart and will ignore it on merge.
Or you can do `git rebase 9e3f4ccff85d1e` to be sure that it contains only your changes.

Luke-Nukem:
Okay, I think we're good now?

EPashkin:
Previous was good (based on 9e3f4ccff85d1e - current upstream master)

EPashkin:
Just to `git rebase 9e3f4ccff85d1e` again

EPashkin:
Yes, now it good.
Thanks again for unions and readme clarification.
I wait for CI just it case and merge

EPashkin:
And excuse me for a lot of nitpicking 😉

Luke-Nukem:
Nitpicking is good. You caught a few things I just didn't think about (or know), or had missed. And it's good practice for me 😉 

I'm looking forward to union stabilization in Rust. And when I get more time (or need it) I'll work on the bit-fields in multitype-structs.

PR #404, opened 2017-07-07
Allow to have a better version prediction for functions

GuillaumeGomez:
And as promised, here it is! However the code isn't really nice so if you have any ideas of improvement, I'll gladly update.

### DIFF
--- a/src/update_version.rs
+++ b/src/update_version.rs
@@ -1,31 +1,32 @@
-use library::{self, Function, Parameter, Type};
+use library::{self, Function, Parameter, Type, MAIN_NAMESPACE};
 use Library;
 use version::Version;
 
 pub fn check_function_real_version(library: &mut Library) {
     // In order to avoid the borrow checker to annoy us...
     let library2 = library as *const Library;
-    for namespace in &mut library.namespaces {
-        for typ in &mut namespace.types {
-            match *typ {
-                Some(Type::Class(ref mut c)) => update_function_version(&mut c.functions, library2),
-                Some(Type::Interface(ref mut i)) => update_function_version(&mut i.functions,
-                                                                            library2),
-                Some(Type::Union(ref mut u)) => update_function_version(&mut u.functions, library2),
-                Some(Type::Record(ref mut r)) => update_function_version(&mut r.functions,
-                                                                         library2),
-                Some(Type::Bitfield(ref mut b)) => update_function_version(&mut b.functions,
-                                                                           library2),
-                Some(Type::Enumeration(ref mut e)) => update_function_version(&mut e.functions,
-                                                                              library2),
-                _ => {}
-            }
+    for typ in &mut library.namespace_mut(MAIN_NAMESPACE).types {
+        match *typ {
+            Some(Type::Class(ref mut c)) => update_function_version(&mut c.functions, library2),
+            Some(Type::Interface(ref mut i)) => update_function_version(&mut i.functions,
+                                                                        library2),
+            Some(Type::Union(ref mut u)) => update_function_version(&mut u.functions, library2),
+            Some(Type::Record(ref mut r)) => update_function_version(&mut r.functions,
+                                                                     library2),
+            Some(Type::Bitfield(ref mut b)) => update_function_version(&mut b.functions,
+                                                                       library2),
+            Some(Type::Enumeration(ref mut e)) => update_function_version(&mut e.functions,
+                                                                          library2),
+            _ => {}
         }
-        update_function_version(&mut namespace.functions, library2);
     }
+    update_function_version(&mut library.namespace_mut(MAIN_NAMESPACE).functions, library2);
 }
 
 fn check_versions(param: &Parameter, current_version: &mut Option<Version>, lib: *const Library) {
+    if param.typ.ns_id != MAIN_NAMESPACE {
+        return;
+    }
     let ty_version = match *unsafe { (*lib).type_(param.typ) } {
         library::Type::Class(ref c) => c.version,
         library::Type::Enumeration(ref c) => c.version,
--- a/src/update_version.rs
+++ b/src/update_version.rs
@@ -1,0 +1,68 @@
+use library::{self, Function, Parameter, Type};
+use Library;
+use version::Version;
+
+pub fn check_function_real_version(library: &mut Library) {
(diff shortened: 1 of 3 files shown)


EPashkin:
PR successfully found version increase `gdk_pixbuf_loader_write_bytes` and many increases in gio-sys.
All of these now same as object where it belong, Ex. g_subprocess_communicate_async, g_test_dbus_add_service_dir.
Not added any new versions to normal crates and detect only `g_notification_set_priority` from version overrides in your `Gir.toml` files.

I have worry how this works with crates with nonuniform versions,
maybe this update need be configured and disabled by default.

src/update_version.rs
```diff
@@ -0,0 +1,61 @@
...
+            (Some(cur_ver), Some(lib_ver)) => cur_ver > lib_ver,
+            (Some(_), None) => true,
+            _ => false,
+        } {
+            function.version = current_version;
+        }
+    }
+}
+
+pub fn check_function_real_version(library: &mut Library) {
```
EPashkin: Please place this public function first

src/update_version.rs
```diff
@@ -0,0 +1,61 @@
...
+                _ => None,
+            } {
+                current_version = Some(new_version);
+            }
+        }
+        if match (current_version, function.version) {
+            (Some(cur_ver), Some(lib_ver)) => cur_ver > lib_ver,
+            (Some(_), None) => true,
+            _ => false,
+        } {
```
EPashkin: Maybe better move match to variable

src/update_version.rs
```diff
@@ -0,0 +1,61 @@
...
+        for parameter in &function.parameters {
+            if let Some(new_version) = match (current_version,
+                                              match *unsafe { (*lib).type_(parameter.typ) } {
+                library::Type::Class(ref c) => c.version,
+                library::Type::Enumeration(ref c) => c.version,
+                library::Type::Bitfield(ref c) => c.version,
+                library::Type::Record(ref c) => c.version,
+                library::Type::Interface(ref c) => c.version,
+                _ => None,
+            }) {
```
EPashkin: This also look tough to understand and better splitted

src/update_version.rs
```diff
@@ -0,0 +1,61 @@
+use library::{self, Function, Type};
+use Library;
+
+fn update_function_version(functions: &mut Vec<Function>, lib: *const Library) {
+    for function in functions {
+        let mut current_version = None;
+        for parameter in &function.parameters {
+            if let Some(new_version) = match (current_version,
+                                              match *unsafe { (*lib).type_(parameter.typ) } {
```
EPashkin: IMHO this `lib` convert better do one time before `for`

EPashkin: Or even do all hacks in `check_function_real_version` and pass normal reference here

EPashkin:
Maybe checking return type also needed (Ex. g_test_dbus_new just luckily detect right version).
Don't see what other can be "improved"

GuillaumeGomez:
Oh right! The return types! Completely forgot about them. Good catch!

src/update_version.rs
GuillaumeGomez: The result is a bit ugly by doing this way. I think it's preferable to keep it it as is.

GuillaumeGomez:
Updated.

GuillaumeGomez:
> I have worry how this works with crates with nonuniform versions,
> maybe this update need be configured and disabled by default.

Do you have an example of nonuniform versions?

EPashkin:
We have https://github.com/gtk-rs/gdk-pixbuf/blob/master/Cargo.toml
that differs from other crates https://github.com/gtk-rs/gdk/blob/master/Cargo.toml#L27.
Cairo and Pango too.

GuillaumeGomez:
We can add a "double-version-check = none" for some functions if you want?

EPashkin:
IMHO better do it on crate level.

EPashkin:
I mean user can disable all system if it starting detecting wrong.

EPashkin:
Or we can limit checking to only main namespace

GuillaumeGomez:
> Or we can limit checking to only main namespace

I think that's the better solution.

GuillaumeGomez:
Limited to main namespace now.

src/update_version.rs
```diff
@@ -0,0 +1,66 @@
...
+    if let Some(new_version) = new_version {
+        *current_version = Some(new_version);
+    }
+}
+
+fn update_function_version(functions: &mut Vec<Function>, lib: *const Library) {
+    for function in functions {
+        let mut current_version = None;
+        for parameter in &function.parameters {
+            check_versions(parameter, &mut current_version, lib);
```
EPashkin: Before this better add check that `parameter.type.ns_id` is same as for function owner,
or nonuniform versions strikes ;)

GuillaumeGomez: Oh right!

EPashkin: Or better do it in `check_version`

GuillaumeGomez: I did it in check_version. Better. ;)

GuillaumeGomez:
Updated!

EPashkin:
Good, now false gdk_pixbuf_loader_write_bytes limit to "v2_32" is gone.
It was wrong gdk as starting with 3.14 using gdk-pixbuf/v2_30, and no use next version.

EPashkin:
@GuillaumeGomez Thanks.
Your prefer me merge as it or will do some squashing?

GuillaumeGomez:
I have no problem with squash. What commits do you want to keep?

EPashkin:
IMHO only last better separated.

GuillaumeGomez:
Squashed! :)

EPashkin:
Travis passed previous Appveyor too, so I merge this.

GuillaumeGomez:
Yeay! Regen time! :D

EPashkin:
Only `sys` actually need regen, maybe plus remove `version` for `Gio.Notification` in gio.

sdroege:
Looks good to me too, even if a bit late :)

EPashkin:
Late is better that never 😉

GuillaumeGomez:
Wise @EPashkin has spoken!

PR #405, opened 2017-07-10
Add configuration for overriding the RefMode

sdroege:
This is mostly useful for manual types where the actually implemented
RefMode is different from what GIR detects automatically.

In my case, I have GstCaps which is a refcounted type. It is basically a subclass of GstMiniObject (not a GObject), and gst_mini_object_ref/unref has to be used on it. There is no gst_caps_ref/unref, so GIR does not know about this and thinks that it's a RecordType::Direct and then uses to_glib_none_mut() when passing out pointers. However this makes no sense for the way how it's implemented.

With the configuration here, I can override the RefMode to ByRef, and everything works fine.


src/analysis/child_properties.rs
```diff
@@ -106,8 +106,8 @@ fn analyze_property(
         if conversion != properties::PropertyConversion::Direct {
             imports.add("std::mem::transmute", None);
         }
-        let get_out_ref_mode = RefMode::of(&env.library, typ, library::ParameterDirection::Return);
-        let mut set_in_ref_mode = RefMode::of(&env.library, typ, library::ParameterDirection::In);
+        let get_out_ref_mode = RefMode::of(&env, typ, library::ParameterDirection::Return);
+        let mut set_in_ref_mode = RefMode::of(&env, typ, library::ParameterDirection::In);
```
EPashkin: &env -> env.
On all uses `RefMode::of` too.

src/analysis/parameter.rs
```diff
@@ -67,7 +67,7 @@ pub fn analyze(
         .matched_parameters(&name)
         .iter()
         .any(|p| p.constant);
-    let ref_mode = RefMode::without_unneeded_mut(&env.library, par, immutable);
+    let ref_mode = RefMode::without_unneeded_mut(&env, par, immutable);
```
EPashkin: just `env` without ampersand too.

src/config/gobjects.rs
```diff
@@ -143,8 +146,25 @@ fn parse_object(toml_object: &Value, concurrency: library::Concurrency) -> GObje
...
+        .lookup("ref_mode")
+        .and_then(|v| v.as_str())
+        .and_then(|v| match v {
+            "none" => Some(ref_mode::RefMode::None),
+            "ref" => Some(ref_mode::RefMode::ByRef),
+            "ref-mut" => Some(ref_mode::RefMode::ByRefMut),
+            "ref-immut" => Some(ref_mode::RefMode::ByRefImmut),
+            "ref-fake" => Some(ref_mode::RefMode::ByRefFake),
+            _ => None,
+        });
```
EPashkin: Please move this match out this function. Just private function in this module IMHO will do as `analysis/ref_mode.rs` use this module.

EPashkin:
Also IMHO mostly `to_glib_none_mut` go from non-const in ffi definition and can be fixed with "const=true" on this parameter. See https://github.com/gtk-rs/gtk/blob/master/Gir.toml#L312-L314.

EPashkin:
Maybe this PR will helps removing need of adding many "const=true".

EPashkin:
And please update readme.md too (maybe will conflict with #406)

sdroege:
All done, thanks for the review. I'm not sure this is the right approach, but all other alternatives for doing what I need here seem equally bad :)

EPashkin:
Thanks, I merge this after CI passed.

PR #407, opened 2017-07-12
Add SetValue/SetValueOptional bounds to property setters and import g…

sdroege:
…object_ffi

This is needed now as glib_wrapper!() implements the Value traits
instead of having generic implementations.

See https://github.com/gtk-rs/glib/pull/196


EPashkin:
Thanks, @sdroege

PR #408, opened 2017-07-14
Add various imports needed by new trait impls and distinguish FromGli…

sdroege:
…bContainer and FromGlibPtrContainer

See https://github.com/gtk-rs/glib/pull/198


EPashkin:
Thanks, @sdroege

PR #410, opened 2017-07-18
Pass get_type() functions, if any, to glib_wrapper!() for shared/boxe…

sdroege:
…d types

See https://github.com/gtk-rs/glib/pull/199

### DIFF
--- a/src/codegen/general.rs
+++ b/src/codegen/general.rs
@@ -105,6 +105,7 @@ pub fn define_boxed_type(
     glib_name: &str,
     copy_fn: &str,
     free_fn: &str,
+    get_type_fn: &Option<String>,
 ) -> Result<()> {
     try!(writeln!(w, ""));
     try!(writeln!(w, "glib_wrapper! {{"));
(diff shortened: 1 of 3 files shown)


src/codegen/general.rs
```diff
@@ -143,6 +148,9 @@ pub fn define_shared_type(
     try!(writeln!(w, "\tmatch fn {{"));
     try!(writeln!(w, "\t\tref => |ptr| ffi::{}(ptr),", ref_fn));
     try!(writeln!(w, "\t\tunref => |ptr| ffi::{}(ptr),", unref_fn));
+    if let &Some(ref get_type_fn) = get_type_fn {
```
EPashkin: Maybe better `Some() = *get_type_fn()`
And in `codegen\translate_from_glib.rs:34:25 if let &Some(_) = array_length` please too .

EPashkin:
Other that minor nit all looks good and only rebase needed.

src/codegen/general.rs
sdroege: Does it make any actual difference, or is it just style?

EPashkin: Just `clippy` warning

sdroege:
Should also be good now

EPashkin:
Thanks, @sdroege

PR #413, opened 2017-07-20
Fix generation of anonymous unions and structs-in-unions

sdroege:
See the individual commits, also cc @Luke-Nukem

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -230,6 +230,7 @@ impl Library {
         let mut impls = Vec::new();
         let mut fields = Vec::new();
         let mut doc = None;
+        let mut union_count = 1;
         loop {
             let event = try!(parser.next());
             match event {
@@ -269,15 +270,43 @@ impl Library {
                         "virtual-method" => try!(ignore_element(parser)),
                         "doc" => doc = try!(read_text(parser)),
                         "union" => {
-                            use self::Type::*;
-                            if let Union(u) = try!(self.read_union(parser, ns_id, &attributes)) {
+                            if let Type::Union(mut u) = try!(self.read_union(
+                                parser,
+                                ns_id,
+                                &attributes,
+                                Some(&class_name),
+                                Some(&c_type)
+                            )) {
+                                let field_name = u.name.clone();
+
+                                u = Union {
+                                    name: format!("{}_u{}", class_name, union_count),
+                                    c_type: Some(format!("{}_u{}", c_type, union_count)),
+                                    ..u
+                                };
+
                                 let u_doc = u.doc.clone();
-                                let type_id = Type::union(self, u, ns_id);
+                                let ctype = u.c_type.clone();
+
+                                let type_id = {
+                                    #[cfg(not(feature = "use_unions"))]
+                                    {
+                                        Type::union(self, u, INTERNAL_NAMESPACE)
+                                    }
+                                    #[cfg(feature = "use_unions")]
+                                    {
+                                        Type::union(self, u, ns_id)
+                                    }
+                                };
+
                                 fields.push(Field {
+                                    name: field_name,
                                     typ: type_id,
                                     doc: u_doc,
+                                    c_type: ctype,
                                     ..Field::default()
                                 });
+                                union_count += 1;
                             }
                         }
                         x => bail!(mk_error!(format!("Unexpected element <{}>", x), parser)),
@@ -316,7 +345,7 @@ impl Library {
         attrs: &Attributes,
     ) -> Result<()> {
 
-        if let Some(typ) = try!(self.read_record(parser, ns_id, attrs)) {
+        if let Some(typ) = try!(self.read_record(parser, ns_id, attrs, None, None)) {
             let name = typ.get_name().clone();
             self.add_type(ns_id, &name, typ);
         }
@@ -328,6 +357,8 @@ impl Library {
         parser: &mut Reader,
         ns_id: u16,
         attrs: &Attributes,
+        parent_name_prefix: Option<&str>,
+        parent_ctype_prefix: Option<&str>,
     ) -> Result<Option<Type>> {
         let mut record_name = try!(
             attrs
@@ -371,67 +402,61 @@ impl Library {
                             ))
                         }
                         "union" => {
-                            if let Type::Union(mut u) =
-                                try!(self.read_union(parser, ns_id, &attributes))
-                            {
-                                // A nested union->struct->union typically has no c_type
-                                // so we iterate over fields to find it. These fields are
-                                // within the nested union->struct if found
-                                let mut nested = true;
-                                for f in &mut u.fields {
-                                    if f.c_type.is_none() || c_type == u.c_type.as_ref().unwrap() {
-                                        nested = true;
-                                        break;
-                                    }
-                                }
-                                if nested {
-                                    u = Union {
-                                        name: format!("{}_u{}", c_type, union_count),
-                                        c_type: Some(format!("{}_u{}", c_type, union_count)),
-                                        fields: u.fields,
-                                        ..Default::default()
-                                    };
-                                }
+                            if let Type::Union(mut u) = try!(self.read_union(
+                                parser,
+                                ns_id,
+                                &attributes,
+                                Some(&record_name),
+                                Some(&c_type)
+                            )) {
+                                let field_name = u.name.clone();
+
+                                u = Union {
+                                    name: format!(
+                                        "{}{}_u{}",
+                                        parent_name_prefix.unwrap_or(""),
+                                        record_name,
+                                        union_count
+                                    ),
+                                    c_type: Some(format!(
+                                        "{}{}_u{}",
+                                        parent_ctype_prefix.unwrap_or(""),
+                                        c_type,
+                                        union_count
+                                    )),
+                                    ..u
+                                };
+
                                 let u_doc = u.doc.clone();
-
-                                #[cfg(not(feature = "use_unions"))]
-                                {
-                                    if nested {
-                                        u.c_type = None;
-                                        let type_id = Type::union(self, u, INTERNAL_NAMESPACE);
-                                        fields.push(Field {
-                                            typ: type_id,
-                                            doc: u_doc,
-                                            ..Field::default()
-                                        });
-                                        continue;
-                                    }
-                                }
-
                                 let ctype = u.c_type.clone();
 
-                                let type_id = Type::union(self, u, ns_id);
-                                if nested {
-                                    fields.push(Field {
-                                        name: format!("u{}", union_count),
-                                        typ: type_id,
-                                        doc: u_doc,
-                                        c_type: ctype,
-                                        ..Field::default()
-                                    });
-                                } else {
-                                    fields.push(Field {
-                                        typ: type_id,
-                                        doc: u_doc,
-                                        c_type: ctype,
-                                        ..Field::default()
-                                    });
-                                }
+                                let type_id = {
+                                    #[cfg(not(feature = "use_unions"))]
+                                    {
+                                        Type::union(self, u, INTERNAL_NAMESPACE)
+                                    }
+                                    #[cfg(feature = "use_unions")]
+                                    {
+                                        Type::union(self, u, ns_id)
+                                    }
+                                };
+
+                                fields.push(Field {
+                                    name: field_name,
+                                    typ: type_id,
+                                    doc: u_doc,
+                                    c_type: ctype,
+                                    ..Field::default()
+                                });
                                 union_count += 1;
                             }
                         }
                         "field" => {
-                            if let Ok(f) = self.read_field(parser, ns_id, &attributes) {
+                            if let Ok(mut f) = self.read_field(parser, ns_id, &attributes) {
+                                // Workaround for wrong GValue c:type
+                                if c_type == "Value" {
+                                    f.c_type = Some("GValue_u1".into());
+                                }
                                 fields.push(f);
                             }
                         }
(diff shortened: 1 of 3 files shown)


sdroege:
This has no effect on any of the sys crates here btw

Luke-Nukem:
I see that Gtk3 gir fails.

In parser.rs, line 366
```rust
    fn read_record(
        &mut self,
        parser: &mut Reader,
        ns_id: u16,
        attrs: &Attributes,
        parent_name_prefix: Option<&str>,
        parent_ctype_prefix: Option<&str>,
    ) -> Result<Option<Type>> {
        let mut record_name = try!(
            attrs
                .by_name("name")
                .ok_or_else(|| mk_error!("Missing union name", parser))
        );
```
Should be `.ok_or_else(|| mk_error!("Missing record name", parser))`. This is probably down to me I think.

I also see that the 2nd level Union (struct->union->struct) in such as GVariantBuilder doesn't label the Struct field with 's'; In gir file it is

```
    <record name="VariantBuilder"
            c:type="GVariantBuilder"
            glib:type-name="GVariantBuilder"
            glib:get-type="g_variant_builder_get_type"
            c:symbol-prefix="variant_builder">
      <doc xml:space="preserve">.</doc>
      <union name="u" c:type="u">
        <record name="s" c:type="s">
```

It currently ends up as
```
pub struct GVariantBuilder {
    pub u: GVariantBuilder_u1,
}
pub union GVariantBuilder_u1 {
    pub u: GVariantBuilderu_s1,
    pub x: [size_t; 16],
}
pub struct GVariantBuilderu_s1 {
    pub partial_magic: size_t,
    pub type_: *const GVariantType,
    pub y: [size_t; 14],
}
```

so `pub u: GVariantBuilderu_s1` is the incorrect field name and struct name.

src/parser.rs
```diff
@@ -537,34 +574,58 @@ impl Library {
...
+                                ns_id,
+                                attrs,
+                                parent_name_prefix,
+                                parent_ctype_prefix
+                            )) {
+                                Some(Type::Record(r)) => r,
+                                _ => continue,
+                            };
+
+                            let field_name = r.name.clone();
```
Luke-Nukem:
```rust
                            let mut field_name = String::new();
                            let mut type_name = String::new();
                            for attr in &attributes {
                                match attr.name.local_name.as_ref() {
                                    "name" => field_name = attr.value.clone(),
                                    "type" => type_name = attr.value.clone(),
                                    _ => {},
                                }
                            }
```
This appears to do the job nicely. Can probably do without the `type_name` though, I don't know where that would apply tbh.

Can now also adjust;
```rust
                                u = Union {
                                    name: format!(
                                        "{}{}_{}{}",
                                        parent_name_prefix.unwrap_or(""),
                                        record_name,
                                        field_name,
                                        union_count
                                    ),
                                    c_type: Some(format!(
                                        "{}{}_{}{}",
                                        parent_ctype_prefix.unwrap_or(""),
                                        c_type,
                                        field_name,
                                        union_count
                                    )),
                                    ..u
                                };
```

Luke-Nukem:
@sdroege I submitted a PR to your `unions_fixup` branch

src/parser.rs
```diff
@@ -37,11 +37,11 @@ macro_rules! xml_next {
...
     pub fn read_file(&mut self, dir: &Path, lib: &str) -> Result<()> {
-        let name = make_file_name(dir, lib);
-        let display_name = name.display();
-        let file = try!(
-            File::open(&name).chain_err(|| format!("Can't read file {}", name.to_string_lossy()))
-        );
+        let file_name = make_file_name(dir, lib);
+        let display_name = file_name.display();
+        let file = try!(File::open(&file_name).chain_err(|| {
+            format!("Can't read file {}", file_name.to_string_lossy())
```
EPashkin: Any reason don't use `display_name` here, other that was it before?

src/parser.rs
```diff
@@ -1227,7 +1284,7 @@ impl Library {
         ns_id: u16,
         attrs: &Attributes,
     ) -> Result<Signal> {
-        let name = try!(
+        let sig_name = try!(
```
EPashkin: Why not `signal_name` ? This IMHO too short

src/parser.rs
```diff
@@ -1568,7 +1625,7 @@ impl Library {
         attrs: &Attributes,
     ) -> Result<(TypeId, Option<String>, Option<u32>)> {
         let start_pos = parser.position();
-        let name = try!(
+        let typ_name = try!(
```
EPashkin: `type_name` IMHO be better

EPashkin:
Just added minor comment and I confirm that gir now can't generate `gtk` with
```
Error in file D:\eap\rust\0\gtk\gir-files\Gtk-3.0.gir
caused by: 106446:7 parser.rs:547 Missing union name
```
other packages seems ok.

sdroege:
Should be all good now then, thanks @Luke-Nukem :)

EPashkin:
Still on gtk generation in both modes
```
Error in file gir-files\Gtk-3.0.gir
caused by: 106446:7 src\parser.rs:555 Missing union name
```

EPashkin:
Strange change for inner struct names in glib_sys:
```diff
 #[repr(C)]
 pub union GDoubleIEEE754 {
     pub v_double: c_double,
-    pub s1: GDoubleIEEE754_s1,
+    pub mpn: _mpn1,
 }
 ```
IMHO it now not unique enough

**Comment by Luke-Nukem** [comment]
@EPashkin I'll see if I can sort that out. Just need to adjust how it gets the attributes and names.

**Comment by EPashkin** [comment]
@Luke-Nukem Don't need, currently naming not unique: two `pub struct _mpn1 {` in glib-sts

**Comment by Luke-Nukem** [comment]
@EPashkin Yeah saw that, fixed 😃.

I'll submit PR to @sdroege so it get's in. [Commit](https://github.com/Luke-Nukem/gir/commit/fe5f3ea32c3bd414bc84d7d680a624a5e541f8c4)

**Comment by sdroege** [comment]
Thanks, will review and merge in a bit. And then fix gtk, that one is easy... And needs a bug report against gtk, I think

**Comment by Luke-Nukem** [comment]
Cool. Once done I'll sync and merge for next PR to remove the feature-gating of Unions; this removes a nice chunk of special-case code.

**Comment by sdroege** [comment]
Let's hope this is all now

**Comment by EPashkin** [comment]
Strange name `GVariantBuilderu_s1` maybe better name it `GVariantBuilder_u_s` with removing counting?
IMHO this will be unique.
```diff
 #[repr(C)]
 pub struct GVariantBuilder {
-    _truncated_record_marker: c_void,
-    //union,
+    pub u: GVariantBuilder_u1,
+}
+
+#[repr(C)]
+pub union GVariantBuilder_u1 {
+    pub s: GVariantBuilderu_s1,
+    pub x: [size_t; 16],
+}
+
+#[repr(C)]
+pub struct GVariantBuilderu_s1 {
+    pub partial_magic: size_t,
+    pub type_: *const GVariantType,
+    pub y: [size_t; 14],
 }
```

sdroege:
For this case yes, for the generic case not. We would need to post-process the fields/types afterwards then, which would make the code quite more complicated.

EPashkin:
Strange, I though that field names on same level unique, so only possible duplicate is struct\union with same name in gir file (IMHO It is almost impossible).
At minimum all compiles now.
I wait for @Luke-Nukem and merge

sdroege:
Ah now I understand what you meant. Yes, it should be possible to omit the counter iff there is a field name. Let's try that

sdroege:
For @Luke-Nukem , he's gone for the night now but wrote "<luke_nukem[m]> slomo_: cool. I dont see any issues with latest PR to you so far. Except the gtk issue" on IRC earlier.

sdroege:
And if we prettify things, I also changed this

```Rust
pub union GVariantBuilder_u {
    pub s: GVariantBuilderu_s,
    pub x: [size_t; 16],
}
```

to
```Rust
pub union GVariantBuilder_u {
    pub s: GVariantBuilder_u_s,
    pub x: [size_t; 16],
}
```

EPashkin:
Yes, it that I meant, thanks

sdroege:
Done :)

src/parser.rs
```diff
@@ -269,15 +270,56 @@ impl Library {
...
+
+                                let type_id = {
+                                    #[cfg(not(feature = "use_unions"))]
+                                    {
+                                        Type::union(self, u, INTERNAL_NAMESPACE)
+                                    }
+                                    #[cfg(feature = "use_unions")]
+                                    {
+                                        Type::union(self, u, ns_id)
+                                    }
```
EPashkin: Seems this switch not needed: `let type_id = Type::union(self, u, ns_id);` will works always

src/parser.rs
```diff
@@ -269,15 +270,56 @@ impl Library {
...
+                                        "name" => {
+                                            field_name = attr.value.clone();
+                                            break;
+                                        }
+                                        _ => {}
+                                    }
+                                }
+
+                                if field_name == "" {
+                                    field_name = format!("u{}", union_count);
```
EPashkin: `if field_name.is_empty() {` (2 cases)
Bad that we can't remove `union_count` completely.

EPashkin: And if previous for replace by `attributes.by_name("name")` then condition can be changed to `if let`

src/parser.rs
```diff
@@ -541,34 +609,83 @@ impl Library {
...
+
+                            let mut field_name = String::new();
+                            for attr in &attributes {
+                                match attr.name.local_name.as_ref() {
+                                    "name" => {
+                                        field_name = attr.value.clone();
+                                        break;
+                                    }
+                                    _ => {}
+                                }
```
EPashkin: The third case for `by_name` and `if let`

src/parser.rs
sdroege: Done

src/parser.rs
sdroege: Done

src/parser.rs
sdroege: I thought adding the anonymous unions to the internal namespace would make sure they don't end up in the generated code later

EPashkin: Don't remember why it was done in first place.

sdroege: Let's try then

src/parser.rs
```diff
@@ -371,67 +407,78 @@ impl Library {
...
+                                let type_id = {
+                                    #[cfg(not(feature = "use_unions"))]
+                                    {
+                                        Type::union(self, u, INTERNAL_NAMESPACE)
+                                    }
+                                    #[cfg(feature = "use_unions")]
+                                    {
+                                        Type::union(self, u, ns_id)
+                                    }
+                                };
```
EPashkin: This too

sdroege: And the Type::record() one, yes

src/parser.rs
sdroege: If you do that, the following happens on glib in non-union mode:

```diff
@@ -1596,6 +1596,8 @@ pub struct GMutex([u32; 2]);
 #[repr(C)]
 pub struct GMutex(*mut c_void);
 pub type GTokenValue = c_void; // union
+pub type GVariantBuilder_u = c_void; // union
+pub type GVariantDict_u = c_void; // union
 
 // Callbacks
 pub type GChildWatchFunc = Option<unsafe extern "C" fn(GPid, c_int, gpointer)>;
@@ -1702,6 +1704,15 @@ pub struct GDebugKey {
 pub struct GDir(c_void);
 
 #[repr(C)]
+pub struct GDoubleIEEE754_mpn {
+    _truncated_record_marker: c_void,
+    //mantissa_low: guint: 32,
+    //mantissa_high: guint: 20,
+    //biased_exponent: guint: 11,
+    //sign: guint: 1,
+}
+
+#[repr(C)]
 pub struct GError {
     pub domain: GQuark,
     pub code: c_int,
@@ -1709,6 +1720,14 @@ pub struct GError {
 }
 
 #[repr(C)]
+pub struct GFloatIEEE754_mpn {
+    _truncated_record_marker: c_void,
+    //mantissa: guint: 23,
+    //biased_exponent: guint: 8,
+    //sign: guint: 1,
+}
+
+#[repr(C)]
 pub struct GHashTable(c_void);
 
 #[repr(C)]
@@ -2110,12 +2129,26 @@ pub struct GVariantBuilder {
 }
 
 #[repr(C)]
+pub struct GVariantBuilder_u_s {
+    pub partial_magic: size_t,
+    pub type_: *const GVariantType,
+    pub y: [size_t; 14],
+}
+
+#[repr(C)]
 pub struct GVariantDict {
     _truncated_record_marker: c_void,
     //union,
 }
 
 #[repr(C)]
+pub struct GVariantDict_u_s {
+    pub asv: *mut GVariant,
+    pub partial_magic: size_t,
+    pub y: [size_t; 14],
+}
+
+#[repr(C)]
 pub struct GVariantIter {
     pub x: [size_t; 16],
 }
```

src/parser.rs
```diff
@@ -269,15 +270,48 @@ impl Library {
...
+                                ns_id,
+                                &attributes,
+                                Some(&class_name),
+                                Some(&c_type)
+                            )) {
+                                let field_name =
+                                    if let Some(field_name) = attributes.by_name("name") {
+                                        field_name.into()
+                                    } else {
+                                        format!("u{}", union_count)
```
EPashkin: Maybe better move `union_count += 1;` here?

EPashkin: I don't insist as we don't have second unnamed union anyway ;)

src/parser.rs
EPashkin: You right, I mixed up `INTERNAL_NAMESPACE` and `MAIN_NAMESPACE`.
All 3 of these better remain

EPashkin:
Only found suspious `pub type GValue_data = c_void; // union` in gobject sys, but I prefer ignore it as `non union` mode will be removed in short time.
@sdroege you want change anything or I merge?

sdroege:
Go ahead :) That GValue_data one is there because it's a union at the top-level. It was only not there previously because it was missing a name/c:type

EPashkin:
Thanks, @sdroege

PR #414, opened 2017-07-21
Unions remove feature-gate

Luke-Nukem:
Removes all feature-gating of `union`, and now requires rust 1.19
Updates readme to reflect this.


Luke-Nukem:
Hmmm, strange list of commits. Will redo.

PR #415, opened 2017-07-21
Unions: Remove feature-gate for stable rustc

Luke-Nukem:
Removes all feature-gating of `union`, and now requires rust 1.19
Updates readme to reflect this.

### DIFF
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -343,7 +343,7 @@ fn generate_unions(w: &mut Write, env: &Env, items: &[&Union]) -> Result<()> {
             if lines.is_empty() {
                 try!(writeln!(
                     w,
-                    "{0}#[repr(C)]\n{0}#[derive(Copy,Clone)]\n{0}pub union {1}(c_void);\n",
+                    "{0}#[repr(C)]\n{0}pub union {1}(c_void);\n",
                     comment,
                     c_type
                 ));
@@ -631,12 +631,9 @@ fn generate_fields(env: &Env, struct_name: &str, fields: &[Field]) -> (Vec<Strin
             lines.push("\tpub priv_: gpointer,".to_owned());
         } else if let Some(ref c_type) = field.c_type {
             let name = mangle_keywords(&*field.name);
-            let mut c_type = ffi_type(env, field.typ, c_type);
+            let c_type = ffi_type(env, field.typ, c_type);
             if c_type.is_err() {
                 commented = true;
-            }
-            if is_gvalue && field.name == "data" {
-                c_type = Ok("[u64; 2]".to_owned());
             }
             lines.push(format!("\tpub {}: {},", name, c_type.into_string()));
         } else {
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -343,16 +343,16 @@ fn generate_unions(w: &mut Write, env: &Env, items: &[&Union]) -> Result<()> {
             if lines.is_empty() {
                 try!(writeln!(
                     w,
-                    "{comment}#[repr(C)]\n{comment}pub union {name}(c_void);\n",
-                    comment = comment,
-                    name = c_type
+                    "{0}#[repr(C)]\n{0}#[derive(Copy,Clone)]\n{0}pub union {1}(c_void);\n",
+                    comment,
+                    c_type
                 ));
             } else {
                 try!(writeln!(
                     w,
-                    "{comment}#[repr(C)]\n{comment}pub union {name} {{",
-                    comment = comment,
-                    name = c_type
+                    "{0}#[repr(C)]\n{0}#[derive(Copy,Clone)]\n{0}pub union {1} {{",
+                    comment,
+                    c_type
                 ));
 
                 for line in lines {
@@ -498,7 +498,7 @@ fn generate_records(w: &mut Write, env: &Env, records: &[&Record]) -> Result<()>
         if lines.is_empty() {
             try!(writeln!(
                 w,
-                "{}#[repr(C)]\n{0}pub struct {}(c_void);\n",
+                "{0}#[repr(C)]\n{0}#[derive(Copy,Clone)]\n{0}pub struct {1}(u8);\n",
                 comment,
                 record.c_type
             ));
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -317,36 +317,7 @@ fn generate_unions(w: &mut Write, env: &Env, items: &[&Union]) -> Result<()> {
 
     for item in items {
         if let Some(ref c_type) = item.c_type {
-            #[cfg(feature = "use_unions")]
-            {
-                let (lines, commented) = generate_fields(env, &item.name, &item.fields);
-
-                let comment = if commented { "//" } else { "" };
-                if lines.is_empty() {
-                    try!(writeln!(
-                        w,
-                        "{comment}#[repr(C)]\n{comment}pub union {name}(c_void);\n",
-                        comment = comment,
-                        name = c_type
-                    ));
-                } else {
-                    try!(writeln!(
-                        w,
-                        "{comment}#[repr(C)]\n{comment}pub union {name} {{",
-                        comment = comment,
-                        name = c_type
-                    ));
-
-                    for line in lines {
-                        try!(writeln!(w, "{}{}", comment, line));
-                    }
-                    try!(writeln!(w, "{}}}\n", comment));
-                }
-                if comment.is_empty() {
-                    try!(generate_debug_with_fields(w, name, &lines));
-                }
-            }
-            #[cfg(not(feature = "use_unions"))]
+           /* #[cfg(not(feature = "use_unions"))]
             {
                 // TODO: GLib/GObject special cases until we have proper union support in Rust
                 if env.config.library_name == "GLib" && c_type == "GMutex" {
@@ -365,19 +336,35 @@ fn generate_unions(w: &mut Write, env: &Env, items: &[&Union]) -> Result<()> {
                          pub struct {0}(*mut c_void);",
                         c_type
                     ));
-                } else {
-                    try!(writeln!(w, "pub type {} = c_void; // union", c_type));
+                }*/
+            let (lines, commented) = generate_fields(env, &item.name, &item.fields);
+
+            let comment = if commented { "//" } else { "" };
+            if lines.is_empty() {
+                try!(writeln!(
+                    w,
+                    "{comment}#[repr(C)]\n{comment}pub union {name}(c_void);\n",
+                    comment = comment,
+                    name = c_type
+                ));
+            } else {
+                try!(writeln!(
+                    w,
+                    "{comment}#[repr(C)]\n{comment}pub union {name} {{",
+                    comment = comment,
+                    name = c_type
+                ));
+
+                for line in lines {
+                    try!(writeln!(w, "{}{}", comment, line));
                 }
+                try!(writeln!(w, "{}}}\n", comment));
+            }
+            if comment.is_empty() {
+                try!(generate_debug_with_fields(w, name, &lines));
             }
         }
     }
-    #[cfg(not(feature = "use_unions"))]
-    {
-        if !items.is_empty() {
-            try!(writeln!(w, ""));
-        }
-    }
-
     Ok(())
 }
 
@@ -531,7 +518,8 @@ fn generate_records(w: &mut Write, env: &Env, records: &[&Record]) -> Result<()>
                 w,
                 "{}#[repr(C)]\n{}{0}pub struct {} {{",
                 comment,
-                if can_generate_fields_debug(&record.fields) { "#[derive(Debug)]\n" } else { "" },
+                if can_generate_fields_debug(&record.fields) { "#[derive(Copy,Clone,Debug)]\n" }
+                else { "" },
                 record.c_type
             ));
             for line in &lines {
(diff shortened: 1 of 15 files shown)


EPashkin:
Travis build failure is legit

src/codegen/sys/statics.rs
```diff
@@ -3,13 +3,8 @@ use std::io::{Result, Write};
...
 
 pub fn begin(w: &mut Write) -> Result<()> {
-    #[cfg(feature = "use_unions")]
-    let u = "#![feature(untagged_unions)]";
-    #[cfg(not(feature = "use_unions"))]
-    let u = "";
-
     let v = vec![
-        u,
+        "#![feature(untagged_unions)]",
```
EPashkin: This line unneeded just "" instead

EPashkin:
Only last problem with `error: unions with non-`Copy` fields are unstable (see issue #32836)`
I thought that we can just add `#[derive(Clone, Copy)]` to child structs, but in case of `GDoubleIEEE754_mpn` it won't works because `_truncated_record_marker: c_void`.
Maybe we need change its type to `u8`.
Also from gdk's "GdkEvent" seems we need derive Copy for many child structs and it inners (GdkRectangle) etc.,
so IMHO simpler derive Copy for all structs and possibly for unions.
@GuillaumeGomez  you agree with this?
@sdroege As I remember you was for derive Copy from start?

src/codegen/sys/statics.rs
Luke-Nukem: It appears that the tag is needed;

```
error: unions with non-`Copy` fields are unstable (see issue #32836)
    --> src/lib.rs:1631:1
     |
1631 | / pub union GVariantDict_u {
1632 | |     pub s: GVariantDict_u_s,
1633 | |     pub x: [size_t; 16],
1634 | | }
     | |_^
     |
     = help: add #![feature(untagged_unions)] to the crate attributes to enable
```

Luke-Nukem:
Not sure how to go about `#[derive(Clone, Copy)]`... To be honest, I don't think we should be adding this manually as this implies that we are checking each case to be sure all fields can be Copy - which we are not doing unless it is at the parser/output level.

Changing `_truncated_record_marker: c_void` to `u8` would be a good idea I think. I'll work on removing the `#![feature(untagged_unions)]` and implementing `#[derive(Copy, Clone)]` for those particular types. Do you have any suggestions for tracking which ones need it?

EPashkin:
@Luke-Nukem Lets wait for other say their options. For `GdkEvent` need or implement recursive check and update for all children or just derive for all structs.

Luke-Nukem:
@EPashkin no problem. I do know we absolutely shouldn't (and can't) implement `#[derive(Copy, Clone)]` for all structs, it would introduce inefficiencies such as copying large data structures automatically instead of using a move. Looking at the structure of what we use to output the bindings, there doesn't seem to be anything easily used for it - unless perhaps we insert some flag at the parse stage?

EPashkin:
Hm, I don't though that some structures can be large and `Copy` for these unreasonable.
Maybe we can add some config for `Copy` and for expanding enum.
Or really check\add Copy flag all unions children and hope that its will not cause problem.

EPashkin:
last commit produce strange diff:
```diff
// Records
 #[repr(C)]
-pub struct GArray {
+GArraypub struct GArray {
     pub data: *mut c_char,
     pub len: c_uint,
 }
```

Luke-Nukem:
Oh dear, looks like I'm making a mess of git. Hang on, I'll try and fix this up.

Luke-Nukem:
The sys crates I've compiled so far seem to be fine now.

src/codegen/sys/lib_.rs
```diff
@@ -434,7 +434,7 @@ fn generate_records(w: &mut Write, env: &Env, records: &[&Record]) -> Result<()>
             if record.derive_copy {
                 try!(writeln!(w, "{}#[derive(Copy,Clone)]", comment));
             }
-            try!(writeln!(w, "{0}pub struct {} {{", record.c_type));
+            try!(writeln!(w, "pub struct {} {{", record.c_type));
```
EPashkin: This not correct too, you lost `comment` text

Luke-Nukem: 🤦‍ 🤦‍ 🤦‍
Added back, squashed last series of commits. Should be good now.

EPashkin:
Only last questionable problems with `GdkEvent` and `GtkTextAppearance_u1`, for this better wait opinions from others.

EPashkin:
With `GtkTextAppearance_u1` is problem that it used `gdk::GdkRGBA` so we can't add Copy for it automaticly.

EPashkin:
Seems `GtkTextAppearance_u1` actually unused as `GtkTextAppearance` is struct with bit-sized fields and truncated before union.

Luke-Nukem:
> Seems GtkTextAppearance_u1 actually unused as GtkTextAppearance is struct with bit-sized fields and truncated before union.

I guess that's just a side effect of how things are parsed then read out. Not sure what we can do about it other than just leave it. Or add in a special case for it? Do you think that is the only problem area?

sdroege:
Just short comments
- u8 or array thereof for bit fields makes sense. Just count the bits
- I think deriving copy for all ffi structs is probably safe. As in, except for very few exceptions you only get them as pointers anyway. Using a configuration seems cleaner though, but OTOH we need to do that automatically for nested structs anyway. Copy for structs with pointers seems dangerous, but this is all unsafe code anyway so maybe OK. And in C, all structs are "Copy" too. Performance-wide I see no problem here in any case: if your code moved before, it will still do that. You might just copy without noticing, but that's only a problem if you have the whole struct and not just a pointer to it.

So in summary, I don't know. Making all Copy seems fine to me, a configuration also. Both are ugly in their own ways and in practice it probably makes no difference

Luke-Nukem:
@sdroege the one problem with making *all* Copy though, is that there will be some things that shouldn't be or are inherently non-copy. I think `c_void` is one of those so if any structs have that as a field type then we'll have issues. Having said that I know `c_void` is `u8` sized, and don't quite understand the purpose of `c_void`.

Copy on pointers... iirc raw pointers are fine and safe for copy, it's only when we deref them that it's unsafe. (?)

sdroege:
For pointers, correct. You just need to be careful to free all copies later, I.e. do double-frees. But that's already with plain pointers

EPashkin:
@GuillaumeGomez what you think about implementing `Copy` for almost all structs in sys crates to make unions work on stable?

GuillaumeGomez:
That's something I proposed not that long ago so I'm totally for it!

EPashkin:
@Luke-Nukem sorry for delay, I thought that I already asked.

Luke-Nukem:
Sorry about the delay here, I'll push through an update soon. Was a bit tied up with study this week.

Luke-Nukem:
I'm not sure about `pub struct Volatile<T>(T);` being copy,clone... Thoughts?

Luke-Nukem:
```
error[E0277]: the trait bound `[f64; 128]: std::clone::Clone` is not satisfied
    --> /home/travis/build/gtk-rs/gir/tests/sys/gdk-sys/src/lib.rs:3743:5
     |
3743 |     pub axes: [c_double; 128],
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::clone::Clone` is not implemented for `[f64; 128]`
     |
     = help: the following implementations were found:
               <[T; 28] as std::clone::Clone>
               <[T; 14] as std::clone::Clone>
               <[T; 7] as std::clone::Clone>
               <[T; 8] as std::clone::Clone>
             and 29 others
     = note: required by `std::clone::AssertParamIsClone`
```

Suggestions?

EPashkin:
About old appveyour version: added fix #420

EPashkin:
As gtk has single problem too, IMHO better add config parameter for object like "dont_copy=true"
```
error[E0277]: the trait bound `unsafe extern "C" fn(*mut GtkStyle, *mut cairo::cairo_t, GtkStateType, GtkShadowType, *mut GtkWidget, *const i8, i32, i32, i32, i32, GtkPositionType, i32, i32): std::clone::Clone` is not satisfied
    --> d:/eap/rust/0/sys/gtk-sys\src\lib.rs:5295:5
     |
5295 |     pub draw_shadow_gap: Option<unsafe extern "C" fn(*mut GtkStyle, *mut cairo::cairo_t, GtkStateType, GtkShadowType, *mut GtkWidget, *const c_char, c_int, c_int, c_int, c_int, GtkPositionType, c_int, c_int)>,
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `std::clone::Clone` is not implemented for `unsafe extern "C" fn(*mut GtkStyle, *mut cairo::cairo_t, GtkStateType, GtkShadowType, *mut GtkWidget, *const i8, i32, i32, i32, i32, GtkPositionType, i32, i32)`
     |
     = note: required because of the requirements on the impl of `std::clone::Clone` for `std::option::Option<unsafe extern "C" fn(*mut GtkStyle, *mut cairo::cairo_t, GtkStateType, GtkShadowType, *mut GtkWidget, *const i8, i32, i32, i32, i32, GtkPositionType, i32, i32)>`
     = note: required by `std::clone::AssertParamIsClone`
```

src/codegen/sys/lib_.rs
```diff
@@ -574,7 +516,7 @@ fn generate_fields(env: &Env, struct_name: &str, fields: &[Field]) -> (Vec<Strin
             if c_type.is_err() {
                 commented = true;
             }
-            if !cfg!(feature = "use_unions") && is_gvalue && field.name == "data" {
+            if is_gvalue && field.name == "data" {
```
EPashkin: Why this don't removed?

src/codegen/sys/lib_.rs
```diff
@@ -565,7 +507,7 @@ fn generate_fields(env: &Env, struct_name: &str, fields: &[Field]) -> (Vec<Strin
             continue 'fields;
         }
 
-        if is_gweakref && !cfg!(feature = "use_unions") {
+        if is_gweakref {
```
EPashkin: Why this don't removed?

src/codegen/sys/lib_.rs
```diff
@@ -444,7 +417,7 @@ fn generate_records(w: &mut Write, env: &Env, records: &[&Record]) -> Result<()>
         if lines.is_empty() {
             try!(writeln!(
                 w,
-                "{}#[repr(C)]\n{0}pub struct {}(c_void);\n",
+                "{0}#[repr(C)]\n{0}#[derive(Copy,Clone)]\n{0}pub struct {1}(u8);\n",
```
EPashkin: This better changed back (it produce only 1 extra "dont_copy")

EPashkin: ..because it just don't right, copy record with unknown size

EPashkin:
About `Volatile`: as it applied only to simple types it can be copy.

src/codegen/sys/lib_.rs
Luke-Nukem: I'm not sure I understand, `pub struct GIConv(u8);` should not be copy?

src/codegen/sys/lib_.rs
```diff
@@ -318,58 +318,31 @@ fn generate_unions(w: &mut Write, env: &Env, items: &[&Union]) -> Result<()> {
...
-                    ));
-                } else {
-                    try!(writeln!(w, "pub type {} = c_void; // union", c_type));
+            let (lines, commented) = generate_fields(env, &item.name, &item.fields);
+
+            let comment = if commented { "//" } else { "" };
+            if lines.is_empty() {
+                try!(writeln!(
+                    w,
+                    "{0}#[repr(C)]\n{0}#[derive(Copy,Clone)]\n{0}pub union {1}(c_void);\n",
```
EPashkin: `Copy` better removed as for records
PS. this code actually unused.

Luke-Nukem:
> config parameter for object like "dont_copy=true"

@EPashkin can you advise where I would look for examples of how to do this?

src/codegen/sys/lib_.rs
EPashkin: Yes, with `GIOChannel`

EPashkin:
Add field to GObject same as https://github.com/gtk-rs/gir/blob/master/src/config/gobjects.rs#L151
then get it in `generate_record` 
```
let full_name = format!("{}.{}", env.namespaces.main().name, record.name);
let config = env.config.objects.get(&full_name);
```

EPashkin:
.. As `config` can be `None` for ffi, use `if-let` or `map(|o| o.dont_copy).unwrap_or(false)` to get value.

src/codegen/sys/lib_.rs
sdroege: Not sure I understand either. What is the broken code generated by this?

EPashkin: Structs without fields only can be used as reference and its size unknown to gir.
For the same reason, it makes no sense to copy these structures.

sdroege: But that's a struct with a single field, a u8?

EPashkin: This in `if lines.is_empty()` so struct don't have fields at all

sdroege: Ah, so this is just for having *anything* in case of opaque structs. Maybe make those `pub struct Meh(())`? That way they can only used as pointers (which they can only be used as correctly anyway).

EPashkin: this case produce `warning: found non-foreign-function-safe member in struct marked #[repr(C)]: found Rust tuple type in foreign module; consider using a struct instead`
just struct produce `warning: found zero-size struct in foreign module, consider adding a member to this struct`

sdroege: I see... a case for the new #[repr(transparent)] I guess

EPashkin: Sorry, now I don't understand how it related to struct without fields.

sdroege: You would want to represent an opaque C type. It's not really a struct without fields, it's just that the struct definition is unknown from the headers. So you can only ever use it as a pointer.

But that's actually `extern type`: https://github.com/rust-lang/rfcs/pull/1861

EPashkin: `extern type` is that we need for this case but it not realized and may cause build error when inserted in other struct directly (as with `GIConv`)
`#[repr(transparent)]` can be used for minimize alignment errors for known fields, so also not for this case, so last `repr(C)`.
I prefer `c_void` over `u8` as it don't implement `Copy` while has same size and alignment.

Luke-Nukem:
I need to put this on the backlog for a week or so while I do a few other things of importance. I may work on it off/on through the week however, we'll see.

sdroege:
What's the status here?

Luke-Nukem:
I've not had much free time. I'll try and get a look at it again this week,
I should be able to juggle my schedule a bit.

Luke-Nukem:
@sdroege @EPashkin I've done a rebase of this on to master and run a couple of small gir tests - seems okay.

Looking back at previous comments it looks like I still have a few things to do such as "dont_copy=true". Is this still a blocker?

sdroege:
IMHO just having copy on all FFI things is fine, configuration can be added later once needed :)

Luke-Nukem:
Okidoki, then this may be good to go if edge cases are okay.

Luke-Nukem:
Tests will pass with this [commit to cairo](https://github.com/Luke-Nukem/cairo/commit/90beb7cca0567393ec2fc62242ff9673de7594f0)

Issue #264, opened 2016-06-23
Function parameters transformations


EPashkin:
For adding posibility parameters type rewrite I proposed changes in `analysis::functions::Info`:
- add unmodified copy of `Vec<library::Parameter>`
- move returned values to both parameter lists (C and rust's) as parameters with `ParameterDirection::return`
- explitic add returned `error` parameter to rust's list (may be with `ParameterDirection::error`)
- add `Vec<Transformation>`, where
```rust
struct Transformation {
  ind_c: i32,  	  //index in `Vec<library::Parameter>`
  ind_rust: i32,  //index in 'Vec<analysis::parameter::Parameter>`
  conversion: ConversionType, //Used in ffi-function call parameters and return converting
  pre_transformation: PreTransformationType,
  post_transformation: PostTransformationType,
}

enum PreTransformationType {
  None,
  VarUninitialized, //for string like "let mut padding_top = mem::uninitialized();" in function body
  VarUnititializedType, //for "let mut aligned_area = gdk::Rectangle::uninitialized();"
  VarNullPointer, 	//for "let mut path = ptr::null_mut();"  
  VarError, 		//for "let mut error = ptr::null_mut();"
}

enum PostTransformationType {
  None,
  Return, 	//for adding to return Tuple
  ReturnOption,	//on original return parameter for returning Option<(...)>
  ReturnError, 	//"if error.is_null() { Ok(()) } else { Err(from_glib_full(error)) }"
}
```
In future we can add transformation for write one `name: String` parameter to 2 ffi parameters: ptr and size, etc.

gkoz:
I'm away, will review on Tuesday.

gkoz:
This seems plausible but I guess we should collect the list of transformations this has to handle. Perhaps two lists: 
 * just rewriting (`i32` to `u32`/`Option<u32>`, `char*` to `TreePath`),
 *  `&str` to `(char*, int)` and the like.

Handling function calls and callbacks independently might be simpler as their needs are different.

EPashkin:
I see `&str` conversion to `(char*, int)` no different from 2 conversions: `&str` to `char*` (which is default) and `&str` to `int` by `.len()`, so I don't currently see reason having "complex" transformation.
For trampolines I found not listed `char*` to `String` in `cell_renderer_text::trampoline`,
and too specific like in `spin_button::input_trampoline` and `editable::insert_trampoline`
(that better do with some "custom" transformation fully defined in config)

[...]

PR #416, opened 2017-07-23
Function parameters transformation with array length support

EPashkin:
Part of https://github.com/gtk-rs/gir/issues/264

TODO:
- [x] Configurable length for string and arrays
- [x] Into transformation


EPashkin:
cc @GuillaumeGomez, @sdroege 
Current version must produce same result as previous.

src/codegen/function_body_chunk.rs
```diff
@@ -339,7 +318,7 @@ impl Builder {
         } else {
             Chunk::FromGlibConversion {
                 mode: parameter.into(),
-                array_length: parameter.array_length.clone(),
+                array_length_name: self.find_array_length_name(&parameter.name),
```
EPashkin: @sdroege In gtk-rs no functions that uses this (out array with out length), you have functions to test it?

src/analysis/function_parameters.rs
```diff
@@ -0,0 +1,235 @@
...
+pub use config::functions::TransformationType;
+
+//TODO: remove unused fields
+#[derive(Clone, Debug)]
+pub struct RustParameter {
+    pub ind_c: usize, //index in `Vec<CParameter>`
+    pub name: String,
+    pub typ: library::TypeId,
+    pub allow_none: bool,
+}
```
EPashkin: Its very strange, but now only needed in this struct is `ind_c` and I plan to remove others.

sdroege:
What's the overall goal of this PR? Cleanup or does it change/add anything?

I had an example with out parameters somewhere, will check tomorrow. And also go through the commits and test it on my stuff :)

EPashkin:
@sdroege Overall goal is add configuration for length for strings and add basis for future parameter transformations (see issue).
First 2 commits will be just refactoring and must don't change produced code.
Maybe also include `Into` conversion as transformation.

EPashkin:
I done all what planned for this PR.

GuillaumeGomez:
It's a bit sad to be forced to specify that the parameter is a length but I guess it's unavoidable...

sdroege:
@EPashkin Makes sense, I'll review tomorrow. See also https://bugzilla.gnome.org/show_bug.cgi?id=784020 for other cases where we need to override/add an array length annotation

sdroege:
@EPashkin About the configuration, it would seem more consistent to do it the other way around. Add a configuration for the array parameter/return-value that gives the name of the length parameter. This is how it's done in GI, but you currently have configuration for the length parameter to specify to which array parameter/return-value (how would you specify a return value btw? "return"?) they relate to.

@GuillaumeGomez I have the feeling that sooner or later we'll have to have configuration for overriding/adding any annotation that GI supports. That's what exists basically for the C# and Vala bindings.

README.md
```diff
@@ -107,6 +107,8 @@ cfg_condition = "mycond"
         nullable = true
         # allow to make parameter immutable
         const = true
+        # parameter is calculated as length of string or array and removed from function declaration
+        length_of = "str"
```
sdroege: Basically what I mean here is that this should be `array_length = "len"`

src/codegen/function_body_chunk.rs
sdroege: Actually I don't, but you could fake one by editing a .gir file accordingly. The only cases that are there for real currently (or so it seems) are return-value & out-length-parameter and in-parameter & in-length-parameter

sdroege:
Generally looks good to me, cleaner than before in any case. And it doesn't cause any changes in the generated code for my things either.

EPashkin:
Updated.
@sdroege Good that it works for out array with out lengths too.
I try `array_length` but I thought that it will need many finds, so changed direction.
PS Currently for return array must work `length_of = ""`

sdroege:
>  I try array_length but I thought that it will need many finds, so changed direction.

But don't you need that anyway for the GI `array length` annotation to work?

EPashkin:
> But don't you need that anyway for the GI `array length` annotation to work?

I collect internal array lengths in function_parameters::analyze https://github.com/gtk-rs/gir/pull/416/files#diff-9d06bce181173b761420826de5e33e89R144 to dictionary (len positions => array name)
and use it to create transformation for

EPashkin:
.. length parameter https://github.com/gtk-rs/gir/pull/416/files#diff-9d06bce181173b761420826de5e33e89R174

sdroege:
Ok, so basically the same code could also be run for the values read from the configuration?

EPashkin:
Yes, on same place https://github.com/gtk-rs/gir/pull/416/files#diff-9d06bce181173b761420826de5e33e89R166
Also as we change `length` parameter (hide and calculate internally) and don't change `array` so IMHO naturally that config applied to `length`

EPashkin:
@sdroege, config can be applied to array with some small change, but there 2 mini problems:
1) find don't used mangle_keywords so there may be misses for some reserved names,
2) "array_length" as config name is not good as it can be applied to string too.
```rust
    //Map: length agrument position => array name
    let array_lengths: HashMap<u32, String> = function_parameters
        .iter()
        .filter_map(|p| {
            if let Some(array_length_name) = configured_functions
                .matched_parameters(&p.name)
                .iter()
                .filter_map(|p| p.array_length.as_ref())
                .next()
            {
                if let Some((pos, _par)) = function_parameters
                    .iter()
                    .enumerate()
                    .find(|&(_pos, par)| par.name == *array_length_name)
                {
                    return Some((pos as u32, p.name.clone()));
                }
            }
            p.array_length.map(|pos| (pos, p.name.clone()))
        })
        .collect();
```

EPashkin:
@GuillaumeGomez, what you think about configuring lengths (both options works):
I think that as only length parameter changed config better applied to it.
@sdroege want config on array.

GuillaumeGomez:
I'd just like our gir to be able to determine which integer is an array's length on its own. But I suppose it's just a dream... :'(

sdroege:
My main point here is consistency with the GI annotations: there you annotate the array (or string), not the length parameter. If that is a good choice, I don't know, but that's how it is ;)

EPashkin:
About consistency: in gir you can't apply array-length to string, value of array-length is int and differs for methods, we configure string in both cases :)

sdroege:
> About consistency: in .gir you can't apply array-length to string, value of array-length is int and differs for methods, we configure string in both cases :)

Value of array length is the name of the array in the annotations. It only gets converted to an integer for the .gir file, but that's not what developers work with usually.

The string vs. array thing is true of course, but that's just silly in GI currently and not changed because of backwards compatibility concerns. In C, all strings are arrays, and the handling on our side here is exactly the same for arrays and strings.

sdroege:
IMHO all good to go except for the way how the configuration works.

sdroege:
@EPashkin Are you also adding support for non-pointer arrays (u/i 8/16/32/64 / f32 / f64) while you're touching that code anyway?

EPashkin:
@sdroege No, I don't know how to do it.
I merge it now as @GuillaumeGomez don't voted for `array-length` config

GuillaumeGomez:
I was fine with the current code.

sdroege:
Ok, I don't care enough to continue discussing that but if a user is confused by it I'll send him your way ;)

sdroege:
> @sdroege No, I don't know how to do it.

I'll do that then. I also don't know yet how to teach gir that, but on the GLib side it's all there already at least.

PR #418, opened 2017-07-27
Add detection of length input parameters

EPashkin:
Part of https://github.com/gtk-rs/gtk/issues/477 and https://github.com/gtk-rs/gir/issues/264.

### DIFF
--- a/src/analysis/function_parameters.rs
+++ b/src/analysis/function_parameters.rs
@@ -138,6 +138,7 @@ pub fn analyze(
     env: &Env,
     function_parameters: &[library::Parameter],
     configured_functions: &[&config::functions::Function],
+    disable_length_detect: bool,
 ) -> Parameters {
     let mut parameters = Parameters::new(function_parameters.len());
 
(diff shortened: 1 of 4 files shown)


src/analysis/function_parameters.rs
```diff
@@ -311,3 +312,62 @@ fn get_length_type(
...
+    );
+    array.map(|p| &p.name)
+}
+
+fn is_length(par: &library::Parameter) -> bool {
+    if par.direction != library::ParameterDirection::In {
+        return false;
+    }
+
+    if par.name.len() >= 3 && &par.name[0..3] == "len" {
```
sdroege: Usually it's ending with 'len', not beginning. 'objects_len', 'string_len', etc. Or with the full word 'length'.

sdroege:
This needs a new configuration so it can be disabled. There will be false positives otherwise, where an argument is detected as a length parameter but actually is not.

src/analysis/function_parameters.rs
EPashkin: Changing this to `if len >= 3 && &par.name[len-3..len] == "len" {`
don't change any functions.
As it don't remove any too and sound reasonable I use new variant.

EPashkin:
Something like `dont_detect_lengths=true` on function level?

EPashkin:
or "detect_lengths=false"?

sdroege:
Yes, either name seems ok

EPashkin:
Added in as "disable_length_detect=true"

EPashkin:
Checked all other crates: PRs needed for glib, gio, pango but as it contains only generated code I postpone it until this merged.

Issue #388, opened 2017-06-19
Add support for non-pointer arrays/containers


sdroege:
Related to https://github.com/gtk-rs/gir/issues/376

Currently only pointer types are supported, i.e. FromGlibPtrContainer. All kinds of integer types would be useful too, including bytes for byte arrays.

sdroege:
I would suggest making a `FromGlibContainer` that has the `from_*_num()` functions, and a `FromGlibPtrContainer` that inherits from that and only has the non-`_num()` functions.

Once that is done, support for that can be added to gir.

The former would then be implemented for `*const T` and `*mut T` for `Vec<T>` where `T` is some suitable type. Now what is a suitable type? `T: Copy` unfortunately not because pointers also seem to be `Copy`. I could add manual implementations (via macro) for u/i 8/16/32/64, GType, and others though.

Any suggestions?

sdroege:
Similarly we need some trait impl for the other way around.

EPashkin:
IMHO better don't mix FromGlibContainer and FromGlibPtrContainer as it have different trait constraints

sdroege:
You mean I should just add a new FromGlibContainer (with the non-`_num()` functions only) and keep FromGlibPtrContainer as is?

What would you suggest for the other way around, ToGlib*?

EPashkin:
Checked `FromGlibPtrContainer` definition. Your idea seems doable, at minimum for concrete types like Vec<u8>

sdroege:
It will only work for `*const T` and `*const glib_ffi::GArray` to `Vec<T>, and the other way around (plus `&[T]` for that direction). Unfortunately this all involves copying for now with how the traits are defined (see https://github.com/gtk-rs/glib/issues/174), but better than not working at all.

sdroege:
For the other way around, ToGlibPtr / ToGlibPtrMut seems to be sufficient. Just needs to be implemented for [T] / Vec<T> where T: Copy and T: ToGlib. Which should cover all basic fundamental types and enums/flags at least.

Generally the main question now is how to implement all this generically in a coherent way, because the trait impl selection detects overlapping impls in these cases. At least for the From* traits. I'll try to figure something useful out.

EPashkin:
Overlapping impls is big problem in glib. I failed even adding GValue conversion for boxed types :(

sdroege:
[...]

PR #419, opened 2017-07-27
Add support for arrays of fundamental types

sdroege:
https://github.com/gtk-rs/gir/issues/388

### DIFF
--- a/src/analysis/rust_type.rs
+++ b/src/analysis/rust_type.rs
@@ -288,9 +288,17 @@ pub fn parameter_rust_type(
             }
         }
 
-        List(..) | SList(..) | CArray(..) => {
+        List(..) | SList(..) => {
             match direction {
                 library::ParameterDirection::In | library::ParameterDirection::Return => rust_type,
+                _ => Err(TypeError::Unimplemented(into_inner(rust_type))),
+            }
+        }
+        CArray(..) => {
+            match direction {
+                library::ParameterDirection::In |
+                library::ParameterDirection::Out |
+                library::ParameterDirection::Return => rust_type,
                 _ => Err(TypeError::Unimplemented(into_inner(rust_type))),
             }
         }
--- a/src/analysis/rust_type.rs
+++ b/src/analysis/rust_type.rs
@@ -147,6 +147,40 @@ fn rust_type_full(
                     format!("Vec<{}>", s)
                 })
         }
+        CArray(inner_tid) if ConversionType::of(&env.library, inner_tid) == ConversionType::Direct => {
+            if let Fundamental(fund) = *env.library.type_(inner_tid) {
+                let array_type = match fund {
+                    Int8 => Some("i8"),
+                    UInt8 => Some("u8"),
+                    Int16 => Some("i16"),
+                    UInt16 => Some("u16"),
+                    Int32 => Some("i32"),
+                    UInt32 => Some("u32"),
+                    Int64 => Some("i64"),
+                    UInt64 => Some("u64"),
+
+                    Int => Some("i32"),      //maybe dependent on target system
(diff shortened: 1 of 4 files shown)


sdroege:
https://github.com/sdroege/glib-rs/commit/2e85eba71c826fc589e53df021f21a0b9ba1a362 for an example how this looks like. g_filename_from_utf8() should probably be ignored.

EPashkin:
`g_key_file_set_double_list` and `g_key_file_set_integer_list` as it don't gir don't have `*` in arrays.
So generated ffi contains wrong `list`
`pub fn g_key_file_set_double_list(key_file: *mut GKeyFile, group_name: *const c_char, key: *const c_char, list: c_double, length: size_t);
`

sdroege:
Yes, see my GLib commit that disables these two. The .gir file is wrong

sdroege:
Also this currently does not seem to (always?) work with out parameters (g_file_load_contents() e.g.).

EPashkin:
`filename_from_utf8` compiles normally under windows but IMHO really don't needed.

EPashkin:
Other compile problems only starts with gtk, so all previous builds fine.

sdroege:
The support for out arrays needs various functions everywhere to be disabled that have missing/wrong annotations.

EPashkin:
From manual gtk\list_store maybe better change `values: &[&glib::Value]` to `values: &[&ToValue]` for input arrays ?

sdroege:
That can be added at a later time. That it is generated now is just a side effect of allowing out-arrays :)

sdroege:
Most remaining compile errors seem to be the following:
1) Wrong annotations
2) "zero-terminated" arrays of fundamental types, we can't really handle them
3) arrays of fundamental types without a length anywhere (some also compile but they would be unsafe as-is)
4) gdk::Atom curiousities

I'm checking now how to handle 2) and 3)

EPashkin:
Hm, now we stumbled upon reserved names names

```rust
pub fn base64_decode_step(in_: &[u8], state: &mut i32, save: &mut u32) -> (usize, Vec<u8>) {
    let len = in.len() as usize;
    unsafe {
        let mut out = ptr::null_mut();
        let ret = ffi::g_base64_decode_step(in_.to_glib_none().0, len, &mut out, state, save);
        (ret, FromGlibPtrContainer::from_glib_full(out))
    }
}
```

sdroege:
That one is not generated anymore

EPashkin:
@sdroege, please add `let array_name = nameutil::mangle_keywords(&array_name[..]).into_owned();` in  body of `if let Some(array_name) = array_name {` in `function_parameters.rs` even if you ignore this function

sdroege:
Sure

EPashkin:
Sorry `.into_owned()` unneeded

src/analysis/rust_type.rs
```diff
@@ -147,6 +147,40 @@ fn rust_type_full(
                     format!("Vec<{}>", s)
                 })
         }
+        CArray(inner_tid) if ConversionType::of(&env.library, inner_tid) == ConversionType::Direct => {
```
EPashkin: Strange, this case under `CArray(inner_tid)` without `if`. It dont skipped?

sdroege: It's also with if. Only pointers are handled above

EPashkin: You right, sorry.

EPashkin:
@sdroege PR ready for merge?

sdroege:
Yes, good to go IMHO. All remaining failures are bugs in the .gir and can be ignored (will do so when doing the regenerate PRs).

EPashkin:
Thanks, @sdroege

PR #424, opened 2017-08-02
Fix borrowed_box clippy warning in generated trampoline code

sdroege:
https://github.com/rust-lang-nursery/rust-clippy/wiki#borrowed_box

### DIFF
--- a/src/codegen/trampoline.rs
+++ b/src/codegen/trampoline.rs
@@ -53,7 +53,7 @@ pub fn generate(
         ));
     }
     try!(writeln!(w, "\tcallback_guard!();"));
-    try!(writeln!(w, "\tlet f: &Box_<{}> = transmute(f);", func_str));
+    try!(writeln!(w, "\tlet f: &&({}) = transmute(f);", func_str));
     try!(transformation_vars(w, analysis));
     let call = trampoline_call_func(env, analysis, in_trait);
     try!(writeln!(w, "\t{}", call));


sdroege:
It also complains about https://github.com/rust-lang-nursery/rust-clippy/wiki#transmute_ptr_to_ref but this seems non-trivial to fix. E.g. in one of my examples this would become the following instead of a simple `transmute`. I'd ignore that warning :)

```Rust
&*(func as *const &for<'r, 'r, 'r> std::ops::Fn(&'r auto::pad::Pad, &'r mut pad::PadProbeInfo<'r>) -> auto::enums::PadProbeReturn + std::marker::Sync + std::marker::Send + 'static)
```

EPashkin:
Thanks.
Affected crates  gio, gdk, gtk.
Gio, Gtk also need same replace in manual part.

sdroege:
You agree that https://github.com/rust-lang-nursery/rust-clippy/wiki#transmute_ptr_to_ref is not worth the pain? Main problem here is updating all the references to `'r`

EPashkin:
Yes, totally, As it need extra type definition for as conversion.

sdroege:
Maybe there's a way how we can silence that one?

sdroege:
Ah yes:
```Rust
#[cfg_attr(feature = "cargo-clippy", allow(transmute_ptr_to_ref))]
```

sdroege:
I'll send a PR for adding that in the relevant places later

EPashkin:
IMHO adding #cfg to each line/trampoline will make the code less understandable.
While adding `#![cfg_attr(feature = "cargo-clippy", allow(transmute_ptr_to_ref))]` will hide some other errors currently I prefer this way.
Also converting completely different types with `transmute` IMHO is better that ``

EPashkin:
.. using `as` with some complex type.

PR #425, opened 2017-08-02
Get rid of some clippy warnings

sdroege:
This makes gstreamer-rs warning free with the default clippy configuration.

### DIFF
--- a/src/codegen/enums.rs
+++ b/src/codegen/enums.rs
@@ -187,6 +187,7 @@ impl ToGlib for {name} {{
     try!(writeln!(
         w,
         "#[doc(hidden)]
+#[cfg_attr(feature = \"cargo-clippy\", allow(unreadable_literal))]
 impl FromGlib<ffi::{ffi_name}> for {name} {{
     fn from_glib(value: ffi::{ffi_name}) -> Self {{
         {assert}match value as i32 {{",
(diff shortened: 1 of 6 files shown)


EPashkin:
I check this latter.
See https://github.com/gtk-rs/gir/pull/424#issuecomment-319863955, but `Default` IMHO good idea.

sdroege:
So you prefer to have no annotations for these errors? Maybe we can just add them top-level to the source files, but I'm not sure how that works with clippy. But would that be more acceptable for you?

EPashkin:
@sdroege Yes, I prefer add `allow`s for clippy on crate levels for `unreadable_literal` and `transmute_ptr_to_ref`. Currently don't checked when `match_same_arms` appeared but maybe this too if it too often.
It works fine, only possible problem that we can miss warning in manual code that better be fixed.
@GuillaumeGomez what you think about this?

sdroege:
> Currently don't checked when match_same_arms appeared but maybe this too if it too often.

It happens for `ErrorDomain` impls if there is a `Failed` case.

>  It works fine, only possible problem that we can miss warning in manual code that better be fixed.

Currently you don't find possible problems in manual code because there's just too much noise in the autogenerated code :) Maybe the `allow`s can be added in src/auto/mod.rs for the whole module? I'm not sure how/if that works though.

EPashkin:
Hm, Interesting idea try adding it before `mod auto;`, maybe it really works

sdroege:
Works indeed, I'm surprised

sdroege:
This does it for me in lib.rs:
```Rust
#[cfg_attr(feature = "cargo-clippy", allow(unreadable_literal))]
#[cfg_attr(feature = "cargo-clippy", allow(transmute_ptr_to_ref))]
#[cfg_attr(feature = "cargo-clippy", allow(too_many_arguments))]
#[cfg_attr(feature = "cargo-clippy", allow(match_same_arms))]
mod auto;
```

EPashkin:
Strange, but good that it works.

PR #427, opened 2017-08-03
Support action signals

sdroege:
These work the other way around than normal signals: We are supposed to
emit them from outside the object instead of connecting to it by using
g_signal_emit*(). It basically is like a dynamic function call.
The object itself is connected to the signal and waiting for us to emit
it, to do something.

### DIFF
--- a/src/analysis/signals.rs
+++ b/src/analysis/signals.rs
@@ -10,12 +10,14 @@ use version::Version;
 
 #[derive(Debug)]
 pub struct Info {
+    pub object_name: String,
     pub connect_name: String,
     pub signal_name: String,
     pub trampoline_name: Result<String, Vec<String>>,
     pub version: Option<Version>,
     pub deprecated_version: Option<Version>,
     pub doc_hidden: bool,
+    pub is_action: bool,
 }
 
 pub fn analyze(
@@ -40,6 +42,7 @@ pub fn analyze(
 
         let info = analyze_signal(
             env,
+            obj,
             signal,
             type_tid,
             in_trait,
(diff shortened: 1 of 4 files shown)


src/analysis/signals.rs
```diff
@@ -10,12 +10,14 @@ use version::Version;
 
 #[derive(Debug)]
 pub struct Info {
+    pub object_name: String,
```
sdroege: This is only added so that we can print it in codegen later for action signals :)

sdroege:
Next step is of course to add support for these, which I don't need currently. It's mostly a matter of making a new function that calls g_signal_emitv() or just ObjectExt::emit() and converts from/to GValues.

EPashkin:
@sdroege, you sure that ignoring action signal is right?
We will lost `gtk::Button::connect_clicked` and many others.

sdroege:
Hrm, so GTK is marking signals as action signals that are expected to be used in both directions. I guess in that case, the solution would be to a) ignore them manually in my stuff and b) change the "ignore" part here to instead generate an emitter function :)

EPashkin:
By https://developer.gnome.org/gobject/unstable/gobject-Signals.html#G-SIGNAL-ACTION:CAPS, action signals is just signals with simplified emitting, so ignoring don't needed.

sdroege:
Not really simplified emitting, but you emit them from the outside usually. Ok, in any case, what is needed here is to add an emitter function for these signals... instead of ignoring them :)

sdroege:
This is fixed since https://github.com/gtk-rs/gir/commit/545158ef37a13cfa33db7c55acb70ddfe481c1fa

PR #428, opened 2017-08-04
Add configuration for concurrency per signal

sdroege:
When the object is Send+Sync it might still make sense to use Send only
for the signal. This is the case if the signal can be called only from a
single thread at a time.

I need this for GStreamer to make usage together with GTK a bit simpler.


EPashkin:
@sdroege LGFM, only only test not fixed.

sdroege:
Now hopefully :)

EPashkin:
Thanks.

PR #429, opened 2017-08-05
Generate version conditions for Default implementation

EPashkin:
Fix `Default` implementation for `PlacesSidebar` that have 
```
#[cfg(feature = "v3_10")]
     pub fn new() -> PlacesSidebar {
```

cc @sdroege

### DIFF
--- a/src/codegen/general.rs
+++ b/src/codegen/general.rs
@@ -251,9 +251,17 @@ pub fn write_vec<T: Display>(w: &mut Write, v: &[T]) -> Result<()> {
     Ok(())
 }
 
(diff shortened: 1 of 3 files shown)


EPashkin:
Maybe for `PlacesSidebar` enough just add version to object in config, but maybe there some classes that get `new` with delay.

EPashkin:
For gdk:
```
"member Gdk.EventType": Unknown key: ["name"]
"ident Gdk.EventType": Unknown key: ["alias"]
"child properties Gdk.EventType": Unknown key: ["member"]
```

sdroege:
Probably better to just do it generically than working around it for this specific object. Makes sense to me in any case.

Issue #431, opened 2017-08-09
Should probably ignore items with introspectable="0" in the .gir


sdroege:
These are the items that are marked as `(skip)` in the in-code annotations, meant to be not used by bindings. Those should probably be omitted from the non-sys bindings.

The annotation can be on objects and functions at least.

EPashkin:
IMHO some usable also has this flag

sdroege:
Which for example? That would be a bug in the .gir file then

sdroege:
Let's close this, there are too many things marked as not introspectable that we can handle just fine.

PR #433, opened 2017-08-09
Don't generate non-sys bindings for introspectable="0" functions/sign…

sdroege:
…als/properties

For classes, records, interfaces, bitfields, enums we ignore that for
now as those have to be explicitly enable in the configuration and
hopefully the user knows what they're doing.

https://github.com/gtk-rs/gir/issues/431

### DIFF
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -68,7 +68,7 @@ pub fn analyze<F: Borrow<library::Function>>(
         if configured_functions.iter().any(|f| f.ignore) {
             continue;
         }
-        if env.is_totally_deprecated(func.deprecated_version) {
+        if env.is_totally_deprecated(func.deprecated_version) || !func.introspectable {
             continue;
         }
         let name = nameutil::mangle_keywords(&*func.name).into_owned();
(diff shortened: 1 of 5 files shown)


sdroege:
Cleans up a lot of noise for varargs/etc in gtk also, in my case it makes a function be ignored for which we otherwise create unsafe bindings

EPashkin:
GLib and Pango don't regen now.
```
thread 'main' panicked at 'Missing memory management functions for GLib.KeyFile', src\codegen\record.rs:40:8
thread 'main' panicked at 'Missing memory management functions for Pango.Attribute', src\codegen\record.rs:40:8
```
On Gtk removed many commented functions.
IMHO wrong changes on `MessageDialog`, `RecentChooserDialog` but it have manual implementations, so no problem.
`IconInfoExt` lost `copy`,
`ClipboardExt` lost `get_selection`.
So better add this functionality under config, that default disabled.

sdroege:
I guess that makes using it for anything useless as the skip annotation is used too much for things it shouldn't have been used for. Let's just close this then? :)

EPashkin:
As you wish.

PR #434, opened 2017-08-09
Add target_os handling

### LINKED ISSUE #421: Add target_os key for toml config
It'd be very convenient in some cases (like X11 interactions with GDK and GTK).

GuillaumeGomez:
Fixes #421.

It's apparently impossible to test it for the moment, so I propose we keep it open, laugh about it from time to time and then forget it unless some unbelievable event makes it testable.

### DIFF
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -47,6 +47,7 @@ pub struct Info {
     pub deprecated_version: Option<Version>,
     pub not_version: Option<Version>,
     pub cfg_condition: Option<String>,
+    pub target_os: Option<String>,
     pub assertion: SafetyAssertionMode,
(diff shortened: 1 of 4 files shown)


EPashkin:
While this can handled with `cfg_condition` this make config more clear.
Thanks, @GuillaumeGomez

GuillaumeGomez:
The only issue is that we don't have any use for it for the moment. :-/

sdroege:
What would be the use case for this, one that actually can work with gobject-introspection as it is now? :)

GuillaumeGomez:
That's exactly my point. I made this change and then wondered: "when could we use this?". And after realising it was completely useless, I decided to open the PR just in case...

Issue #437, opened 2017-08-17
Add support for property notify signals


sdroege:
GObject has a "notify" signal, which is emitted for property changes. "notify::property-name" is usually used if you're interested only in a specific property.

I would suggest that we generate `connect_property_XXX_notify` functions for every property (even for one that we currently ignore because there is already a setter/getter function for them) that take a `Fn(&Self) + 'static`, plus `Send` or `Sync` depending on the concurrency configuration.

sdroege:
@EPashkin How would we best go about implementing this? Add them to the signals as a special case in analysis/properties? Generate them as part of codegen/properties and somehow re-use the signal stuff? Ideally this shouldn't require refactoring everything :)

EPashkin:
Seems for this both `signals::Info` and `Trampoline` can be used directly.
So just adding `pub nofity_signals: Vec<signals::Info>` to `object::Info` and add trampoline to same vector is enough.

PR #441, opened 2017-08-18
Add support for property notifications (notify signal)

sdroege:
https://github.com/gtk-rs/gir/issues/437

### DIFF
--- a/src/analysis/properties.rs
+++ b/src/analysis/properties.rs
@@ -3,6 +3,8 @@ use analysis::imports::Imports;
 use analysis::ref_mode::RefMode;
 use analysis::rust_type::*;
 use analysis::signatures::{Signature, Signatures};
+use analysis::signals;
+use analysis::trampolines;
 use config;
 use config::gobjects::GObject;
 use env::Env;
@@ -32,12 +34,15 @@ pub fn analyze(
     env: &Env,
     props: &[library::Property],
     type_tid: library::TypeId,
+    generate_trait: bool,
+    trampolines: &mut trampolines::Trampolines,
     obj: &GObject,
     imports: &mut Imports,
     signatures: &Signatures,
     deps: &[library::TypeId],
-) -> Vec<Property> {
+) -> (Vec<Property>, Vec<signals::Info>) {
     let mut properties = Vec::new();
+    let mut notify_signals = Vec::new();
 
     for prop in props {
         let configured_properties = obj.properties.matched(&prop.name);
@@ -49,14 +54,23 @@ pub fn analyze(
             continue;
         }
 
-        let (getter, setter) = analyze_property(
+        let (getter, setter, notify_signal) = analyze_property(
             env,
             prop,
             type_tid,
             &configured_properties,
+            generate_trait,
+            trampolines,
+            obj,
+            imports,
             signatures,
             deps,
         );
+
+        if let Some(notify_signal) = notify_signal {
+            notify_signals.push(notify_signal);
+        }
+
         if getter.is_none() && setter.is_none() {
             continue;
         }
@@ -93,7 +107,7 @@ pub fn analyze(
         }
     }
 
-    properties
+    (properties, notify_signals)
 }
 
 fn analyze_property(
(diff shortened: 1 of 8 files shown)


src/analysis/properties.rs
```diff
@@ -199,7 +217,64 @@ fn analyze_property(
         None
     };
 
-    (getter, setter)
+    let mut used_types: Vec<String> = Vec::with_capacity(4);
+    let trampoline_name = trampolines::analyze(
+        env,
+        &library::Signal {
+            name: format!("notify::{}", name),
+            parameters: Vec::new(),
```
sdroege: Somehow we need to tell it to generate a `gpointer` parameter for the `extern "C"` trampoline but ignore it for anything else except for just having it there.

sdroege: Done

sdroege:
With those commits I get proper property notification connectors for everything in the gstreamer bindings.

sdroege:
https://github.com/sdroege/gstreamer-rs/commit/74c9f9145a028623187a45ac3c67427c4ad4e508 for how this looks like

sdroege:
@GuillaumeGomez Would be nice to get this into the release, even if *I* don't need it for anything in glib/gtk/etc. But I won't be able to do any fixups for the PR in the next days

GuillaumeGomez:
Hum indeed. Anything against it @EPashkin?

src/analysis/ref_mode.rs
```diff
@@ -27,7 +27,11 @@ impl RefMode {
             ..
         }) = env.config.objects.get(&tid.full_name(&env.library))
         {
-            return ref_mode;
+            if direction == library::ParameterDirection::In {
+                return ref_mode;
+            } else {
+                return RefMode::None;
```
EPashkin: Can you give sample, when this needed.

sdroege: Caps in gstreamer. Manual but ref-counted, so `ref_mode = "ref"` is configured. And if you look below in that function, the behaviour is the same. It only returns something else than `None` for `In` parameters, as everything else makes no sense.

Without this, e.g. a property returning Caps will return `Option<&Caps>`, which makes no sense and is nothing that could be constructed.

EPashkin: if direction== Return { return None } will be cleaner

EPashkin:
I has only one question, but can merge without it solves as this change don't affect gtk-rs.

sdroege:
It should affect every crate that has properties. They will get a notification signal for those properties now.

src/analysis/ref_mode.rs
sdroege: It would also be for Out parameters (and InOut I guess, but we don't really support them).

EPashkin:
I meant `return None` question. If you can't fix it fast, I merge.

sdroege:
`return None` would return `Option::None` instead of `RefMode::None`. I'm not sure I understand what you mean then :)

src/analysis/ref_mode.rs
EPashkin:
```
if direction == library::ParameterDirection::Return {
      return RefMode::None;
} else {
      return ref_mode;
}
```

EPashkin:
Ok, then I merge now.
@sdroege Thanks

sdroege:
Thanks :)

src/analysis/ref_mode.rs
sdroege: It should also be None for ::Out at least

sdroege:
Now only someone needs to regenerate gtk and gio at least, not sure which others have properties

EPashkin:
I will do it tomorrow

GuillaumeGomez:
I'll do gtk this evening, I leave the rest to @EPashkin. ;)

EPashkin:
There path that better added for gtk (I will add it to others)
```diff
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,9 @@ script:
   - ./check_init_asserts
   - cargo install clippy || touch clippy_failed
   - if [ ! -f clippy_failed ]; then cargo clippy; fi
-  - cargo test --features embed-lgpl-docs
+  - cargo test --features embed-lgpl-docs --jobs 1
   # catch any sneaked in lgpl docs
-  - cargo build --features purge-lgpl-docs
+  - cargo build --features purge-lgpl-docs --jobs 1
   - git diff -R --exit-code
   - mkdir .cargo
   - echo 'paths = ["."]' > .cargo/config
```

GuillaumeGomez:
Why the `--jobs 1` option btw?

EPashkin:
Your libs take too much memory when compiling.
See https://github.com/gtk-rs/examples/pull/138.

GuillaumeGomez:
Erf. Do you want PRs to add this change or do you prefer I do it?

EPashkin:
As you do regen on gtk, I prefer that you add this fix in your PR

GuillaumeGomez:
Yes sir!

PR #450, opened 2017-09-01
Rename binary to 'gir-rs'

### LINKED ISSUE #449: Slightly more description name for binary
'gir' is way too vague a name if the binary is to be installed on the systems (which it should if we want external bindings to be possible). I'd suggest renaming it to 'gir-rs' or 'gir-rust'.

On that note, I'd also suggest changing name of the repository. Currently when you clone the repo, you'll get a repo named 'gir'. Today I might remember what that is about, but not in a month's time. :) Also, if any other language bindings for gir decided to name their repo like this, we'll have a conflict.

zeenix:
If this binary is to be installed, we really need to give it a more
descriptive name to avoid any conflicts or confusions.

Fixes #449.


zeenix:
Seems i ended up forced pushing to a branch that was already part of a PR. :(

GuillaumeGomez:
This PR doesn't do what it says it did so I'll rename and merge then. Thanks!

PR #451, opened 2017-09-04
Remove crate time

EPashkin:
It was not used


EPashkin:
Strange `Cargo.lock` don't changed when I lower `git2` version to 0.6.6.

EPashkin:
Note: there was second commit with results `cargo update`,
but it causes appveor failure due git2 update to 6.8.
Seems it need greater version that appveyor's mingw-w64-i686-libgit2-0.24.0-1 (latest is 0.26.0).
As result this commit was removed.

PR #453, opened 2017-09-07
Use a newtype wrapper for the signal handler IDs

sdroege:
https://github.com/gtk-rs/glib/issues/229

### DIFF
--- a/src/analysis/properties.rs
+++ b/src/analysis/properties.rs
@@ -258,6 +258,7 @@ fn analyze_property(
             imports.add("glib::object::Downcast", prop_version);
         }
         imports.add("glib::signal::connect", prop_version);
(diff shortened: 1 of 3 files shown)


EPashkin:
Sorry, I forgot about this PR.

EPashkin:
Note: now regen on gtk produced many `ERROR:gir::analysis::trampoline_parameters: Wrong conversion_type for borrow transformation Borrow`, but code seems good.

sdroege:
Do you understand why that warning happens and exists?

Will send regens in a bit, lunch first

EPashkin:
It was manually changed conversion type for `TransformationType::Borrow` when `ConversionType::Pointer` passed to `ConversionType::Borrow`,
As you changed passed to `ConversionType::Borrow` it think that you used unsupported type and nags.
I fix it.

PR #454, opened 2017-09-09
Borrow Boxed/Shared/Object/Interface arguments in signal trampolines

sdroege:
They are valid for the whole scope of the trampoline, obviously, and by
using from_glib_none() and passing them by reference we would in the
best case only do atomic reference counting but in the worst case
actually copy the data just to pass on a reference.

### DIFF
--- a/src/analysis/trampoline_parameters.rs
+++ b/src/analysis/trampoline_parameters.rs
@@ -119,7 +119,7 @@ pub fn analyze(
         library::Transfer::None,
         library::Nullable(false),
         RefMode::ByRef,
-        ConversionType::Pointer,
+        ConversionType::Borrow,
     );
     parameters.transformations.push(transform);
 
@@ -135,7 +135,14 @@ pub fn analyze(
             .next();
         let nullable = nullable_override.unwrap_or(par.nullable);
 
-        let conversion_type = ConversionType::of(&env.library, par.typ);
+        let conversion_type = {
+            match *env.library.type_(par.typ) {
+                library::Type::Record(..) |
+                library::Type::Interface(..) |
+                library::Type::Class(..) => ConversionType::Borrow,
+                _ => ConversionType::of(&env.library, par.typ),
+            }
+        };
 
         let new_name = configured_signals
             .matched_parameters(&name)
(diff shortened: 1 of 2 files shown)


sdroege:
For interfaces this is not entirely correct though:
```diff
@@ -95,12 +95,12 @@ unsafe extern "C" fn notify_mute_trampoline<P>(this: *mut ffi::GstStreamVolume,
 where P: IsA<StreamVolume> {
     callback_guard!();
     let f: &&(Fn(&P) + Send + Sync + 'static) = transmute(f);
-    f(&StreamVolume::from_glib_none(this).downcast_unchecked())
+    f(&StreamVolume::from_glib_borrow(this).downcast_unchecked())
 }
```

The downcast would steal the borrowed reference, it internally uses `from_glib_none`. How should we solve this?

sdroege:
Maybe having a cast variant that just borrows for this specific case?

sdroege:
Ah, never mind. It actually does the right thing all along the way. Good to merge :)

```rust
    unsafe fn downcast_unchecked(self) -> Sub {
        // Just gets the pointer without any ref/unref
        debug_assert!(types::instance_of::<Sub>(self.to_glib_none().0 as *const _));
        // Into<ObjectRef>:
        // Just returns the ObjectRef that contains the borrowed==true, so it's preserved
        // UnsafeFrom::from:
        // Just wraps the ObjectRef again, preserving borrowed==true
        Sub::from(self.into())
    }
```

EPashkin:
Seems glib need fix beforehand, as no `from_glib_borrow` for `GVariant`
```
error[E0277]: the trait bound `std::option::Option<glib::Variant>: glib::translate::FromGlibPtrBorrow<*mut glib_ffi::GVariant>` is not satisfied
   --> D:/eap/rust/0/gio\src\auto\simple_action.rs:130:68
    |
130 |     f(&SimpleAction::from_glib_borrow(this).downcast_unchecked(), &from_glib_borrow(parameter))
    |                                                                    ^^^^^^^^^^^^^^^^ the trait `glib::translate::FromGlibPtrBorrow<*mut glib_ffi::GVariant>` is not implemented for `std::option::Option<glib::Variant>`
    |
    = note: required by `glib::translate::from_glib_borrow`
```

sdroege:
`glib_wrapper!` is defining one here, but the Option<_> impl is missing. Find it here: https://github.com/gtk-rs/glib/pull/232

GuillaumeGomez:
Ok, so now the fix has been merged. Anything else to update in here?

EPashkin:
IMHO all good now.

GuillaumeGomez:
I let you merge then!

PR #456, opened 2017-09-10
Use Path for shared library name extraction

tdgne:
First commit makes it panic when `shared-library` has a wrong form.
(Thought it'd be helpful to do so instead of generating uncompilable code.)

Second commit combines `Path` with `Regex` to extract the library name, which will support extraction from absolute paths and macOS `.dylib`s as well.

Sorry if I'm misunderstanding the intention of the present code...

### DIFF
--- a/src/codegen/sys/build.rs
+++ b/src/codegen/sys/build.rs
@@ -1,4 +1,5 @@
 use std::io::{Result, Write};
+use std::path::Path;
 
 use env::Env;
 use file_saver::save_to_file;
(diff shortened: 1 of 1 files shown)


EPashkin:
Can you give example for `.gir` file with `dylib`?

tdgne:
Here it is!

[Gdl-3.gir.zip](https://github.com/gtk-rs/gir/files/1291031/Gdl-3.gir.zip)

EPashkin:
@tdgne Thanks

PR #459, opened 2017-09-14
Emit ffi::FooClass whenever we emit a ffi::Foo Object

federicomenaquintero:
This is for https://github.com/gtk-rs/glib/issues/233

For gnome-class (the procedural macro to create GObject implementations in Rust easily), I need to be able to obtain both the Instance and the Class structs for a GObject type.

This pull request adds support for parsing this info out of .gir.  We'll require an API change in the glib_wrapper!() macro from the glib crate.


src/codegen/general.rs
```diff
@@ -62,21 +63,26 @@ pub fn define_object_type(
         })
         .collect();
 
+    let class_name = glib_class_name.map_or ("()".to_string(),
+                                             |s| format!("ffi::{}", s));
```
sdroege: Space before the `(`. Please run everything through latest rustfmt-nightly :)

src/codegen/general.rs
```diff
@@ -62,21 +63,26 @@ pub fn define_object_type(
...
             type_name,
-            glib_name
+            glib_name,
+            class_name
         ));
     } else if external_parents {
         try!(writeln!(
             w,
-            "\tpub struct {}(Object<ffi::{}>): [",
+            "\tpub struct {}(Object<ffi::{}, {}>): [",
```
sdroege: You could also implement the macro so that it can have one or two parameters for `Object<..>`. Not sure which is nicer but with having the macro in two variants there would be less noise in the generated code for things that have no class struct

federicomenaquintero: You are absolutely correct!  I started with the assumption that everything had a class struct, then it turned out that that wasn't true, I added the (), and left it at that.

But with your scheme, there wouldn't be an API break, just things that cannot be subclassed with gnome-class.

I like the idea of being able to not specify a class struct; I'll change the code to do that.

federicomenaquintero:
Updated per @sdroege's comments.

EPashkin:
Maybe postprocess step don't need as types has `glib:type-struct="ActionableInterface"`

EPashkin:
Also `src/auto/actionable.rs` don't updated

federicomenaquintero:
Good catch on actionable.rs.  I'm debugging it.

Hmm, any ideas on how to start writing tests for gir?  Maybe make it process little .gir files and grep the output to see if it contains... something?

EPashkin:
Note, seems all other interfaces don't updated too.

Yes theoretically test will looks like you say.
Not sure that external `tests` can be done when executable is main,
so IMHO need some library extraction.

federicomenaquintero:
Duuuh, I was storing c_class_type in analysis::object::class(), but not analysis::object::interface().  Fixed; will update the repository ASAP.

(BTW, rr-project.org is awesome for this!)

EPashkin:
@federicomenaquintero Can you change parser to use `glib:type-struct` instead `is-gtype-struct-for`?

EPashkin:
@federicomenaquintero Can you "fix" generator to use `glib:type-struct` instead `is-gtype-struct-for`?

federicomenaquintero:
Yes, I will look into this.  Thanks for being patient - I had this in the state of "it works for me" and used it to keep working on gnome-class.  I'll do the forward lookups instead of backward ones as you suggest.

federicomenaquintero:
OK, I made it read glib::type-struct as well.  In the postprocessing stage, there is now a sanity check to see that glib:type-struct and glib:is-gtype-struct-for actually match.

EPashkin:
Please fix https://travis-ci.org/gtk-rs/gir/jobs/298687669#L687-L705

Gir_Gtk.toml
```diff
@@ -28,6 +28,7 @@ generate = [
     "Gtk.ButtonBox",
     "Gtk.Calendar",
     "Gtk.CheckButton",
+    "Gtk.Clipboard",
```
EPashkin: This really needed?

src/codegen/object.rs
```diff
@@ -14,11 +14,13 @@ use super::trampoline;
 pub fn generate(w: &mut Write, env: &Env, analysis: &analysis::object::Info) -> Result<()> {
     try!(general::start_comments(w, &env.config));
     try!(general::uses(w, env, &analysis.imports));
+
     try!(general::define_object_type(
         w,
         env,
         &analysis.name,
         &analysis.c_type,
+        &analysis.c_class_type.as_ref().map(|s| &*s as &str),
```
EPashkin: `&s[..]` not better?

src/library_postprocessing.rs
```diff
@@ -168,4 +170,107 @@ impl Library {
...
+    fn resolve_class_structs(&mut self) {
+        // stores pairs of (gtype-struct-c-name, type-name)
+        let mut structs_and_types = Vec::new();
+
+        for (ns_id, ns) in self.namespaces.iter().enumerate() {
+            for type_ in ns.types.iter() {
+                let type_ = type_.as_ref().unwrap(); //Always contains something
+
+                match *type_ {
+                    Type::Record(ref record) => {
```
EPashkin: `if let Type::Record` ?

src/library_postprocessing.rs
```diff
@@ -168,4 +170,107 @@ impl Library {
...
+
+                if let &Some(ref type_struct) = type_struct {
+                    let type_struct_tid = self.find_type(ns_id as u16, type_struct);
+                    if type_struct_tid.is_none() {
+                        panic!("\"{}\" has glib:type-struct=\"{}\" but there is no such record",
+                               name,
+                               type_struct);
+                    }
+
+                    let type_ = self.type_(*type_struct_tid.as_ref().unwrap());
```
EPashkin: Too many `type_` in one function

EPashkin: `.as_ref()` seems unneeded

src/library_postprocessing.rs
```diff
@@ -168,4 +170,107 @@ impl Library {
...
+
+                    match *type_ {
+                        Type::Record(ref r) => {
```
EPashkin: Also `if let` ?

src/library_postprocessing.rs
```diff
@@ -168,4 +170,107 @@ impl Library {
...
+                                           r.name);
+                                }
+                            }
+                        },
+
+                        _ => { panic!("Element with name=\"{}\" should be a record but it isn't",
+                                      type_struct); }
+                    }
+                } else {
+                    if !(type_struct.is_none() && c_class_type.is_none()) {
```
EPashkin: IMHO Something is not right with this condition: is equivalent to `type_struct.is_some() || c_class_type.is_some()`,
as first part covered in `if let` maybe second is enough

EPashkin: And `else if` will works in one line

src/main.rs
```diff
@@ -84,35 +84,57 @@ fn do_main() -> Result<()> {
...
+    }
+
     {
         let _watcher = statistics.enter("Analysing");
         analysis::run(&mut env);
     }
 
     {
-        let _watcher_generating = statistics.enter("Generating");
+        let _watcher = statistics.enter("Generating");
```
EPashkin: Unneeded empty line. Thanks for more clean statistics.

EPashkin:
@federicomenaquintero Thanks for fix. There only small nits.

src/library_postprocessing.rs
```diff
@@ -168,4 +170,107 @@ impl Library {
...
+                    Type::Interface(ref iface) => {
+                        name         = &iface.name;
+                        type_struct  = &iface.type_struct;
+                        c_class_type = &iface.c_class_type;
+                    },
+
+                    _ => { continue; }
+                }
+
+                if let &Some(ref type_struct) = type_struct {
```
EPashkin: Please `if let Some(ref type_struct) = *type_struct {` if compiler allow it.

Gir_Gtk.toml
federicomenaquintero: For eyeball tests, yes :-P  I wanted to check that GtkClipboard is emitted without the class struct (it doesn't have a public one).

src/codegen/object.rs
federicomenaquintero: Of course, you are right.

src/library_postprocessing.rs
federicomenaquintero: Done.

src/library_postprocessing.rs
federicomenaquintero: Oh, it is Copy, you are right.

src/library_postprocessing.rs
federicomenaquintero: Done.

src/library_postprocessing.rs
federicomenaquintero: You are right.  I've redone it.

src/main.rs
federicomenaquintero: Done

src/library_postprocessing.rs
federicomenaquintero: Done.  Good catch.

Gir_Gtk.toml
federicomenaquintero: At some point I'd like to refactor the various stages of gir so that small-scale tests are possible.  Currently it must run from the beginning, with "real" input files.

EPashkin: Ok

src/config/gobjects.rs
```diff
@@ -1,4 +1,3 @@
-use std::ascii::AsciiExt;
```
EPashkin: It used.

EPashkin:
@federicomenaquintero Thanks, only last is removing last commit (about `use std::ascii::AsciiExt`),
and fix https://travis-ci.org/gtk-rs/gir/jobs/298687669#L687-L705 (strange that only stable give this warning, but it works without `mut` in it too)

src/library_postprocessing.rs
```diff
@@ -168,4 +170,102 @@ impl Library {
...
+            match *self.type_mut(struct_for_tid) {
+                Type::Class(ref mut klass) => {
+                    klass.c_class_type = Some(gtype_struct_c_type);
+                },
+
+                Type::Interface(ref mut iface) => {
+                    iface.c_class_type = Some(gtype_struct_c_type);
+                },
+
+                ref x @ _ => unreachable!("Something other than a class or interface has a class struct: {:?}", x)
```
EPashkin: Clippy says `the 'x @ _' pattern can be written as just 'x'`

src/library_postprocessing.rs
```diff
@@ -168,4 +170,102 @@ impl Library {
             _ => false,
         }
     }
+
+    fn resolve_class_structs(&mut self) {
+        // stores pairs of (gtype-struct-c-name, type-name)
+        let mut structs_and_types = Vec::new();
+
+        for (ns_id, ns) in self.namespaces.iter().enumerate() {
+            for type_ in ns.types.iter() {
```
EPashkin: `&ns.types` ?

src/config/gobjects.rs
federicomenaquintero: Uhh, where? :)  The compiler told me it was unused, and it built fine without it.

src/library_postprocessing.rs
federicomenaquintero: Nice, done.

src/library_postprocessing.rs
federicomenaquintero: Of course, thanks.

src/config/gobjects.rs
federicomenaquintero: Never mind, I see it now.  It's only required on stable; nightly doesn't need it.  I've added it back.

EPashkin:
@federicomenaquintero  Thanks

federicomenaquintero:
Thank you for merging this!

PR #460, opened 2017-09-14
Add doc for signals

GuillaumeGomez:
I'll add for properties a bit later. I found a bug on rustdoc-stripper and I'm trying to fix it.

PR #462, opened 2017-09-15
Add property doc generation

GuillaumeGomez:
Fixes #460.

And here is the second part!

cc @EPashkin 

(I'll regenerate docs for crates once merged.)


EPashkin:
@GuillaumeGomez, thanks

Issue #465, opened 2017-09-18
Add support for C bitfields


sdroege:
Blocked by https://github.com/rust-lang/rfcs/issues/314 but we need that for the remaining types missing in glib-sys, etc.

GuillaumeGomez:
Is it done?

PR #466, opened 2017-09-18
Generate non-truncated versions for GLib's GHookList

sdroege:
I need this for GStreamer, as `GstPad` contains a `GHookList` and I need access to its fields after the `GHookList`. Did an awful (and not entirely correct) workaround for that for now: https://github.com/sdroege/gstreamer-rs/blob/f4958b71d71d7d6a71fddaed71535887e19e44c0/gstreamer/src/pad.rs#L672

This generates
```rust
#[repr(C)]
pub struct GHookList {
    pub seq_id: c_ulong,
    #[cfg(any(not(windows), not(target_pointer_width = "64")))]
    pub hook_size_and_setup: gpointer,
    #[cfg(all(windows, target_pointer_width = "64"))]
    pub hook_size_and_setup: c_ulong
    pub hooks: *mut GHook,
    pub dummy3: gpointer,
    pub finalize_hook: GHookFinalizeFunc,
    pub dummy: [gpointer; 2],
}
```

I *think* this is correct on 64 bit Windows but I'm not entirely sure. Can someone (@nirbheek) print `sizeof(GHookList)` from C on 32 bit and 64 bit Windows so we can be sure? My expectation is 28 bytes (32 bit Windows, and 32 bit everywhere) and 48 bytes (64 bit Windows), and 56 bytes (64 bit everywhere else).

### DIFF
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -491,6 +491,9 @@ fn generate_fields(env: &Env, struct_name: &str, fields: &[Field]) -> (Vec<Strin
     // instead guint64
     let is_gvalue = env.config.library_name == "GObject" && struct_name == "Value";
 
+    // TODO: Workaround for GHookList using bitfields
+    let is_ghooklist = env.config.library_name == "GLib" && struct_name == "HookList";
+
     // TODO: GLib/GObject special cases until we have proper union support in Rust
     let is_gweakref = env.config.library_name == "GObject" && struct_name == "WeakRef";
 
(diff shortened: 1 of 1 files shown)


sdroege:
Related: https://github.com/gtk-rs/gir/issues/465

nirbheek:
> Can someone (@nirbheek) print sizeof(GHookList) from C on 32 bit and 64 bit Windows so we can be sure? My expectation is 28 bytes (32 bit Windows, and 32 bit everywhere) and 48 bytes (64 bit Windows)

Your expectation is correct. I tested this by building a C program with 64-bit MSVC against 64-bit MinGW glib and 32-bit MSVC against 32-bit MinGW glib. To reiterate:

sizeof GHookList on 64-bit is 48
sizeof GHookList on 32-bit is 28

I can test against a glib built with MSVC, but that will take longer.

sdroege:
Great, thanks!

EPashkin:
Thanks, @sdroege

PR #467, opened 2017-09-20
Add dox feature in order to build all docs, whatever the platform

GuillaumeGomez:
Since [rustc/43348](https://github.com/rust-lang/rust/pull/43348) has been merged, we can now do this!

cc @EPashkin @sdroege

### DIFF
--- a/src/codegen/general.rs
+++ b/src/codegen/general.rs
@@ -184,7 +184,8 @@ pub fn version_condition_string(
     match version {
         Some(v) if v > env.config.min_cfg_version => {
             let comment = if commented { "//" } else { "" };
-            Some(format!("{}{}#[cfg({})]", tabs(indent), comment, v.to_cfg()))
+            Some(format!("{}{}#[cfg(any({}, feature = \"dox\"))]",
+                         tabs(indent), comment, v.to_cfg()))
         }
         _ => None,
     }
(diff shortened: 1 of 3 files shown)


EPashkin:
`sys/cargo_toml.rs` don't need changed too with adding feature? Or `dox` internal config and always present?

GuillaumeGomez:
Hum, good question. I thought it was a "provided" feature. Let me check.

Also, we'll need to add into the libs the following line:

```rust
#![cfg_attr(dox, feature(doc_cfg)]
```

So that people don't get struggle while trying to build it in stable.

QuietMisdreavus:
Scratch that on the `cfg_attr`. You're not using the thing that's behind that feature gate (the thing introduced in that PR linked at the top), so it's not necessary to add.

GuillaumeGomez:
@gkoz: So you were right, we'll need to add a new feature:

```toml
dox = []
```

It's completely unrelated to the other PR, we can just play with this. No need for `cfg_attr` either. I updated the PR.

EPashkin:
So it ready for merge?

GuillaumeGomez:
Yep, I'll need to update the crates though.

EPashkin:
@GuillaumeGomez thanks

EPashkin:
@GuillaumeGomez sys have problem with this: hook_size_and_setup, GMutex not have 2 definition when I try run `cargo doc --features="dox"`.
Not sure how to fix it :(

EPashkin:
*now have 2 definition

GuillaumeGomez:
Ah. I suppose it's coming from the double target... I didn't think about such a case. Maybe I should add a check to see if a type is defined in two different cfgs?

EPashkin:
I was too lazy and don't check from where this code go.
As it come from string I just add PR to fix it

GuillaumeGomez:
Fine by me.

PR #469, opened 2017-09-28
Update various dependencies to their latest versions

sdroege:


EPashkin:
Bad that appveyor still have `mingw-w64-i686-libgit2-0.24.0-1` instead of `mingw-w64-i686-libgit2-0.26.0-1`

sdroege:
What does that mean? We can't update because appveyor is old? :)

EPashkin:
Yes 😢 
At minimum we need lock git2 to 0.6.6 until appveyour updated.

sdroege:
Ah good old windows preventing progress everywhere else again :) I assume we don't know when or if appveyor will be updated?

EPashkin:
Yes 😭

sdroege:
Ok so what do you suggest? Update everything except for git2?

EPashkin:
Yes, maybe only add comment to git2 about appveyor

sdroege:
Done

EPashkin:
Thanks, just waiting CI

PR #472, opened 2017-09-30
Implement generation of emitter functions for action signals

sdroege:
This is all WIP for now, but it generates valid emitter functions for the GTK bindings at least.

Remaining issues:
- Proper nullability of return values
- Option usage for arguments correctly
- Bounds

All this should ideally come from the same configuration as for the signals, but the types don't match up with the function analysis/code generator that already does the same.

Do you have any suggestions how to do this best without huge changes? Any other review comments already?

cc @EPashkin


sdroege:
In theory the function signature from the trampolines should be possible to re-use here.

sdroege:
Ok, I have it using the trampoline's information now. That should make it complete, except for possibly ugly code. Any suggestions? :)

sdroege:
Works in GTK and gstreamer-app, not sure where else there are action signals.

GuillaumeGomez:
I'll update once merged then. :)

sdroege:
Great, I'm out of battery now :) I'm going to do the other regens (also for new bitflags) once this is merged, and the -sys change.

sdroege:
Don't worry, I did and I am :p

EPashkin:
@sdroege Big thanks.

sdroege:
We could implement signal connectors similarly now, around Closure. No unsafe code or trampolines needed anymore

EPashkin:
Not sure that it possible to signals like https://github.com/gtk-rs/gtk/blob/master/src/signal.rs#L268-L281 and others in this file, but it was good if it done to simpler cases.

sdroege:
Uh what is that!? What an awful and Hindi g unfriendly API :) yes that won't be handled because G_TYPE_POINTER. That always needs manual bindings

PR #481, opened 2017-11-04
Allow marking enums and bitflags as #[must_use]

sdroege:
Useful for error/status types that the user must handle to not ignore
errors.


sdroege:
See https://github.com/sdroege/gstreamer-rs/issues/44 for a potential use

EPashkin:
Thanks, @sdroege

Issue #480, opened 2017-10-28
Represent enums as constants instead of enums


sdroege:
From the bindgen 0.31.0 release notes

> C/C++ enums are now translated into constants by default, rather than Rust enums. The old behavior was a big footgun because rustc assumes that the only values of an enum are its variants, whereas a lot of C/C++ code uses random values as enums. Put these two things and it leads to undefined behavior. Translating C/C++ enums into Rust enums is still available with the --rustified-enum <regex> CLI flag and bindgen::Builder::rustified_enum("<regex>") builder method. https://github.com/rust-lang-nursery/rust-bindgen/issues/758

EPashkin:
@sdroege  You don't understand, how it solves problem with function accepts enum values? Now it accept any int?

sdroege:
Basically what has to be done to make this safe is that the sys bindings use integers and constants. And the non-sys bindings use enums with the `Unknown` variant we have already, and convert from/to integers with the conversion traits.

The important part is that we should never have an out of range value anywhere, currently we can at the sys layer.

We should try to fix that before the next release. We already have the constants in the sys bindings, just the enums would go away.

EPashkin:
If it only for sys, then IMHO this good, while slightly loss in declarability it removes UB

sdroege:
OK, I'll prepare a PR :)

EPashkin:
As pangocairo don't contains enums, seems gtk last PR.

PR #482, opened 2017-11-04
Generate enums in -sys crates as integer constants

sdroege:
The C code might be extended by additional values without the bindings
being updated, which then is considered undefined behaviour on the Rust
side. As such we must use plain integers here.

The enums stay as enums in the non-sys crates and the conversion traits
ensure that the conversion from the integers to the actual enum values
is done correctly, while storing unknown values in a hidden enum
variant.

Fixes https://github.com/gtk-rs/gir/issues/480

### DIFF
--- a/src/codegen/enums.rs
+++ b/src/codegen/enums.rs
@@ -75,7 +75,6 @@ pub fn generate(env: &Env, root_path: &Path, mod_rs: &mut Vec<String>) {
             }
             try!(writeln!(w, "use glib::translate::*;"));
         }
-        try!(writeln!(w, "use std;"));
         try!(writeln!(w, ""));
 
         if has_any {
@@ -166,7 +165,7 @@ impl ToGlib for {name} {{
     }
     try!(writeln!(
         w,
-        "\t\t\t{}::__Unknown(value) => unsafe{{std::mem::transmute(value)}}",
+        "\t\t\t{}::__Unknown(value) => value",
         enum_.name
     ));
     try!(writeln!(
@@ -190,7 +189,7 @@ impl ToGlib for {name} {{
         "#[doc(hidden)]
 impl FromGlib<ffi::{ffi_name}> for {name} {{
     fn from_glib(value: ffi::{ffi_name}) -> Self {{
-        {assert}match value as i32 {{",
+        {assert}match value {{",
         name = enum_.name,
         ffi_name = enum_.c_type,
         assert = assert
@@ -231,7 +230,7 @@ impl FromGlib<ffi::{ffi_name}> for {name} {{
     }}
 
     fn code(self) -> i32 {{
-        self.to_glib() as i32
+        self.to_glib()
     }}
 
     fn from(code: i32) -> Option<Self> {{
@@ -302,11 +301,10 @@ impl FromGlib<ffi::{ffi_name}> for {name} {{
             w,
             "impl<'a> FromValue<'a> for {name} {{
     unsafe fn from_value(value: &Value) -> Self {{
-        from_glib(std::mem::transmute::<i32, ffi::{ffi_name}>(gobject_ffi::g_value_get_enum(value.to_glib_none().0)))
+        from_glib(gobject_ffi::g_value_get_enum(value.to_glib_none().0))
     }}
 }}",
             name = enum_.name,
-            ffi_name = enum_.c_type,
         ));
         try!(writeln!(w, ""));
 
@@ -315,7 +313,7 @@ impl FromGlib<ffi::{ffi_name}> for {name} {{
             w,
             "impl SetValue for {name} {{
     unsafe fn set_value(value: &mut Value, this: &Self) {{
-        gobject_ffi::g_value_set_enum(value.to_glib_none_mut().0, this.to_glib() as i32)
+        gobject_ffi::g_value_set_enum(value.to_glib_none_mut().0, this.to_glib())
     }}
 }}",
             name = enum_.name,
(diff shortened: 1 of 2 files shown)


sdroege:
The changes in the non-sys codegen are optional, it's just that the transmutes and casts are not needed anymore now.

EPashkin:
@sdroege Gdk will need fix in manual part mostly on `event_subtype!` usage due removal `gdk_sys::GdkEventType`

@GuillaumeGomez What you think?

Changes in sys:
```diff
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
-#[repr(C)]
-pub enum GBookmarkFileError {
-    InvalidUri = 0,
-    InvalidValue = 1,
-    AppNotRegistered = 2,
-    UriNotFound = 3,
-    Read = 4,
-    UnknownEncoding = 5,
-    Write = 6,
-    FileNotFound = 7,
-}
-pub const G_BOOKMARK_FILE_ERROR_INVALID_URI: GBookmarkFileError = GBookmarkFileError::InvalidUri;
-pub const G_BOOKMARK_FILE_ERROR_INVALID_VALUE: GBookmarkFileError = GBookmarkFileError::InvalidValue;
-pub const G_BOOKMARK_FILE_ERROR_APP_NOT_REGISTERED: GBookmarkFileError = GBookmarkFileError::AppNotRegistered;
-pub const G_BOOKMARK_FILE_ERROR_URI_NOT_FOUND: GBookmarkFileError = GBookmarkFileError::UriNotFound;
-pub const G_BOOKMARK_FILE_ERROR_READ: GBookmarkFileError = GBookmarkFileError::Read;
-pub const G_BOOKMARK_FILE_ERROR_UNKNOWN_ENCODING: GBookmarkFileError = GBookmarkFileError::UnknownEncoding;
-pub const G_BOOKMARK_FILE_ERROR_WRITE: GBookmarkFileError = GBookmarkFileError::Write;
-pub const G_BOOKMARK_FILE_ERROR_FILE_NOT_FOUND: GBookmarkFileError = GBookmarkFileError::FileNotFound;
+pub type GBookmarkFileError = c_int;
+pub const G_BOOKMARK_FILE_ERROR_INVALID_URI: GBookmarkFileError = 0;
+pub const G_BOOKMARK_FILE_ERROR_INVALID_VALUE: GBookmarkFileError = 1;
+pub const G_BOOKMARK_FILE_ERROR_APP_NOT_REGISTERED: GBookmarkFileError = 2;
+pub const G_BOOKMARK_FILE_ERROR_URI_NOT_FOUND: GBookmarkFileError = 3;
+pub const G_BOOKMARK_FILE_ERROR_READ: GBookmarkFileError = 4;
+pub const G_BOOKMARK_FILE_ERROR_UNKNOWN_ENCODING: GBookmarkFileError = 5;
+pub const G_BOOKMARK_FILE_ERROR_WRITE: GBookmarkFileError = 6;
+pub const G_BOOKMARK_FILE_ERROR_FILE_NOT_FOUND: GBookmarkFileError = 7;
```

In gdk/src/enums.rs:
```diff
@@ -111,7 +110,7 @@ impl ToGlib for ByteOrder {
         match *self {
             ByteOrder::LsbFirst => ffi::GDK_LSB_FIRST,
             ByteOrder::MsbFirst => ffi::GDK_MSB_FIRST,
-            ByteOrder::__Unknown(value) => unsafe{std::mem::transmute(value)}
+            ByteOrder::__Unknown(value) => value
         }
     }
 }
@@ -120,7 +119,7 @@ impl ToGlib for ByteOrder {
impl FromGlib<ffi::GdkByteOrder> for ByteOrder {
     fn from_glib(value: ffi::GdkByteOrder) -> Self {
         skip_assert_initialized!();
-        match value as i32 {
+        match value {
             0 => ByteOrder::LsbFirst,
             1 => ByteOrder::MsbFirst,
             value => ByteOrder::__Unknown(value),
@@ -142,13 +141,13 @@ impl<'a> FromValueOptional<'a> for ByteOrder {
 
 impl<'a> FromValue<'a> for ByteOrder {
     unsafe fn from_value(value: &Value) -> Self {
-        from_glib(std::mem::transmute::<i32, ffi::GdkByteOrder>(gobject_ffi::g_value_get_enum(value.to_glib_none().0)))
+        from_glib(gobject_ffi::g_value_get_enum(value.to_glib_none().0))
     }
 }
 
 impl SetValue for ByteOrder {
     unsafe fn set_value(value: &mut Value, this: &Self) {
-        gobject_ffi::g_value_set_enum(value.to_glib_none_mut().0, this.to_glib() as i32)
+        gobject_ffi::g_value_set_enum(value.to_glib_none_mut().0, this.to_glib())
     }
 }
```

GuillaumeGomez:
I don't like this. :-/ The point with an enum was that we were kinda able to *force* users to use one specific enum as an argument. Now you can just send an `i32` and here we go.

sdroege:
That's still the case with the proper Rust enums, only in the -sys crate their are plain integers now. And you can't get around that without going into undefined behaviour, see #480 and the linked bindgen change.

GuillaumeGomez:
Oh, I see. Sow we have non-sys crate having constant instead of direct values now, right?

sdroege:
The -sys crate only has `c_int` constants (it had constants before that mapped to the enum values). The non-sys crate has the same enums as before, and the conversion traits map to the constants (as before, just that the constants are now `c_int` instead of enums).

Nothing changes for the user of the non-sys crates, and for users of the -sys crates *usually* nothing changes (as long as the constants were used before already).

GuillaumeGomez:
Then I'm all good with this change. My bad for the misunderstanding.

EPashkin:
@sdroege  Thanks

PR #487, opened 2017-11-10
Generate Rust wrappers for class structs too

sdroege:


sdroege:
Next step: generating class methods for them. Which will mean adding a FooClassExt trait, and an IsAClass trait for class inheritance

sdroege:
More work needed here also, just comment on it already if you have any comments

vhdirk:
Picking this up again in #604

PR #490, opened 2017-11-12
Add async method code generation

antoyo:
For now, it's usable (tested with gio, gtk and sourceview), but we could improve the heuristic as mentionned in the TODO comment.

The other two things that will be implemented in next PRs will be the annotation to specify what is the finish function when it cannot be guessed and writing Unimplemented next to Cancellable when this type does not exist (i.e. when the user forgets to add it to `Gir.toml`).

I believe some code should be improved, especially the trampoline code generation.
If you have any other improvements, I'd be glad to hear them.

### DIFF
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -116,19 +116,17 @@ impl Bounds {
         if !par.instance_parameter && par.direction != ParameterDirection::Out {
             if let Some(bound_type) = Bounds::type_for(env, par.typ, par.nullable) {
                 ret = Some(Bounds::get_to_glib_extra(&bound_type));
-                let mut param_trait_bound = String::new();
                 if async && par.name == "callback" {
                     let finish_func_name = replace_async_by_finish(&func.name);
                     if let Some(function) = find_function(env, &finish_func_name) {
                         let out_parameters = find_out_parameters(env, &function);
                         let parameters = format_out_parameters(&out_parameters);
                         let error_type = find_error_type(env, &function);
-                        type_string = format!("Fn(Result<{}, {}>)", parameters, error_type);
-                        param_trait_bound = " + Send + Sync".to_string();
+                        type_string = format!("Fn(Result<{}, {}>) + Send + Sync + 'static", parameters, error_type);
                         trampoline_type = Some(type_string.clone());
                     }
                 }
-                if !self.add_parameter(&par.name, &format!("{}{}", type_string, param_trait_bound), bound_type, async) {
+                if !self.add_parameter(&par.name, &type_string, bound_type, async) {
                     panic!(
                         "Too many type constraints for {}",
                         func.c_identifier.as_ref().unwrap()
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -5,13 +5,15 @@ use std::vec::Vec;
 use consts::TYPE_PARAMETERS_START;
 use env::Env;
 use analysis::imports::Imports;
-use analysis::function_parameters::CParameter;
-use analysis::rust_type::bounds_rust_type;
+use analysis::functions::{find_function, find_index_to_ignore, replace_async_by_finish};
+use analysis::function_parameters::{CParameter, async_param_to_remove};
+use analysis::rust_type::{bounds_rust_type, rust_type};
 use library::{Function, Fundamental, Nullable, ParameterDirection, Type, TypeId};
 use traits::IntoString;
 
 #[derive(Clone, Eq, Debug, PartialEq)]
 pub enum BoundType {
+    NoWrapper,
     // lifetime
     IsA(Option<char>),
     // lifetime <- shouldn't be used but just in case...
@@ -30,6 +32,7 @@ impl BoundType {
 
     fn with_lifetime(ty_: BoundType, lifetime: char) -> BoundType {
         match ty_ {
+            BoundType::NoWrapper => unreachable!(),
             BoundType::IsA(_) => BoundType::IsA(Some(lifetime)),
             BoundType::AsRef(_) => BoundType::AsRef(Some(lifetime)),
             BoundType::Into(_, x) => BoundType::Into(Some(lifetime), x),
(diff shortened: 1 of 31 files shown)


antoyo:
cc @GuillaumeGomez @EPashkin @sdroege

src/analysis/bounds.rs
```diff
@@ -99,21 +102,42 @@ impl Bounds {
         env: &Env,
         func: &Function,
         par: &CParameter,
-    ) -> Option<String> {
+        async: bool,
+    ) -> (Option<String>, Option<String>) {
+        let type_name = bounds_rust_type(env, par.typ);
+        let mut type_string =
+            if async {
```
GuillaumeGomez: I suppose you could write it like this:

```
if async && async_param_to_remove(&par.name) {
    return (None, None);
} else {
    type_name.into_string()
}
```

src/analysis/bounds.rs
```diff
@@ -149,7 +173,23 @@ impl Bounds {
...
+                if let Some(alias) = self.unused.pop_front() {
+                    self.used.push(Bound {
+                        bound_type: BoundType::NoWrapper,
+                        parameter_name: name.to_owned(),
+                        alias,
+                        type_str: type_str.to_string(),
+                        info_for_next_type: false,
+                    });
+                    return true;
+                } else {
```
GuillaumeGomez: This `else` condition is unnecessary since the `if` returns.

src/analysis/bounds.rs
```diff
@@ -196,18 +236,20 @@ impl Bounds {
     pub fn get_parameter_alias_info(&self, name: &str) -> Option<(char, BoundType)> {
         self.used
             .iter()
-            .find(move |n| if n.parameter_name == name {
+            .find(move |n| {
+                if n.parameter_name == name {
                 !n.info_for_next_type
```
GuillaumeGomez: The indent is broken in here. :p

src/analysis/function_parameters.rs
```diff
@@ -161,9 +166,15 @@ pub fn analyze(
         let mut add_rust_parameter = match par.direction {
             library::ParameterDirection::In | library::ParameterDirection::InOut => true,
             library::ParameterDirection::Return => false,
-            library::ParameterDirection::Out => !can_as_return(env, par),
+            library::ParameterDirection::Out => !can_as_return(env, par) && !async_func,
         };
 
+        if async_func {
```
GuillaumeGomez:
```
if async_func && async_param_to_remove(&par.name) {
    add_rust_parameter = false;
}
```

src/analysis/rust_type.rs
```diff
@@ -241,11 +241,13 @@ pub fn parameter_rust_type(
     direction: library::ParameterDirection,
     nullable: Nullable,
     ref_mode: RefMode,
+    async: bool,
 ) -> Result {
     use library::Type::*;
     let type_ = env.library.type_(type_id);
     let rust_type = rust_type_full(env, type_id, nullable, ref_mode);
     match *type_ {
+        Fundamental(super::library::Fundamental::Pointer) if async => Ok("TODO4".to_string()), // FIXME: remove
```
GuillaumeGomez: Wut?

GuillaumeGomez:
What I'm seeing for the moment seems correct (even if it seems incomplete). However I'm not the maintainer in here hehe.

sdroege:
Can you create a PR for gio with the changes this would add, i.e. all the implementations for the async methods in GFile?

src/codegen/function_body_chunk.rs
```diff
@@ -128,12 +120,100 @@ impl Builder {
...
+                    "_source_object as *mut _, "
+                } else {
+                    ""
+                };
+            let output_vars = trampoline.output_params.iter()
+                .filter(|param| param.direction == ParameterDirection::Out && param.name != "error")
+                .map(|param| (param, type_mem_mode(env, param)))
+                .map(|(param, mode)| format!("let mut {} = {};\n", param.name, mode_to_string(mode)))
+                .collect::<Vec<_>>()
+                .join("                ");
```
EPashkin: Please change it to "\t\t\t\t"

src/codegen/function_body_chunk.rs
```diff
@@ -128,12 +120,100 @@ impl Builder {
...
+            unsafe {{
+                let mut error = ptr::null_mut();
+                {output_vars}
+                let _ = ffi::{finish_func_name}({object_arg} res, {output_args} &mut error);
+                let result =
+                    if error.is_null() {{
+                        Ok(({result}))
+                    }} else {{
+                        Err(from_glib_full(error))
+                    }};
```
EPashkin: IMHO `let result = ` better be oneliner.

src/codegen/function_body_chunk.rs
```diff
@@ -431,3 +511,75 @@ impl Builder {
...
+}
+
+fn crate_name(name: &str, env: &Env) -> String {
+    let id = env.library.find_namespace(name).expect("namespace from crate name");
+    let namespace = env.library.namespace(id);
+    let name = nameutil::crate_name(&namespace.name);
+    if id == namespaces::MAIN {
+        return "ffi".to_string();
+    }
+    format!("{}_ffi", name)
```
EPashkin: `Namespace.ffi_crate_name` is not same?

EPashkin:
@antoyo Overall is good idea, I slightly not like "internal trampoline" but it not so big problem. Thanks.

src/analysis/rust_type.rs
antoyo: Yeah, some TODO that are not generated.
I'll have to figure out why.

antoyo:
@sdroege I'll do it when I get back to Canada, so maybe in one week.
@EPashkin I didn't know how to do it (not sure if I can reuse the trampoline logic), so I did it this way because I wanted to have something working during the HackFest.
But the overall trampoline code generation should be redone before we merge this PR.

Thanks to all of you for your comments.

EPashkin:
@antoyo don't worry about trampolines.

antoyo:
@sdroege Here's the [PR](https://github.com/gtk-rs/gio/pull/62).

src/codegen/function_body_chunk.rs
antoyo: I'll instead change that to use `Chunk` directly, to avoid having code as strings.

src/codegen/function_body_chunk.rs
antoyo: `ffi_crate_name` is only available on the other `Namespace` struct.

antoyo:
I've fixed all the issues you mentioned and I've fixed the async trampoline code generation.
I think this PR is ready to be tested.
Do we have any infrastructure to test `gir` against every GNOME library that we support?

EPashkin:
@antoyo No we don't have :(
I tested building of examples on updated gtk, it builds fine.

EPashkin:
@GuillaumeGomez we add async function it this release?

sdroege:
I added some comments about the generated code here https://github.com/gtk-rs/gio/pull/63

sdroege:
IMHO this is good to go, judging from the generated code.

GuillaumeGomez:
Seems good to me as well. Is it good for you @EPashkin or do you want to add something else @antoyo?

EPashkin:
@GuillaumeGomez Its good for me too.
You wan't add this before release?

GuillaumeGomez:
Yes if possible so I'll handle the release tomorrow as the latest (won't be able to do it before next week-end otherwise).

antoyo:
@sdroege @GuillaumeGomez I wanted to fix the `Cancellable` issue mentionned by @sdroege.
I think I can do it today.

antoyo:
This will require to add `Gio.Cancellable` in `Gir.toml` files when needed.
I'll commit the fix soon.

EPashkin:
Gtk-rs examples still builds.
@sdroege, @antoyo there still any problem? Or I can merge?

GuillaumeGomez:
For me it was already good so as you wish!

antoyo:
It's ready to be merged.
Just waiting for @sdroege to see the latest change (`&&Fn`).

sdroege:
See my comment there

antoyo:
I've fixed this issue in the latest commit.

antoyo:
@sdroege I've fixed the other issues mentionned in the gio PR.

EPashkin:
@antoyo Thanks.
I add regen PR for sourceview. Gtk, glib has only commented changes and IMHO can be ignored.

EPashkin:
Actually gtk also have async functions in `IconInfo`, but current version not works good outside `gio` as it don't add `use gio_ffi`

EPashkin:
Created fix, will add regen for Gtk too.

sdroege:
Looks good to me, just need to add support for the "progress-callback" pattern in the future, as used by various async operations

PR #496, opened 2017-11-26
Get rid of some unneeded transmutes and other special cases for enum/…

sdroege:
…flags properties

They implement the Value traits now, no special handling needed here.

### DIFF
--- a/src/codegen/child_properties.rs
+++ b/src/codegen/child_properties.rs
@@ -54,7 +54,7 @@ fn generate_func(
     let pub_prefix = if in_trait { "" } else { "pub " };
     let decl_suffix = if only_declaration { ";" } else { " {" };
     let type_string = rust_type(env, prop.typ);
-    let comment_prefix = if type_string.is_err() || (prop.default_value.is_none() && is_get) {
+    let comment_prefix = if type_string.is_err() {
         "//"
     } else {
         ""
@@ -75,7 +75,7 @@ fn generate_func(
     ));
 
     if !only_declaration {
-        let body = body(prop, is_get).to_code(env);
+        let body = body(env, prop, is_get).to_code(env);
         for s in body {
             try!(writeln!(w, "{}{}{}", tabs(indent), comment_prefix, s));
         }
@@ -119,7 +119,7 @@ fn declaration(env: &Env, prop: &ChildProperty, is_get: bool) -> String {
     )
 }
 
-fn body(prop: &ChildProperty, is_get: bool) -> Chunk {
+fn body(env: &Env, prop: &ChildProperty, is_get: bool) -> Chunk {
     let mut builder = property_body::Builder::new_for_child_property();
     let prop_name = nameutil::signal_to_snake(&*prop.name);
     builder
@@ -129,10 +129,11 @@ fn body(prop: &ChildProperty, is_get: bool) -> Chunk {
         .is_ref(prop.set_in_ref_mode.is_ref())
         .is_nullable(*prop.nullable)
         .is_into(prop.is_into);
-    if let Some(ref default_value) = prop.default_value {
-        builder.default_value(default_value);
+
+    if let Ok(type_) = rust_type(env, prop.typ) {
+        builder.type_(&type_);
     } else {
-        builder.default_value("/*Unknown default value*/");
+        builder.type_("/*Unknown type */");
     }
 
     builder.generate()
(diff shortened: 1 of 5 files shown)


EPashkin:
This [runtime](https://github.com/EPashkin/gtk-rs-runtime-tests/blob/master/tests/child_properties.rs#L25-L28) test is failing (at minimum on Windows)

EPashkin:
But setters seems work, at minimum change https://github.com/gtk-rs/examples/blob/master/src/bin/child-properties.rs#L48 give effect.

sdroege:
How does the child-properties example fail for you? It works just fine here on Linux
If it fails, that would suggest that the `Value` traits implementations for enums is broken generally on Windows.

sdroege:
Ah, not that example fails for you but when using the getters. Yes, I can confirm that!

sdroege:
The returned `Value` is of type `gint` instead of `GtkPackType`.

sdroege:
The problem is https://github.com/gtk-rs/gir/blob/master/src/codegen/property_body.rs#L123 . This is wrong even before my change, and if you compile GObject with further runtime checks it will also cause errors.

This should instead use something like
```
        let mut value = Value::uninitialized();
        gobject_ffi::g_value_init(value.to_glib_none_mut().0, gtk::PackType::static_type().to_glib());
```

EPashkin:
Agree, and then calculating `default_value` need removed too.
Not sure that it can be normally generated :(

sdroege:
It can, yes

sdroege:
Fixed, related to the `Vec<String>` special case here there's https://github.com/gtk-rs/glib/pull/264

sdroege:
It is however *not* a dependency for this PR.

EPashkin:
@sdroege At first glance it looks fine, I check at evening.

sdroege:
Thanks!

src/codegen/properties.rs
```diff
@@ -127,19 +127,19 @@ fn declaration(env: &Env, prop: &Property) -> String {
...
-        .conversion(prop.conversion);
-    if let Some(ref default_value) = prop.default_value {
-        builder.default_value(default_value);
+        .is_nullable(*prop.nullable);
+
+    if let Ok(type_) = rust_type(env, prop.typ) {
+        builder.type_(&type_);
     } else {
-        builder.default_value("/*Unknown default value*/");
+        builder.type_("/*Unknown type */");
```
EPashkin: IMHO unneeded space at end of string.

src/codegen/child_properties.rs
```diff
@@ -128,12 +128,12 @@ fn body(prop: &ChildProperty, is_get: bool) -> Chunk {
...
-        .conversion(prop.conversion);
-    if let Some(ref default_value) = prop.default_value {
-        builder.default_value(default_value);
+        .is_into(prop.is_into);
+
+    if let Ok(type_) = rust_type(env, prop.typ) {
+        builder.type_(&type_);
     } else {
-        builder.default_value("/*Unknown default value*/");
+        builder.type_("/*Unknown type */");
```
EPashkin: Extra space here too.

EPashkin:
@sdroege Runtime test now passed, only small nit last.

EPashkin:
Hm, there many `Value::uninitialized/g_value_init` maybe we need add `Value::withType` at some time

src/codegen/child_properties.rs
sdroege: Removed the two spaces :)

sdroege:
> Hm, there many Value::uninitialized/g_value_init maybe we need add Value::with_type at some time

That could make sense, but I'd prefer to wait with that until after everything is re-gen'd. This here should ideally go into the next bugfix release as it has the potential to blow up things.

EPashkin:
@sdroege Thanks

EPashkin:
By way, I tried to use `ObjectExt::get_property` in `get_property_xxx`.
Strange that it need  `glib::value::SetValue` constraint on impl,
and as it always check property existence,
seems current version is better.

sdroege:
We could probably get rid of that trait bound somehow, and also add an unsafe variant that does not check for the property's existence... But there's not much difference in the code overall then. Unlike with signal connectors :)

sdroege:
I'll send regent PRs for everything later unless someone else wants too

EPashkin:
About `SetValue`: as this constraint seems need only for one function we can extract it to separate trait.

I have free time so I add PR for gtk at minimum.

sdroege:
Thanks, all looking good to me

PR #497, opened 2017-11-26
Fix invalid [c_void] debug implementation

GuillaumeGomez:
cc @sdroege @EPashkin


EPashkin:
It has no side effect on gtk-rs, so I leave all decision to @sdroege

sdroege:
Works for me, thanks 👍

EPashkin:
@GuillaumeGomez thanks.

PR #498, opened 2017-11-28
Use new Value::from_type() function for property getters

sdroege:
See https://github.com/gtk-rs/glib/pull/266

### DIFF
--- a/src/analysis/properties.rs
+++ b/src/analysis/properties.rs
@@ -92,6 +92,7 @@ pub fn analyze(
                 imports.add_used_type(s, prop.version);
             }
             if type_string.is_ok() {
+                imports.add("gobject_ffi", prop.version);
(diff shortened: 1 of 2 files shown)


EPashkin:
Please move https://github.com/gtk-rs/gir/blob/master/src/analysis/properties.rs#L83 to setter branch.
This don't change generated code object `glib_wrapper!` won't work without `use gobject_ffi;`,
but will make more clear that getter don't need it.

sdroege:
Why? We need it for the setter *and* getter, or not? For calling `g_object_get_property()`. It should be in both branches, is that what you mean?

EPashkin:
I meant to move it, was wrong that it not used :(
You right that safer add it to setter too.

sdroege:
Done :)

EPashkin:
Thanks

Issue #512, opened 2017-12-15
Allow use custom modules in sys mode


EPashkin:
Currently `sys` module contains only `lib.rs` that always recreated.
This will broke gtk-sys after rebuild as it need manual module `xlib` added in https://github.com/gtk-rs/sys/pull/76.

cc @GuillaumeGomez, @sdroege

EPashkin:
I see at minimum 2 way to fix this:
1. Add array parameter to `Gir.toml` with list of modules,
2. When generating `lib.rs` check folder and add `pub use ` for all found *.rs files.

GuillaumeGomez:
I prefer the most automated way if possible but otherwise, I don't really have an opinion in here...

EPashkin:
Ok, I do 2,
we can add 1 to it, if needed

PR #514, opened 2017-12-16
Add detecting and including custom modules on sys mode

EPashkin:
Fix https://github.com/gtk-rs/gir/issues/512

cc @GuillaumeGomez, @sdroege

### DIFF
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -1,4 +1,5 @@
 use std::collections::HashMap;
+use std::fs;
 use std::io::{Result, Write};
 
 use analysis::c_type::rustify_pointers;
@@ -29,6 +30,7 @@ fn generate_lib(w: &mut Write, env: &Env) -> Result<()> {
     try!(statics::begin(w));
 
     try!(generate_extern_crates(w, env));
+    try!(include_custom_modules(w, env));
     try!(statics::after_extern_crates(w));
 
     if env.config.library_name != "GLib" {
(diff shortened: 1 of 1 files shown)


src/codegen/sys/lib_.rs
```diff
@@ -90,6 +92,36 @@ fn get_extern_crate_string(library: &ExternalLibrary) -> String {
     )
 }
 
+fn include_custom_modules(w: &mut Write, env: &Env) -> Result<()> {
+    let modules = try!(find_modules(env));
+    if !modules.is_empty() {
+        try!(writeln!(w, ""));
+        for module in modules {
+            try!(writeln!(w, "pub mod {};", module));
```
sdroege: Maybe
```
mod foo;
pub use foo::*;
```
instead?

EPashkin: You right, this give better flexibility.
@GuillaumeGomez You agree?

sdroege: Should there also be something to ignore autogenerated pieces then, so they can be manually implemented?

EPashkin: [ignoring](https://github.com/gtk-rs/gir/blob/master/README.md#the-ffi-mode-toml-config) not enough?

sdroege: That's exactly what I thought of. For some reason I didn't notice that it is already implemented for FFI :)

sdroege: Ah but it can't be used for individual functions, or can it?

EPashkin: Can same as in normal version as [code](https://github.com/gtk-rs/gir/blob/master/src/codegen/sys/functions.rs#L191) still here.

GuillaumeGomez: > Maybe
>
> ```
> mod foo;
> pub use foo::*;
> ```
>
> instead?

I agree on it. :)

EPashkin:
Updated.

src/codegen/sys/lib_.rs
```diff
@@ -90,6 +92,40 @@ fn get_extern_crate_string(library: &ExternalLibrary) -> String {
...
+fn find_modules(env: &Env) -> Result<Vec<String>> {
+    let path = env.config.target_path.join("src");
+
+    let mut vec = Vec::<String>::new();
+    for entry in try!(fs::read_dir(path)) {
+        let path = try!(entry).path();
+        let ext = path.extension().unwrap();
+        if ext != "rs" { continue }
+        let file_stem = path.file_stem().unwrap();
+        if file_stem == "lib" { continue; }
```
GuillaumeGomez: `continue;` or `continue`? 😛

src/codegen/sys/lib_.rs
```diff
@@ -90,6 +92,40 @@ fn get_extern_crate_string(library: &ExternalLibrary) -> String {
...
+    Ok(())
+}
+
+fn find_modules(env: &Env) -> Result<Vec<String>> {
+    let path = env.config.target_path.join("src");
+
+    let mut vec = Vec::<String>::new();
+    for entry in try!(fs::read_dir(path)) {
+        let path = try!(entry).path();
+        let ext = path.extension().unwrap();
```
GuillaumeGomez: Also, can you replace all `.unwrap()` with `.expect(...)` please? We should try to keep this as a good habit.

src/codegen/sys/lib_.rs
EPashkin: Thanks, I use `cargo fmt` to this function.

src/codegen/sys/lib_.rs
EPashkin: Updated but IMHO Problem with expect, that it need include "path" to be clear and `.unwrap_or_else(|| panic!(...))` take much space.

EPashkin:
Updated

sdroege:
👍

PR #515, opened 2017-12-17
Generate const_override for objects trait functions that have const p…

EPashkin:
…ointer for self parameter

Part of https://github.com/gtk-rs/glib/pull/249

cc @GuillaumeGomez , @sdroege


GuillaumeGomez:
👍 as well!

PR #518, opened 2018-01-15
Generate deprecated attribute on deprecated functions

### LINKED ISSUE #517: Set functions deprecated when compilation version >= deprecation version
Something like:

```rust
#[cfg_attr(version = "3.20", deprecated)]
```

should work just fine. Doing it in the next days or so. :)

GuillaumeGomez:
Fixes #517.

cc @EPashkin @sdroege

### DIFF
--- a/src/codegen/function.rs
+++ b/src/codegen/function.rs
@@ -7,7 +7,9 @@ use analysis::namespaces;
 use chunk::{ffi_function_todo, Chunk};
 use env::Env;
 use super::function_body_chunk;
-use super::general::{cfg_condition, doc_hidden, not_version_condition, version_condition};
+use super::general::{
+    cfg_condition, cfg_deprecated, doc_hidden, not_version_condition, version_condition,
+};
 use super::parameter::ToParameter;
 use super::return_value::{out_parameters_as_return, ToReturnValue};
 use writer::primitives::tabs;
@@ -49,6 +51,9 @@ pub fn generate(
     let suffix = if only_declaration { ";" } else { " {" };
 
     try!(writeln!(w, ""));
+    if !in_trait || only_declaration {
+        try!(cfg_deprecated(w, env, analysis.deprecated_version, commented, indent));
+    }
     try!(cfg_condition(w, &analysis.cfg_condition, commented, indent));
     try!(version_condition(
         w,
--- a/src/codegen/general.rs
+++ b/src/codegen/general.rs
@@ -177,6 +177,40 @@ pub fn define_shared_type(
     Ok(())
 }
 
+pub fn cfg_deprecated(
+    w: &mut Write,
+    env: &Env,
+    deprecated: Option<Version>,
+    commented: bool,
+    indent: usize,
+) -> Result<()> {
+    if let Some(s) = cfg_deprecated_string(deprecated, env, commented, indent) {
+        try!(writeln!(w, "{}", s));
+    }
+    Ok(())
+}
+
+pub fn cfg_deprecated_string(
+    deprecated: Option<Version>,
+    env: &Env,
+    commented: bool,
+    indent: usize,
+) -> Option<String> {
+    let comment = if commented { "//" } else { "" };
+    if env.is_too_low_version(deprecated) {
+        Some(format!("{}{}#[deprecated]", tabs(indent), comment))
+    } else if let Some(v) = deprecated {
+        Some(format!(
+            "{}{}#[cfg_attr({}, deprecated)]",
+            tabs(indent),
+            comment,
+            v.to_cfg()
+        ))
+    } else {
+        None
+    }
+}
+
 pub fn version_condition(
     w: &mut Write,
     env: &Env,
@@ -291,6 +325,7 @@ pub fn declare_default_from_new(
         !f.visibility.hidden() && f.name == "new" && f.parameters.rust_parameters.is_empty()
     }) {
         try!(writeln!(w, ""));
+        try!(cfg_deprecated(w, env, func.deprecated_version, false, 0));
         try!(version_condition(w, env, func.version, false, 0));
         try!(writeln!(w, "impl Default for {} {{", name));
         try!(writeln!(w, "    fn default() -> Self {{"));
--- a/src/codegen/properties.rs
+++ b/src/codegen/properties.rs
@@ -5,7 +5,7 @@ use analysis::properties::Property;
 use analysis::rust_type::{parameter_rust_type, rust_type};
 use chunk::Chunk;
 use env::Env;
-use super::general::version_condition;
+use super::general::{cfg_deprecated, version_condition};
 use library;
 use writer::primitives::tabs;
 use super::property_body;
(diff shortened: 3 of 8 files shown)


src/codegen/general.rs
```diff
@@ -177,6 +177,36 @@ pub fn define_shared_type(
...
+
+pub fn cfg_deprecated_string(
+    deprecated: Option<Version>,
+    commented: bool,
+    indent: usize,
+) -> Option<String> {
+    if let Some(v) = deprecated {
+        let comment = if commented { "//" } else { "" };
+        Some(format!(
+            "{}{}#[cfg_attr({}, deprecated)]",
```
sdroege: I don't fully understand how this should work. The optional "deprecated-version" attribute in the gir file is for the native library version, so we would have to e.g. generate a `feature = "v2_40"`. Is that what `v.to_cfg()` does?
What for the case where there is no version, we just ignore that case?

I don't see you parsing the deprecated-version field from the gir file, only something from the toml configuration.

sdroege: AFAIU it would work as `#[cfg_attr(feature = "v2_40", deprecated)]`, not just `#[cfg_attr("v2_40", deprecated)]`.

GuillaumeGomez: `to_cfg` generates `feature = "v2_40"`. Also, I added the parse of the deprecated field [here](https://github.com/gtk-rs/gir/pull/518/files#diff-08faf151bfcc6c456f5db4c647591a76R42). But more globally, we already get the deprecated version when parsing the gir files (when I generated for gtk and gio, it got the information). But maybe I'm wrong or misunderstood you?

sdroege: That line is parsing it from the toml config, not from the gir file. You mean the code to parse it from the gir file was already in src/parser.rs before? All good then

GuillaumeGomez: Yes, we already got this information before (I didn't add much `deprecated_version` fields as you can see). :)

sdroege: Lucky you :) All good then

GuillaumeGomez:
Ok ok, I'll update this evening. Thanks for the review!

GuillaumeGomez:
Ok, should be good now. We'll need to add `#![allow(deprecated)]` in gtk-rs non-`sys` crates to avoid the warnings when building them but otherwise that should be all. :)

src/codegen/function.rs
```diff
@@ -49,6 +51,9 @@ pub fn generate(
     let suffix = if only_declaration { ";" } else { " {" };
 
     try!(writeln!(w, ""));
+    if only_declaration {
```
EPashkin: Need be `!in_trait || only_declaration`, or `g_settings_list_relocatable_schemas` will be not deprecated in gio.

src/codegen/general.rs
```diff
@@ -177,6 +177,40 @@ pub fn define_shared_type(
...
+    Ok(())
+}
+
+pub fn cfg_deprecated_string(
+    deprecated: Option<Version>,
+    env: &Env,
+    commented: bool,
+    indent: usize,
+) -> Option<String> {
+    if env.is_totally_deprecated(deprecated) {
```
EPashkin: This functions always return false in gdk, due `options.deprecate_by_min_version=false`.
IMHO better add to `Env` function `is_too_low_version`, and write `#[deprecated]` in this case

src/codegen/properties.rs
```diff
@@ -50,6 +50,9 @@ fn generate_prop_func(
     try!(writeln!(w, ""));
 
     let decl = declaration(env, prop);
+    if only_declaration {
```
EPashkin: Same as other

src/codegen/signal.rs
```diff
@@ -29,6 +29,9 @@ pub fn generate(
     let suffix = if only_declaration { ";" } else { " {" };
 
     try!(writeln!(w, ""));
+    if only_declaration {
```
EPashkin: This too

EPashkin:
Agree about `#![allow(deprecated)]`, then we can even restore deprecation on sys, but IMHO better not do this

GuillaumeGomez:
I prefer not in sys. Let's keep it as "simple" as possible.

GuillaumeGomez:
Updated!

src/codegen/signal.rs
```diff
@@ -73,6 +76,9 @@ pub fn generate(
 
     if let Some(ref emit_name) = analysis.action_emit_name {
         try!(writeln!(w, ""));
+        if only_declaration {
```
EPashkin: Maybe this also need "fixed"

src/codegen/general.rs
```diff
@@ -177,6 +177,40 @@ pub fn define_shared_type(
...
+}
+
+pub fn cfg_deprecated_string(
+    deprecated: Option<Version>,
+    env: &Env,
+    commented: bool,
+    indent: usize,
+) -> Option<String> {
+    if env.is_too_low_version(deprecated) {
+        None
```
EPashkin: Why not return `#[deprecated]` here?

EPashkin:
Code looks good, only last final condition fix,  README.md update for `[[object.member]]`
and maybe "#[deprecated]" generation

GuillaumeGomez:
About the `#[deprecated]` generation, don't you to just keep it outside of the automatic generation? Having it at the top of the crates seems better, no? Instead of having it repeated in every file having some deprecated content generated.

EPashkin:
@GuillaumeGomez You seems mistake with `#![allow(deprecated)]`: agree that it will be added manually.
But if function, for example, deprecated in v3.0, then we better add `#[deprecated]` for it, if `min_cfg_version = "3.4"`

GuillaumeGomez:
Aaaaaaaah! Ok now I got it.

EPashkin:
`analysis::trampolines::Trampoline` don't use `deprecated_version` field and it can be safely removed.

GuillaumeGomez:
Updated!

src/codegen/general.rs
```diff
@@ -177,6 +177,40 @@ pub fn define_shared_type(
...
+}
+
+pub fn cfg_deprecated_string(
+    deprecated: Option<Version>,
+    env: &Env,
+    commented: bool,
+    indent: usize,
+) -> Option<String> {
+    if env.is_too_low_version(deprecated) {
+        Some("#[deprecated]".to_owned())
```
EPashkin: Comment info and indent will be lost

GuillaumeGomez: Sorry, quick update on the way!

GuillaumeGomez:
Added missing info and indent.

EPashkin:
Now it ready from my point of view.
@sdroege How it for you?

sdroege:
Seems fine to me but I'd like to look at the output of GIR for e.g. gtk. Is there are PR already?

EPashkin:
As I know there no PR for gtk with this :(

EPashkin:
@sdroege https://github.com/EPashkin/gtk/tree/deprecate with gtk changes

sdroege:
Looks good to me. But do deprecated attributes work in traits? Also are deprecated flags/enum members handled?

EPashkin:
At minimum we know now that deprecated functions only allowed in direct impl and in trait, but not in trait implementation.
I see many warnings about flags and enums, but don't looked at it.

GuillaumeGomez:
Deprecated attributes work in traits *declaration*. For what I tested, they work in enums, didn't check for flags.

EPashkin:
Checked that deprecation works for enum members and trait functions.

EPashkin:
Ok, I merge this.
@GuillaumeGomez Thanks

Issue #511, opened 2017-12-15
Fix property ignoring


EPashkin:
`gio::PermissionExt` have both getters and property getters for some properties:
`allowed`, `can-acquire`, `can_release`.
Same with other objects in gio.
Gir normally don't generate property, if getter present.

EPashkin:
Fixed.

PR #524, opened 2018-01-27
Fix property parent check

EPashkin:
Part of https://github.com/gtk-rs/gtk/issues/608, #511

cc @GuillaumeGomez, @sdroege

### DIFF
--- a/src/analysis/properties.rs
+++ b/src/analysis/properties.rs
@@ -126,7 +126,8 @@ fn analyze_property(
         .iter()
         .filter_map(|f| f.version)
         .min()
-        .or(prop.version);
+        .or(prop.version)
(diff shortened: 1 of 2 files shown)


EPashkin:
It only generated new property "alignment" for `GtkCellRendererText`
No other crates affected

EPashkin:
Alternative solution: add `force=true` to property config, that disable dublicate check for it.

EPashkin:
Second commit fix adding properties when same function already present in class.

Extra crates affected: gdk, sourceview

GuillaumeGomez:
Looks good to me.

sdroege:
👍 but I didn't check if this has any effect on GStreamer. I'll create an issue if it does, will check in the next days

PR #525, opened 2018-01-27
Remove unnecessary parentheses in async trampoline result

EPashkin:
cc @GuillaumeGomez, @sdroege 

Removes warning "unnecessary parentheses around function argument"
which appeared in nightly for gtk and gio.

```diff
--- a/src/auto/icon_info.rs
+++ b/src/auto/icon_info.rs
@@ -214,7 +214,7 @@ impl<O: IsA<IconInfo>> IconInfoExt for O {
             let mut error = ptr::null_mut();
             let mut was_symbolic = mem::uninitialized();
             let _ = ffi::gtk_icon_info_load_symbolic_finish(_source_object as *mut _, res, &mut was_symbolic, &mut error);
-            let result = if error.is_null() { Ok((from_glib(was_symbolic))) } else { Err(from_glib_full(error)) };
+            let result = if error.is_null() { Ok(from_glib(was_symbolic)) } else { Err(from_glib_full(error)) };
             let callback: Box<Box<T>> = Box::from_raw(user_data as *mut _);
             callback(result);
         }
```

### DIFF
--- a/src/chunk/chunk.rs
+++ b/src/chunk/chunk.rs
@@ -86,5 +86,7 @@ pub fn chunks(ch: Chunk) -> Vec<Chunk> {
 pub enum TupleMode {
     Auto, // "", "1", "(1,2)"
     WithUnit, // "()", "1", "(1,2)"
+    #[deprecated]
+    #[allow(dead_code)]
     Simple,    // "()", "(1)", "(1,2)"
 }
--- a/src/codegen/function_body_chunk.rs
+++ b/src/codegen/function_body_chunk.rs
@@ -159,7 +159,7 @@ impl Builder {
                     }
                 }
             }).collect();
-        let result = Chunk::Tuple(result, TupleMode::Simple);
+        let result = Chunk::Tuple(result, TupleMode::WithUnit);
         let gio_crate_name = crate_name("Gio", env);
         let gobject_crate_name = crate_name("GObject", env);
         let glib_crate_name = crate_name("GLib", env);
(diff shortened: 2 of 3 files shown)


src/codegen/function_body_chunk.rs
```diff
@@ -159,7 +159,7 @@ impl Builder {
                     }
                 }
             }).collect();
-        let result = Chunk::Tuple(result, TupleMode::Simple);
+        let result = Chunk::Tuple(result, TupleMode::WithUnit);
```
EPashkin: @antoyo Maybe you have some other reason to uncomment and use `TupleMode::Simple`

GuillaumeGomez:
👍

src/codegen/function_body_chunk.rs
antoyo: That should be okay.

sdroege:
👍

src/chunk/chunk.rs
```diff
@@ -86,5 +86,7 @@ pub fn chunks(ch: Chunk) -> Vec<Chunk> {
 pub enum TupleMode {
     Auto, // "", "1", "(1,2)"
     WithUnit, // "()", "1", "(1,2)"
+    #[deprecated]
+    #[allow(dead_code)]
     Simple,    // "()", "(1)", "(1,2)"
```
sdroege: Why don't you remove it yet? Is it still in use elsewhere?

EPashkin: It was commented before @antoyo decommented it when adding asyncs,
so I decided on my whim that better not comment it but deprecate 😉

PR #526, opened 2018-01-28
 Fill missing c:type in fields in postprocessing phase.

tmiasko:
Reduce field related logic in code generation phase by moving it to library
postprocessing phase. ~~This PR also includes sockaddr which allows to
successfully generate bindings for Soup (which didn't work correctly
previously).~~

Tested with bindings hosted under gtk-rs in FFI and API mode.

### DIFF
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -660,12 +660,11 @@ fn generate_fields(env: &Env, struct_name: &str, fields: &[Field]) -> (Vec<Strin
                             || union_field.name.contains("padding")
                         {
                             if let Some(ref c_type) = union_field.c_type {
-                                let name = mangle_keywords(&*union_field.name);
                                 let c_type = ffi_type(env, union_field.typ, c_type);
                                 if c_type.is_err() {
                                     commented = true;
                                 }
-                                lines.push(format!("\tpub {}: {},", name, c_type.into_string()));
+                                lines.push(format!("\tpub {}: {},", &union_field.name, c_type.into_string()));
                                 continue 'fields;
                             }
                         }
@@ -728,7 +727,6 @@ fn generate_fields(env: &Env, struct_name: &str, fields: &[Field]) -> (Vec<Strin
             // union containing a single pointer
             lines.push("\tpub priv_: gpointer,".to_owned());
         } else if let Some(ref c_type) = field.c_type {
-            let name = mangle_keywords(&*field.name);
             let mut c_type = ffi_type(env, field.typ, c_type);
             if c_type.is_err() {
                 commented = true;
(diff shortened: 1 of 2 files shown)


EPashkin:
LGFM
Thanks, @tmiasko.
cc @GuillaumeGomez, @sdroege just in case

src/codegen/sys/lib_.rs
```diff
@@ -660,12 +660,11 @@ fn generate_fields(env: &Env, struct_name: &str, fields: &[Field]) -> (Vec<Strin
                             || union_field.name.contains("padding")
                         {
                             if let Some(ref c_type) = union_field.c_type {
-                                let name = mangle_keywords(&*union_field.name);
                                 let c_type = ffi_type(env, union_field.typ, c_type);
                                 if c_type.is_err() {
                                     commented = true;
                                 }
-                                lines.push(format!("\tpub {}: {},", name, c_type.into_string()));
+                                lines.push(format!("\tpub {}: {},", &union_field.name, c_type.into_string()));
```
GuillaumeGomez: We don't need the mangling anymore?

src/codegen/sys/lib_.rs
```diff
@@ -728,41 +727,27 @@ fn generate_fields(env: &Env, struct_name: &str, fields: &[Field]) -> (Vec<Strin
...
-            let name = mangle_keywords(&*field.name);
             let mut c_type = ffi_type(env, field.typ, c_type);
             if c_type.is_err() {
                 commented = true;
             }
             if !cfg!(feature = "use_unions") && is_gvalue && field.name == "data" {
                 c_type = Ok("[u64; 2]".to_owned());
             }
-            lines.push(format!("\tpub {}: {},", name, c_type.into_string()));
+            lines.push(format!("\tpub {}: {},", &field.name, c_type.into_string()));
```
GuillaumeGomez: Same question.

src/library_postprocessing.rs
```diff
@@ -268,4 +271,70 @@ impl Library {
...
+                            fields[fid].c_type = Some(c_type);
+                        }
+                        Action::SetName(name) => fields[fid].name = name,
+                    }
+                },
+                _ => unreachable!("Expected class, record or union"),
+            }
+        }
+    }
+
```
GuillaumeGomez: Extra empty line.

src/codegen/sys/lib_.rs
tmiasko: Now mangling is done once in library postprocessing phase, instead of doing it
in every single place where field name is used. This approach doesn't retain
the original field name, but I don't think those will be useful either way.
Especially, if they had name like "priv".

EPashkin:
I merge this.
@tmiasko Thanks again

PR #530, opened 2018-02-02
Allow non_snake_case to silence warnings in generated code.

tmiasko:
Useful especially in the case of WebKit crates.


EPashkin:
Thanks

PR #531, opened 2018-02-02
Include default value in to_bool function.

tmiasko:


EPashkin:
@tmiasko  Thanks for this too

PR #537, opened 2018-02-07
Fix use declarations when generating code for glib.

tmiasko:
Fix for https://github.com/gtk-rs/glib/pull/290.

It seems we are missing some layer of abstraction, or general consistency in
generated code. I thinking about all checks of the form:  `* == namespaces::MAIN`.
It shouldn't really be necessary.

### DIFF
--- a/src/codegen/enums.rs
+++ b/src/codegen/enums.rs
@@ -50,12 +50,12 @@ pub fn generate(env: &Env, root_path: &Path, mod_rs: &mut Vec<String>) {
                 try!(writeln!(w, "use error::ErrorDomain;"));
             }
             if has_get_type {
-                try!(writeln!(w, "use Type;"));
-                try!(writeln!(w, "use StaticType;"));
-                try!(writeln!(w, "use Value;"));
-                try!(writeln!(w, "use SetValue;"));
-                try!(writeln!(w, "use FromValue;"));
-                try!(writeln!(w, "use FromValueOptional;"));
+                try!(writeln!(w, "use types::Type;"));
+                try!(writeln!(w, "use types::StaticType;"));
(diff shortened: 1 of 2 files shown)


sdroege:
This is fixing https://github.com/gtk-rs/glib/pull/290 ? Elsewhere we check if the namespace is GLib, and only then do the special paths.

tmiasko:
Yes, this was intended as a fix for https://github.com/gtk-rs/glib/pull/290, as
I have mentioned in the description. Did you have any further issues with it?

Regarding names, I though we could do something like `use crate as glib` or
`use super as glib`, and then use exactly the same qualified name.
Unfortunately this is not yet implemented: https://github.com/rust-lang/rust/issues/37156#issuecomment-362917486

sdroege:
I would prefer if we consistently did the special casing for the glib crate the same way everywhere, apart from that the changes look good to me. Thanks :)

tmiasko:
Could you point me to one such example? When I look for code that solves
analogous problem in codegen, the solution looks like above. So it is not
readily apparent to me what else this is supposed to be consistent with.

sdroege:
```rust
        if env.namespaces.glib_ns_id != namespaces::MAIN {
            return;
        }
```

in imports.rs for example

sdroege:
Ah I missed that the first part of the commit is in such a block, the second also is. Nevermind then

EPashkin:
@tmiasko Thanks.

EPashkin:
@tmiasko About checks for `namespaces::MAIN`:
I don't worry about it. 
Most checks need as we have different logic of "foreign" types,
only exceptions is `is we generating glib?` check that also needed, as it source for most base types.
But if you have some good idea I welcome it.

PR #538, opened 2018-02-08
Refactor imports.

tmiasko:
* Make Imports aware of current crate and names defined in module.
    
  This move responsibility for special handling of names defined within
  the same crate or module inside the Imports struct. Users of Imports
  merely declare what they use.

* Use Imports to generate use declarations in enum / flags codegen.

### DIFF
--- a/src/codegen/flags.rs
+++ b/src/codegen/flags.rs
@@ -1,3 +1,4 @@
+use analysis::imports::Imports;
 use analysis::namespaces;
 use codegen::general::{self, cfg_deprecated, version_condition, version_condition_string};
 use config::gobjects::GObject;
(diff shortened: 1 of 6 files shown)


src/codegen/flags.rs
```diff
@@ -21,42 +22,27 @@ pub fn generate(env: &Env, root_path: &Path, mod_rs: &mut Vec<String>) {
...
-                try!(writeln!(w, "use glib::StaticType;"));
-                try!(writeln!(
-                    w,
-                    "use glib::value::{{Value, SetValue, FromValue, FromValueOptional}};"
-                ));
-                try!(writeln!(w, "use gobject_ffi;"));
-            }
-            try!(writeln!(w, "use glib::translate::*;"));
-        }
+        try!(general::uses(w, &env, &imports));
```
EPashkin: just `env` without reference

EPashkin:
@tmiasko Thanks, now use's in fully sorted

src/codegen/flags.rs
tmiasko: Done.

PR #539, opened 2018-02-09
doc: allow specifying a separate target path

MathieuDuponchelle:
This reduce the amount of manual tweaking required when
regenerating the doc of the gstreamer bindings, as it is
output in a different path with a different filename.

When the option is not specified, the current behaviour
is preserved.

### DIFF
--- a/src/config/config.rs
+++ b/src/config/config.rs
@@ -19,13 +19,14 @@ Usage: gir [options] [<library> <version>]
        gir --help
 
 Options:
-    -h, --help          Show this message.
-    -c CONFIG           Config file path (default: Gir.toml)
-    -d GIRSPATH         Directory for girs
-    -m MODE             Work mode: doc, normal, sys or not_bound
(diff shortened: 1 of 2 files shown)


src/config/config.rs
```diff
@@ -108,6 +110,16 @@ impl Config {
             a => a.into(),
         };
 
+        let doc_target_path: PathBuf = match args.get_str("--doc-target-path") {
+            "" => {
+                match toml.lookup("options.doc_target_path") {
+                    Some(p) => config_dir.join(try!(p.as_result_str("options.doc_target_path"))),
```
EPashkin: Not sure that it need be relative to config

EPashkin:
@MathieuDuponchelle Thanks. allowing specify file name too is good idea.

EPashkin:
I was too fast with merge:
Section "Generating documentation" in https://github.com/gtk-rs/gir/blob/master/README.md need update too.

PR #541, opened 2018-02-10
Add repositories urls in head comment

GuillaumeGomez:
cc @EPashkin


sdroege:
Good idea, just the name gir is a bit general

EPashkin:
@sdroege Agree
@GuillaumeGomez Thanks

EPashkin:
Hm, actually maybe this not so good idea:
at minimum someone can use custom fork of gir-files (maybe with complete different name)

GuillaumeGomez:
Then they can update the url themselves? I'm not sure this is really an issue. Before we didn't indicate the repository at all. The situation is a bit better at least. :)

EPashkin:
You mean they need fork and maintain fork of gir too? :wink:
If that problem really arises we can add options for gir-files link.

GuillaumeGomez:
Yes, that's what I meant. And yes, but I'm not sure such a problem will ever appear...

EPashkin:
IMHO as we don't include all .gir  in your gir-files (like https://github.com/gtk-rs/gir-files/pull/12),
creator wrapper for that .gir will need fork.

PR #542, opened 2018-02-12
record: use g_boxed_* if a get_type function exists

MathieuDuponchelle:
If a record has a get_type function, it is better to use
g_boxed_copy and g_boxed_free to make sure the appropriate
functions are called, for example:

struct Foo {};
foo_copy(void); // unrelated copy function
foo_free(void); // unrelated free function
foo_boxed_copy(Foo *);
foo_boxed_free(Foo *);
foo_get_type();

G_DEFINE_BOXED_TYPE (Foo, foo, foo_boxed_copy, foo_boxed_free)

### DIFF
--- a/src/codegen/general.rs
+++ b/src/codegen/general.rs
@@ -120,9 +120,9 @@ pub fn define_boxed_type(
     w: &mut Write,
     type_name: &str,
     glib_name: &str,
-    copy_fn: &str,
-    free_fn: &str,
-    get_type_fn: &Option<String>,
+    copy_fn: Option<&String>,
+    free_fn: Option<&String>,
+    get_type_fn: Option<&String>,
 ) -> Result<()> {
     try!(writeln!(w, ""));
     try!(writeln!(w, "glib_wrapper! {{"));
@@ -134,14 +134,26 @@ pub fn define_boxed_type(
     ));
     try!(writeln!(w, ""));
     try!(writeln!(w, "\tmatch fn {{"));
-    try!(writeln!(
-        w,
-        "\t\tcopy => |ptr| ffi::{}(mut_override(ptr)),",
-        copy_fn
-    ));
-    try!(writeln!(w, "\t\tfree => |ptr| ffi::{}(ptr),", free_fn));
-    if let Some(ref get_type_fn) = *get_type_fn {
+    if let Some(get_type_fn) = get_type_fn {
+        try!(writeln!(
+            w,
+            "\t\tcopy => |ptr| gobject_ffi::g_boxed_copy(ffi::{}(), ptr as *mut _) as *mut ffi::{},",
+            get_type_fn, glib_name
+        ));
+        try!(writeln!(w, "\t\tfree => |ptr| gobject_ffi::g_boxed_free(ffi::{}(), ptr as *mut _),", get_type_fn));
         try!(writeln!(w, "\t\tget_type => || ffi::{}(),", get_type_fn));
+    } else {
+        if let Some(copy_fn) = copy_fn {
+            try!(writeln!(
+                w,
+                "\t\tcopy => |ptr| ffi::{}(mut_override(ptr)),",
+                copy_fn
+            ));
+        }
+
+        if let Some(free_fn) = free_fn {
+            try!(writeln!(w, "\t\tfree => |ptr| ffi::{}(ptr),", free_fn));
+        }
     }
     try!(writeln!(w, "\t}}"));
     try!(writeln!(w, "}}"));
--- a/src/codegen/record.rs
+++ b/src/codegen/record.rs
@@ -24,17 +24,24 @@ pub fn generate(w: &mut Write, env: &Env, analysis: &analysis::record::Info) -> 
             unref_fn,
             &analysis.glib_get_type,
         ));
-    } else if let (Some(copy_fn), Some(free_fn)) = (
-        analysis.specials.get(&Type::Copy),
-        analysis.specials.get(&Type::Free),
-    ) {
+    } else if analysis.glib_get_type.is_some() {
         try!(general::define_boxed_type(
             w,
             &analysis.name,
             &type_.c_type,
-            copy_fn,
-            free_fn,
-            &analysis.glib_get_type,
+            None,
+            None,
+            analysis.glib_get_type.as_ref(),
+        ));
+    } else if analysis.specials.get(&Type::Copy).is_some() &&
+        analysis.specials.get(&Type::Free).is_some() {
+        try!(general::define_boxed_type(
+            w,
+            &analysis.name,
+            &type_.c_type,
+            analysis.specials.get(&Type::Copy),
+            analysis.specials.get(&Type::Free),
+            None,
         ));
     } else {
         panic!(


sdroege:
Generally makes sense, however it involves another function call to get the GType. Should be fine

Maybe would also make sense to make the copy/free functions configurable?

sdroege:
Also this of course would only work if it's actually registered as a boxed type, which is not the case for all things that we can currently handle with the existing scheme.

MathieuDuponchelle:
> Also this of course would only work if it's actually registered as a boxed type, which is not the case for all things that we can currently handle with the existing scheme.

Right, this is why I preserved the current behaviour when the record does not have a get_type function, however in the long run I don't see a reason why libraries that want records to be exposed shouldn't be expected to register them as boxed types?

sdroege:
They probably should, but you would then immediately depend on the version that started registering as boxed type as your minimum required version.

MathieuDuponchelle:
> They probably should, but you would then immediately depend on the version that started registering as boxed type as your minimum required version.

ack

src/codegen/record.rs
```diff
@@ -24,17 +24,24 @@ pub fn generate(w: &mut Write, env: &Env, analysis: &analysis::record::Info) ->
             unref_fn,
             &analysis.glib_get_type,
         ));
-    } else if let (Some(copy_fn), Some(free_fn)) = (
-        analysis.specials.get(&Type::Copy),
-        analysis.specials.get(&Type::Free),
-    ) {
+    } else if !analysis.glib_get_type.is_none() {
```
EPashkin: .is_some()

src/codegen/record.rs
```diff
@@ -24,17 +24,24 @@ pub fn generate(w: &mut Write, env: &Env, analysis: &analysis::record::Info) ->
...
             &analysis.name,
             &type_.c_type,
-            copy_fn,
-            free_fn,
-            &analysis.glib_get_type,
+            None,
+            None,
+            analysis.glib_get_type.as_ref(),
+        ));
+    } else if !analysis.specials.get(&Type::Copy).is_none() &&
```
EPashkin: same*2

EPashkin: Actually this is `.contains_key(&Type::Copy)`

src/codegen/general.rs
```diff
@@ -134,14 +134,26 @@ pub fn define_boxed_type(
...
     try!(writeln!(w, ""));
     try!(writeln!(w, "\tmatch fn {{"));
-    try!(writeln!(
-        w,
-        "\t\tcopy => |ptr| ffi::{}(mut_override(ptr)),",
-        copy_fn
-    ));
-    try!(writeln!(w, "\t\tfree => |ptr| ffi::{}(ptr),", free_fn));
-    if let Some(ref get_type_fn) = *get_type_fn {
+    if let Some(ref get_type_fn) = get_type_fn {
```
EPashkin: Clippy nags about this:
```
warning: this pattern creates a reference to a reference
   --> src\codegen\general.rs:137:17
    |
137 |     if let Some(ref get_type_fn) = get_type_fn {
    |                 ^^^^^^^^^^^^^^^ help: change this to: `get_type_fn`
    |
    = note: #[warn(needless_borrow)] on by default
    = help: for further information visit https://rust-lang-nursery.github.io/rust-clippy/v0.0.186/index.html#needless_borrow
```
same in other 2 here

MathieuDuponchelle: How do you run clippy btw?

EPashkin:
@MathieuDuponchelle thanks for idea.
@GuillaumeGomez, @sdroege,
I think that this is safer in theory, despite minor slowing.
Problem that we detect `RecordType::Boxed` for presence "copy" and "free" functions,
so this unsafe by same reasoning as PR started.
In current state we get compiler error if these functions has wrong arguments,
but after this PR we can get only runtime error, maybe even AV.

src/codegen/general.rs
EPashkin: Through `cargo clippy`.
`cargo install -f clippy` had sometimes build error, but with my current nightly it works.

sdroege:
It's not really safer, just allows for binding types that have non standard copy/free (like GstSDPMessage) or that don't have any at all. OTOH it does not work with types that don't register a boxed type (fallback code in the PR) and it raises the min version requirements in various cases.

Also IIRC going through the generic functions requires going through a global mutex all the time :(

EPashkin:
Then maybe better solution is add config that declares type boxed with copy\free as defined in this PR?

PS. I hear nothing about `going through the generic functions requires going through a global mutex all the time` :(

src/codegen/general.rs
sdroege: You need nightly in any case. If you use rustup, install as second toolchain, then cargo +nightly install clippy and cargo +nightly clippy for running it.

Stable can stay your default toolchain

MathieuDuponchelle:
> Also IIRC going through the generic functions requires going through a global mutex all the time :(

Just had a look, that is indeed the case (in g_type_value_table_peek), I assume this should not be needed for boxed copy / free functions as these cannot change, so it could probably be optimized.

I tend to think using g_object_boxed_copy / g_object_boxed_free is the correct approach here but up to you guys :)

EPashkin:
@MathieuDuponchelle Added manual alternative, please check for your records (GstSDPMessage ?)

@GuillaumeGomez, @sdroege Which version do you prefer?

sdroege:
I prefer the current behaviour and fall-back to the boxed functions if no usable direct functions can be found, plus configuration for the copy/free function name

EPashkin:
@sdroege Currently we fall-back to direct mode and not generate record type at all, as we don't know if record is actually boxed or shared.
You think that record without ref\unref is actually boxed?

MathieuDuponchelle:
For the record, I must that looking up the functions by name still feels a bit too much like guessing to me, and using g_boxed_* functions is the correct mechanism. If this needs optimizing, then it should be, that would also profit pygobject, which does use these functions.

EPashkin:
@MathieuDuponchelle, @sdroege You right, all records defined with `G_DEFINE_BOXED_TYPE` or `G_DEFINE_BOXED_TYPE_WITH_CODE` so it safe to use boxed for records as default.

sdroege:
Note that this applies to "Shared" and "Boxed" types (we currently distinguish between the two based on whether ref/unref or copy/free functions are found, IIRC). As long as they have a GType attached to them, and as long as that GType is a boxed type, then this will work (and go through that global mutex...).

Now problem is to detect if a GType is registered as boxed or not.

> looking up the functions by name still feels a bit too much like guessing to me

The Boxed/Shared distinction is also done that way. And if you have a GObject-style API that uses ref/unref or copy/free in another way, then it's violating the API design principles and will also fall apart in other bindings (C# and Vala calls the functions directly too, for example, instead of going through the whole GType infrastructure).

MathieuDuponchelle:
> Now problem is to detect if a GType is registered as boxed or not.

This is done with <https://developer.gnome.org/gobject/stable/gobject-Type-Information.html#G-TYPE-FUNDAMENTAL:CAPS> in pygobject fwiw

sdroege:
Yes that's the difference between dynamic and static bindings :) We have to do this decision here statically based on the information from the .gir file

MathieuDuponchelle:
> Yes that's the difference between dynamic and static bindings :) We have to do this decision here statically based on the information from the .gir file

Right, but one could statically generate code that checks that at runtime :)

The information whether a type is boxed is not carried over to the gir, though it is present in the intermediary dump file generated by gi.

sdroege:
> Right, but one could statically generate code that checks that at runtime :)

Even more runtime costs :) For dynamic bindings that all does not really matter, they do everything at runtime anyway... but we should try to be efficient here.

MathieuDuponchelle:
>  Even  more runtime costs :) For dynamic bindings that all does not really  matter, they do everything at runtime anyway... but we should try to be  efficient here.

rust must have some sort of static initializer :)

sdroege:
> rust must have some sort of static initializer :)

lazy_static could be used here, yes.  But how would this work at runtime? You would at runtime decide whether to call the proper ref/unref/copy/free functions or if it's actually registered as a boxed type call the generic ones? That seems weird.

MathieuDuponchelle:
> lazy_static could be used here, yes.
ack
> You would at runtime decide whether to call the proper ref/unref/copy/free functions or if it's actually registered as a boxed type call the generic ones?
yes
> That seems weird.
It's not weird if it works! :)

sdroege:
You would then always have the calls to the "guessed" functions in the code anyway, so could as well just call them directly. They would fail to compile if they're wrong (like GstSDPMessage) :)

MathieuDuponchelle:
> They would fail to compile if they're wrong (like GstSDPMessage) 

No answer to that, that kills the idea :)

Adding is-boxed to records in the gir should be very easy ofc

EPashkin:
@sdroege, @GuillaumeGomez Any vote for this or #545 ?

sdroege:
#545 makes sense to me as it only changes behaviour when configured to do so

GuillaumeGomez:
#545 seems more complete.

EPashkin:
Then I merge #545 

@MathieuDuponchelle Thanks again for idea, I close this PR.

MathieuDuponchelle:
@EPashkin, no problem, good that we have some way to do that :)

Issue #529, opened 2018-02-02
Fix generation of disguised records


EPashkin:
As example `IConv` has `<record name="IConv" c:type="GIConv" disguised="1">` in .gir file and defined as `typedef struct _GIConv *GIConv;` in C sources,
so instead generate as incomplete type `pub struct GIConv(c_void);`,
it need be `struct GIConv(*mut c_void)` with showing in contained types like `GIOChannel`.

Note: as https://bugzilla.gnome.org/show_bug.cgi?id=721481 says, there many cases when this flag also used for `typedef struct _Foo Foo;` so fixing with https://github.com/gtk-rs/gir-files/blob/master/fix.sh may also needed.

PS: We need trait `glib::translate::Ptr` for disguised `struct GdkAtom(*mut c_void)`,
but can't implement it in sys module without separate crate for trait,
so better generate as before. with only removing unusable `Debug` implementation and internal aliasing.
```
#[repr(C)]
pub struct _GdkAtom(c_void);
pub type GdkAtom = *mut _GdkAtom;
```

EPashkin:
I tried to fix this issue and found that it almost can't be done until upstream fixed:
- all disguised in gio, gobject is not pointers
- all in glib too except GIOConv
- GdkDevicePadInterface is not pointer
- there many disguised xxClass and xxPrivate but all of it not pointers
- GdkAtom is pointer, but it already has special handling

GIConv used only in GIOChannel but it still incomplete due bitflag field use_buffer,
so patch in gir as for `GHookList` pointless

tmiasko:
Could we push more generalized solution to master (compared to one which we
have now for Atom), but enable it only for GdkAtom and GIConv?

Once upstream GIR XML is fixed we will remove special check for GdkAtom and
GIConv.

BTW, the types that are disguised in the original meaning seems to be quite
rare. In fact, those two seems to be the only cases.

EPashkin:
@tmiasko IMHO It not worth do general solution for one example.
`GdkAtom` defined as `const char *` but actually contains int convertible values that not points to valid memory (like 1, 2 etc.), so general rules almost not applied to it.

tmiasko:
I would to generate similar code as we currently do for GdkAtom. This commit
demonstrates what I had in mind in more detail: https://github.com/tmiasko/gir/commit/953272d9705925f5a8b73faaeb8d84d062b52295.

For GIConv this generates code below (note that I prefixed type with
[...]

PR #544, opened 2018-02-13
General support for disguised types, but enabled only for GdkAtom and…

EPashkin:
… GIConv.

Part of #529

cc @tmiasko


tmiasko:
Thanks. LGTM ;-).

EPashkin:
Thanks @tmiasko

PR #545, opened 2018-02-13
Allow generate record with g_boxed_copy and g_boxed_free

EPashkin:
Alternative of https://github.com/gtk-rs/gir/pull/542
Add `record_boxed` config for records.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,18 @@ version = "3.12"
     version = "3.18"
 ```
 
+Some records have non-standard `copy` or `free` functions or even do not have it.
+You can define these records with functions for boxed types.
+
+```toml
+[[object]]
+name = "GstSdp.SDPMessage"
+status = "generate"
+# generates `copy` and `free` function by `g_boxed_copy` and `g_boxed_free`
+# only works if record has `glib:get-type` defined
+record_boxed = true
+```
+
 For global functions, the members can be configured by configuring the `Gtk.*` object:
 
 ```toml
--- a/README.md
+++ b/README.md
@@ -203,8 +203,9 @@ version = "3.12"
     version = "3.18"
 ```
 
-Some records have non-standard `copy` or `free` functions or even do not have it.
-You can define these records with functions for boxed types.
+Gir auto-detects copy and free functions by looking up `type_name_copy` and `type_name_free` functions.
+If the relevant functions are not named that way, and the type is defined with `G_DEFINE_BOXED_TYPE`,
+`use_boxed_functions=true` can be used to call `g_boxed_copy` and `g_boxed_free` instead.
 
 ```toml
 [[object]]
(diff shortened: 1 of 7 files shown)


src/codegen/general.rs
```diff
@@ -149,6 +149,48 @@ pub fn define_boxed_type(
...
+    try!(writeln!(w, ""));
+    try!(writeln!(w, "glib_wrapper! {{"));
+    try!(writeln!(
+        w,
+        "\tpub struct {}(Boxed<ffi::{}>);",
+        type_name,
+        glib_name
+    ));
+    try!(writeln!(w, ""));
+    try!(writeln!(w, "\tmatch fn {{"));
```
GuillaumeGomez: We have tabulations now? :O

EPashkin: And long long time ;)

GuillaumeGomez: Oh damn. I'm so late!

EPashkin: It replaced with 4 spaces by https://github.com/gtk-rs/gir/blob/master/src/writer/untabber.rs

GuillaumeGomez: Just saw it a few minutes ago (I was very surprised). Getting old, my memory can't follow up...

README.md
```diff
@@ -203,6 +203,18 @@ version = "3.12"
     version = "3.18"
 ```
 
+Some records have non-standard `copy` or `free` functions or even do not have it.
+You can define these records with functions for boxed types.
```
MathieuDuponchelle: I think the name does not really reflect the intent here, a record might be defined as a boxed type but would not need to use g_boxed functions, how about `use_boxed_functions` maybe?

As for the documentation comment I would then go for:

```
gir auto-detects copy and free functions by looking up `type_name_copy` and `type_name_free` functions.
If the relevant functions are not named that way, and the type is defined with `G_DEFINE_BOXED_TYPE`,
`use_boxed_functions=true` can be used to call `g_boxed_copy` and `g_boxed_free` instead.

> In that case, it is required to declare the `get_type` function in the headers.
```

That is a bit verbose but would make me less confused :)

EPashkin: Thanks, I make change

EPashkin:
Now gir generate `AutoBoxed` record as fall-back.
`RecordType::Direct` commented until we have type to generate for it

sdroege:
Can you put a diff somewhere for the effect this has on one of the crates?

EPashkin:
No changes in gtk-rs crates by default as old fall-back was panic.
Changes for `use_boxed_functions = true`
```diff
--- a/src/auto/paper_size.rs
+++ b/src/auto/paper_size.rs
@@ -15,8 +16,8 @@ glib_wrapper! {
     pub struct PaperSize(Boxed<ffi::GtkPaperSize>);
 
     match fn {
-        copy => |ptr| ffi::gtk_paper_size_copy(mut_override(ptr)),
-        free => |ptr| ffi::gtk_paper_size_free(ptr),
+        copy => |ptr| gobject_ffi::g_boxed_copy(ffi::gtk_paper_size_get_type(), ptr as *mut _) as *mut ffi::GtkPaperSize,
+        free => |ptr| gobject_ffi::g_boxed_free(ffi::gtk_paper_size_get_type(), ptr as *mut _),
         get_type => || ffi::gtk_paper_size_get_type(),
     }
 }
```

sdroege:
Seems good to me, also the changes in GIR itself.

PR #546, opened 2018-02-15
Alternative of General support for disguised types

EPashkin:
Part of #529
Followup of https://github.com/gtk-rs/gir/pull/544

cc @GuillaumeGomez,  @sdroege, @tmiasko

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -215,8 +215,8 @@ impl Library {
         parent_name_prefix: Option<&str>,
         parent_ctype_prefix: Option<&str>,
     ) -> Result<Option<Type>> {
-        let mut record_name = elem.attr_required("name")?.to_owned();
-        let mut c_type = elem.attr_required("type")?.to_owned();
+        let record_name = elem.attr_required("name")?;
+        let c_type = elem.attr_required("type")?;
         let get_type = elem.attr("get-type").map(|s| s.to_owned());
         let gtype_struct_for = elem.attr("is-gtype-struct-for");
         let version = self.read_version(parser, ns_id, elem)?;
@@ -233,7 +233,7 @@ impl Library {
                 self.read_function_to_vec(parser, ns_id, elem, &mut fns)
             }
             "union" => {
-                self.read_union(parser, ns_id, elem, Some(&record_name), Some(&c_type)).map(|mut u| {
+                self.read_union(parser, ns_id, elem, Some(record_name), Some(c_type)).map(|mut u| {
                     let field_name = if let Some(field_name) = elem.attr("name") {
                         field_name.into()
                     } else {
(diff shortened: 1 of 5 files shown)


src/parser.rs
```diff
@@ -295,39 +295,17 @@ impl Library {
...
-                    target_c_type,
-                    doc: None, //TODO: temporary
-                })
-            );
-        }
+        // A 'disguised' structure is one where the c:type is a typedef that
+        // doesn't look like a pointer, but is internally: typedef struct _X *X;
+        //
+        // Currently g-ir-sanner is too eager to mark all typedef to non-complete
+        // types as disguised, so we limit this behaviour two most common cases.
```
GuillaumeGomez: "two" -> "to"?

src/parser.rs
```diff
@@ -295,39 +295,17 @@ impl Library {
...
-                    typ: tid,
-                    target_c_type,
-                    doc: None, //TODO: temporary
-                })
-            );
-        }
+        // A 'disguised' structure is one where the c:type is a typedef that
+        // doesn't look like a pointer, but is internally: typedef struct _X *X;
+        //
+        // Currently g-ir-sanner is too eager to mark all typedef to non-complete
```
GuillaumeGomez: "sanner"?

src/parser.rs
EPashkin: 2

GuillaumeGomez:
👍 (and thanks for removing all extra whitespaces!)

src/parser.rs
GuillaumeGomez: Then it's "to the two". 😃

src/parser.rs
EPashkin: Hm, Don't know.
@tmiasko It must be "g-ir-scanner" ?

EPashkin:
Not all whitespaces, only in files that I touch.

GuillaumeGomez:
Let me reword my sentence then: "and thanks for removing all extra whitespaces in those files!". Better? 😆

EPashkin:
Do not thank me, the editor (emacs) does it himself while saving ;)

GuillaumeGomez:
Just take the compliment and stop debating! Can't even thank people anymore now...

src/parser.rs
EPashkin: Thanks, I will fix

src/parser.rs
tmiasko: Yes, it should be g-ir-scanner. I made so many typos there, and now you have to fix them ;-).

EPashkin:
Yes sir!

src/parser.rs
```diff
@@ -295,39 +295,17 @@ impl Library {
...
-                    name: disguised_record_name.clone(),
-                    c_identifier: disguised_c_type,
-                    typ: tid,
-                    target_c_type,
-                    doc: None, //TODO: temporary
-                })
-            );
-        }
+        // A 'disguised' structure is one where the c:type is a typedef that
+        // doesn't look like a pointer, but is internally: typedef struct _X *X;
```
tmiasko: What about using this as a comment for Record::disguised field?
Seems like a more appropriate place now.

EPashkin: Good idea, it will be first comment in `library::Record` :wink:

PR #547, opened 2018-02-17
Remove adding callback_guard

EPashkin:
Part of https://github.com/gtk-rs/glib/pull/250


EPashkin:
@sdroege, @GuillaumeGomez as we all agreed do this and changes is minimal, I merge this without reviews.

sdroege:
Now just need to regen everything and remove from manual code :) thanks!

PR #548, opened 2018-02-18
Add support for async functions that have a non-trivial ..

### LINKED ISSUE #535: Add support for async functions that have a non-trivial return type of the finish() function
Lots of GIO functions actually return the result instead of having it as an out parameter. Fixing this would remove the need of lots of manual implementations.

MathieuDuponchelle:
return type of the finish() function

Fixes #535 

I've got a WIP patch for Gio here: https://github.com/MathieuDuponchelle/gio/commit/babfd3a287f4050227b681aa45946c4fc95482ac

I also fixed a bug in the same commit, where we would look up the wrong finish function in gio, because we looked it up by rust name, and there were multiple unrelated functions with that name (connect_finish, maybe others). It was a bit awkward to split up, just let me know if you'd rather I split it up!

### DIFF
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -7,6 +7,7 @@ use env::Env;
 use analysis::imports::Imports;
 use analysis::functions::{find_function, find_index_to_ignore, replace_async_by_finish};
 use analysis::function_parameters::{CParameter, async_param_to_remove};
+use analysis::out_parameters::use_function_return_for_result;
 use analysis::rust_type::{bounds_rust_type, rust_type};
 use library::{Function, Fundamental, Nullable, ParameterDirection, Type, TypeId};
 use traits::IntoString;
@@ -117,9 +118,13 @@ impl Bounds {
             if let Some(bound_type) = Bounds::type_for(env, par.typ, par.nullable) {
                 ret = Some(Bounds::get_to_glib_extra(&bound_type));
                 if async && par.name == "callback" {
-                    let finish_func_name = replace_async_by_finish(&func.name);
+                    let func_name = func.c_identifier.as_ref().unwrap();
+                    let finish_func_name = replace_async_by_finish(func_name);
                     if let Some(function) = find_function(env, &finish_func_name) {
-                        let out_parameters = find_out_parameters(env, function);
+                        let mut out_parameters = find_out_parameters(env, function);
+                        if use_function_return_for_result(env, &function.ret) {
+                            out_parameters.insert(0, rust_type(env, function.ret.typ).expect("get rust type from return"));
+                        }
                         let parameters = format_out_parameters(&out_parameters);
                         let error_type = find_error_type(env, function);
                         type_string = format!("FnOnce(Result<{}, {}>) + Send + 'static", parameters, error_type);
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -10,6 +10,7 @@ use std::vec::Vec;
 
 use analysis::bounds::Bounds;
 use analysis::function_parameters::{self, Parameters, Transformation, TransformationType};
+use analysis::out_parameters::use_function_return_for_result;
 use analysis::imports::Imports;
 use analysis::out_parameters;
 use analysis::ref_mode::RefMode;
@@ -47,6 +48,7 @@ pub struct AsyncTrampoline {
     pub callback_type: String,
     pub bound_name: char,
     pub output_params: Vec<Parameter>,
+    pub ffi_ret: Option<Parameter>,
 }
 
 #[derive(Debug)]
@@ -192,9 +194,13 @@ fn analyze_function(
         if let Some((callback_type, bound_name)) = type_string {
             let func_name = func.c_identifier.as_ref().unwrap();
             let finish_func_name = replace_async_by_finish(func_name);
-            let rust_finish_func_name = replace_async_by_finish(&func.name);
             let mut output_params = vec![];
-            if let Some(function) = find_function(env, &rust_finish_func_name) {
+            let mut ffi_ret = None;
+            if let Some(function) = find_function(env, &finish_func_name) {
+                if use_function_return_for_result(env, &function.ret) {
+                    ffi_ret = Some(function.ret.clone());
+                }
+
                 output_params.extend(function.parameters.clone());
                 for param in &mut output_params {
                     if nameutil::needs_mangling(&param.name) {
@@ -209,6 +215,7 @@ fn analyze_function(
                 callback_type,
                 bound_name,
                 output_params,
+                ffi_ret,
             });
         }
         let type_error =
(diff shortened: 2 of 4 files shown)


src/analysis/bounds.rs
```diff
@@ -117,9 +118,13 @@ impl Bounds {
             if let Some(bound_type) = Bounds::type_for(env, par.typ, par.nullable) {
                 ret = Some(Bounds::get_to_glib_extra(&bound_type));
                 if async && par.name == "callback" {
-                    let finish_func_name = replace_async_by_finish(&func.name);
+                    let func_name = func.c_identifier.as_ref().unwrap();
+                    let finish_func_name = replace_async_by_finish(&func_name);
```
EPashkin: Unneeded &

src/codegen/function_body_chunk.rs
```diff
@@ -182,9 +200,13 @@ impl Builder {
                      type_: None,
                  });
         body.extend(output_vars);
+        let ret_name = match trampoline.use_ret {
+            true => "retval",
+            false => "_",
+        };
```
EPashkin: Please just "ret" and use if

EPashkin:
Big thanks.
Looks good, I added small nits, will look thoroughly later.

src/analysis/bounds.rs
MathieuDuponchelle: done

src/codegen/function_body_chunk.rs
MathieuDuponchelle: done, though I'm not sure using if looks much better?

src/codegen/function_body_chunk.rs
```diff
@@ -182,9 +200,15 @@ impl Builder {
                      type_: None,
                  });
         body.extend(output_vars);
+
+        let mut ret_name = "_";
+        if trampoline.use_ret {
+            ret_name = "ret";
+        }
```
EPashkin: Why not `let ret_name = if trampoline.use_ret { "ret" } else { "_" };`

MathieuDuponchelle: ah ! googled `let if else` but could only find `if let` documentation, tnx! (if you can't tell I'm not exactly fluent :)

src/codegen/function_body_chunk.rs
```diff
@@ -159,6 +159,24 @@ impl Builder {
                     }
                 }
             }).collect();
+
+        if trampoline.use_ret {
+            if let Some(ref return_value) = trampoline.return_value {
+                let mem_mode = c_type_mem_mode_lib(env, return_value.typ, return_value.caller_allocates, return_value.transfer);
+                let value = Chunk::Name("retval".to_string());
```
EPashkin: still `retval`

EPashkin: Still "retval"

src/codegen/function_body_chunk.rs
```diff
@@ -159,6 +159,24 @@ impl Builder {
                     }
                 }
             }).collect();
+
+        if trampoline.use_ret {
+            if let Some(ref return_value) = trampoline.return_value {
```
EPashkin: Hm, there way to have use_ret=true and return_value = None ?
Maybe we can combine this fields to one return_value?

EPashkin:
As I understand @sdroege already approved results.
@GuillaumeGomez You see any problem?

src/codegen/function_body_chunk.rs
MathieuDuponchelle: the opposite could be true yes (return_value that shouldn't be used), I could pass only the return_value when it needs to be used, but then that would make things a bit unintuitive.

GuillaumeGomez:
I'm not much into async so apart from the code, I don't have much to say. So if it's good for you and @sdroege, it's good for me as well.

src/codegen/function_body_chunk.rs
EPashkin: As I know it normal in languages with optional types that if `return_value = Some(parameter)` then it used and `return_value = None` meant that nothing to return, so separate boolean field on the contrary makes it more confusing

sdroege:
> As I understand @sdroege already approved results.

I only looked at the generated output, that seems great. And only shortly looked over this PR, nothing bad caught my eye. If the code generator changes look good to you, go for it :)

src/codegen/function_body_chunk.rs
MathieuDuponchelle: in that context, return_value is the return_value of the finish function, not something we have to "return", that's consistent with the other fields in the trampoline object (`output_params` in particular)

EPashkin: Yes, from trampoline point of view it parameter for result for ffi-function (xxx_finish) and from other side it main type for good case. So IMHO we can use `ffi_ret` as field name, as I don't see normal way to separate parts of `Ok` content

MathieuDuponchelle: I don't understand what you're saying here :)

EPashkin: I says that this problem field in async trampoline (currently named `return_value`) is parameter that returns from FFI-function (xx_finish) in your case, so we can name it `ffi_ret` or `finish_ret` and use it in same way as now but without `use_ret` flag

EPashkin: .. but set `ffi_ret` to `Some` only if `use_function_return_for_result` return true

MathieuDuponchelle: Well if you insist on that I don't really mind, still don't like it though ;)

src/analysis/functions.rs
```diff
@@ -192,9 +194,13 @@ fn analyze_function(
         if let Some((callback_type, bound_name)) = type_string {
             let func_name = func.c_identifier.as_ref().unwrap();
             let finish_func_name = replace_async_by_finish(func_name);
-            let rust_finish_func_name = replace_async_by_finish(&func.name);
             let mut output_params = vec![];
-            if let Some(function) = find_function(env, &rust_finish_func_name) {
+            let mut ffi_ret = None;
+            if let Some(function) = find_function(env, &finish_func_name) {
```
EPashkin: We still add trampoline when no finish function found, its normal?
If not then `output_params` and `ffi_ret` can be moved inside this if and created without `mut` modifier.

src/codegen/function_body_chunk.rs
```diff
@@ -182,9 +198,12 @@ impl Builder {
                      type_: None,
                  });
         body.extend(output_vars);
+
+        let ret_name = if let Some(_) = trampoline.ffi_ret { "ret" } else { "_" };
```
EPashkin: Better write `let ret_name = if trampoline.ffi_ret.is_some() { "ret" } else { "_" };`

src/analysis/functions.rs
MathieuDuponchelle: Well I don't really know if that's normal, that's not really related to this PR though is it?

EPashkin: Yes

EPashkin:
@MathieuDuponchelle Thanks, I merging this

tmiasko:
Since this was merged I have panic when generating sys bindings: 
```
thread 'main' panicked at 'get rust type from return: Ignored("PixbufFormat")', libcore/result.rs:945:5
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at libstd/sys_common/backtrace.rs:59
             at libstd/panicking.rs:380
   3: std::panicking::default_hook
             at libstd/panicking.rs:396
   4: std::panicking::rust_panic_with_hook
             at libstd/panicking.rs:576
   5: std::panicking::begin_panic
             at libstd/panicking.rs:537
   6: std::panicking::begin_panic_fmt
             at libstd/panicking.rs:521
   7: rust_begin_unwind
             at libstd/panicking.rs:497
   8: core::panicking::panic_fmt
             at libcore/panicking.rs:71
   9: core::result::unwrap_failed
  10: gir::analysis::bounds::Bounds::add_for_parameter
  11: gir::analysis::functions::analyze_function
  12: gir::analysis::functions::analyze
  13: gir::analysis::object::class
  14: gir::analysis::run
  15: gir::do_main
  16: gir::main
  17: std::rt::lang_start::{{closure}}
  18: std::panicking::try::do_call
             at libstd/rt.rs:59
             at libstd/panicking.rs:479
  19: __rust_maybe_catch_panic
             at libpanic_unwind/lib.rs:102
  20: std::rt::lang_start_internal
             at libstd/panicking.rs:458
             at libstd/panic.rs:358
             at libstd/rt.rs:58
  21: main
  22: __libc_start_main
  23: _start
```

PR #549, opened 2018-02-20
Bail out on errors returned from rust_type instead of panicking.

tmiasko:


MathieuDuponchelle:
Ah this makes sense, I copied the expect code from `find_out_parameters` for what it's worth, so it means you may also end up with a panic if one of the out parameters is also ignored :)

src/analysis/bounds.rs
```diff
@@ -123,7 +123,10 @@ impl Bounds {
                     if let Some(function) = find_function(env, &finish_func_name) {
                         let mut out_parameters = find_out_parameters(env, function);
                         if use_function_return_for_result(env, &function.ret) {
-                            out_parameters.insert(0, rust_type(env, function.ret.typ).expect("get rust type from return"));
+                            match rust_type(env, function.ret.typ) {
+                                Ok(rt) => out_parameters.insert(0, rt),
+                                Err(_) => return (None, None),
```
EPashkin: This way we silently lost that function return something.
Better use `rust_type(...).into_string()`

EPashkin: At minimum it worth to try

tmiasko: This was mostly intended as a hack to avoid panic for now. It doesn't really
change anything with respect to generated code, because function will be
omitted anyway and return value doesn't matter. Though, now I see that panic
can be easily avoided with config change, so this no longer feels necessary.

EPashkin:
@tmiasko What you tried to generate?

tmiasko:
It panics when generating gdk-pixbuf-sys. Following config change avoids the problem:

```
[[object]]
name = "GdkPixbuf.PixbufFormat"
status = "generate"
```

EPashkin:
@tmiasko Thanks. My version generate uncommented function in normal mode (if I remove generate for `PixbufFormat`) but it better that just skip it by unknown reason.
Minor problem is how to comment function.
IMHO better not add "generate" to sys crate as all objects must be generated by default.
```rust
    #[cfg(any(feature = "v2_32", feature = "dox"))]
    pub fn get_file_info_async<'a, P: Into<Option<&'a gio::Cancellable>>, Q: FnOnce(Result<(/*Ignored*/PixbufFormat, i32, i32), Error>) + Send + 'static>(filename: &str, cancellable: P, callback: Q) {
        let cancellable = cancellable.into();
        let cancellable = cancellable.to_glib_none();
        let user_data: Box<Box<Q>> = Box::new(Box::new(callback));
        unsafe extern "C" fn get_file_info_async_trampoline<Q: FnOnce(Result<(/*Ignored*/PixbufFormat, i32, i32), Error>) + Send + 'static>(_source_object: *mut gobject_ffi::GObject, res: *mut gio_ffi::GAsyncResult, user_data: glib_ffi::gpointer)
        {
            let mut width = mem::uninitialized();
            let mut height = mem::uninitialized();
            let ret = ffi::gdk_pixbuf_get_file_info_finish(res, &mut width, &mut height, &mut error);
            let result = if error.is_null() { Ok((from_glib_none(ret), width, height)) } else { Err(from_glib_full(error)) };
            let callback: Box<Box<Q>> = Box::from_raw(user_data as *mut _);
            callback(result);
        }
        let callback = get_file_info_async_trampoline::<Q>;
        unsafe {
            ffi::gdk_pixbuf_get_file_info_async(filename.to_glib_none().0, cancellable.0, Some(callback), Box::into_raw(user_data) as *mut _);
        }
    }

    //pub fn get_formats() -> /*Ignored*/Vec<PixbufFormat> {
    //    unsafe { TODO: call ffi::gdk_pixbuf_get_formats() }
    //}
```

PR #551, opened 2018-02-20
Fix panic on generating gdk-pixbuf-sys

EPashkin:
Fix [panic when generating gdk-pixbuf-sys](https://github.com/gtk-rs/gir/pull/548#issuecomment-366976512)
Also see #549

@tmiasko please, can you confirm that it works without adding "generate" to sys?

### DIFF
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -123,7 +123,7 @@ impl Bounds {
                     if let Some(function) = find_function(env, &finish_func_name) {
                         let mut out_parameters = find_out_parameters(env, function);
                         if use_function_return_for_result(env, &function.ret) {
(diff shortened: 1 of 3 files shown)


tmiasko:
Yes, this works as well. BTW, you don't have the same problem, or just don't generate gdk-pixbuf-sys?

EPashkin:
@tmiasko I usually generate all syses in one go and just missed that one panics.
Thanks for reporting.

EPashkin:
@GuillaumeGomez, @sdroege Strange that `analysis::run` was called in sys mode,
its result not used in sys generation.
So I disabled call.

GuillaumeGomez:
Fine by me.

PR #552, opened 2018-02-23
Prioritize RecordType::Refcounted over other record types

EPashkin:
as only this record type is referenced without `mut`

cc @GuillaumeGomez, @sdroege

### DIFF
--- a/src/analysis/record_type.rs
+++ b/src/analysis/record_type.rs
@@ -34,10 +34,10 @@ impl RecordType {
             has_free = true;
         }
 
-        if has_copy && has_free {
-            RecordType::Boxed
-        } else if has_ref && has_unref {
+        if has_ref && has_unref {
             RecordType::Refcounted
+        } else if has_copy && has_free {
+            RecordType::Boxed
         } else {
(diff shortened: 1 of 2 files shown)


EPashkin:
Added regen example

EPashkin:
Record `GtkWidgetPath` has both ref/unref and copy/free pairs, so detected as `RecordType::Boxed` before.
It usage has many const access and this breaks build (but maybe https://github.com/gtk-rs/glib/pull/291 helps fix too)

sdroege:
Doesn't change anything for me and makes sense. If there is the possibility of refcounting, we should use it to prevent copies. The disadvantages of refcounted types exist already anyway (e.g. we can't make this Send without Sync) as C code might already make use of the refcounting.

It might make sense to expose the copy function independently though to be able to create a real copy.

EPashkin:
We already expose `copy` for shared(refcounted) records, try find `gtk_widget_path_copy` in regen example

sdroege:
Ok then :)

EPashkin:
Thanks, for review

PR #553, opened 2018-02-23
codegen/build.rs: disable print_system_libs in calls to pkg-config

MathieuDuponchelle:
By default, pkg-config-rs sets the PKG_CONFIG_ALLOW_SYSTEM_LIBS
environment variable
(alexcrichton/pkg-config-rs#35), which leads
to eg /usr/lib being present in the middle of the final command,
in turn leading to issues when working in an uninstalled environment,
where libraries that are also present system-wide do not get linked
against.


EPashkin:
Thanks

PR #554, opened 2018-02-24
config/gobject.rs: use AsciiExt so to_ascii_lowercase() will be found.

federicomenaquintero:
This makes gir build on Rust 1.21.

### DIFF
--- a/src/config/gobjects.rs
+++ b/src/config/gobjects.rs
@@ -1,3 +1,6 @@
+#[allow(unused_imports)]
+use std::ascii::AsciiExt;
+
 use std::collections::BTreeMap;
 use std::str::FromStr;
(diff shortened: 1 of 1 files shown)


EPashkin:
As I remember we support only current stable and previous

sdroege:
And with newer version this would give a deprecation warning iirc

federicomenaquintero:
Just pushed an update with allow(unused_imports) for that line.  Hopefully this makes it work without warnings on Rust 1.23 and later.

Sounds like we need to formalize the minimum versions of rustc that all of gtk-rs needs?

bilelmoussaoui:
I suppose this can be closed as we only support Rust 1.51+ nowadays?

PR #557, opened 2018-02-28
Add support for string constants

MathieuDuponchelle:
Output based on the manual bindings in the GLib:
https://github.com/gtk-rs/glib/blob/master/src/lib.rs#L139-L168

The output for GIO can be seen @ <https://github.com/MathieuDuponchelle/gio/blob/master/src/auto/mod.rs> and <https://github.com/MathieuDuponchelle/gio/blob/master/src/auto/constants.rs>

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,20 @@ status = "generate"
 Note that you must not place `Gtk.*` into the `generate` array and
 additionally configure its members.
 
+You can control the generation of constants in a similar fashion:
+
+```
+[[object]]
+name = "Gtk.*"
+status = "generate"
+    [[object.constant]]
+    pattern = "*"
+    # No constants will be generated
+    ignore = true
+```
+
+Constants also support `version` and `cfg_condition` fields.
+
 In various cases, GObjects or boxed types can be used from multiple threads
 and have certain concurrency guarantees. This can be configured with the
 `concurrency` setting at the top-level options or per object. It will
--- a/src/analysis/constants.rs
+++ b/src/analysis/constants.rs
@@ -1,0 +1,69 @@
+use std::borrow::Borrow;
+
+use config;
+use env::Env;
+use library;
+use traits::*;
+use nameutil;
+use version::Version;
+
+#[derive(Debug)]
+pub struct Info {
+    pub name: String,
+    pub glib_name: String,
+    pub typ: library::TypeId,
+    pub version: Option<Version>,
+    pub deprecated_version: Option<Version>,
+    pub cfg_condition: Option<String>,
+}
+
+pub fn analyze<F: Borrow<library::Constant>>(
+    env: &Env,
+    constants: &[F],
+    obj: &config::gobjects::GObject,
+) -> Vec<Info> {
+    let mut consts = Vec::new();
+
+    for constant in constants {
+        let constant = constant.borrow();
+        let configured_constants = obj.constants.matched(&constant.name);
+
+        if configured_constants.iter().any(|c| c.ignore) {
+            continue;
+        }
+
+        if env.is_totally_deprecated(constant.deprecated_version) {
+            continue;
+        }
+
+        match *env.type_(constant.typ) {
+            library::Type::Fundamental(library::Fundamental::Utf8) => (),
+            _ => continue,
+        }
+
+        let version = configured_constants
+            .iter()
+            .filter_map(|c| c.version)
+            .min()
+            .or(constant.version);
+        let version = env.config.filter_version(version);
+        let deprecated_version = constant.deprecated_version;
+        let cfg_condition = configured_constants
+            .iter()
+            .filter_map(|c| c.cfg_condition.clone())
+            .next();
+
+        let name = nameutil::mangle_keywords(&*constant.name).into_owned();
+
+        consts.push(Info {
+            name: name,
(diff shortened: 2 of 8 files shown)


EPashkin:
Some crates (like gdk) don't have constant, so better don't generate file for it

src/analysis/mod.rs
```diff
@@ -125,6 +129,38 @@ fn analyze_global_functions(env: &mut Env) {
...
+
+    let obj = match env.config.objects.get(&*full_name) {
+        Some(obj) if obj.status.need_generate() => obj,
+        _ => return,
+    };
+
+    let constants: Vec<_> = ns.constants
+        .iter()
+        .collect();
+    if constants.is_empty() {
```
EPashkin: Strange why this check not works for gdk

EPashkin: Seems need early return after `constants::analyze` too

src/analysis/mod.rs
```diff
@@ -39,6 +40,7 @@ pub struct Analysis {
     pub objects: BTreeMap<String, object::Info>,
     pub records: BTreeMap<String, record::Info>,
     pub global_functions: Option<info_base::InfoBase>,
+    pub constants: Option<info_base::InfoBase>,
```
EPashkin: Maybe better just use `Vec<constants::Info>` and remove `constants` from `InfoBase` ?

README.md
```diff
@@ -231,6 +231,20 @@ status = "generate"
...
+You can control the generation of constants in a similar fashion:
+
+```
+[[object]]
+name = "Gtk.*"
+status = "generate"
+    [[object.constant]]
+    pattern = "*"
+    # No constants will be generated
+    ignore = true
```
sdroege: Can the pattern also contain something useful and then only those constants that match the pattern will be ignored?

sdroege:
> Some crates (like gdk) don't have constants, so better don't generate file for it

We also generate empty flags/enums.rs for crates without :) Should also be fixed I guess

Are there any bigger changes required for non-string constants? Other than just putting them outside lazy_static directly?

README.md
MathieuDuponchelle: yes, I tried, you can use `name` as well

src/analysis/mod.rs
MathieuDuponchelle: makes sense

src/analysis/mod.rs
MathieuDuponchelle: I'll take a look :)

EPashkin:
There will be minor warning with `glib::CSET_a_2_z` constant, it also have `CSET_A_2_Z`.
IMHO we better just ignore it.

src/codegen/constants.rs
```diff
@@ -0,0 +1,61 @@
...
+
+pub fn generate(env: &Env, root_path: &Path, mod_rs: &mut Vec<String>) {
+    let path = root_path.join("constants.rs");
+    let mut imports = Imports::new(&env.library);
+
+    let has_bindable_constants = env.analysis.constants.
+        iter().
+        any(|c| *env.type_(c.typ) == library::Type::Fundamental(library::Fundamental::Utf8));
+
+    if !has_bindable_constants {
```
EPashkin: Don't better check type in analysis phase and only check length here?

EPashkin:
Thanks, only minor possible nit remains

src/codegen/constants.rs
MathieuDuponchelle: I think you're right, updated!

src/analysis/constants.rs
```diff
@@ -0,0 +1,66 @@
...
+        let version = env.config.filter_version(version);
+        let deprecated_version = constant.deprecated_version;
+        let cfg_condition = configured_constants
+            .iter()
+            .filter_map(|c| c.cfg_condition.clone())
+            .next();
+
+        let name = nameutil::mangle_keywords(&*constant.name).into_owned();
+
+        if *env.type_(constant.typ) == library::Type::Fundamental(library::Fundamental::Utf8) {
```
EPashkin: Sorry for nags but can you move check after `is_totally_deprecated` and change it to empy good case and continue in else, or use match with empty good cases. This will minimize changes for other constants.

MathieuDuponchelle: Done

EPashkin:
@MathieuDuponchelle Thanks, waiting CI just in case

MathieuDuponchelle:
And CI is broken :)

EPashkin:
Just download timeout. I restarted it

MathieuDuponchelle:
Thanks!

README.md
```diff
@@ -231,6 +231,20 @@ status = "generate"
...
 additionally configure its members.
 
+You can control the generation of constants in a similar fashion:
+
+```
+[[object]]
+name = "Gtk.*"
+status = "generate"
+    [[object.constant]]
+    pattern = "*"
```
EPashkin: Just not to forget about problem:
I missed that wrong pattern format used: it not glob pattern but regex, so better use ".+" here.
It works because actual pattern looks "^*$"

MathieuDuponchelle: hrm, then I'm not sure how it even worked at all :)

PR #558, opened 2018-03-01
[WIP] Cross validate Rust ABI with C ABI.

tmiasko:
This is still work in progress. Just in case you have any comments already,
or want to test it out.

Note: I am not using cc and pkg-config crates because they are ill-suited for
usage outside build.rs. For example, cc expects various env variables to be
preconfigured and errors out otherwise, nor does it supports building
executables. Pkg-config prints out cargo specific things on stdout as a side
effect of its usage, and it doesn't even provide access to complete cflags.


EPashkin:
Glib test buils but failed
```
running 1 test
test test_cross_validate_abi_with_c ... FAILED

failures:

---- test_cross_validate_abi_with_c stdout ----
	thread 'test_cross_validate_abi_with_c' panicked at 'called `Result::unwrap()` on an `Err` value: ParseIntError { kind: InvalidDigit }', libcore\result.rs:945:5
stack backtrace:
   0: std::sys_common::backtrace::_print
             at libstd\sys\windows\backtrace/mod.rs:65
             at libstd\sys_common/backtrace.rs:71
   1: std::panicking::default_hook::{{closure}}
             at libstd\sys_common/backtrace.rs:59
             at libstd/panicking.rs:206
   2: std::panicking::default_hook
             at libstd/panicking.rs:216
   3: std::panicking::rust_panic_with_hook
             at libstd/panicking.rs:400
   4: std::panicking::begin_panic_fmt
             at libstd/panicking.rs:347
   5: rust_begin_unwind
             at libstd/panicking.rs:323
   6: core::panicking::panic_fmt
             at libcore/panicking.rs:71
   7: core::result::unwrap_failed
             at C:\projects\rust\src\libcore/macros.rs:23
   8: <core::result::Result<T, E>>::unwrap
             at C:\projects\rust\src\libcore/result.rs:782
   9: abi::get_c_abi
             at tests/abi.rs:140
  10: abi::test_cross_validate_abi_with_c
             at tests/abi.rs:95
  11: abi::__test::TESTS::{{closure}}
             at tests/abi.rs:84
  12: core::ops::function::FnOnce::call_once
             at C:\projects\rust\src\libcore\ops/function.rs:223
  13: <F as alloc::boxed::FnBox<A>>::call_box
             at libtest/lib.rs:1471
             at C:\projects\rust\src\libcore\ops/function.rs:223
             at C:\projects\rust\src\liballoc/boxed.rs:783
  14: _rust_maybe_catch_panic
             at libpanic_unwind/lib.rs:102
```
It normal that generated ABI executable print "\nzu\nzu" ?

tmiasko:
Not expected at all. If you are running this on Windows, I am surprised it got
that far. Currently %zu is used to print size_t, but this requires at least C99.
Changed to G_GSIZE_FORMAT from glib, it should be more portable.

I wonder how exactly do you run this on Windows, under msys2?

EPashkin:
Msys2x64 with gcc from it.
Nothing changed after usage `G_GSIZE_FORMAT` :(

EPashkin:
`gcc.EXE (Rev1, Built by MSYS2 project) 7.3.0`

tmiasko:
There is really nice writeup why it doesn't work: https://stackoverflow.com/questions/10678124/mingw-gcc-unknown-conversion-type-character-h-snprintf#10680635. As a workaround one can use following before running tests `export CPPFLAGS=-D__USE_MINGW_ANSI_STDIO=1`. I will have to think about proper long-term solution that would work out of the box.

On Windows the glib-sys test fails with following:
```
ABI mismatch for GPid
Rust: ABI { size: 4, alignment: 4 }
C:    ABI { size: 8, alignment: 8 }
ABI mismatch for GPollFD
Rust: ABI { size: 8, alignment: 4 }
C:    ABI { size: 16, alignment: 8 }
```

EPashkin:
Really error for GPid
`pub type GPid = c_int;` va `typedef void * GPid` for Windows64, can't find linux definition.

GPollFD seems also have different sizes
```
pub struct GPollFD {
    pub fd: c_int,
    pub events: c_ushort,
    pub revents: c_ushort,
}
```
vs
```
struct _GPollFD
{
#if defined (G_OS_WIN32) && GLIB_SIZEOF_VOID_P == 8
#ifndef __GTK_DOC_IGNORE__
  gint64	fd;
#endif
#else
  gint		fd;
#endif
  gushort 	events;
  gushort 	revents;
};
```

EPashkin:
Confirming that set `CPPFLAGS=-D__USE_MINGW_ANSI_STDIO=1` helps on tests
and results same 2 errors.

tmiasko:
* Updated template code so that _D__USE_MINGW_ANSI_STDIO=1 should no longer be necessary on Windows. 
* Test codegen now understands sys ignores.
* It works best with following fixes to gir files https://github.com/gtk-rs/gir-files/pull/16.

EPashkin:
Still don't work without `D__USE_MINGW_ANSI_STDIO=1`
Strange as it present in comand line
```
        compilation command "cc" "-D__USE_MINGW_ANSI_STDIO=1" "-mms-bitfields" "-ID:/P/msys64/mingw64/include/glib-2.0" "-ID:/P/msys64/mingw64
/lib/glib-2.0/include" "-ID:/P/msys64/mingw64/include" "-DABI_TEST_TYPE=GTypeCValue" "tests/abi.c" "-oabi" failed, exit code: 1
```
GIO don't compiled until i leave only `#include <gio/gio.h>` form <gio/*>
Glib has same 2 errors.
GObject failed on `GTypeCValue`
```
tests/abi.c: In function 'main':
<command-line>:0:15: error: invalid application of 'sizeof' to incomplete type 'GTypeCValue {aka union _GTypeCValue}'
tests/abi.c:8:31: note: in expansion of macro 'ABI_TEST_TYPE'
   printf("%zu\n%zu\n", sizeof(ABI_TEST_TYPE), alignof(ABI_TEST_TYPE));
                               ^~~~~~~~~~~~~
```
Gtk too:
```
tests/abi.c: In function 'main':
<command-line>:0:15: error: 'GtkPlug' undeclared (first use in this function); did you mean 'AtkPlug'?
tests/abi.c:9:31: note: in expansion of macro 'ABI_TEST_TYPE'
   printf("%zu\n%zu\n", sizeof(ABI_TEST_TYPE), alignof(ABI_TEST_TYPE));
                               ^~~~~~~~~~~~~
<command-line>:0:15: note: each undeclared identifier is reported only once for each function it appears in
tests/abi.c:9:31: note: in expansion of macro 'ABI_TEST_TYPE'
   printf("%zu\n%zu\n", sizeof(ABI_TEST_TYPE), alignof(ABI_TEST_TYPE));
                               ^~~~~~~~~~~~~
tests/abi.c: In function 'main':
<command-line>:0:15: error: 'GtkSocket' undeclared (first use in this function); did you mean 'AtkSocket'?
tests/abi.c:9:31: note: in expansion of macro 'ABI_TEST_TYPE'
   printf("%zu\n%zu\n", sizeof(ABI_TEST_TYPE), alignof(ABI_TEST_TYPE));
                               ^~~~~~~~~~~~~
<command-line>:0:15: note: each undeclared identifier is reported only once for each function it appears in
tests/abi.c:9:31: note: in expansion of macro 'ABI_TEST_TYPE'
   printf("%zu\n%zu\n", sizeof(ABI_TEST_TYPE), alignof(ABI_TEST_TYPE));
                               ^~~~~~~~~~~~~
tests/abi.c: In function 'main':
<command-line>:0:15: error: 'GtkStackAccessible' undeclared (first use in this function); did you mean 'GtkScaleAccessible'?
tests/abi.c:9:31: note: in expansion of macro 'ABI_TEST_TYPE'
   printf("%zu\n%zu\n", sizeof(ABI_TEST_TYPE), alignof(ABI_TEST_TYPE));
                               ^~~~~~~~~~~~~
<command-line>:0:15: note: each undeclared identifier is reported only once for each function it appears in
tests/abi.c:9:31: note: in expansion of macro 'ABI_TEST_TYPE'
   printf("%zu\n%zu\n", sizeof(ABI_TEST_TYPE), alignof(ABI_TEST_TYPE));
                               ^~~~~~~~~~~~~
tests/abi.c: In function 'main':
<command-line>:0:15: error: 'GtkStackAccessibleClass' undeclared (first use in this function); did you mean 'GtkScaleAccessibleClass'?
tests/abi.c:9:31: note: in expansion of macro 'ABI_TEST_TYPE'
   printf("%zu\n%zu\n", sizeof(ABI_TEST_TYPE), alignof(ABI_TEST_TYPE));
                               ^~~~~~~~~~~~~
<command-line>:0:15: note: each undeclared identifier is reported only once for each function it appears in
tests/abi.c:9:31: note: in expansion of macro 'ABI_TEST_TYPE'
   printf("%zu\n%zu\n", sizeof(ABI_TEST_TYPE), alignof(ABI_TEST_TYPE));
```
Pango also have many errors.
Pango-cairo just don't have types to check

tmiasko:
Thanks for testing @EPashkin.

* GTypeCValue needs additional include gobject/gvaluecollector.h, I will update gir-files.
* Compiling Gio is problematic as it currently contains unix specific headers. 
  Should we remove them from gir and ignore non-portable types, if any?
* GtkPlug, GtkSocket is X11 specific. Ignore them in sys crates?
* GtkStackAccessible is a bug in gtk headers, it should be included as part of gtk/gtk-a11y.h, but currently isn't.
  Neither can it be included directly, because it contains special guard. I will report this upstream.

AFAIR I was compiling pango with additional define: -DPANGO_ENABLE_ENGINE,
it reduced the amount of errors in pango.

EPashkin:
IMHO no help for Gio on windows. How you plan detect non-portable types?
GtkPlug, GtkSocket used in Gtk with `cfg_condition = "not(windows)"`. Bad that sys don't use this config now.

tmiasko:
Latest results for Windows when using MSYS2 MinGW 64-bit with following flags
for pango: `export CFLAGS='-DPANGO_ENABLE_BACKEND -DPANGO_ENABLE_ENGINE'`

* atk-sys - OK
* gdk-pixbuf-sys - OK
* gdk-sys - OK
* glib-sys - everything compiles; ABI mismatch for GPid, GPollFd
* gobject-sys - OK
* pango-sys - PangoScriptForLang undeclared, everything else OK
* secret-sys - OK

EPashkin:
Pango still failed for me
```
running 1 test
cc: error: '': No such file or directory
cc: error: '-DPANGO_ENABLE_BACKEND: No such file or directory
test test_cross_validate_abi_with_c ... FAILED

failures:

---- test_cross_validate_abi_with_c stdout ----
        thread 'test_cross_validate_abi_with_c' panicked at 'C ABI for char: StringError("compilation command \"cc\" \"\\\'\\\'\" \"\\\'-DPANG
O_ENABLE_BACKEND\" \"-DPANGO_ENABLE_ENGINE\\\'\" \"-mms-bitfields\" \"-ID:/P/msys64/mingw64/include/pango-1.0\" \"-ID:/P/msys64/mingw64/includ
e/glib-2.0\" \"-ID:/P/msys64/mingw64/lib/glib-2.0/include\" \"-ID:/P/msys64/mingw64/include\" \"-Wno-deprecated-declarations\" \"-DABI_TEST_TY
PE=char\" \"tests/abi.c\" \"-oabi\" failed, exit code: 1")', libcore\result.rs:945:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

tmiasko:
It looks like a problem with exporting environment variable.
Value of environment variable CFLAGS should be:
```
-DPANGO_ENABLE_BACKEND -DPANGO_ENABLE_ENGINE
```
instead, it is:
```
'-DPANGO_ENABLE_BACKEND -DPANGO_ENABLE_ENGINE'
```

Are you using cmd.exe instead of posix shell? 
Escaping rules are somewhat different there.

EPashkin:
I build under cmd.exe.
In msys2x64 shell it failed with old
```
---- test_cross_validate_abi_with_c stdout ----
        thread 'test_cross_validate_abi_with_c' panicked at 'called `Result::unwrap()` on an `Err` value: ParseIntError { kind: InvalidDigit }', libcore\result.rs:945:5
```
Glib failed with same error here too, while it works fine under cmd.exe

EPashkin:
@tmiasko Rechecked again, previous problem self fixed.
Others as you says in https://github.com/gtk-rs/gir/pull/558#issuecomment-370134109 with exception that glib now be ok after https://github.com/gtk-rs/sys/pull/81 😉 
Gio still problem on Windows, but we can test it on Linux only or/and make fix later.
Gtk has problem in linux too, but again we can fix it separately.
And files generated don't have any changes on current crate as test is external and need special launch.
So I vote for merging this PR (it already was help us find 3 problem in glib).

@GuillaumeGomez, @sdroege You see any problem.

src/codegen/sys/tests.rs
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    writeln!(w, "{}", r##"    abi
+}
+
+"##)
+
+}
+
+
```
EPashkin: Please, remove empty lines

EPashkin:
I think I came up with a solution to the problem with uncheckable types:
gir also generate second file `abi_filter.rs` (only if this not exists) with:
```
pub fn filter(_name: &str) -> bool {
    true
}
```
in `abi.rs` it import this function and use it to skipping unwanted C check.
`abi_filter.rs` can be edited manually, for example for skipping types, not available on Windows.

EPashkin:
Also IMHO `abi.c` and `abi.rs` need standard generated comment on beginning

tmiasko:
I tried to address the uncheckable types using somewhat different approach.
In particular, I generate cfg conditions in the same way we do in user crates.
Of course, you also need to put those in the config to use this feature.

Unfortunately, this is only a partial solution. The main problem is that C
headers inclusion should be conditional as well. Doing that would be even more
involved.

EPashkin:
@tmiasko Yes, we already have `cfg_condition` option for this, sadly it don't used in sys mode,
but can be used in `get_rust_abi`.
Agree about adding manual header file.

tmiasko:
Changes since last version:

* Rebased.
* Parse CC, CFLAGS and CPPFLAGS using shell parsing rules for compatibility with GNU Make and other build tools.
* Generate "DO NOT EDIT" headers.
* Use cfg_condition in ABI tests.
* Show ABI tests results summary on failure.
* Include header for manual customization of ABI tests.

src/codegen/sys/tests.rs
```diff
@@ -0,0 +1,290 @@
...
+/// Checks if type name is unlikely to correspond to a real C type name.
+fn is_name_made_up(name: &str) -> bool {
+    // Unnamed types are assigned name during parsing, those names contain an underscore.
+    name.contains('_')
+}
+
+fn generate_abi_c(env: &Env, path: &Path, w: &mut Write) -> io::Result<()> {
+    info!("Generating file {:?}", path);
+    general::start_comments(w, &env.config)?;
```
EPashkin: Please add `try!(writeln!(w, ""));` after this, and for `abi.rs` too.

EPashkin:
IMHO you can ignore secret-sys: it was added on initial stage and we decide don't publish it

EPashkin:
Maybe we need special `cfg_condition` function for tests as current version always add `feature="dox"` that unneeded and may cause problem for tests.

src/codegen/sys/tests.rs
```diff
@@ -0,0 +1,290 @@
...
+            => {
+                let full_name = format!("{}.{}", &ns.name, t.get_name());
+                if env.type_status_sys(&full_name).ignored() {
+                    return None;
+                }
+                let name = match t.get_glib_name() {
+                    None => return None,
+                    Some(name) => name,
+                };
+                if is_name_made_up(&name) {
```
EPashkin: & unneeded

src/codegen/sys/tests.rs
```diff
@@ -0,0 +1,290 @@
...
+                    results.failed += 1;
+                    eprintln!("ABI mismatch for {}\nRust: {:?}\nC:    {:?}",
+                              name, rust_abi, c_abi);
+                }
+            }
+        };
+    }
+
+    if results.failed + results.failed_compile != 0 {
+        panic!("FAILED\nABI test results: {} passed; {} failed (compilation errors: {})",
```
EPashkin: Good idea

EPashkin:
Maybe better don't include "includes" from .gir file to `abi.c` at all, and only add it to `manual.h`?

tmiasko:
Regarding secret-sys, I wanted to test test generation :-) and selected
libsecret as a test subject, because it is small. Though, version on Ubuntu
Trusty, which is used on Travis CI, is not compatible with our GIR. I will try
using atk-sys now.

Regarding headers, they don't change too often, so putting them into separate
file but never overwriting them again (thus allowing for manual customization),
could work quite nicely. I will write that version later.

src/codegen/sys/tests.rs
tmiasko: Done

src/codegen/sys/tests.rs
tmiasko: Done

tmiasko:
* Added commit that put all headers into manual.h, which won't be regenerated unless removed.
* Regarding feature="dox" in generated condition, I would leave it for now. It will compile fine, and you really shouldn't be running tests with dox.

src/codegen/sys/tests.rs
```diff
@@ -0,0 +1,297 @@
...
+    name.contains('_')
+}
+
+fn generate_manual_h(env: &Env, path: &Path, w: &mut Write) -> io::Result<()> {
+    info!("Generating file {:?}", path);
+    writeln!(w, "// Feel free to edit this file, it won't be regenerated by gir generator unless removed.")?;
+    writeln!(w, "")?;
+
+    let ns = env.library.namespace(MAIN_NAMESPACE);
+    for include in ns.c_includes.iter() {
```
EPashkin: Please, change this to `&ns.c_includes`

tmiasko: Missed due to all those warnings. Is there some option in clippy to show lints only for files that have some changes in version control?

EPashkin: As I know currently this impossible

tmiasko:
Fixed clippy warnings and rebased on top of #564.

PR #559, opened 2018-03-02
Revert "Remove adding callback_guard"

EPashkin:
This reverts commit a61ccc16be2e0add02ec7b7953f55ee04317cbf5.

Part of https://github.com/gtk-rs/glib/pull/250

cc @GuillaumeGomez, @sdroege

### DIFF
--- a/src/codegen/function_body_chunk.rs
+++ b/src/codegen/function_body_chunk.rs
@@ -242,11 +242,15 @@ impl Builder {
             Param { name: "res".to_string(), typ: format!("*mut {}::GAsyncResult", gio_crate_name) },
             Param { name: "user_data".to_string(), typ: format!("{}::gpointer", glib_crate_name) },
         ];
+        let body = Chunk::Chunks(vec![
+            Chunk::Custom("callback_guard!();".into()),
+            Chunk::Chunks(body),
+        ]);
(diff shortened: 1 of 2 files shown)


EPashkin:
Or better add config for this?

GuillaumeGomez:
A default `cfg!` maybe?

EPashkin:
@GuillaumeGomez Sorry, I don't understand.

GuillaumeGomez:
I meant we added a new feature (in `Cargo.toml`) and it was a default one so it can be disabled. Better? :)

EPashkin:
This way gir with from same commit can produce different result.
Not sure that it good idea.
Hm. I wrong, you can't change features in Cargo.toml without adding new commit (ignoring --no-default-features flag).
I don't see difference without this and just adding commit.
I meant adding option to `Gir.toml`.

EPashkin:
.. but as "my" option in "Gir.toml" will be used only year or half maybe it unnecessary work.

GuillaumeGomez:
I suppose... Let's just revert everything and avoid any unnecessary work.

sdroege:
Yes, revert simply and then we revert the reverts in 1-2 months again.

PR #560, opened 2018-03-02
Use info instead of println in codegen.

tmiasko:
The default log level is warn, thus this change also intentionally hides
those messages by default, avoiding excessive output when generating
code for multiple crates in succession.


EPashkin:
I not against this.
@GuillaumeGomez you agree?

GuillaumeGomez:
Perfectly fine for me as well. Thanks!

EPashkin:
@tmiasko I merge this tomorrow, when finish restoring CallbackGuards

PR #562, opened 2018-03-03
Add docs links

GuillaumeGomez:
Since https://github.com/rust-lang/rust/issues/43466 is in a good enough working state, I suppose we can now just link to types like this.

cc @EPashkin @sdroege


EPashkin:
@GuillaumeGomez Simple change. It works on stable too?

GuillaumeGomez:
I always forget about stable... Then let's just wait for this feature to hit stable.

PR #563, opened 2018-03-09
Use GObject status when generating types

EPashkin:
Helper for https://github.com/gtk-rs/sys/issues/82

### DIFF
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -59,7 +59,7 @@ fn generate_lib(w: &mut Write, env: &Env) -> Result<()> {
     try!(functions::generate_callbacks(w, env, &prepare(ns)));
     try!(generate_records(w, env, &records));
     try!(generate_classes_structs(w, env, &classes));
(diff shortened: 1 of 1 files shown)


EPashkin:
Main effect that it removes ignored constants instead comment it (https://github.com/gtk-rs/sys/pull/81),
I done this to be same for other types.

@sdroege It not cause problem in gstreamer?
cc @GuillaumeGomez too

tmiasko:
Good idea. For consistency, I think we should consider types with manual
implementation as complete in type analysis. Then, we could just remove the
hack for GHookList and implement it outside the gir.

EPashkin:
Overall is good idea, but `is_incomplete` then additional config parameter

sdroege:
@EPashkin causes no changes at all in gstreamer

EPashkin:
@sdroege thanks

GuillaumeGomez:
Good for me as well. Thanks!

EPashkin:
Then I merge this and update sys's PR

PR #565, opened 2018-03-11
Cross validate constants in XML with those from C.

tmiasko:


EPashkin:
IMHO only problem last is long run time without any output, Ex. travis don't like that and count build as failed.
Maybe we need write something each 100 classes or constants?

tmiasko:
Sure, I can make it a little bit more verbose. You would need to run it with --nocapture, of course.

The general situation with Travis is somewhat unsatisfactory. The packages there are quite outdated, so I didn't have much success in running tests there either way. The AppVeyor works much better in this respect.

EPashkin:
Usually we use custom package for gtk3.18: https://github.com/gtk-rs/examples/blob/master/build.sh#L14
But IMHO we better don't do this in sys.

tmiasko:
Now with some partial tests results after every 25th test.

EPashkin:
@tmiasko Thanks

PR #569, opened 2018-03-16
 Support libraries with customized crate name when generating tests.

tmiasko:
The mechanism of crate renaming that is available for external libraries
unfortunately does not cover the crate currently being generated.
Generated integration tests need to know the crate name to generate
proper extern crate declaration.

Obtain crate name from Cargo.toml file or auto-generate as before when
generating code for the very first time.

Also, support printing const char* in constants ABI test.

### DIFF
--- a/src/codegen/sys/cargo_toml.rs
+++ b/src/codegen/sys/cargo_toml.rs
@@ -4,12 +4,13 @@ use std::io::prelude::*;
 use toml::{self, Value};
 use toml::value::Table;
 
+use config::Config;
 use env::Env;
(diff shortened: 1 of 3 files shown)


EPashkin:
@tmiasko All looks good for me.
Waiting confirmation from @sdroege just in case.

sdroege:
If it works for you, get it in. I'll test it with gstreamer at a later time, don't wait for me :) I'll create a ticket if there are any problems.

EPashkin:
Ok

EPashkin:
@tmiasko Thansk

PR #570, opened 2018-03-16
Sort types and constants used in ABI tests by name at generation time.

tmiasko:
Also use constant arrays instead of BTreeMap, now that they are already
ordered. Constant functions are unstable feature, replace calls to
Layout::from_type, with manual construction of Layout struct.


EPashkin:
@tmiasko Thanks

PR #571, opened 2018-03-21
Include pkg-config --libs and LDFLAGS in the compiler command-line

sdroege:
Without this it fails in gstreamer-sys as some of the macros (which are actually mis-detected as constants...) are calling into GLib

### DIFF
--- a/src/codegen/sys/tests.rs
+++ b/src/codegen/sys/tests.rs
@@ -295,7 +295,7 @@ fn pkg_config_flags(packages: &[&str]) -> Result<Vec<String>, Box<Error>> {
                            &cmd, out.status).into());
     }
     let stdout = str::from_utf8(&out.stdout)?;
-    Ok(shell_words::split(stdout)?)
+    Ok(shell_words::split(stdout.trim())?)
 }
 
 
@@ -422,7 +422,7 @@ fn get_c_layout(dir: &Path, cc: &Compiler, name: &str) -> Result<Layout, Box<Err
     }
 
     let stdout = str::from_utf8(&output.stdout)?;
-    let mut words = stdout.split_whitespace();
+    let mut words = stdout.trim().split_whitespace();
     let size = words.next().unwrap().parse().unwrap();
     let alignment = words.next().unwrap().parse().unwrap();
     Ok(Layout {size, alignment})
@@ -441,7 +441,7 @@ fn get_c_value(dir: &Path, cc: &Compiler, name: &str) -> Result<String, Box<Erro
                            &abi_cmd, &output).into());
     }
 
-    Ok(str::from_utf8(&output.stdout)?.to_owned())
+    Ok(str::from_utf8(&output.stdout)?.trim().to_owned())
 }
 
 const RUST_LAYOUTS: &[(&str, Layout)] = &["##)?;
--- a/src/codegen/sys/tests.rs
+++ b/src/codegen/sys/tests.rs
@@ -240,7 +240,8 @@ impl Compiler {
         args.push("-Wno-deprecated-declarations".to_owned());
         args.extend(get_var("CFLAGS", "")?);
         args.extend(get_var("CPPFLAGS", "")?);
-        args.extend(pkg_config_cflags(PACKAGES)?);
+        args.extend(get_var("LDFLAGS", "")?);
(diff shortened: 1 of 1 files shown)


sdroege:
It might also make sense to use the pkg-config crate here, or not?
Unrelated, is there a way to ignore constants in gir's sys mode?

@EPashkin @tmiasko

EPashkin:
@sdroege You can add constants to `ignore` array, see https://github.com/gtk-rs/sys/pull/81/files.
About pkg-config IMHO it good idea.

EPashkin:
Appveyor error repeated for me too:
```
$ pkg-config --libs atk
-LD:/P/msys64/mingw64/lib -latk-1.0 -lgobject-2.0 -lglib-2.0 -lintl

eap@EAPCOMP3 MINGW64 ~
$ pkg-config --cflags atk
-mms-bitfields -ID:/P/msys64/mingw64/include/atk-1.0 -ID:/P/msys64/mingw64/include/glib-2.0 -ID:/P/msys64/mingw64/lib/glib-2.0/include -ID:/P/msys64/mingw64/include

$ cargo test -- --nocapture
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running D:\eap\rust\0\./target\debug\deps\atk_sys-043a3443263a3cf9.exe

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

     Running D:\eap\rust\0\./target\debug\deps\abi-1a9410588c5c5b1c.exe

running 2 tests
D:/P/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/7.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lintl
collect2.exe: error: ld returned 1 exit status
D:/P/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/7.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lintl
collect2.exe: error: ld returned 1 exit status

```

sdroege:
Why is the library not found, where is it?

EPashkin:
Don't know
I have `libintl-8.dll` in `bin` folder and `libintl.a` and `libintl.dll.a` in `lib` folder

EPashkin:
Also command runs fine from both command prompt and msys2x64 prompt.
```
cc -Wno-deprecated-declarations -mms-bitfields -ID:/P/msys64/mingw64/include/atk-1.0 -ID:/P/msys64/mingw64/include/glib-2.0 -ID:/P/msys64/mingw64/lib/glib-2.0/include -ID:/P/msys64/mingw64/include -LD:/P/msys64/mingw64/lib -latk-1.0 -lgobject-2.0 -lglib-2.0 -lintl -DABI_TYPE_NAME=char layout.c
```

EPashkin:
@sdroege I found problem: `output` need right trim to remove last '\r'

EPashkin:
Maybe using pkg-config package also fix this issue

sdroege:
Let's see

sdroege:
Not sure if `\r` is considered whitespace. Need to use the explicit trimming with it in the patterns otherwise.

tmiasko:
I really wanted to avoid linking in libraries. Mostly because of additional
linking time, but also because so many additional things could go wrong.

I wonder what cases are those, could you given any examples? Are they actually
worth testing? The only case like this I found was related to g_strerror,
in a macro, but it wasn't something you could test either way.

sdroege:
Yes exactly that one, `g_strerror` in a macro and that one should be ignored anyway (it's not a constant after all).

tmiasko:
Using pkg-config crate outside build.rs is somewhat problematic: it expects various environment variable to be set, it also prints things out to screen as a side effect of its work, and it doesn't even provide access to cflags.

EPashkin:
@sdroege This PR still needed or you decide to ignore `g_strerror` macros?

sdroege:
The second commit to strip off \r is needed, the other part is done

EPashkin:
@sdroege Then I merge PR. Thank.

sdroege:
Not the whole PR, only the second patch :)

EPashkin:
:(

EPashkin:
I added https://github.com/gtk-rs/gir/pull/573 to revert first commid

PR #576, opened 2018-04-01
Remove _POSIX_C_SOURCE feature test macro.

tmiasko:
When feature test macro is explicitly defined, the implicit defaults are no
longer provided. While we don't need any more features than requested, this is
not necessarily true of other libraries which may in fact assume that defaults
will be always available.

Instead of defining _POSIX_C_SOURCE to get appropriate functionality from
printf on MinGW, just define __USE_MINGW_ANSI_STDIO directly.

Also include manual header first so that it can be free to customize macro
definitions, before other headers are processed.


EPashkin:
@tmiasko Thanks

Issue #575, opened 2018-03-31
Generating constants for bitfields works only with types from the same crate.


tmiasko:
The bits field in structure generated by bitflags! macro is not public, and thus when generating a constant using a type from other sys crate, the code will fail to compile. For example, when generating code for libdazzle one currently gets following error at build time:

```
error[E0451]: field `bits` of struct `glib::GLogLevelFlags` is private
  --> src/lib.rs:78:78
   |
78 | pub const DZL_LOG_LEVEL_TRACE: glib::GLogLevelFlags = glib::GLogLevelFlags { bits: 1 };
   |                                                                              ^^^^^^^ field `bits` is private
```

I don't really care about DZL_LOG_LEVEL_TRACE in particular, but it would be nice to handle this in general, if possible.

cc: @alatiera

EPashkin:
Usually you can rexport sys's DZL_LOG_LEVEL_TRACE, it must be public

EPashkin:
Oh, gir really generate `{ bits: }` on some cases, but not one of these in gtk-rs.
cc @sdroege

EPashkin:
Hm, Seems I still don't understand problem:
`{ bits: 1 }` generated only in sys crate,
how you get it generated in normal crate (I think it normal as it use `glib::`)?,
also in normal mode IMHO we currently generate only string constants.

tmiasko:
It is a sys crate for libdazzle, `extern crate glib_sys as glib;` :-).

The bitflags crate don't support public bits field unfortunately.

EPashkin:
Why you don't use `glib::GLogLevelFlags::FLAG_RECURSION`  or `G_LOG_FLAG_RECURSION` ?

EPashkin:
I still can't find repo, so I don't understand why this const attempts to use `bits` as base const don't use it.

sdroege:
This is in a -sys crate, and the problem here is that dazzle has a bitflags constant that is of a type from another crate (glib-sys). I'm not sure how to generate a correct constant for that until we have stable const_fn, or we don't generate an actual constant but something in lazy_static.

EPashkin:
From Dazzle-1.0.gir
```xml
    <constant name="LOG_LEVEL_TRACE"
              value="1"
              c:type="DZL_LOG_LEVEL_TRACE"
              introspectable="0">
      <type name="GLib.LogLevelFlags" c:type="GLogLevelFlags"/>
    </constant>
```
Seems this only one constant form other crate.
[...]

PR #577, opened 2018-04-02
Uint flags in sys

EPashkin:
Part of https://github.com/gtk-rs/gir/issues/575

@sdroege Please, check changes in gstreamer, especially in GST_BUFFER_COPY_ALL and other constants based on flags

cc @GuillaumeGomez,, @tmiasko

### DIFF
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -187,13 +187,7 @@ fn generate_bitfields(w: &mut Write, env: &Env, items: &[&Bitfield]) -> Result<(
         if let Some(false) = config.map(|c| c.status.need_generate()) {
             continue;
         }
-        let mut vals: HashMap<String, (String, Option<Version>)> = HashMap::new();
-
-        try!(writeln!(
-            w,
-            "bitflags! {{\n\t#[repr(C)]\n\tpub struct {}: c_uint {{",
-            item.c_type
-        ));
+        try!(writeln!(w, "pub type {} = c_uint;", item.c_type));
         for member in &item.members {
             let member_config = config
                 .as_ref()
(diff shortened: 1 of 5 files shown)


sdroege:
The results look good to me but I still have to test it with gstreamer later. Thanks!

tmiasko:
Looks fine to me as well. 

BTW, do you know if this cast  `val as u32` is necessary and if so why?

EPashkin:
IMHO We parse string value as i64 first to make sure that it fully parsed (in case introspection error etc),
and then make it unsigned.
Seems this need be `as c_uint` currently as we changed flags type from `u32` long ago.

EPashkin:
.. but this will need add `libc` as reference to gir. So I not sure that about doing this change

tmiasko:
The commit that introduced this described this as "convert negative bitflags members". Now I see one example in GLib, G_LOG_LEVEL_MASK is -4, but that's about it.

sdroege:
> Now I see one example in GLib, G_LOG_LEVEL_MASK is -4, but that's about it.

That's just a shortcut and exactly what I meant in the other issue about signedness being basically meaningless in GLib-style C. That's `0xfffffffc`, i.e. all bits set except for the lowest 2 (which are `G_LOG_FLAG_RECURSION` and `G_LOG_FLAG_FATAL` and have a special meaning compared to all others).

GuillaumeGomez:
That's quite the removal of code. I'm all for it!

src/codegen/sys/lib_.rs
```diff
@@ -265,29 +244,11 @@ fn generate_constants(w: &mut Write, env: &Env, constants: &[Constant]) -> Resul
...
-                constant.c_identifier,
-                type_,
-                value
-            ));
-        }
+        try!(writeln!(
+            w,
+            "{}pub const {}: {} = {};",
+            comment, constant.c_identifier, type_, value
+        ));
```
EPashkin: There will be problem with this for negative const's like `G_LOG_LEVEL_MASK`,
so maybe better restore special case for `Bitfield` with same action as in bitfield generation?

sdroege: ack

tmiasko: This was actually never handled with casting, and G_LOG_LEVEL_MASK is not
generated by this code but rather as a part of generate_bitfields, so there is
no pressing need to fix this as a part of this PR, but in general it would be
nice for consistency.

sdroege:
@EPashkin Looks all good with GStreamer and seems to speed up compilation even.

EPashkin:
@sdroege Thanks

EPashkin:
Updated generating constants, based on bitfield

PR #579, opened 2018-04-17
Remove volatile wrapper struct.

tmiasko:
Using additional struct could potentially change the ABI, in similar way
wrapper for bitflags did. Generate volatile types without wrapper.


src/codegen/sys/statics.rs
```diff
@@ -45,10 +45,6 @@ pub fn only_for_glib(w: &mut Write) -> Result<()> {
         "pub type gconstpointer = *const c_void;",
         "pub type gpointer = *mut c_void;",
         "",
-        "#[repr(C)]",
-        "#[derive(Copy, Clone)]",
-        "pub struct Volatile<T>(T);",
```
sdroege: This is not equivalent to `T` at the FFI level?

tmiasko: Usually, but not always. For some examples see discussion about bitflags in #575.
It would be equivalent with `#[repr(transparent)]`, which is another nice but unstable thing.

EPashkin:
@tmiasko Sorry, currently I have bad internet access. I look at this PR later.

src/codegen/sys/statics.rs
sdroege: Ok, so this is only about wrappers around "fundamental" types and not structs

sdroege:
Good to go IMHO

EPashkin:
@tmiasko You sure that `Volatile` can cause problem?
I see 2 usage cases:
- field of struct:
```
#[repr(C)]
#[derive(Copy, Clone)]
pub struct GDBusAnnotationInfo {
    pub ref_count: Volatile<c_int>,
    pub key: *mut c_char,
    pub value: *mut c_char,
    pub annotations: *mut *mut GDBusAnnotationInfo,
}
```
- pointer function parameter:
```
    pub fn g_dbus_error_register_error_domain(error_domain_quark_name: *const c_char, quark_volatile: *mut Volatile<size_t>, entries: *const GDBusErrorEntry, num_entries: c_uint);
```

EPashkin:
Then maybe better fix these cases (in .gir file)?
Reason about gir not supporting these types in normal mode is strong,
but I still want think about this.

EPashkin:
Ok, lets remove `Volatile` wrapper.
But IMHO better leave marks of it,
so @tmiasko please restore if in `src/codegen/sys/ffi_type.rs` with changing format to `"/*Volatile*/{}"`

sdroege:
Seems good to me to keep it in a comment. But the volatile wrapper itself should really disappear, it makes not much sense in the FFI bindings. It's only an inconvenience and has no effect

tmiasko:
I restored comment about volatile. A few examples from generated code:

```
pub fn g_atomic_int_exchange_and_add(atomic: *mut /*volatile*/c_int, val: c_int) -> c_int;
pub fn g_bit_lock(address: *mut /*volatile*/c_int, lock_bit: c_int);
pub fn g_dbus_connection_send_message(connection: *mut GDBusConnection, message: *mut GDBusMessage, flags: GDBusSendMessageFlags, out_serial: *mut /*volatile*/u32, error: *mut *mut glib::GError) -> gboolean;
pub fn g_dbus_connection_send_message_with_reply(connection: *mut GDBusConnection, message: *mut GDBusMessage, flags: GDBusSendMessageFlags, timeout_msec: c_int, out_serial: *mut /*volatile*/u32, cancellable: *mut GCancellable, callback: GAsyncReadyCallback, user_data: gpointer);
pub fn g_dbus_connection_send_message_with_reply_sync(connection: *mut GDBusConnection, message: *mut GDBusMessage, flags: GDBusSendMessageFlags, timeout_msec: c_int, out_serial: *mut /*volatile*/u32, cancellable: *mut GCancellable, error: *mut *mut glib::GError) -> *mut GDBusMessage;
```

This includes clippy cleanup from #587 now.

PR #581, opened 2018-04-18
Generate bindings for global asynchronous functions.

tmiasko:
* Drop requirement that asynchronous functions have to end in `_async`.
  D-Bus related functions tend to have `_sync` version instead.
* Include global functions in search for finish function.
* Import modules used by trampolines, glib_ffi for gpointer and
  gobject_ffi for GObject.

### DIFF
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -5,7 +5,7 @@ use std::vec::Vec;
 use consts::TYPE_PARAMETERS_START;
 use env::Env;
 use analysis::imports::Imports;
-use analysis::functions::{find_function, find_index_to_ignore, replace_async_by_finish};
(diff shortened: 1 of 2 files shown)


sdroege:
There's also gtk::print_run_page_setup_dialog_async and gdk_pixbuf::new_from_stream_async

tmiasko:
The new_from_stream_async case should be handled by this. Things like
print_run_page_setup_dialog_async that don't have explicit finish call
shouldn't be affected either way.

EPashkin:
@tmiasko Thanks

2 functions now generated in gio: `poll_mountable`, `dbus_address_get_stream`, maybe some more after deignoring.
Commented changes in: glib, gtk, sourceview
Minor change in gdk-pixbuf

PR #582, opened 2018-04-18
Generate futures variants of GIO async functions

sdroege:
@EPashkin If you have any suggestions how to make this less ugly, please tell me :)

This depends on https://github.com/gtk-rs/glib/pull/314

### DIFF
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -51,6 +51,14 @@ pub struct AsyncTrampoline {
     pub ffi_ret: Option<Parameter>,
 }
 
+#[derive(Clone, Debug)]
+pub struct AsyncFuture {
+    pub is_method: bool,
+    pub name: String,
+    pub success_parameters: String,
+    pub error_parameters: String,
+}
+
 #[derive(Debug)]
 pub struct Info {
     pub name: String,
@@ -70,6 +78,7 @@ pub struct Info {
     pub doc_hidden: bool,
     pub async: bool,
     pub trampoline: Option<AsyncTrampoline>,
+    pub async_future: Option<AsyncFuture>,
 }
 
 impl Info {
@@ -143,6 +152,7 @@ fn analyze_function(
     let mut to_glib_extras = HashMap::<usize, String>::new();
     let mut used_types: Vec<String> = Vec::with_capacity(4);
     let mut trampoline = None;
+    let mut async_future = None;
 
     let version = configured_functions
         .iter()
@@ -191,7 +201,7 @@ fn analyze_function(
         if let Some(to_glib_extra) = to_glib_extra {
             to_glib_extras.insert(pos, to_glib_extra);
         }
-        if let Some((callback_type, bound_name)) = type_string {
+        if let Some((callback_type, success_parameters, error_parameters, bound_name)) = type_string {
             // Checks for /*Ignored*/ or other error comments
             if callback_type.find("/*").is_some() {
                 commented = true;
@@ -220,6 +230,13 @@ fn analyze_function(
                 bound_name,
                 output_params,
                 ffi_ret,
+            });
+
+            async_future = Some(AsyncFuture {
+                is_method: func.kind == FunctionKind::Method,
+                name: format!("{}_future", func.name),
+                success_parameters,
+                error_parameters,
             });
         }
         let type_error =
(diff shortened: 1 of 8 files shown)


sdroege:
Example for what it generates is at https://github.com/gtk-rs/gio/pull/108

src/codegen/function.rs
```diff
@@ -187,3 +294,88 @@ pub fn body_chunk(env: &Env, analysis: &analysis::functions::Info) -> Chunk {
...
+                try!(writeln!(body, "let {} = {}.cloned();", par.name, par.name));
+            }
+        } else if is_str {
+            try!(writeln!(body, "let {} = String::from({});", par.name, par.name));
+        } else if c_par.ref_mode != RefMode::None {
+            try!(writeln!(body, "let {} = {}.clone();", par.name, par.name));
+        }
+    }
+
+    try!(writeln!(body, "GioFuture::new(self, move |obj, send| {{"));
```
sdroege: This would currently not work if the async function is a global one (https://github.com/gtk-rs/gir/pull/581) and also not for methods not taking `self` as first argument (`gdk_pixbuf::new_from_stream_async`).

Any suggestions would be welcome how to make this more generic. I guess the object could become an `Option` for the `GioFuture`, only stored internally (we need to keep it alive until the operation is done), and the object is then passed via a `SendCell` directly into the closure.

src/codegen/object.rs
```diff
@@ -172,12 +181,16 @@ fn generate_trait(w: &mut Write, env: &Env, analysis: &analysis::object::Info) -
...
     }
-    if analysis.has_signals() || !analysis.properties.is_empty() {
+    if analysis.has_signals() || !analysis.properties.is_empty() || has_async {
         extra_isa.push(" + IsA<glib::object::Object>");
     }
     if analysis.has_action_signals() {
         extra_isa.push(" + glib::object::ObjectExt");
     }
+    if has_async {
+        extra_isa.push(" + Clone + 'static");
```
sdroege: We should probably add these directly to the `IsA` trait. @GuillaumeGomez?

src/analysis/functions.rs
```diff
@@ -264,6 +281,12 @@ fn analyze_function(
 
     if async && !commented && env.config.library_name != "Gio" {
         imports.add("gio_ffi", version);
+        imports.add("gio", version);
+    }
+
+    if async && !commented {
+        imports.add("futures_core", version);
```
tmiasko: This probably needs to be conditional: `#[cfg(feature = "futures")]`.

sdroege: Yes but that's currently not supported by the `imports`. I hacked around it where the imports are written out to add that cfg directive

tmiasko: FYI: https://github.com/sdroege/gio-rs/blob/886afc3d001ae10d8f2e39c56112575ff1a6a2f9/src/auto/file_i_o_stream.rs#L11, not sure if this already includes this hack or not.

sdroege: Ah my hack failed. Thanks! Will fix

sdroege: done

EPashkin:
@sdroege Thanks.
Please rebase and squash on last master

Now many unused warnings in gio.
Seems `analysis::Imports` need `map: BTreeMap<String, (Option<Version>, Option<String>)>,`
to store features, maybe even `Vec<String>` if we plan adding other.

EPashkin:
Other affected crates: gtk, sourceview
Commented changes: gdk_pixbuf

sdroege:
This will need some further changes now that @tmiasko's changes are merged.

@EPashkin does the approach look good to you otherwise? Any opinions on adding `Clone + 'static` bounds to `IsA` directly?

EPashkin:
Currently results looks good, I don't see problem `Clone + 'static` currently.
So only last is problem `cfg` for imports

sdroege:
All cleaned up, now have to do the imports part

sdroege:
Also done

EPashkin:
On gio result is much better, 
but in gtk it add `use gio;use gio_ffi;` to almost all generated file,
same problem in pango, pangocairo, glib, gdk-pixbuf, gdk, sourceview

sdroege:
>  but in gtk it add use gio;use gio_ffi; to almost all generated file,

Oops. Should be better now

src/codegen/function.rs
```diff
@@ -86,6 +90,54 @@ pub fn generate(
         }
     }
 
+    if analysis.async_future.is_some() {
+        let declaration = declaration_futures(env, analysis);
+        let suffix = if only_declaration { ";" } else { " {" };
+
+        try!(writeln!(w, ""));
```
EPashkin: Please `try!(writeln!(w));` and in other 3 cases here

EPashkin:
Yes, no stray `use` now, only small nit remain.

src/codegen/function.rs
sdroege: done

EPashkin:
@sdroege Thanks.

sdroege:
Now just to regen all the crates :)

PR #584, opened 2018-04-18
Split gir into lib and binary

GuillaumeGomez:
cc @EPashkin @sdroege @federicomenaquintero


EPashkin:
@GuillaumeGomez Thanks. Slightly bad that it mixed with `errorchain` removal, but it looks good for me.

Sorry but after #583 it need another fix
```
  --> src\xmlparser.rs:95:51
   |
95 |     pub fn attr_from_str<T>(&self, name: &str) -> Result<Option<T>>
   |                                                   ^^^^^^^^^^^^^^^^^ expected 2 type arguments
```

GuillaumeGomez:
Ah right. At some time I just removed a few things and then I went up until removing all of `error-chain`... More like a side-effect than a main focus. :laughing:

GuillaumeGomez:
I'd prefer to not use anything for error handling until we have a real need for it. :)

PR #586, opened 2018-04-18
RFC: Not so twisted snakes.

ghost:
Improve to_snake implementation. For example:

type name    | old module name | new module name
-------------|-----------------|----------------
DBusMessage  | d_bus_message   | dbus_message
FileIOStream | file_i_o_stream | file_iostream
IMContext    | i_m_context     | imcontext
IOStream     | i_o_stream      | iostream
URIHandler   | u_r_i_handler   | urihandler
UnixFDList   | unix_f_d_list   | unix_fdlist

Also remove dependency on case crate now that it is no longer needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,11 +27,6 @@ version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
-name = "case"
-version = "0.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-
-[[package]]
 name = "cc"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -91,7 +86,6 @@ dependencies = [
 name = "gir"
 version = "0.0.1"
 dependencies = [
- "case 0.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "docopt 0.8.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "env_logger 0.5.3 (registry+https://github.com/rust-lang/crates.io-index)",
  "git2 0.6.6 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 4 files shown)


EPashkin:
IMHO very good idea, even if it produces big changes.
Need rebase to solve merge conflict.

cc @GuillaumeGomez, @sdroege and @antoyo

sdroege:
Seems good to me but if we change this let's please follow what is done in C a bit closer. It's `io_stream`, `im_context`, `uri_handler`, `unix_fd_list`. DBus is the odd one here with `GDbus` but `g_dbus` instead of `g_d_bus`

tmiasko:
In this case, some further ideas:

1. Insert underscore in-between characters when going from lowercase to uppercase.
2. Version 1, but also insert underscore in-between second last and last
   uppercase character when going from sequence of three or more uppercase
   characters to lowercase.

Version 1 is what was proposed first, version 2 is some further refinement of
what I think @sdroege had in mind. The part "three or more uppercase characters"
is intended to avoid the case of single letter delimited with underscores.
If this don't feel to special casey to you, I would propose version 2.

sdroege:
Do we have other examples of two uppercase characters in a row, other than `DBus`?

tmiasko:
In trailing position: SoupCookieJarDB

sdroege:
That's `soup_cookie_jar_db`, all good then. 2) seems like a good approach

tmiasko:
I pushed updated implementation. It includes some additional test cases that
you might find interesting, especially those from GStreamer plugins. I think it
does agreeable job there, even in cases where it doesn't match what is used in
C code exactly.

For the record, new comparison table:

Type name    | Old module name | New module name
-------------|-----------------|----------------
DBusMessage  | d_bus_message   | dbus_message
FileIOStream | file_i_o_stream | file_io_stream
IMContext    | i_m_context     | im_context
IOStream     | i_o_stream      | io_stream
URIHandler   | u_r_i_handler   | uri_handler
UnixFDList   | unix_f_d_list   | unix_fd_list

sdroege:
Looks good to me :+1:

sdroege:
I would prefer if we could get my glib/gio PRs merged before moving around all the files. The merge conflicts will be annoying to solve :)

GuillaumeGomez:
Fine by me. :)

EPashkin:
Affected crates: gtk, gio, gdk

PR #588, opened 2018-04-21
Specify gio::Cancellable by full path in futures code if not inside GIO

sdroege:


EPashkin:
Thanks, waiting CI just in case

EPashkin:
Travis IMHO is enough

PR #589, opened 2018-04-23
Use ToOwned and Borrow traits when generating futures.

tmiasko:
Replace cloned() with to_owned() to handle situations where borrowed
type is different from the owned one. The cloned() method supports only
the case of &T -> T.

Borrow trait is used to perform analogous work in other direction.


sdroege:
Looks good to me and a bit simpler/more generic indeed. Not sure how I forgot about that :+1:

EPashkin:
@tmiasko Thanks,

Issue #592, opened 2018-04-28
Filename type in GObject Introspection vs Path and OsStr in Rust.


tmiasko:
GLib now uses filename types not only for passing filenames, but also for
environment variables and process arguments:

> The filename type represents an utf-8 string on Windows and a zero terminated
> guint8 array on Unix. It should be used for filenames, environment variables
> and process arguments.

This doesn't match exactly types used in Rust which makes more nuanced
distinction and uses Path for filenames and OsStr for environment variables and
process arguments.

It would be nice to have possibility to select desired interpretation for
given parameter / return type.

EPashkin:
Related to #264

tmiasko:
I was of the impression that intention of the filename type now is to represent
what we call OsString in Rust. Some of those are also paths, but not all of
them, so I would consider this change to be correct but quite unsatisfactory
from Rust perspective which does make the disticion between the two.

Depending on your priorities the current outcome might be worse, or an
improvement, say in the case that you care about non-UTF-8 arguments (note that
there is 1-to-1 mapping between OsStrings and PathBufs in Rust).

sdroege:
That seems to be the intention but it's a backwards incompatible change and breaks bindings as-such. GLib has to add a new type annotation for this or at least intentionally break API and document this properly.

tmiasko:
Given that we agree about os string as optimal solution for those new cases,
are paths actually that different? They all have no-op from / into conversions
and AsRef implementations. Consider following example, you can replace paths
with os strings, and everything will continue to work. The most problematic
cases are arrays.

```rust
use std::ffi::*;
use std::path::*;

// With Path
fn arg<P: AsRef<Path>>(_: P) {}
fn ret() -> PathBuf { PathBuf::new() }

// With OsStr
// fn arg<P: AsRef<OsStr>>(_: P) {}
// fn ret() -> OsString { OsString::new() }

fn as_argument() {
        let string = String::new();
        let osstring = OsString::new();
        let str: &str = string.as_ref();
        let osstr: &OsStr = osstring.as_ref();

        arg(str);
        arg(osstr);
        arg(&string);
        arg(&osstring);
}

fn as_return_value() {
    let _: OsString = ret().into();
    let _: String = ret().to_string_lossy().into_owned();
}

fn main() {
    as_argument();
    as_return_value();
}
```

sdroege:
They would work exactly like paths, yes. Just a different pair of types

tmiasko:
[...]

PR #597, opened 2018-05-09
WIP: Support string type override

EPashkin:
Part of https://github.com/gtk-rs/gir/issues/592
All works fine, but need last attempt to remove ugly code in src/analysis/out_parameters.rs

cc @GuillaumeGomez


EPashkin:
@sdroege You was right: need only one options to configure switch string between `String`, `PathBuf` and `OsString`

GuillaumeGomez:
Looks good to me. That's a great change! :)

sdroege:
Same, thanks :)

EPashkin:
Then I merge and update PR for glib, prepare PR for gio

PR #598, opened 2018-05-10
Rename "string" in string_type to "utf8" to match glib type

### LINKED ISSUE #592: Filename type in GObject Introspection vs Path and OsStr in Rust.

EPashkin:
.. and update readme
Closes #592

cc @GuillaumeGomez, @sdroege

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,9 @@ cfg_condition = "mycond"
         # parameter is calculated as length of string or array and removed from function declaration
         #  ( for length of return value use "return" )
         length_of = "str"
+        # change string type. Variants: "utf8", "filename", "os_string"
+        string_type = "os_string"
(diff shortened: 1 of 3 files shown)


sdroege:
:+1:

GuillaumeGomez:
Thanks!

PR #599, opened 2018-05-12
Make single_version_file accept path for "versions.txt"

EPashkin:
@GuillaumeGomez, @sdroege This will solve unneeded changes in gtk-rs/sys.


sdroege:
:+1:

EPashkin:
Updated: single_version_file works with filename too.

PR #603, opened 2018-05-15
Fix generating function body when out parameter has reserved name

### LINKED ISSUE #601: Function parameters named `type` appear not to be processed
The function `gst_mpegts_descriptor_parse_dvb_subtitling_idx` in the GStreamer MPEG-TS library has a parameter called `type` in the GIR file. This seems to be output as `type_` but with no processing of the GIR file data. This appears true for the sys mode as well as normal mode.

EPashkin:
Closes #601

@russel Please try this


russel:
@EPashkin I normally build gir from master/HEAD. However this pull request is not yet merged. What process would you like me to use to experiment for you?

EPashkin:
@russel See https://help.github.com/articles/checking-out-pull-requests-locally

russel:
@EPashkin I am not sure that works. I'll do something the non-GitHub way.

russel:
I cloned @EPashkin GitHub repository, built the fix_out-parameter_type branch and it seems to have fixed the problem I had with the parameter named `type_`.

EPashkin:
@russel Thanks for checking, then I merge

PR #610, opened 2018-06-17
Add `use glib;` if class has async method

EPashkin:
Fix error found in https://github.com/gtk-rs/gio/pull/124

cc @GuillaumeGomez


GuillaumeGomez:
Cool, thanks! I'll update my PR as soon as this one is merged.

PR #611, opened 2018-06-18
Enum / Flags: generate user defined `derive`

fengalin:
Allow using API mode TOML config such as:

``` toml
[[object]]
name = "Gst.BufferFlags"
status = "generate"
    [[object.derive]]
    name = "Serialize, Deserialize"
    cfg_condition = "feature = \"ser_de\""

[[object]]
name = "Gst.Format"
status = "generate"
    [[object.derive]]
    name = "Serialize, Deserialize"
    cfg_condition = "feature = \"ser_de\""
```

to generate:

``` rust
 bitflags! {
     #[cfg_attr(feature = "ser_de", derive(Serialize, Deserialize))]
     pub struct BufferFlags: u32 {
         const LIVE = 16;
         [...]
     }
 }
```

and

``` rust
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
 #[cfg_attr(feature = "ser_de", derive(Serialize, Deserialize))]
 pub enum Format {
     Undefined,
     [...]
 }
```

### DIFF
--- a/src/codegen/flags.rs
+++ b/src/codegen/flags.rs
@@ -1,6 +1,6 @@
 use analysis::imports::Imports;
 use analysis::namespaces;
-use codegen::general::{self, cfg_deprecated, version_condition, version_condition_string};
+use codegen::general::{self, cfg_deprecated, derives, version_condition, version_condition_string};
 use config::gobjects::GObject;
 use env::Env;
 use file_saver;
@@ -74,6 +74,7 @@ fn generate_flags(env: &Env, w: &mut Write, flags: &Bitfield, config: &GObject) 
             "    #[must_use]"
         ));
     }
+    try!(derives(w, &config.derives, 1));
     try!(writeln!(w, "    pub struct {}: u32 {{", flags.name));
     for member in &flags.members {
         let member_config = config.members.matched(&member.name);
--- a/src/codegen/general.rs
+++ b/src/codegen/general.rs
@@ -6,6 +6,7 @@ use analysis::general::StatusedTypeId;
 use analysis::imports::Imports;
 use analysis::namespaces;
 use config::Config;
+use config::derives::Derive;
 use env::Env;
 use gir_version::VERSION;
 use version::Version;
@@ -351,6 +352,25 @@ pub fn cfg_condition_string(
         }
         None => None,
     }
+}
+
+pub fn derives(
+    w: &mut Write,
+    derives: &[Derive],
+    indent: usize,
+) -> Result<()> {
+    for derive in derives {
+        let s = match &derive.cfg_condition {
+            Some(condition) => format!(
+                "#[cfg_attr({}, derive({}))]",
+                condition,
+                derive.name
+            ),
+            None => format!("#[derive({})]", derive.name),
+        };
+        try!(writeln!(w, "{}{}", tabs(indent), s));
+    }
+    Ok(())
 }
 
 pub fn doc_hidden(
--- a/src/config/derives.rs
+++ b/src/config/derives.rs
@@ -1,0 +1,39 @@
+use super::parsable::Parse;
+use toml::Value;
+use super::error::TomlHelper;
+
+#[derive(Clone, Debug)]
+pub struct Derive {
+    pub name: String,
+    pub cfg_condition: Option<String>,
+}
+
+impl Parse for Derive {
+    fn parse(toml: &Value, object_name: &str) -> Option<Derive> {
+        let name = match toml.lookup("name").and_then(|v| v.as_str()) {
+            Some(name) => name.to_owned(),
+            None => {
+                error!(
+                    "No 'name' given for derive for object {}",
+                    object_name
+                );
+                return None;
+            }
+        };
+        toml.check_unwanted(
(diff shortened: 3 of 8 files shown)


EPashkin:
@fengalin Code looks good, but problem that in normal struct cfg applied to it, not to next derive.
You sure that it works with bitflags?

EPashkin:
Seems not works: https://play.rust-lang.org/?gist=169b98323d26e8c6412214710d25ee8e&version=stable&mode=debug

Very bad.

EPashkin:
try remove `not`

fengalin:
Yes, you're right! Let me take a look at this...

EPashkin:
It seems possible: 
https://stackoverflow.com/questions/42551113/is-it-possible-to-conditionally-derive-with-features

src/config/gobjects.rs
```diff
@@ -199,6 +203,18 @@ fn parse_object(toml_object: &Value, concurrency: library::Concurrency) -> GObje
     };
     let members = Members::parse(toml_object.lookup("member"), &name);
     let properties = Properties::parse(toml_object.lookup("property"), &name);
+    let derives = {
+        let mut v = Vec::new();
+        if let Some(configs) = toml_object.lookup("derive").and_then(|val| val.as_array()) {
```
EPashkin: Maybe this better be as `Derives::parse(`, same as previous?

fengalin:
> It seems possible

Thanks. Implementing it now.

src/config/gobjects.rs
fengalin: Yes indeed

src/config/derives.rs
```diff
@@ -0,0 +1,53 @@
+use super::ident::Ident;
+use super::parsable::Parse;
+use toml::Value;
+use super::error::TomlHelper;
+
+#[derive(Clone, Debug)]
+pub struct Derive {
+    pub ident: Ident,
```
EPashkin: Better `pub name: String` instead, but then possible you need add manual parse.

EPashkin:
Working playground for watchers: https://play.rust-lang.org/?gist=0b9aff53230ae4fe6d2248ba3d4d578e&version=stable&mode=debug

src/config/derives.rs
EPashkin: `impl<T: Parse + AsRef<Ident>> Parsable for Vec<T> {` just don't need `AsRef` so it IMHO will be applied automatically.

fengalin:
Thanks for the review @EPashkin! It should be better now.

src/codegen/flags.rs
```diff
@@ -74,6 +74,17 @@ fn generate_flags(env: &Env, w: &mut Write, flags: &Bitfield, config: &GObject)
...
+        let derive_str = match &derive.cfg_condition {
+            Some(condition) => format!(
+                "#[cfg_attr(any({}, feature = \"dox\"), derive({}))]",
+                condition,
+                derive.name
+            ),
+            None => format!("#[derive({})]", derive.name),
+        };
+        try!(writeln!(w, "    {}", derive_str));
+    }
```
EPashkin: This is duplicate, so please extract it to `codegen\general.rs` with usage `tabs(indent)` instead prefix spaces

EPashkin:
Also better check doc generation, maybe `feature="dox"` will cause error without "ser_de" in real project

EPashkin:
.. and IMHO `[[object.derive]]name = "Serialize, Deserialize" will works too, so readme can be updated to reflect this

fengalin:
> Also better check doc generation, maybe feature="dox" will cause error without "ser_de" in real project

I couldn't find a way to detect doc generation without `feature="dox"` and for some reasons, doc is never generated for `Serialize` and `Deserialize` when used as `#[derive()]`, even without a condition. So I decided to keep only the condition and remove `feature="dox"`. I hope this is acceptable.

src/codegen/general.rs
```diff
@@ -353,6 +354,25 @@ pub fn cfg_condition_string(
     }
 }
 
+pub fn derives(
```
EPashkin: Please add `#[cfg_attr(feature = "cargo-clippy", allow(ptr_arg))]`

EPashkin:
@fengalin Thanks, only small nit remain.
Yes, without 'dox' is acceptable, as we can write in `cfg_condition` if needed.

Just in case: your old real code compiles with `dox` but without `ser-de` ?
I worried that it just don't compiles as `serde*` crates not included but `#[derive(Serialise)]` is.

src/codegen/general.rs
fengalin: > I fixed the clippy warning.

fengalin:
> your old real code compiles with dox but without ser-de ?

Sorry, my comment was not clear. The code compiles and behaves as expected. It's just the documentation for the `Serialize` and `Deserialize` `Trait` implementations for `enum`s and `struct`s that use `#[derive(Serialize, Deserialize)]` that is not generated.

EPashkin:
Strange.
Thank again, PR is great.

Issue #608, opened 2018-06-07
Many warnings from external libraries


jsparber:
I'm creating [libhandy rust bindings](https://gitlab.gnome.org/jsparber/libhandy-rs). I use gtk as external library therefore gir shows many warnings.
` WARN 2018-06-07T10:20:04Z: libgir::analysis::functions: Function gtk_widget_get_allocated_size has unsupported outs
 WARN 2018-06-07T10:20:04Z: libgir::analysis::functions: Function gtk_widget_get_allocation has unsupported outs
 WARN 2018-06-07T10:20:04Z: libgir::analysis::functions: Function gtk_widget_get_child_requisition has unsupported outs
 WARN 2018-06-07T10:20:04Z: libgir::analysis::functions: Function gtk_widget_get_clip has unsupported outs
 WARN 2018-06-07T10:20:04Z: libgir::analysis::functions: Function gtk_widget_get_preferred_size has unsupported outs
 WARN 2018-06-07T10:20:04Z: libgir::analysis::functions: Function gtk_widget_get_requisition has unsupported outs
 WARN 2018-06-07T10:20:04Z: libgir::analysis::functions: Function gtk_widget_intersect has unsupported outs
 WARN 2018-06-07T10:20:04Z: libgir::analysis::functions: Function gtk_widget_size_request has unsupported outs
 WARN 2018-06-07T10:20:04Z: libgir::analysis::trampolines: Can't generate Gtk.Widget trampoline for signal 'button-press-event'
 WARN 2018-06-07T10:20:04Z: libgir::analysis::trampolines: Can't generate Gtk.Widget trampoline for signal 'button-release-event'
 WARN 2018-06-07T10:20:04Z: libgir::analysis::trampolines: Can't generate Gtk.Widget trampoline for signal 'child-notify'
 WARN 2018-06-07T10:20:04Z: libgir::analysis::trampolines: Can't generate Gtk.Widget trampoline for signal 'configure-event'
 WARN 2018-06-07T10:20:04Z: libgir::analysis::trampolines: Can't generate Gtk.Widget trampoline for signal 'damage-event'
 WARN 2018-06-07T10:20:04Z: libgir::analysis::trampolines: Can't generate Gtk.Widget trampoline for signal 'delete-event'
 WARN 2018-06-07T10:20:04Z: libgir::analysis::trampolines: Can't generate Gtk.Widget trampoline for signal 'destroy-event'
[...]

PR #612, opened 2018-06-20
Show analysis warnings only for generated library

EPashkin:
Closes https://github.com/gtk-rs/gir/issues/608

cc @GuillaumeGomez, @jsparber


GuillaumeGomez:
:+1:

PR #617, opened 2018-07-01
Fix generating array with c:type="gpointer*" and other actual type

EPashkin:
Part of https://github.com/gtk-rs/gio/pull/131

cc @sdroege

### DIFF
--- a/src/codegen/sys/ffi_type.rs
+++ b/src/codegen/sys/ffi_type.rs
@@ -132,7 +132,9 @@ fn ffi_inner(env: &Env, tid: library::TypeId, mut inner: String) -> Result {
         Type::HashTable(..) => fix_name(env, tid, &inner),
         _ => if let Some(glib_name) = env.library.type_(tid).get_glib_name() {
             if inner != glib_name {
-                if implements_c_type(env, tid, &inner) {
+                if inner == "gpointer" {
+                    fix_name(env, tid, glib_name).map_any(|s| format!("*mut {}", s))
+                } else if implements_c_type(env, tid, &inner) {
                     info!(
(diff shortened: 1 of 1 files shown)


EPashkin:
In gtk-rs crates only affect gio-sys.
```diff
--- a/gio-sys/src/lib.rs
+++ b/gio-sys/src/lib.rs
@@ -8237,8 +8237,8 @@ extern "C" {
     pub fn g_list_store_remove_all(store: *mut GListStore);
     #[cfg(any(feature = "v2_46", feature = "dox"))]
     pub fn g_list_store_sort(store: *mut GListStore, compare_func: glib::GCompareDataFunc, user_data: gpointer);
-    //#[cfg(any(feature = "v2_44", feature = "dox"))]
-    //pub fn g_list_store_splice(store: *mut GListStore, position: c_uint, n_removals: c_uint, additions: /*Metadata mismatch*/*mut [c:type mismatch gpointer != GObject of Object], n_additions: c_uint);
+    #[cfg(any(feature = "v2_44", feature = "dox"))]
+    pub fn g_list_store_splice(store: *mut GListStore, position: c_uint, n_removals: c_uint, additions: *mut *mut gobject::GObject, n_additions: c_uint);
```

sdroege:
Nice, that's what I tried except for adding the other `*mut` :) is that always correct?

Thanks!

EPashkin:
IMHO is correct: `glib_type` is type name and it not contains `"*"`, but `c:type` says that it passed as `gpointer` so we add `"*"`.
Anyway its very rare occurrence.

sdroege:
Great! You regen sys?

PR #620, opened 2018-07-03
Prevent some useless generations

GuillaumeGomez:
However before merging, I'd like to be able to confirm that it works (on GObject) but I wasn't able to generate the GObject files for some reason... Am I missing something?

### DIFF
--- a/src/codegen/enums.rs
+++ b/src/codegen/enums.rs
@@ -40,6 +40,10 @@ pub fn generate(env: &Env, root_path: &Path, mod_rs: &mut Vec<String>) {
         }
     }
 
+    if !has_any {
+        return
+    }
+
     let mut imports = Imports::new(&env.library);
     imports.add("ffi", None);
     if has_get_quark {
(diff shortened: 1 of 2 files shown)


src/codegen/enums.rs
```diff
@@ -57,6 +57,10 @@ pub fn generate(env: &Env, root_path: &Path, mod_rs: &mut Vec<String>) {
     }
     imports.add("glib::translate::*", None);
 
+    if !has_any {
```
EPashkin: `if has_any` for "mod enums;" don't need then.

EPashkin: in flags `let mut first = true` unneeded to for "mod flags;"

src/codegen/enums.rs
```diff
@@ -57,6 +57,10 @@ pub fn generate(env: &Env, root_path: &Path, mod_rs: &mut Vec<String>) {
     }
     imports.add("glib::translate::*", None);
```
sdroege: Should probably also not any imports when not generating this file anyway

EPashkin: You right, `if !has_any {return}` need be sooner

GuillaumeGomez:
Updated! Once merged, I'll send the update for glib.

EPashkin:
@GuillaumeGomez Thanks, 👍

GuillaumeGomez:
@EPashkin Should we merge it then?

EPashkin:
Yes

PR #621, opened 2018-07-04
Fix panic in generating `emit` for incomplete signal

EPashkin:
Fix panic on GObject generation (see https://github.com/gtk-rs/gir/pull/620#issue-199102491).

cc @sdroege

### DIFF
--- a/src/codegen/signal.rs
+++ b/src/codegen/signal.rs
@@ -75,6 +75,11 @@ pub fn generate(
         }
     }
 
+    if function_type.is_none() {
+        // Signal incomplete, can't generate emit
+        return Ok(());
+    }
+
     if let Some(ref emit_name) = analysis.action_emit_name {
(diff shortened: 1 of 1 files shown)


EPashkin:
Also can be fixed by just adding "GObject.ParamSpec" to manual in Gir_GObject.toml,
but with this we can find what type need be deignored

src/codegen/signal.rs
```diff
@@ -75,6 +75,11 @@ pub fn generate(
         }
     }
 
+    if function_type.is_none() {
+        // Signal incomplete, can't generate emit
+        return Ok(());
```
sdroege: Makes sense, but which signal is that exactly?

sdroege: I see, makes sense

EPashkin:
Ok, let merge this then

PR #622, opened 2018-07-05
fix invalid multiple flag import

GuillaumeGomez:
cc @EPashkin

### DIFF
--- a/src/codegen/flags.rs
+++ b/src/codegen/flags.rs
@@ -58,9 +58,9 @@ pub fn generate(env: &Env, root_path: &Path, mod_rs: &mut Vec<String>) {
         try!(general::uses(w, env, &imports));
         try!(writeln!(w));
 
+        mod_rs.push("\nmod flags;".into());
         for config in &configs {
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Fixes https://github.com/gtk-rs/glib/compare/master...GuillaumeGomez:regen

src/codegen/flags.rs
```diff
@@ -58,9 +58,13 @@ pub fn generate(env: &Env, root_path: &Path, mod_rs: &mut Vec<String>) {
         try!(general::uses(w, env, &imports));
         try!(writeln!(w));
 
+        let mut first = true;
         for config in &configs {
             if let Type::Bitfield(ref flags) = *env.library.type_(config.type_id.unwrap()) {
-                mod_rs.push("\nmod flags;".into());
+                if first {
+                    mod_rs.push("\nmod flags;".into());
```
EPashkin: Just move it before for

GuillaumeGomez: Ah true, now the check is before... Good catch!

GuillaumeGomez:
Updated.

EPashkin:
Thanks.

PR #623, opened 2018-07-19
Bump pkg-config version

ferjm:
This [version](https://github.com/alexcrichton/pkg-config-rs/pull/70) adds support for MSVC builds.


sdroege:
That should not be necessary, we build fine with older versions but it would take the latest version it finds from the `0.3` releases. And if someone wants to build with MSVC they will have to get the newer version (and will most likely get it automatically already).

ferjm:
Ah, ok, got it. It should be a matter of changing the Cargo.lock on my side then. Thanks

sdroege:
Yeah, or simply `cargo update` :)

PR #627, opened 2018-08-02
Improve deriving of trait impls

sdroege:
See https://github.com/gtk-rs/glib/pull/367 for context

This is WIP

### DIFF
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -20,7 +20,7 @@ use analysis::safety_assertion_mode::SafetyAssertionMode;
 use analysis::signatures::{Signature, Signatures};
 use config;
 use env::Env;
-use library::{self, Function, FunctionKind, Nullable, Parameter, ParameterScope, Type};
+use library::{self, Function, FunctionKind, Nullable, Parameter, ParameterScope, Type, Transfer};
 use nameutil;
 use std::borrow::Borrow;
 use traits::*;
@@ -137,6 +137,61 @@ pub fn analyze<F: Borrow<library::Function>>(
     funcs
 }
 
+fn fixup_gpointer_parameter(
+    env: &Env,
+    type_tid: library::TypeId,
+    parameters: &mut Parameters,
+    idx: usize
+) {
+    use analysis::ffi_type;
+
+    let instance_parameter = idx == 0;
+
+    let glib_name = env.library.type_(type_tid).get_glib_name().unwrap();
+    let ffi_name = ffi_type::ffi_type(env, type_tid, &glib_name).unwrap();
+    parameters.rust_parameters[idx].typ = type_tid;
+    parameters.c_parameters[idx].typ = type_tid;
+    parameters.c_parameters[idx].instance_parameter = instance_parameter;
+    parameters.c_parameters[idx].ref_mode = RefMode::ByRef;
+    parameters.c_parameters[idx].transfer = Transfer::None;
+    parameters.transformations[idx] = Transformation {
+        ind_c: idx,
+        ind_rust: Some(idx),
+        transformation_type: TransformationType::ToGlibPointer {
+            name: parameters.rust_parameters[idx].name.clone(),
+            instance_parameter: instance_parameter,
+            transfer: Transfer::None,
+            ref_mode: RefMode::ByRef,
+            to_glib_extra: String::new(),
+            explicit_target_type: format!("*mut {}", ffi_name),
+            pointer_cast: " as glib_ffi::gconstpointer".into(),
+        },
+    };
+}
+
+fn fixup_special_functions(
+    env: &Env,
+    name: &str,
+    type_tid: library::TypeId,
+    parameters: &mut Parameters
+) {
+    // Workaround for some _hash() / _compare() / _equal() functions taking
+    // "gconstpointer" as arguments instead of the actual type
+    if name == "hash" && parameters.c_parameters.len() == 1 {
+        if parameters.c_parameters[0].c_type == "gconstpointer" {
+            fixup_gpointer_parameter(env, type_tid, parameters, 0);
+        }
+    }
+
+    if (name == "compare" || name == "equal" || name == "is_equal") && parameters.c_parameters.len() == 2 {
+        if parameters.c_parameters[0].c_type == "gconstpointer" &&
+           parameters.c_parameters[1].c_type == "gconstpointer" {
+            fixup_gpointer_parameter(env, type_tid, parameters, 0);
+            fixup_gpointer_parameter(env, type_tid, parameters, 1);
+        }
+    }
+}
+
 fn analyze_function(
     env: &Env,
     name: String,
@@ -188,6 +243,8 @@ fn analyze_function(
     );
     parameters.analyze_return(env, &ret.parameter);
 
+    fixup_special_functions(env, name.as_str(), type_tid, &mut parameters);
+
     for (pos, par) in parameters.c_parameters.iter().enumerate() {
         assert!(
             !par.instance_parameter || pos == 0,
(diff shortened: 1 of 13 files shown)


src/analysis/functions.rs
```diff
@@ -188,6 +188,67 @@ fn analyze_function(
     );
     parameters.analyze_return(env, &ret.parameter);
 
+    // Workaround for some _hash() / _compare() / _equal() functions taking
+    // "gconstpointer" as arguments instead of the actual type
```
sdroege: This is all a bit ugly (should be moved to a separate function too) but the main problem is that we use `to_glib_none().0` but the FFI functions expects a `*const c_void` and not a `*mut WhateverFFIType`.

@EPashkin How would we solve this best? Apart from that this is working fine

EPashkin: IMHO this solution fine if separated to function,
other option is replace type parameter on preprocess stage if mode normal (or in sys too to remove "*const void" from member functions)

sdroege: The problem is how to to make the FFI function accept the right type, or how to add the correct cast after `to_glib_none().0`.

Would you prefer to have those functions fixed up in FFI to take the actual type instead of gconstpointer?

EPashkin: I thought that `to_glib_extra` help with ` as *const _` but I was wrong.
Currently I don't see way to add something after `.to_glib_none().0"`.
Fixing in FFI IMHO removed uncertainness  about parameter meaning,
but it will don't works good for traited `compare`, so also not solution.
So actual solution seems add analog `to_glib_extra` to `TransformationType::ToGlibPointer` and fill it here

sdroege: ok thanks :)

sdroege: Works for GLib but not for `GObjects`

```
error[E0282]: type annotations needed
   --> src/auto/file.rs:816:30
    |
816 |             ffi::g_file_hash(self.to_glib_none().0 as glib_ffi::gconstpointer)
    |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type for `P`
```

Need to call a specific `to_glib_none()` here for the target type. Do we have infrastructure for that?

EPashkin: We don't have, I even can't write oneline here that return right type.
Seems for function in trait we need use 2 transformation:
TransformationType::ToGlibStash, 
new transformation that was part of TransformationType::Into for with_stash = true
Last need changed too to allow define type.
Name for new transformation IMHO better be "ToGlibStash" with old renamed to "UseGlibStash"
Problem with name of this variable, and I can't get working code :(
Last try
```
    fn hash(&self) -> u32 {
        unsafe {
            let this: Stash<*mut ffi::GFileIface, _> = self.to_glib_none();
            ffi::g_file_hash(this.0 as *const _)
        }
    }
```

sdroege:
```rust
    fn hash(&self) -> u32 {
        unsafe {
            ffi::g_file_hash(ToGlibPtr::<*mut ffi::GFile>::to_glib_none(self).0 as glib_ffi::gconstpointer)
        }
    }
```

This works. I guess that can relatively easily be special-cased

sdroege: @EPashkin How do I get a string of the FFI type name (`*mut ffi::GFile`) from a type ID in GIR?

EPashkin: You need add it too, all function in ffi_type.rs need "inner" string.
Use `Type::glib_get_name()` https://github.com/gtk-rs/gir/blob/master/src/analysis/ffi_type.rs#L126
and `fix_name` here.

sdroege: implemented that now and it seems to work. To the next step! :)

sdroege: Feel free to review the existing commits already btw, everything else is only on top of this

src/analysis/functions.rs
```diff
@@ -137,6 +137,89 @@ pub fn analyze<F: Borrow<library::Function>>(
...
+        if parameters.c_parameters[0].c_type == "gconstpointer" {
+            use analysis::ffi_type;
+
+            parameters.rust_parameters[0].typ = type_tid;
+            parameters.c_parameters[0].typ = type_tid;
+            parameters.c_parameters[0].instance_parameter = true;
+            parameters.c_parameters[0].ref_mode = RefMode::ByRef;
+            parameters.c_parameters[0].transfer = Transfer::None;
+            let glib_name = env.library.type_(type_tid).get_glib_name().unwrap();
+            let ffi_name = ffi_type::fix_name(env, type_tid, &glib_name).unwrap();
```
EPashkin: This 2 lines IMHO better be function in `ffi_type` without making `ffi_type::fix_name` public.

EPashkin:
Looks very good.
My only worry, that I don't understand why `impl hash::Hash` and other don't generated for traited Gio.File, Gio.Icon

src/analysis/functions.rs
```diff
@@ -137,6 +137,89 @@ pub fn analyze<F: Borrow<library::Function>>(
...
+            parameters.transformations[0] = Transformation {
+                ind_c: 0,
+                ind_rust: Some(0),
+                transformation_type: TransformationType::ToGlibPointer {
+                    name: parameters.rust_parameters[0].name.clone(),
+                    instance_parameter: true,
+                    transfer: Transfer::None,
+                    ref_mode: RefMode::ByRef,
+                    to_glib_extra: String::new(),
+                    explicit_target_type: format!("*mut {}", ffi_name),
```
EPashkin: Maybe better use "*const" here, as we IMHO always implemented it too

EPashkin: and maybe use this only if function in trait

sdroege: > Maybe better use "*const" here, as we IMHO always implemented it too

I think that's not implemented for `Shared`, not sure. Will check

sdroege:
> My only worry, that I don't understand why impl hash::Hash and other don't generated for traited Gio.File, Gio.Icon

See https://github.com/gtk-rs/glib/pull/367#discussion_r207158491 . It's intentional and expected

EPashkin:
@sdroege Thanks for reminder

sdroege:
> @sdroege Thanks for reminder

Basically trait based, non-dynamic-dispatch `Eq` etc mechanism are conceptually incompatible with subclassing

src/analysis/functions.rs
sdroege: Not implemented for `Object` at least, so let's keep it `*mut`

src/analysis/functions.rs
```diff
@@ -137,6 +137,89 @@ pub fn analyze<F: Borrow<library::Function>>(
...
+                    pointer_cast: " as glib_ffi::gconstpointer".into(),
+                },
+            };
```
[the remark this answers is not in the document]
sdroege: Fixed

src/analysis/functions.rs
sdroege: I can actually directly call `ffi_type::ffi_type()` here, no changes needed. Fixed

sdroege:
Ok fixed all the comments, now just have to bring it to the final decision about which traits to implement where. Will do so tomorrow

sdroege:
This should be all good now, and should be merged together with https://github.com/gtk-rs/glib/pull/367 and then regens everywhere

EPashkin:
@sdroege Thanks, code looks good.
I'm a little worried about `let instance_parameter = idx == 0;` but IMHO this can be fixed if caused problem.

After this merged https://github.com/gtk-rs/glib/pull/367 need update.

PR #628, opened 2018-08-19
List not bound parents in not bound mode

### LINKED ISSUE #595: Show when parent isn't bound/available in current crate
We had this issue on gtk-rs/sourceview#51: `View` is supposed to implement `Scrollable`, but since it wasn't listed as a manual dependency, it wasn't generated (which is logical). However, we had no warning from gir about this. `not_bound` should have at least reported it.

GuillaumeGomez:
Fixes #595.

cc @sdroege @EPashkin

### DIFF
--- a/src/library.rs
+++ b/src/library.rs
@@ -829,8 +829,8 @@ impl Library {
         for x in &self.namespace(MAIN_NAMESPACE).types {
             if let Some(ref x) = *x {
                 let name = x.get_name();
+                let full_name = format!("{}.{}", namespace_name, name);
                 if !not_allowed_ending.iter().any(|s| name.ends_with(s)) {
-                    let full_name = format!("{}.{}", namespace_name, name);
                     let version = x.get_deprecated_version();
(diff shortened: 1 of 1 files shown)


EPashkin:
@GuillaumeGomez Thanks

src/library.rs
```diff
@@ -844,6 +844,20 @@ impl Library {
                         }
                     }
                 }
+                if let Some(tid) = env.library.find_type(0, &full_name) {
+                    let gobject_id = env.library.find_type(0, "GObject.Object").unwrap();
+
+                    for &super_tid in env.class_hierarchy.supertypes(tid) {
+                        if super_tid != gobject_id &&
+                           env.type_status(&super_tid.full_name(&env.library)).ignored() {
+                            println!("[NOT GENERATED PARENT] {}.{} for {}.{}",
```
sdroege: Didn't we have macros for errors, warnings, debug, etc?

GuillaumeGomez: We do but I just did like I did previously for this mode.

PR #629, opened 2018-08-20
Remove parent duplicates for not bound mode

GuillaumeGomez:
The current output is a bit too noisy (especially on gtk haha).

cc @sdroege @EPashkin


EPashkin:
@GuillaumeGomez Thanks

PR #631, opened 2018-08-29
Remove Windows pkg-config hack from -sys build.rs

sdroege:
We can simply use pkg-config to list us all required shared libraries
instead of only taking the library paths.

----

pkg-config-rs works properly on MSVC and MinGW now

### DIFF
--- a/src/codegen/sys/build.rs
+++ b/src/codegen/sys/build.rs
@@ -103,28 +103,11 @@ fn find() -> Result<(), Error> {
         return Ok(())
     }
 
-    let target = env::var("TARGET").expect("TARGET environment variable doesn't exist");
-    let hardcode_shared_libs = target.contains("windows");
-
     let mut config = Config::new();
     config.atleast_version(version);
     config.print_system_libs(false);
-    if hardcode_shared_libs {
-        config.cargo_metadata(false);
-    }
     match config.probe(package_name) {
-        Ok(library) => {
-            if hardcode_shared_libs {
-                for lib_ in shared_libs.iter() {
-                    println!("cargo:rustc-link-lib=dylib={}", lib_);
-                }
-                for path in library.link_paths.iter() {
-                    println!("cargo:rustc-link-search=native={}",
-                             path.to_str().expect("library path doesn't exist"));
-                }
-            }
-            Ok(())
-        }
+        Ok(_) => Ok(()),
         Err(Error::EnvNoPkgConfig(_)) | Err(Error::Command { .. }) => {
             for lib_ in shared_libs.iter() {
                 println!("cargo:rustc-link-lib=dylib={}", lib_);


sdroege:
The way how it fails on AppVeyour now is caused by a misconfigured environment. Those Windows libraries must be in the library path so that the linker can actually find them as needed.

@EPashkin You run Windows, right? Can you check what's wrong there?

EPashkin:
@sdroege Have same error, investigating.

src/codegen/sys/build.rs
```diff
@@ -103,28 +103,11 @@ fn find() -> Result<(), Error> {
...
     match config.probe(package_name) {
-        Ok(library) => {
-            if hardcode_shared_libs {
-                for lib_ in shared_libs.iter() {
-                    println!("cargo:rustc-link-lib=dylib={}", lib_);
-                }
-                for path in library.link_paths.iter() {
-                    println!("cargo:rustc-link-search=native={}",
-                             path.to_str().expect("library path doesn't exist"));
-                }
```
EPashkin: For me only this for can be removed even after upgrading pkg-config to 0.3.14, but not sure that it works globally as I have msys64/bin with gtk's dlls in path.
I tested on examples\cairo_threads.rs, do this change works fine for cairo-sys.
But on gtk it not works:
removing dylib caused many
```
          D:\eap\rust\0\target\debug\deps\libgtk-d3ee1e1d9d9104f6.rlib(gtk-d3ee1e1d9d9104f6.1consvywiv34sg52.rcgu.o): In function `_$LT$gtk..border..MemoryManager$u20$as$u20$glib..boxed..BoxedMemoryManager$LT$gtk_sys..GtkBorder$GT$$GT$::free::h324bcaf1ec9c7496':
          D:/eap/rust/0/gtk\src/border.rs:20: undefined reference to `gtk_border_free'
```
removing `config.cargo_metadata(false);` caused
```
  = note: ld: cannot find -limm32
          ld: cannot find -ldwmapi
          ld: cannot find -lcfgmgr32
```

sdroege: Any idea why? Can you provide the corresponding (for each issue) cargo output (with `-vv`) of a clean build?

EPashkin: IMHO at minimum no working gtk static lib on windows, about others I don't have any idea.
Output here: https://gist.github.com/EPashkin/a07f84a48f1b791e382e22716e2da144

sdroege: Why would it try static libs to begin with?

The GTK undefined symbols happen because there's no `-lgtk-3` in the linker commandline. The gtk-sys build.rs does *not* output this (or anything) for whatever reason. Can you check why? The other -sys crates output the correct things.

The missing Windows libraries, can you check where they are on your system and why those directories (the ones with the correspnding .lib files) are not in the linker search path?

EPashkin:
Continue talk from https://github.com/gtk-rs/gir/pull/631#discussion_r213929575
There no `cargo:rustc-link-lib=gtk-3` in `_no_dylib.txt` as it has `config.cargo_metadata(false);`.
In `_no_dylib_metadata.txt` it present without `dylib`,
`libgtk-3.dll.a` also present in location by `"-L" "D:/P/msys64/mingw64/lib"`
and it don't contains string `imm32`, `dwmapi`, `cfgmgr32`

sdroege:
Ok, the cargo metadata is needed for proper functioning so let's ignore that case.

So the only actual problem is that the imm32, etc libraries are not found, correct? Where are they (and their corresponding .lib files) on your system and are they part of the linker search path?

EPashkin:
All 3 of those libraries is 32bit (only on 64bit system) and lies in different lib folder.
If I manually add `println!("cargo:rustc-link-search=native=D:/P/msys64/mingw64/x86_64-w64-mingw32/lib");` to gtk-sys's build-rs, then it builds and works too.
Not sure what to do with this.

sdroege:
rustc should've added that path to the linker path already (when using the mingw toolchain)

EPashkin:
This path not from rust installation, but from separate msys2x64, so rustc can't add it.
Set "LIBRARY_PATH" environment to "D:/P/msys64/mingw64/x86_64-w64-mingw32/lib"
helps me locally, but with it we need set it for all appveyor.yml on x64 build.
Long term solution is fix pgkconfix's `gtk+-3.0.pc` for msys2x64, but it will take time even if accepted

sdroege:
It's wrong that those libraries are in the .pc file actually... if they are not needed for linking they should not be there.

EPashkin:
Libraries not in .pc file, it also not in libgtk-3.dll.a as I can see (tried find strings),
but libgtk-3-0.dll depends on them
```
prefix=/mingw64
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include
targets=win32 broadway

gtk_binary_version=3.0.0
gtk_host=x86_64-w64-mingw32

Name: GTK+
Description: GTK+ Graphical UI Library
Version: 3.22.30
Requires: gdk-3.0 atk >= 2.15.1 cairo >= 1.14.0 cairo-gobject >= 1.14.0 gdk-pixbuf-2.0 >= 2.30.0 gio-2.0 >= 2.49.4
Requires.private: atk   epoxy >= 1.0
Libs: -L${libdir} -lgtk-3 
Cflags: -I${includedir}/gtk-3.0 -mms-bitfields
```

EPashkin:
But pkg-config get it from somewhere
```
> pkg-config --libs gtk+-3.0
-LD:/P/msys64/mingw64/lib -lgtk-3 -lgdk-3 -lgdi32 -limm32 -lshell32 -lole32 -Wl,-luuid -lwinmm -ldwmapi -lsetupapi -lcfgmgr32 -lz -lpangowin32
-1.0 -lpangocairo-1.0 -lpango-1.0 -latk-1.0 -lcairo-gobject -lcairo -lgdk_pixbuf-2.0 -lgio-2.0 -lgobject-2.0 -lglib-2.0 -lintl
```

EPashkin:
Also strange for 32bit mode:
It don't builds without LIBRARY_PATH with same error,
but builds with it even it links to 32 for 64 proxy, but failed to run.
Msys2x64 for it don't have separate libimm32.a for 32bit.

sdroege:
> But pkg-config get it from somewhere

grep for them in all pkg-config files you have. It's probably from GDK or PangoWin32

sdroege:
In any case, the problem here are a) broken pkg-config files (they should not include those libraries unless their API is used from any *headers*, they should only be in Libs.private here) and b) those libraries not in the correct places or environment variables not set correctly for GTK.

It's not a problem of pkg-config

EPashkin:
Yes, all 3 dlls present in gdk-3.0.pc, gdk-broadway-3.0.pc, gdk-win32-3.0.pc,
all 3 identifical:
```
prefix=/mingw64
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include
targets=win32 broadway

Name: GDK
Description: GTK+ Drawing Kit
Version: 3.22.30
Requires: pangowin32 pangocairo gdk-pixbuf-2.0 >= 2.30.0 cairo >= 1.14.0 cairo-gobject >= 1.14.0
Requires.private: gio-2.0 >= 2.49.4     cairo-win32 cairo epoxy >= 1.0 
Libs: -L${libdir} -lgdk-3  -lgdi32 -limm32 -lshell32 -lole32 -Wl,-luuid -lwinmm -ldwmapi -lsetupapi -lcfgmgr32 -lz
Cflags: -I${includedir}/gtk-3.0 
```

sdroege:
Ok, so what I said above then :) GDK and environment setup bug

EPashkin:
@sdroege This seems can be closed too.
Bad that it not works.

sdroege:
Because of bugs elsewhere, yes. We should report those at least :)

@nirbheek do you have any input here?

EPashkin:
Seems problem with rust version of gcc, there output for it (I added line endings for libs)
```
C:\Documents and Settings\eap\.rustup\toolchains\nightly-x86_64-pc-windows-gnu\lib\rustlib\x86_64-pc-windows-gnu\bin\gcc.exe --print-search-dirs

install: C:/Documents and Settings/eap/.rustup/toolchains/nightly-x86_64-pc-windows-gnu/lib/rustlib/x86_64-pc-windows-gnu/bin/../lib/gcc/x86_64-w64-mingw32/6.3.0/
programs: =C:/Documents and Settings/eap/.rustup/toolchains/nightly-x86_64-pc-windows-gnu/lib/rustlib/x86_64-pc-windows-gnu/bin/../libexec/gcc/x86_64-w64-mingw32/6.3.0/;C:/Documents and Settings/eap/.rustup/toolchains/nightly-x86_64-pc-windows-gnu/lib/rustlib/x86_64-pc-windows-gnu/bin/../libexec/gcc/;C:/Documents and Settings/eap/.rustup/toolchains/nightly-x86_64-pc-windows-gnu/lib/rustlib/x86_64-pc-windows-gnu/bin/../lib/gcc/x86_64-w64-mingw32/6.3.0/../../../../x86_64-w64-mingw32/bin/x86_64-w64-mingw32/6.3.0/;C:/Documents and Settings/eap/.rustup/toolchains/nightly-x86_64-pc-windows-gnu/lib/rustlib/x86_64-pc-windows-gnu/bin/../lib/gcc/x86_64-w64-mingw32/6.3.0/../../../../x86_64-w64-mingw32/bin/
libraries: =C:/Documents and Settings/eap/.rustup/toolchains/nightly-x86_64-pc-windows-gnu/lib/rustlib/x86_64-pc-windows-gnu/bin/../lib/gcc/x86_64-w64-mingw32/6.3.0/;
C:/Documents and Settings/eap/.rustup/toolchains/nightly-x86_64-pc-windows-gnu/lib/rustlib/x86_64-pc-windows-gnu/bin/../lib/gcc/;
C:/Documents and Settings/eap/.rustup/toolchains/nightly-x86_64-pc-windows-gnu/lib/rustlib/x86_64-pc-windows-gnu/bin/../lib/gcc/x86_64-w64-mingw32/6.3.0/../../../../x86_64-w64-mingw32/lib/x86_64-w64-mingw32/6.3.0/;
C:/Documents and Settings/eap/.rustup/toolchains/nightly-x86_64-pc-windows-gnu/lib/rustlib/x86_64-pc-windows-gnu/bin/../lib/gcc/x86_64-w64-mingw32/6.3.0/../../../../x86_64-w64-mingw32/lib/../lib/;
C:/Documents and Settings/eap/.rustup/toolchains/nightly-x86_64-pc-windows-gnu/lib/rustlib/x86_64-pc-windows-gnu/bin/../lib/gcc/x86_64-w64-mingw32/6.3.0/../../../x86_64-w64-mingw32/6.3.0/;
C:/Documents and Settings/eap/.rustup/toolchains/nightly-x86_64-pc-windows-gnu/lib/rustlib/x86_64-pc-windows-gnu/bin/../lib/gcc/x86_64-w64-mingw32/6.3.0/../../../../lib/;
C:/mingw630/x86_64-630-posix-seh-rt_v5-rev2/mingw64/mingw/lib/x86_64-w64-mingw32/6.3.0/;
C:/mingw630/x86_64-630-posix-seh-rt_v5-rev2/mingw64/mingw/lib/../lib/;
C:/Documents and Settings/eap/.rustup/toolchains/nightly-x86_64-pc-windows-gnu/lib/rustlib/x86_64-pc-windows-gnu/bin/../lib/gcc/x86_64-w64-mingw32/6.3.0/../../../../x86_64-w64-mingw32/lib/;
C:/Documents and Settings/eap/.rustup/toolchains/nightly-x86_64-pc-windows-gnu/lib/rustlib/x86_64-pc-windows-gnu/bin/../lib/gcc/x86_64-w64-mingw32/6.3.0/../../../;
C:/mingw630/x86_64-630-posix-seh-rt_v5-rev2/mingw64/mingw/lib/
```
I have no subfolders in C:/Documents and Settings/eap/.rustup/toolchains/nightly-x86_64-pc-windows-gnu/lib/rustlib/x86_64-pc-windows-gnu/lib/ so all of these path not exists

nirbheek:
Those paths seem to be completely broken. When configured correctly, gcc sets those paths relative to its current location, so you can move the toolchain directory and the dirs will still be correct.

sdroege:
This sounds like it (also) needs a bug reports against the Rust MinGW toolchain then: for what @nirbheek said as well as for not including the MinGW .libs for the system libraries

EPashkin:
Note: despite warning near rust gcc.exe it not done its primary task: if I build program until I get link error and then rename this gcc.exe link (only step that last) will work fine with msys2x64 gcc
```
gcc.exe contained in this folder cannot be used for compiling C files - it is onlyused as a linker. In order to be able to compile projects containing C code usethe GCC provided by MinGW or Cygwin.
```

sdroege:
Happened in the meantime

PR #633, opened 2018-08-30
Get real parent name

GuillaumeGomez:
That should be the last commit for this serie. Just realised that the first letter of the namespace wasn't capitalised for parents.


EPashkin:
😉 Thanks

PR #634, opened 2018-09-01
Show not bound functions and methods

### LINKED ISSUE #607: More helpful warning messages
When a function cannot be generated because a type is missing, it could be nice to have more helpful warning messages like:

```
cannot generate binding for XXX because gdk::YY is missing; try adding it in your Gir.toml as "Gdk.YY"
```

GuillaumeGomez:
Fixes #607.


EPashkin:
@GuillaumeGomez Thanks

PR #635, opened 2018-09-01
Fix not bound functions and methods filtering

GuillaumeGomez:
I completely failed my previous PR (only tested it on sourceview, shame on me...). However I still have an issue: some types have a `ns_id` equal to 0. Therefore, the filtering is failing because the full name doesn't match (`*` instead of `Gtk` for example). For example we have this issue for:

```
Gtk.TextIter::get_marks because of *.TextMark
```

Any idea @EPashkin?

### DIFF
--- a/src/library.rs
+++ b/src/library.rs
@@ -688,6 +688,28 @@ impl Type {
             _ => &[],
         }
     }
+
+    pub fn is_fundamental(&self) -> bool {
+        match *self {
+            Type::Fundamental(_) => true,
+            _ => false,
+        }
+    }
+
+    pub fn get_inner_type<'a>(&'a self, env: &'a Env) -> Option<(&'a Type, u16)> {
+        match *self {
+            Type::Array(t) |
+            Type::CArray(t) |
+            Type::FixedArray(t, _) |
+            Type::PtrArray(t) |
+            Type::List(t) |
+            Type::SList(t) => {
+                let ty = env.type_(t);
+                ty.get_inner_type(env).or_else(|| Some((ty, t.ns_id)))
+            }
+            _ => None,
+        }
+    }
 }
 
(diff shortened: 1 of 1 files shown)


src/library.rs
```diff
@@ -688,6 +688,26 @@ impl Type {
...
+
+    pub fn get_real_type<'a>(&'a self, env: &'a Env) -> &'a Type {
+        match *self {
+            Type::Array(t) |
+            Type::CArray(t) |
+            Type::FixedArray(t, _) |
+            Type::PtrArray(t) |
+            Type::List(t) |
+            Type::SList(t) => return env.type_(t).get_real_type(env),
+            _ => {}
```
EPashkin: This function need return not only `&'a Type` but `t.ns_id` as well.

GuillaumeGomez: Testing!

GuillaumeGomez:
@EPashkin: All fixed, thanks a lot, wonderful idea!

src/library.rs
```diff
@@ -688,6 +688,44 @@ impl Type {
...
+                            return inner(t, env)
+                        }
+                        _ => {}
+                    }
+                    let ns_id = t.ns_id;
+                    Some((ty, ns_id))
+                }
+                return inner(t, env);
+            }
+            _ => {}
```
EPashkin: Why not?
```
            Type::SList(t) => {
                let ty = env.type_(t);
                ty.get_inner_type(env).or(Some((ty, t.ns_id)))
            }
            _ => None
```

EPashkin: Although clippy want `.or_else(|| Some((ty, t.ns_id)))`

GuillaumeGomez:
Updated.

EPashkin:
@GuillaumeGomez Thanks

PR #636, opened 2018-09-05
Move from send-cell to fragile

sdroege:
The former is deprecated in favour of the latter and just a wrapper
around fragile nowadays.

----

This needs regens of all crates that have futures. I'll send those once this is merged

### DIFF
--- a/src/codegen/function.rs
+++ b/src/codegen/function.rs
@@ -314,7 +314,7 @@ pub fn body_chunk_futures(env: &Env, analysis: &analysis::functions::Info) -> St
     } else {
         try!(writeln!(body, "use GioFuture;"));
     }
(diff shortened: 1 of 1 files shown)


EPashkin:
IMHO `Fragile` from https://crates.io/crates/fragile is not replacement of `SendCell` as last allow use contained in other thread (if it have `reference_count==1`)

sdroege:
> IMHO Fragile from https://crates.io/crates/fragile is not replacement of SendCell as last allow use contained in other thread (if it have reference_count==1)

You're mixing `SendCell` from my crate with the `SendUnique` stuff we merged to `glib` last week :)

EPashkin:
Oh, you right.

Issue #639, opened 2018-09-16
Add support for generating only connect_property_xxx_notify without getter and setter


EPashkin:
See https://github.com/gtk-rs/gtk/pull/705#pullrequestreview-155750851

sdroege:
This is closely related to https://github.com/gtk-rs/gir/issues/510

Issue #510, opened 2017-12-15
Don't generate connect_xxx_notify for Read or ConstructOnly properties


EPashkin:
Example: `gio::IOStream`, `gio::FilterInputStream::base-stream`, `gio::FilterOutputStream::base-stream`

sdroege:
There are cases where read-only properties are changing from inside the object, and you want to be notified of those changes from the outside. If that also exists for construct-only properties, I don't know.

EPashkin:
Bad, then this need be done manually.
There way to check is readonly property can be changed from inside the object?

sdroege:
Only by checking its code (or the documentation) :)

EPashkin:
@GuillaumeGomez, @sdroege 
I tried to add warning in https://github.com/gtk-rs/gir/compare/master...EPashkin:partial_properties_warn_construct_only?expand=1, checked some in gtk, seems usable `notify` is rare case (still don't found any).
Sometimes (Gtk.MessageDialog.buttons) notify [fired](https://github.com/GNOME/gtk/blob/master/gtk/gtkmessagedialog.c#L843), but property not readable,
so I don't see any reason to fire as `gtk_message_dialog_add_buttons` is private, so will be fired only once before object returned to user.

Maybe we better just don't generate notify by default without warning,
but allow generation if needed.

sdroege:
What do you mean with "usable notify" being rare? It's a bug if they the property can change but notify is not triggering.

Do you mean not generating notify by default only for construct-only/read-only properties, or in general?

EPashkin:
Yes, I propose don't generate connect_xxx_notify for ConstructOnly properties by default.

sdroege:
That seems fine to me

EPashkin:
All PR in gtk-rs done and linked to this issue

PR #640, opened 2018-09-25
Partial properties generation

EPashkin:
Part of #639 and #510

cc @GuillaumeGomez, @sdroege 

IMHO needed other name instead "object.property.generate", maybe for `PropertyGenerateFlags` too.

WIP as I don't like current parser implementation and try to rework it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,11 @@ cfg_condition = "mycond"
     name = "baseline-position"
     version = "3.10"
     ignore = true
+    [[object.property]]
+    name = "events"
+    # generate only `connect_property_events_notify`, without `get_property_events` and `set_property_events`
+    # supported values: "get", "set", "notify"
+    generate = ["notify"]
 ```
 
 Since there are no child properties in `.gir` files, it needs to be added for classes manually:
--- a/src/analysis/properties.rs
+++ b/src/analysis/properties.rs
@@ -130,8 +130,10 @@ fn analyze_property(
     let generate = configured_properties
         .iter()
         .filter_map(|f| f.generate)
-        .next()
-        .unwrap_or_else(PropertyGenerateFlags::all);
+        .next();
+    let generate_set = generate.is_some();
+    let generate = generate.unwrap_or_else(PropertyGenerateFlags::all);
+
     let name_for_func = nameutil::signal_to_snake(&name);
     let var_name = nameutil::mangle_keywords(&*name_for_func).into_owned();
     let get_func_name = format!("get_property_{}", name_for_func);
@@ -139,12 +141,26 @@ fn analyze_property(
     let check_get_func_name = format!("get_{}", name_for_func);
     let check_set_func_name = format!("set_{}", name_for_func);
 
-    let mut readable = prop.readable && generate.contains(PropertyGenerateFlags::GET);
+    let mut readable = prop.readable;
     let mut writable = if prop.construct_only {
         false
     } else {
-        prop.writable && generate.contains(PropertyGenerateFlags::SET)
+        prop.writable
     };
+    if generate_set && generate.contains(PropertyGenerateFlags::GET) && !readable {
+        warn!(
+            "Attempt to generate getter for notreadable property \"{}\"",
+            name
+        );
+    }
+    if generate_set && generate.contains(PropertyGenerateFlags::SET) && !writable {
+        warn!(
+            "Attempt to generate setter for nonwritable property \"{}\"",
+            name
+        );
+    }
+    readable &= generate.contains(PropertyGenerateFlags::GET);
+    writable &= generate.contains(PropertyGenerateFlags::SET);
     let notifable = generate.contains(PropertyGenerateFlags::NOTIFY);
 
     if readable {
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -189,6 +189,12 @@ cfg_condition = "mycond"
     name = "baseline-position"
     version = "3.10"
     ignore = true
+    [[object.property]]
+    name = "events"
+    # generate only `connect_property_notify_event`, without `get_property_events` and `set_property_events`
```
GuillaumeGomez: Shouldn't it be `connect_property_notify_eventS`?

EPashkin: Strange, but is not:
```
    fn connect_property_notify_event<F: Fn(&Self, &gdk::EventProperty) -> Inhibit + 'static>(&self, f: F) -> SignalHandlerId {
        unsafe {
            let f: Box_<Box_<Fn(&Self, &gdk::EventProperty) -> Inhibit + 'static>> = Box_::new(Box_::new(f));
            connect(self.to_glib_none().0, "property-notify-event",
                transmute(property_notify_event_trampoline::<Self> as usize), Box_::into_raw(f) as *mut _)
        }
    }

```

GuillaumeGomez: Let's call it dark magic haha.

EPashkin: You was right, I looked to wrong function: `connect_property_events_notify`

GuillaumeGomez:
Seems good to me. Thanks a lot for adding it!

README.md
GuillaumeGomez: Ah now it seems more logical! (And explain why I couldn't find it...)

sdroege: > IMHO needed other name instead "object.property.generate", maybe for `PropertyGenerateFlags` too.

generate seems fine to me

GuillaumeGomez: Oh sorry, forgot to answer this question: seems fine to me as well.

README.md
```diff
@@ -189,6 +189,12 @@ cfg_condition = "mycond"
     name = "baseline-position"
     version = "3.10"
     ignore = true
+    [[object.property]]
+    name = "events"
+    # generate only `connect_property_events_notify`, without `get_property_events` and `set_property_events`
+    # supported values: "get", "set", "notify",
+    # also supported union "get|set" and inverting "not(notify)"
```
sdroege: Why so complicated that we even need nom? :)

I would simply list the things that should be generated. What is the use-case of negations?

EPashkin: For 3 only cases `not` really unusable, I was do it automatically.
Maybe better only support union with simple list splitting and parsing.
I also thought that we need other generate flags in future (for signal), but seems I mixed up with something.

sdroege: Yes better just have a list :)

EPashkin:
Changed expression parsing

README.md
```diff
@@ -189,6 +189,11 @@ cfg_condition = "mycond"
     name = "baseline-position"
     version = "3.10"
     ignore = true
+    [[object.property]]
+    name = "events"
+    # generate only `connect_property_events_notify`, without `get_property_events` and `set_property_events`
+    # supported values: "get", "set", "notify", also allowed union "get|set"
+    generate = "notify"
```
sdroege: Why not a toml list of strings?

sdroege:
Looks good to me otherwise

README.md
EPashkin: Then we need accept both array and single value: too much work on Toml types,
and only array IMHO redundantly.
Can change if @GuillaumeGomez prefer array too.

EPashkin:
@GuillaumeGomez Which config format for flags you prefer:
`generate="get|set"` or `generate=["get","set"]`?

EPashkin:
Ok, then I do it tomorrow

EPashkin:
Updated

src/analysis/properties.rs
```diff
@@ -128,19 +127,25 @@ fn analyze_property(
...
+        .unwrap_or_else(PropertyGenerateFlags::all);
     let name_for_func = nameutil::signal_to_snake(&name);
     let var_name = nameutil::mangle_keywords(&*name_for_func).into_owned();
     let get_func_name = format!("get_property_{}", name_for_func);
     let set_func_name = format!("set_property_{}", name_for_func);
     let check_get_func_name = format!("get_{}", name_for_func);
     let check_set_func_name = format!("set_{}", name_for_func);
 
-    let mut readable = prop.readable;
+    let mut readable = prop.readable && generate.contains(PropertyGenerateFlags::GET);
```
sdroege: It should probably warn if `GET` is configured but the property is write-only (and same for writable)

sdroege:
Seems good to me except for some cosmetic detail https://github.com/gtk-rs/gir/pull/640#pullrequestreview-159362116

src/analysis/properties.rs
EPashkin: Good idea

EPashkin:
Added commit with warning

EPashkin:
Maybe it need be error for faster detection

sdroege:
Fine with me either way :) Thanks!

EPashkin:
Then I merge it

PR #642, opened 2018-10-02
Add underscore to enum and bitflags member starting with digits

### LINKED ISSUE #600: Enum generation error for GstMpegts
In the file GstMpegts-1.0.gir there is an enum with specification starting:
```xml
    <enumeration name="DVBCodeRate" c:type="GstMpegtsDVBCodeRate">
      <member name="none" value="0" c:identifier="GST_MPEGTS_FEC_NONE">
      </member>
      <member name="1_2" value="1" c:identifier="GST_MPEGTS_FEC_1_2">
      </member>
      <member name="2_3" value="2" c:identifier="GST_MPEGTS_FEC_2_3">
      </member>
      <member name="3_4" value="3" c:identifier="GST_MPEGTS_FEC_3_4">
      </member>
      <member name="4_5" value="4" c:identifier="GST_MPEGTS_FEC_4_5">
      </member>
```
this generates the output starting:
```rust
#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
pub enum DVBCodeRate {
    None,
    12,
    23,
    34,
    45,
```
which isn't valid Rust code. :-(

Am I just missing an option in using `gir`:
```
gir -d ../gir-files -o . -c ../Gir_GstMpegts.toml
```

EPashkin:
Fix #600

@ceyusa can you test it?

cc @GuillaumeGomez, @sdroege, @russel

### DIFF
--- a/src/nameutil.rs
+++ b/src/nameutil.rs
@@ -5,7 +5,6 @@ use std::string::String;
 
 use case::*;
 
-
 pub fn split_namespace_name(name: &str) -> (Option<&str>, &str) {
     let mut parts = name.split('.');
     let name = parts.next_back().unwrap();
@@ -43,6 +42,22 @@ pub fn crate_name(name: &str) -> String {
 
 pub fn module_name(name: &str) -> String {
     mangle_keywords(name.to_snake()).into_owned()
+}
+
+pub fn enum_member_name(name: &str) -> String {
+    if name.starts_with(char::is_alphabetic) {
+        name.to_camel()
+    } else {
+        format!("_{}", name.to_camel())
+    }
+}
+
+pub fn bitfield_member_name(name: &str) -> String {
+    if name.starts_with(char::is_alphabetic) {
+        name.to_uppercase()
+    } else {
+        format!("_{}", name.to_uppercase())
+    }
 }
 
 pub fn needs_mangling(name: &str) -> bool {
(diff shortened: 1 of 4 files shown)


src/nameutil.rs
```diff
@@ -45,6 +44,22 @@ pub fn module_name(name: &str) -> String {
     mangle_keywords(name.to_snake()).into_owned()
 }
 
+pub fn enum_member_name(name: &str) -> String {
+    if name.starts_with(|c: char| c.is_digit(10)) {
```
sdroege: Maybe make this the other way around? If it does not start with a letter

src/nameutil.rs
```diff
@@ -45,6 +44,22 @@ pub fn module_name(name: &str) -> String {
...
+        format!("_{}", name.to_camel())
+    } else {
+        name.to_camel()
+    }
```
sdroege: And maybe error if name is not only containing alpha-num-underscore

src/nameutil.rs
EPashkin: You right

src/nameutil.rs
EPashkin: IMHO this unneeded, these members better ignored manually

sdroege: ok :)

ceyusa:
it works! or at least, it generates a compileable code :) thanks!

sdroege:
Let's go for that then :)

GuillaumeGomez:
Let's go for it then!

Issue #448, opened 2017-09-01
Generate special trait impls (Display specifically) for enums and flags


sdroege:
There are often to_string() functions available for them.

EPashkin:
Not sure if this really needed,
but as we already generating `Display` for classes. then we can do it too.

PR #643, opened 2018-10-09
Generate Display impl for enums

GuillaumeGomez:
Part of #448.

I have still troubles to fully understand how to correctly do it for flags but it'll come as well.

cc @EPashkin @sdroege


src/codegen/enums.rs
```diff
@@ -150,6 +151,25 @@ fn generate_enum(env: &Env, w: &mut Write, enum_: &Enumeration, config: &GObject
...
+    for member in &members {
+        try!(cfg_deprecated(w, env, member.deprecated_version, false, 3));
+        try!(version_condition(w, env, member.version, false, 3));
+        try!(writeln!(w, "\t\t\t{0}::{1} => \"{1}\",", enum_.name, member.name));
+    }
+    try!(writeln!(w,
+                  "\t\t\t_ => \"Unknown\",
+        }})
+    }}
+}}"));
```
EPashkin: IMHO extra new line needed here before next `#[doc(hidden)] impl ToGlib for xx {`

src/codegen/enums.rs
```diff
@@ -65,6 +65,7 @@ pub fn generate(env: &Env, root_path: &Path, mod_rs: &mut Vec<String>) {
     file_saver::save_to_file(path, env.config.make_backup, |w| {
         try!(general::start_comments(w, &env.config));
         try!(general::uses(w, env, &imports));
+        try!(writeln!(w, "use std::fmt;"));
```
EPashkin: This better be done with `imports.add`

EPashkin:
@GuillaumeGomez Thanks, 👍

src/codegen/enums.rs
```diff
@@ -150,6 +151,25 @@ fn generate_enum(env: &Env, w: &mut Write, enum_: &Enumeration, config: &GObject
...
+    try!(cfg_deprecated(w, env, enum_.deprecated_version, false, 0));
+    try!(version_condition(w, env, enum_.version, false, 0));
+    try!(writeln!(w,
+                  "impl fmt::Display for {0} {{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {{
+        write!(f, \"{0}::{{}}\", match *self {{", enum_.name));
+    for member in &members {
+        try!(cfg_deprecated(w, env, member.deprecated_version, false, 3));
+        try!(version_condition(w, env, member.version, false, 3));
+        try!(writeln!(w, "\t\t\t{0}::{1} => \"{1}\",", enum_.name, member.name));
```
sdroege: Two issues here
1) It should be configurable if this generated, just like what I added for other types. So that it can be manually overridden.
2) Use the "special functions" stuff from gir (just like we do for the various trait impls of other kinds of types). If there's e.g. a `to_string()` function, then this can be used here. I actually only thought (in the issue I created months ago) about creating the `Display` impl *only* if there's a special function for it.

EPashkin: Note: currently we don't generate enums and flags functions, so this need be done before, if we decide generate only for some enums.

sdroege: That should happen because otherwise this is going to break the GStreamer bindings (where I have some custom `Display` impls) :)

GuillaumeGomez:
Updated.

src/codegen/enums.rs
```diff
@@ -24,6 +24,7 @@ pub fn generate(env: &Env, root_path: &Path, mod_rs: &mut Vec<String>) {
     let mut has_get_quark = false;
     let mut has_any = false;
     let mut has_get_type = false;
+    let mut generate_display_trait = true;
```
EPashkin: IMHO this better be "= false":
If no enums then don't need add "use std::fmt",
also assignment will always "true"

src/codegen/enums.rs
```diff
@@ -33,6 +34,7 @@ pub fn generate(env: &Env, root_path: &Path, mod_rs: &mut Vec<String>) {
             if enum_.glib_get_type.is_some() {
                 has_get_type = true;
             }
+            generate_display_trait = config.generate_display_trait || generate_display_trait;
```
EPashkin: Maybe better `generate_display_trait |= config.generate_display_trait;`

src/codegen/object.rs
```diff
@@ -249,6 +262,19 @@ fn generate_trait(w: &mut Write, env: &Env, analysis: &analysis::object::Info) -
...
+        try!(writeln!(
+            w,
+            "impl fmt::Display for {} {{",
+            analysis.name,
+        ));
+        // Generate Display trait implementation.
+        try!(writeln!(w, "\tfn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {{\n\
+                            \t\twrite!(f, \"{}\")\n\
+                          \t}}\n\
+            }}\n", analysis.name));
```
EPashkin: For objects no empty line before and 2 empty lines at end

EPashkin: "gtk/auto/check_button.rs" has `use std::fmt` but don't have `impl fmt::Display` as it don't have `CheckButtonExt`

src/codegen/object.rs
```diff
@@ -11,9 +11,17 @@ use super::signal;
...
+    w: &mut Write,
+    env: &Env,
+    analysis: &analysis::object::Info,
+    generate_display_trait: bool,
+) -> Result<()> {
     try!(general::start_comments(w, &env.config));
     try!(general::uses(w, env, &analysis.imports));
+    if generate_display_trait {
+        writeln!(w, "use std::fmt;")?;
+    }
```
EPashkin: Maybe this better added to `analysis.imports` on analysis stage

src/codegen/enums.rs
GuillaumeGomez: Oh right! I had the fact that by default everything has an implemention of Display... Good catch!

src/codegen/enums.rs
GuillaumeGomez: I'm shared on this: it doesn't bring much and makes code a bit more difficult to read. I don't have strong opinion about it though... @sdroege any opinion on this?

GuillaumeGomez:
Updated.

src/codegen/enums.rs
sdroege: I agree with @EPashkin

sdroege:
Seems fine to me

GuillaumeGomez:
@EPashkin I'm merging then (so I can regenerate everything).

src/config/gobjects.rs
```diff
@@ -250,6 +253,10 @@ fn parse_object(toml_object: &Value, concurrency: library::Concurrency) -> GObje
         .lookup("use_boxed_functions")
         .and_then(|v| v.as_bool())
         .unwrap_or(false);
+    let generate_display_trait = toml_object
+        .lookup("generate_display_trait")
+        .and_then(|v| v.as_bool())
+        .unwrap_or(true);
```
sdroege: @GuillaumeGomez Why does this default to true still?

GuillaumeGomez: By "default" we want to generate the `Display` trait. However, if we have an invalid value, would you prefer instead?

sdroege: I would prefer for it to be opt-in

GuillaumeGomez: @EPashkin opened #668. You might want to take a look at it. :)

PR #647, opened 2018-10-27
Into wrapping

GuillaumeGomez:
This code is currently a huge mess and generate stuff like:

```
fn set_property_cell_background<'a, P: Into<Option<&'a Option<&str>>>>(&self, cell_background: P) {
```

Which is pretty funny but not what I want. So my main question is:

Why does it generate two `Option` wrapping instead of one? (I assume the `Into` wrapping generates one over a type which already has one)

A bit of help would be very appreciated here (I think I did it completely wrong). :)

Once this part done, I'll just have to generate the `.into` call and we'll be good to go I think.

### DIFF
--- a/src/codegen/properties.rs
+++ b/src/codegen/properties.rs
@@ -1,9 +1,11 @@
 use std::io::{Result, Write};
 
-use analysis::bounds::Bound;
+use analysis;
+use analysis::bounds::{Bound, Bounds};
 use analysis::properties::Property;
 use analysis::rust_type::{parameter_rust_type, rust_type};
 use chunk::Chunk;
+use codegen;
 use env::Env;
 use super::general::{cfg_deprecated, version_condition};
 use library;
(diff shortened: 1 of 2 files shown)


src/codegen/properties.rs
```diff
@@ -83,28 +84,39 @@ fn declaration(env: &Env, prop: &Property) -> String {
...
         {
-            let value_bound = if !prop.is_get {
+            if bound_type.is_into() {
+                let mut bounds = Bounds::default();
+                bounds.add_parameter(&prop.var_name,
+                                     &parameter_rust_type(env, prop.typ, dir, prop.nullable, prop.set_in_ref_mode).into_string(),
+                                     bound_type.clone(),
+                                     false);
+                bound = codegen::function::bounds(&bounds, &[], false);
+                format!("{}", bounds.iter().next().unwrap().alias)
```
EPashkin: Alias from inner bounds used while bounds destroyed,
maybe just change to TYPE_PARAMETERS_START ?

EPashkin: Ignore this, was wrong.

EPashkin: May be this helps: fix for non-trait cases:
```diff
--- a/src/codegen/properties.rs
+++ b/src/codegen/properties.rs
@@ -1,5 +1,6 @@
 use std::io::{Result, Write};
 
+use analysis;
 use analysis::bounds::{Bound, Bounds};
 use analysis::properties::Property;
 use analysis::rust_type::{parameter_rust_type, rust_type};
@@ -89,9 +90,10 @@ fn declaration(env: &Env, prop: &Property) -> String {
         }) = prop.bound
         {
             if bound_type.is_into() {
+                let type_name = parameter_rust_type(env, prop.typ, dir, library::Nullable(false), analysis::ref_mode::RefMode::ByRefFake).into_string();
                 let mut bounds = Bounds::default();
                 bounds.add_parameter(&prop.var_name,
-                                     &parameter_rust_type(env, prop.typ, dir, prop.nullable, prop.set_in_ref_mode).into_string(),
+                                     &type_name,
                                      bound_type.clone(),
                                      false);
                 bound = codegen::function::bounds(&bounds, &[], false);
```

GuillaumeGomez:
I was able to get a working version (thanks @EPashkin!).

EPashkin:
In enums.rs no newline between Display and ToGlib

src/codegen/properties.rs
```diff
@@ -83,28 +85,63 @@ fn declaration(env: &Env, prop: &Property) -> String {
...
-                } else {
-                    " + glib::value::SetValue"
+            use library::Type::*;
+            use analysis::bounds::BoundType;
+
+            let type_ = env.library.type_(prop.typ);
+            let bound_type = match *type_ {
+                Fundamental(_) => Some(bound_type.clone()),
+                _ => match bound_type {
+                    BoundType::Into(_, Some(ref x)) => Some(*x.clone()),
```
EPashkin: Despite this branch only string (`&str`) properties converted.
This strange.
Ex. for me `fn set_property_cell_background_rgba<'a, P: Into<Option<&'a gdk::RGBA>>>(&self, cell_background_rgba: P) {` works in fine, and don't needed extra '+SetValueOptional` etc.

GuillaumeGomez: Yes but for now we only use `Into` on `Fundamental` types and I'd like to keep it this way until we fully switch all `IsA` implementations.

EPashkin: Ok

GuillaumeGomez:
So I think it's ready. Tell me if anything needs to be updated. Once merged, I'll send PRs to other repositories.

EPashkin:
If you think that this enough for this PR, please, squash to single commit.

GuillaumeGomez:
Done!

EPashkin:
@GuillaumeGomez Thanks.
Affected crates: gio, gtk, sourceview.

PR #649, opened 2018-10-30
sys: generate cfg_condition for records

ceyusa:
Since GObject class structures are records from GIR perspective.

I just forgot to add this too :disappointed:


EPashkin:
Thanks, just waiting CI

PR #651, opened 2018-11-03
Fix adding `use IsA` for properties

EPashkin:
Fix generating https://github.com/sdroege/gstreamer-rs

Remove adding unused
```
use glib;
use glib::object::IsA;
```
to `gstreamer/src/auto/stream.rs`,
`gstreamer/src/auto/stream_collection.rs`,
`gstreamer-player/src/auto/player.rs` 

@sdroege Please check this.

### DIFF
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -31,6 +31,14 @@ impl BoundType {
         }
     }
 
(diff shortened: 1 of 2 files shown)


sdroege:
Makes sense, thanks

Issue #572, opened 2018-03-21
ABI checks have problems comparing signed/unsigned numbers


sdroege:
See e.g. in GStreamer
```
Constant value mismatch for GST_MESSAGE_ANY
Rust: "4294967295"
C:    "-1"
constants ... 272 passed; 3 failed (compilation errors: 0)
Constant value mismatch for GST_MESSAGE_DEVICE_ADDED
Rust: "2147483649"
C:    "-2147483647"
Constant value mismatch for GST_MESSAGE_DEVICE_REMOVED
Rust: "2147483650"
C:    "-2147483646"
Constant value mismatch for GST_MESSAGE_EXTENDED
Rust: "2147483648"
C:    "-2147483648"
Constant value mismatch for GST_MESSAGE_PROPERTY_NOTIFY
Rust: "2147483651"
C:    "-2147483645"
Constant value mismatch for GST_MESSAGE_REDIRECT
Rust: "2147483654"
C:    "-2147483642"
constants ... 292 passed; 8 failed (compilation errors: 0)
Constant value mismatch for GST_MESSAGE_STREAMS_SELECTED
Rust: "2147483653"
C:    "-2147483643"
Constant value mismatch for GST_MESSAGE_STREAM_COLLECTION
Rust: "2147483652"
C:    "-2147483644"
```

These are all the same value if casting the Rust value to a `c_int` (instead of `c_uint`), or casting the C value the other way around.

CC @EPashkin @tmiasko

tmiasko:
I would consider this to be a bug in g-ir-scanner. It should produce XML with
values in correct range. Otherwise you have no way to determine the underlying
type of enum because it depends on values in enumeration. In this case the
underlying type would be unsigned int when looking at XML, but int when looking
at C (the code actually goes out of the way to prevent changing the underlying
type to unsigned one).

I wonder if all actually works correctly, given that g-ir-repository determines
underlying integer type based on values in enumeration, see giroffsets.c. Though, I
wouldn't be completely surprised it values gets wrapped around before reaching
there.

sdroege:
C enums are always signed or not? Just what integer size is used is implementation defined

tmiasko:
From System V Application Binary Interface for AMD64: 

> C++ and some implementations of C permit enums larger than an int. The underlying
type is bumped to an unsigned int, long int or unsigned long int, in that order.

EPashkin:
Maybe we can add `dont_test` options for object?

[...]

PR #658, opened 2018-11-10
Cast enum/bitfield values to gint/guint for sys tests

sdroege:
GLib assumes that these are gint/guint internally so sign differences
don't really make a difference here, but cause test failures otherwise.

Fixes https://github.com/gtk-rs/gir/issues/572

### DIFF
--- a/src/codegen/sys/tests.rs
+++ b/src/codegen/sys/tests.rs
@@ -131,11 +131,22 @@ fn prepare_cconsts(env: &Env) -> Vec<CConstant> {
             continue;
         }
         match *typ {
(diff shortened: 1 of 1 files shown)


EPashkin:
Looks good,
but not better add field `signed:bool` to `CConstant` and print "(g[u]?int)" when writing `RUST_CONSTANTS` ?

EPashkin:
Hm, I missed that we have normal constants too.

EPashkin:
@sdroege I can merge current version, if you don't want bother with new field (it need be enum).

Issue #644, opened 2018-10-16
Trim in get_c_value causes gstreamer-sys-rs tests to fail.


alatiera:
While running the [gstreamer-rs-sys][1] tests, we stumbled in an issue with the `get_c_value` function which is auto generated in `tests/abi.rs`. In the log below, there an assertion of 2 Strings that is failing cause the leading space is [being trimmed][2]. The space exists in both the `.gir` file, the C headers and the generated C test file.

```
running 2 tests
test cross_validate_layout_with_c ... ok
test cross_validate_constants_with_c ... FAILED

failures:

---- cross_validate_constants_with_c stdout ----
constants ... 25 passed; 0 failed (compilation errors: 0)
constants ... 50 passed; 0 failed (compilation errors: 0)
constants ... 75 passed; 0 failed (compilation errors: 0)
Constant value mismatch for GST_AUDIO_FORMATS_ALL
Rust: " { S8, U8, S16LE, S16BE, U16LE, U16BE, S24_32LE, S24_32BE, U24_32LE, U24_32BE, S32LE, S32BE, U32LE, U32BE, S24LE, S24BE, U24LE, U24BE, S20LE, S20BE, U20LE, U20BE, S18LE, S18BE, U18LE, U18BE, F32LE, F32BE, F64LE, F64BE }"
C:    "{ S8, U8, S16LE, S16BE, U16LE, U16BE, S24_32LE, S24_32BE, U24_32LE, U24_32BE, S32LE, S32BE, U32LE, U32BE, S24LE, S24BE, U24LE, U24BE, S20LE, S20BE, U20LE, U20BE, S18LE, S18BE, U18LE, U18BE, F32LE, F32BE, F64LE, F64BE }"
constants ... 99 passed; 1 failed (compilation errors: 0)
constants ... 124 passed; 1 failed (compilation errors: 0)
constants ... 149 passed; 1 failed (compilation errors: 0)
constants ... 174 passed; 1 failed (compilation errors: 0)
thread 'cross_validate_constants_with_c' panicked at 'FAILED: 192 passed; 1 failed (compilation errors: 0)', gstreamer-audio-sys/tests/abi.rs:126:13
note: Run with `RUST_BACKTRACE=1` for a backtrace.


failures:
    cross_validate_constants_with_c

test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out

error: test failed, to rerun pass '-p gstreamer-audio-sys --test abi'
```

[...]

PR #659, opened 2018-11-10
Don't trim constants themselves when testing for their value

sdroege:
Otherwise tests will fail if the constants start/end with whitespace.

Fixes https://github.com/gtk-rs/gir/issues/644


EPashkin:
@sdroege Thanks

PR #660, opened 2018-11-15
Remove useless files

GuillaumeGomez:
Neither @sdroege nor @federicomenaquintero (nor me) remember why we have these files. Do you have any clue @EPashkin or should we just remove them?


EPashkin:
@GuillaumeGomez You right, it was used only on early stages.

PR #666, opened 2018-11-18
Generate code using CStringHolder instead of String

philn:
Generated code relying on CStringHolder should be more efficient because it
reduces the amount of string copies.

### DIFF
--- a/src/codegen/return_value.rs
+++ b/src/codegen/return_value.rs
@@ -8,14 +8,17 @@ use traits::*;
 use nameutil;
 
 pub trait ToReturnValue {
-    fn to_return_value(&self, env: &Env) -> String;
+    fn to_return_value(&self, env: &Env, is_trampoline: bool) -> String;
 }
 
 impl ToReturnValue for library::Parameter {
-    fn to_return_value(&self, env: &Env) -> String {
+    fn to_return_value(&self, env: &Env, is_trampoline: bool) -> String {
         let rust_type =
             parameter_rust_type(env, self.typ, self.direction, self.nullable, RefMode::None);
-        let name = rust_type.into_string();
+        let mut name = rust_type.into_string();
+        if is_trampoline && self.direction == library::ParameterDirection::Return && name == "GString" {
+            name = "String".to_owned();
+        }
         let type_str = match ConversionType::of(env, self.typ) {
             ConversionType::Unknown => format!("/*Unknown conversion*/{}", name),
             //TODO: records as in gtk_container_get_path_for_child
@@ -26,9 +29,9 @@ impl ToReturnValue for library::Parameter {
 }
 
 impl ToReturnValue for analysis::return_value::Info {
-    fn to_return_value(&self, env: &Env) -> String {
+    fn to_return_value(&self, env: &Env, is_trampoline: bool) -> String {
         match self.parameter {
-            Some(ref par) => par.to_return_value(env),
+            Some(ref par) => par.to_return_value(env, is_trampoline),
             None => String::new(),
         }
     }
--- a/src/analysis/child_properties.rs
+++ b/src/analysis/child_properties.rs
@@ -62,7 +62,7 @@ pub fn analyze(
 
     if !properties.is_empty() {
         imports.add("glib::object::IsA", None);
-        if let Some(s) = child_type.and_then(|typ| used_rust_type(env, typ).ok()) {
+        if let Some(s) = child_type.and_then(|typ| used_rust_type(env, typ, true).ok()) {
             imports.add_used_type(&s, None);
         }
     }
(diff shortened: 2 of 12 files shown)


philn:
CI fails on Pango for which I have a branch. Along with most of gtk-rs modules for which bindings need to be regenerated, I have branches for most of those at this point. I can open PRs for them now or after this one + the glib one is merged. Please advise ;)

EPashkin:
Seems it use need fixed as 'use glib::cstringingholder;`. Just add it to all imports (for glib it will be removed automatically).
```
error[E0433]: failed to resolve. Use of undeclared type or module `cstringholder`
  --> D:/eap/rust/0/gio\src\auto\settings_schema_source.rs:38:57
   |
38 |     pub fn list_schemas(&self, recursive: bool) -> (Vec<cstringholder::CStringHolder>, Vec<cstringholder::CStringHolder>) {
   |                                                         ^^^^^^^^^^^^^ Use of undeclared type or module `cstringholder`
```
Also adding it unneeded in pango\src\auto\coverage.rs, atk\src\auto\socket.rs, gio\src\auto\settings_backend.rs, gio\src\auto\subprocess_launcher.rs

philn:
Ah, I didn't test gio with the `v2_40` feature indeed, I'll fix it. I wanted to avoid "unused" warnings so I tried to import the module only when it's needed in the generated code.

philn:
gio-rs regen there, https://github.com/philn/gio/commit/ea028eb4d0f819345b3ea62ff8fc5352b4b2ff1d

philn:
PR updated!

src/codegen/return_value.rs
```diff
@@ -8,14 +8,17 @@ use traits::*;
...
 }
 
 impl ToReturnValue for library::Parameter {
-    fn to_return_value(&self, env: &Env) -> String {
+    fn to_return_value(&self, env: &Env, is_trampoline: bool) -> String {
         let rust_type =
             parameter_rust_type(env, self.typ, self.direction, self.nullable, RefMode::None);
-        let name = rust_type.into_string();
+        let mut name = rust_type.into_string();
+        if is_trampoline && self.direction == library::ParameterDirection::Return && name == "gstring::GString" {
```
EPashkin: "GString"

src/codegen/return_value.rs
```diff
@@ -8,14 +8,17 @@ use traits::*;
...
+            name = std::string::String::from("String");
```
EPashkin: "String".to_owned() or .into()

src/library.rs
```diff
@@ -277,6 +277,14 @@ impl TypeId {
...
         TypeId { ns_id: 0, id: 33 }
     }
+
+    pub fn tid_gstrv() -> TypeId {
+        TypeId { ns_id: 0, id: 34 }
+    }
+
+    pub fn tid_utf8_list() -> TypeId {
+        TypeId { ns_id: 0, id: 258 }
+    }
```
EPashkin: How these 2 types looks in .gir file?
Property `ScaleButton::get_property_icons(&self) -> Vec<GString>` don't detected by it :(
So maybe its don't have constant id and need be added in Library::new after fundamentals.

EPashkin:
@philn After this paths, you can remove all manual imports, at minimum on gtk-rs 😉 
https://gist.github.com/EPashkin/4780324a59f986be68028f152ce7cf07

philn:
Awesome! I'll check this tomorrow, thanks!

src/library.rs
philn: The icons property has the following type in the gir file:
```
        <array>
          <type name="utf8"/>
        </array>
```

philn: I already see `Type::c_array(&mut library, TypeId::tid_utf8(), None, None);` in Library::new()  :/

EPashkin: Strange that you get this big number then,
if this really same type, then `Library::new` need assert.
PS. And what about second added type?

philn: I don't remember for which module I added `gstrv` support actually. I commented it out and no new error appeard :confused:

philn:
PR updated, much nicer thanks to your patch @EPashkin :)

EPashkin:
@philn Now it really much shorter and without new warning. Thanks.
Can you add this patch as final touch?
It use `&GString::from_glib_borrow` in trampolines
```diff
+++ b/src/analysis/trampoline_parameters.rs
@@ -137,6 +137,7 @@ pub fn analyze(
 
         let conversion_type = {
             match *env.library.type_(par.typ) {
+                library::Type::Fundamental(library::Fundamental::Utf8) |
                 library::Type::Record(..) |
                 library::Type::Interface(..) |
                 library::Type::Class(..) => ConversionType::Borrow,
```

philn:
Done. Thanks for the reviews and help

EPashkin:
@philn Thanks again, Lets merge to allow update glib's PR.
Don't forget use `glib/Gir_GObject.toml` too 😉

PR #668, opened 2018-11-18
Default generate_display_trait

EPashkin:
Fix problem in https://github.com/gtk-rs/gir/pull/643#pullrequestreview-176087849

cc @GuillaumeGomez , @sdroege


EPashkin:
There also possible to separate defaults for enum and objects, but IMHO it unnecessary,
and will need more complex change

GuillaumeGomez:
You added a global option for it? Sounds actually like a good idea. :) But then, could you precise in the example that if it's not set on an object, then it's set for all objects please?

EPashkin:
Updated readme.txt

sdroege:
Seems fine to me but I'd still prefer it to be disabled by default :) These `Display` impls we generate are generally not what `Display` is meant for apart from very few cases

GuillaumeGomez:
I'd prefer to keep it on by default. However, let's vote.

@GuillaumeGomez votes for it.
@sdroege  votes against it.

What about you @EPashkin?

sdroege:
I should probably add my rationale also :) `Display` is meant for things to show to the user, but these impls are not providing anything that is useful for users. They are useful for programmers, and that's what `Debug` is for

GuillaumeGomez:
Maybe but it also allows to use the `ToString` trait, which is very convenient whereas `Debug` doesn't. :)

And both `Displayt` and `Debug` are for programmers in GUI applications development. In here, they just don't have the same purpose.

EPashkin:
I also think that `Display` don't usable for end users for GUI elements anyway,
maybe with exception for some row-like objects or other rare cases,
so generating `Display` instead `Debug` is acceptable in this case.
So I vote for generating by default too.

PR #669, opened 2018-11-19
Implement basics for subclassing

sdroege:
This passes the Rust class struct to glib_wrapper! if configured to do
so but still requires it to be manually implemented for the time being.

See https://github.com/gtk-rs/glib/pull/392

### DIFF
--- a/src/analysis/object.rs
+++ b/src/analysis/object.rs
@@ -17,6 +17,7 @@ pub struct Info {
     pub base: InfoBase,
     pub c_type: String,
     pub c_class_type: Option<String>,
+    pub rust_class_type: Option<String>,
     pub get_type: String,
     pub supertypes: Vec<general::StatusedTypeId>,
     pub generate_trait: bool,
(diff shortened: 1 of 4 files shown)


sdroege:
Example usage is here https://gitlab.freedesktop.org/slomo/gstreamer-rs/commit/735a09e2224d390d9253b24d52e69156d72b4d41

Later I'd like to automatically generate the class structs but that will have to wait a while. It's on my list after I've ported all the GStreamer things though, so next week or the week after probably.

Generally, I'd like to do the next steps gradually, one by one, so it would be good to have this merged before everything is in place already.

src/analysis/object.rs
```diff
@@ -97,6 +98,9 @@ pub fn class(env: &Env, obj: &GObject, deps: &[library::TypeId]) -> Option<Info>
     if obj.generate_display_trait {
         imports.add("std::fmt", None);
     }
+    if obj.subclassing {
+        imports.add(&format!("{}Class", name), None);
+    }
```
EPashkin: Maybe move down to `rust_class_type` and combine?

EPashkin:
Looks very good, thanks

src/analysis/object.rs
sdroege: Thanks, done!

src/analysis/object.rs
```diff
@@ -207,6 +208,13 @@ pub fn class(env: &Env, obj: &GObject, deps: &[library::TypeId]) -> Option<Info>
         imports.add("glib::ObjectExt", None);
     }
 
+    let rust_class_type = if obj.subclassing {
+        imports.add(&format!("{}Class", name), None);
+        Some(format!("{}Class", name))
```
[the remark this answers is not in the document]
sdroege: Because it was late :) Thanks, fixed!

PR #671, opened 2018-11-27
Remove now unneeded use statements for the glib_wrapper! macro

sdroege:
See https://github.com/gtk-rs/glib/pull/398

This is not complete. There's still the `std::ptr` added from `src/analysis/out_parameters.rs:analyze_imports()`. It seems like the condition there is wrong but I don't know what it should be. It seems to be related to `src/codegen/function_body_chunk.rs:c_type_mem_mode_lib()`.

@EPashkin any ideas?

### DIFF
--- a/src/analysis/record.rs
+++ b/src/analysis/record.rs
@@ -8,6 +8,7 @@ use nameutil::*;
 use super::*;
 use super::imports::Imports;
 use super::info_base::InfoBase;
+use super::record_type::RecordType;
 use traits::*;
 
 #[derive(Debug, Default)]
@@ -75,13 +76,13 @@ pub fn new(env: &Env, obj: &GObject) -> Option<Info> {
     let use_boxed_functions = obj.use_boxed_functions;
 
     let mut imports = Imports::with_defined(&env.library, &name);
-    imports.add("glib::translate::*", None);
     imports.add("ffi", None);
-    imports.add("glib_ffi", None);
-    imports.add("std::mem", None);
-    imports.add("std::ptr", None);
     if record.glib_get_type.is_some() {
-        imports.add("gobject_ffi", None);
+        if use_boxed_functions {
+            imports.add("gobject_ffi", None);
+        } else if let RecordType::AutoBoxed = RecordType::of(&record) {
+            imports.add("gobject_ffi", None);
+        }
     }
 
     let mut functions = functions::analyze(
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -171,6 +171,7 @@ fn fixup_gpointer_parameter(
 
 fn fixup_special_functions(
     env: &Env,
+    imports: &mut Imports,
     name: &str,
     type_tid: library::TypeId,
     parameters: &mut Parameters
@@ -180,6 +181,7 @@ fn fixup_special_functions(
     if name == "hash" && parameters.c_parameters.len() == 1 {
         if parameters.c_parameters[0].c_type == "gconstpointer" {
             fixup_gpointer_parameter(env, type_tid, parameters, 0);
+            imports.add("glib_ffi", None);
         }
     }
 
@@ -188,6 +190,7 @@ fn fixup_special_functions(
            parameters.c_parameters[1].c_type == "gconstpointer" {
             fixup_gpointer_parameter(env, type_tid, parameters, 0);
             fixup_gpointer_parameter(env, type_tid, parameters, 1);
+            imports.add("glib_ffi", None);
         }
     }
 }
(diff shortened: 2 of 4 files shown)


EPashkin:
IMHO all in analyze_imports is good.
It used for in functions like https://github.com/gtk-rs/gtk/blob/master/src/auto/builder.rs#L123-L129

sdroege:
It's the only place that adds std::ptr and in gstreamer (e.g. GstClock) it's added without being used

EPashkin:
Maybe there commented function with out parameter?

EPashkin:
Found one problem function `gst_clock_id_wait`.
We don't checks for `Type::Alias` here.

EPashkin:
But also `caller-allocates="0"` for `gint64` out parameter is IMHO wrong anyway,
there also at minimum `gst_clock_add_observation` with out `double` that also have that mark,
so maybe gir don't need change for this part,

sdroege:
So what do you suggest for fixing the GStreamer clock part exactly? The check there is apparently wrong but it's not clear to me what the correct one would be :)

For the deletion breaking things in other places, that means that the code generator needs to re-add the use statements in the other places where it actually makes use of the modules. I'll look into that at later.

EPashkin:
I checked Gtk-3.0.gir: seems it normal to use `caller-allocates="0"` for normal types :(
Then we have simply solution: in https://github.com/gtk-rs/gir/blob/master/src/analysis/out_parameters.rs#L124 calculate `ConversionType` and add import if it `Direct` or `Scalar`, no special case for `Type::Alias` needed.

EPashkin:
* add import if it NOT Direct or Scalar.

EPashkin:
@sdroege Only last errors in gdk and atk.
Agree that we don't need hurry with regens.

Errors can be fixed with something like:
```diff
+++ b/src/analysis/record.rs
@@ -8,6 +8,7 @@ use nameutil::*;
 use super::*;
 use super::imports::Imports;
 use super::info_base::InfoBase;
+use super::record_type::RecordType;
 use traits::*;
 
 #[derive(Debug, Default)]
@@ -77,6 +78,13 @@ pub fn new(env: &Env, obj: &GObject) -> Option<Info> {
     let mut imports = Imports::with_defined(&env.library, &name);
     imports.add("glib::translate::*", None);
     imports.add("ffi", None);
+    if use_boxed_functions {
+        imports.add("gobject_ffi", None);
+    } else if record.glib_get_type.is_some() {
+        if let RecordType::AutoBoxed = RecordType::of(&record) {
+            imports.add("gobject_ffi", None);
+        }
+    }
 
     let mut functions = functions::analyze(
         env,
```

EPashkin:
.. was slightly wrong, `glib_get_type` check need be outer if

sdroege:
Next try :) This works in atk and gdk for me

src/analysis/record.rs
```diff
@@ -77,11 +78,12 @@ pub fn new(env: &Env, obj: &GObject) -> Option<Info> {
...
     imports.add("ffi", None);
-    imports.add("glib_ffi", None);
-    imports.add("std::mem", None);
-    imports.add("std::ptr", None);
     if record.glib_get_type.is_some() {
-        imports.add("gobject_ffi", None);
+        if let RecordType::AutoBoxed = RecordType::of(&record) {
+            imports.add("gobject_ffi", None);
+        } else if use_boxed_functions {
+            imports.add("gobject_ffi", None);
```
EPashkin: Not better do fast check "use_boxed_function" first and then do slow "RecordTypeOf" ?

EPashkin:
This patch removes last warning in atk/gdk.
```diff
diff --git a/src/analysis/functions.rs b/src/analysis/functions.rs
index 90f9c4f..f6e57d1 100644
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -362,6 +362,7 @@ fn analyze_function(
         if ret.base_tid.is_some() {
             imports.add("glib::object::Downcast", None);
         }
+        imports.add("glib::translate::*", version);
         bounds.update_imports(imports);
     }
 
diff --git a/src/analysis/record.rs b/src/analysis/record.rs
index f7fbaf5..796f309 100644
--- a/src/analysis/record.rs
+++ b/src/analysis/record.rs
@@ -76,7 +76,6 @@ pub fn new(env: &Env, obj: &GObject) -> Option<Info> {
     let use_boxed_functions = obj.use_boxed_functions;
 
     let mut imports = Imports::with_defined(&env.library, &name);
-    imports.add("glib::translate::*", None);
     imports.add("ffi", None);
     if record.glib_get_type.is_some() {
         if use_boxed_functions {
```

src/analysis/record.rs
sdroege: Done

sdroege:
Thanks, done! Should be good to go now :)

EPashkin:
Ok, lets merge then.

PR #672, opened 2018-11-29
WIP: Don't generate any but the main IsA<_> bound for extension traits

sdroege:
Closes https://github.com/gtk-rs/gir/issues/670

PR #673, opened 2018-12-08
Various minor fixes

sdroege:
commit bcfdfe6584cb9d934eb22ac150889a7d8b1c9924 (HEAD -> raw-strings)
Author: Sebastian Dröge <sebastian@centricular.com>
Date:   Sat Dec 8 10:58:23 2018 +0200

    Also generate display trait for shorthand types only if the global default says so

commit 6178a527b14e9f226db2b7f7758e23bf3cdc3488
Author: Sebastian Dröge <sebastian@centricular.com>
Date:   Sat Dec 8 10:48:03 2018 +0200

    Only insert casts in property getters/setters if we're generating them for a trait
    
    Otherwise compilation fails because type annotations would be needed.

commit 83a8280c1b9cbdedee5cc57323ca1f424f94383f
Author: Sebastian Dröge <sebastian@centricular.com>
Date:   Sat Dec 8 10:33:38 2018 +0200

    Use raw 0-terminated strings for connecting to signals and accessing properties
    
    Going through `s.to_glib_none().0` involves another allocation and copy.

### DIFF
--- a/src/analysis/properties.rs
+++ b/src/analysis/properties.rs
@@ -273,7 +273,7 @@ fn analyze_property(
             if generate_trait {
                 imports.add("glib::object::Downcast", prop_version);
             }
-            imports.add("glib::signal::connect", prop_version);
+            imports.add("glib::signal::connect_raw", prop_version);
             imports.add("glib::signal::SignalHandlerId", prop_version);
             imports.add("std::mem::transmute", prop_version);
             imports.add("std::boxed::Box as Box_", prop_version);
--- a/src/analysis/signals.rs
+++ b/src/analysis/signals.rs
@@ -103,7 +103,7 @@ fn analyze_signal(
         if in_trait {
             imports.add("glib::object::Downcast", version);
         }
-        imports.add("glib::signal::connect", version);
+        imports.add("glib::signal::connect_raw", version);
         imports.add("glib::signal::SignalHandlerId", version);
         imports.add("std::mem::transmute", version);
         imports.add("std::boxed::Box as Box_", version);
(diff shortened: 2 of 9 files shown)


sdroege:
Depends on https://github.com/gtk-rs/glib/pull/406

EPashkin:
@sdroege Thanks, looks good and result build fine locally.
I hope https://github.com/gtk-rs/gir/pull/672#issuecomment-443424649 fixed too.

sdroege:
>  I hope [#672 (comment)](https://github.com/gtk-rs/gir/pull/672#issuecomment-443424649) fixed too.

Yes, that's the change to handle `in_trait`

EPashkin:
Ok, then lets merge, thanks again

EPashkin:
Gtk-rs fails with same error on 1.28.0 too.
Next fix forks and even don't produce "newer used "use glib" warnings",
but it don't equal generating code in https://github.com/gtk-rs/gir/blob/master/src/codegen/properties.rs#L120

```+++ b/src/analysis/properties.rs
@@ -108,6 +108,7 @@ pub fn analyze(
             if let Some(ref bound, ..) = prop.bound {
                 if bound.bound_type.need_isa() {
                     imports.add("glib::object::IsA", prop.version);
+                    imports.add("glib", prop.version);
                 }
             }
```

Note: current version of glib/src/gstring.rs fails in 1.28.0 to without "use std;" which is strange for me.
cc @philn

sdroege:
> Note: current version of glib/src/gstring.rs fails in 1.28.0 to without "use std;" which is strange for me.

You need to `use std` to be able to do `std::bla`. `::std::bla` is what works without.

sdroege:
Thanks for the fix above, I'll take a look at that (and removal of the additional bound) in a bit. Ideally this should only have the bound for `IsA` and nothing else.

PR #674, opened 2018-12-08
Add missing "glib" use statement for property setters

sdroege:
And remove unneeded IsA<Object> bound from them

----

See https://github.com/gtk-rs/gir/pull/673#issuecomment-445453829

I'll get rid of the other unneeded bounds at a later time


sdroege:
https://gitlab.freedesktop.org/slomo/gstreamer-rs/-/jobs/59177 looks fine with this

EPashkin:
And gtk-rs too.
Thanks.

PR #677, opened 2018-12-20
Revert "They weren't meant to be bit fields"

tmiasko:
This reverts commit 69ecbe911e8b3491a8d98e20ba98c8232f5e398d.

We assumed those changed would be covered by new tests, but since neither
@sdroege  nor @GuillaumeGomez are interested in maintaining two test suites,
revert those changes at least until after migration to ctest is complete.


GuillaumeGomez:
I suppose we can keep them meanwhile. At least I don't personnaly see any troubles doing that.

sdroege:
Yes, let's just get the tests done soonish and not add more back and forth

PR #679, opened 2018-12-20
Mark autogenerated files as such

kornelski:


EPashkin:
Thanks notice this, but better use `try!(general::start_comments(w, &env.config));` instead.

sdroege:
:+1: once @EPashkin's change is done :) Thanks!

EPashkin:
@kornelski If you don't have time to fix PR, I will add replacement PR on next Monday.

kornelski:
I've updated the PR

src/codegen/sys/build.rs
```diff
@@ -22,10 +23,13 @@ pub fn generate(env: &Env) {
 }
 
 fn generate_build_script(w: &mut Write, env: &Env) -> Result<()> {
+    try!(general::start_comments(w, &env.config));
     try!(write!(
         w,
         "{}",
-        r##"extern crate pkg_config;
+        r##"
+
```
EPashkin: Please make only one empty line with `try!(writeln!(w));` instead

kornelski: That's weird. Why? 

(also why use try! when the crate requires Rust 1.19?)

kornelski: Updated PR

EPashkin:
@kornelski Thanks, waiting for CI just in case.
After merge I will add PR in sys

PR #680, opened 2018-12-25
Fix some clippy warnings

EPashkin:
also add clippy check in travis

### DIFF
--- a/src/analysis/function_parameters.rs
+++ b/src/analysis/function_parameters.rs
@@ -141,7 +141,7 @@ impl Parameters {
     }
 }
 
-#[cfg_attr(feature = "cargo-clippy", allow(useless_let_if_seq))]
+#[allow(clippy::useless_let_if_seq)]
 pub fn analyze(
     env: &Env,
     function_parameters: &[library::Parameter],
--- a/src/analysis/signatures.rs
+++ b/src/analysis/signatures.rs
@@ -65,11 +65,11 @@ impl Signature {
         (false, None)
     }
 
-    pub fn eq(&self, other: &Signature) -> bool {
+    fn eq(&self, other: &Signature) -> bool {
         other.1 == self.1 && other.0[1..] == self.0[1..]
     }
 
-    pub fn property_eq(&self, other: &Signature, is_get: bool) -> bool {
+    fn property_eq(&self, other: &Signature, is_get: bool) -> bool {
         if self.eq(other) {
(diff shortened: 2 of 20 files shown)


src/analysis/function_parameters.rs
```diff
@@ -141,7 +141,7 @@ impl Parameters {
     }
 }
 
-#[cfg_attr(feature = "cargo-clippy", allow(useless_let_if_seq))]
+#[cfg_attr(feature = "cargo-clippy", allow(clippy::useless_let_if_seq))]
```
sdroege: You don't need the feature anymore

src/analysis/signatures.rs
```diff
@@ -65,12 +65,12 @@ impl Signature {
         (false, None)
     }
 
-    pub fn eq(&self, other: &Signature) -> bool {
+    pub fn equals(&self, other: &Signature) -> bool {
```
sdroege: Or implement Eq here?

src/analysis/function_parameters.rs
EPashkin: You right, it already on stable too.

src/analysis/signatures.rs
EPashkin: Hm, if I remove unneeded `pub` clippy don't nags about old name

src/codegen/sys/statics.rs
```diff
@@ -6,7 +6,7 @@ pub fn begin(w: &mut Write) -> Result<()> {
     let v = vec![
         "",
         "#![allow(non_camel_case_types, non_upper_case_globals, non_snake_case)]",
-        "#![cfg_attr(feature = \"cargo-clippy\", allow(approx_constant, type_complexity, unreadable_literal))]",
+        "#![cfg_attr(feature = \"cargo-clippy\", allow(clippy::approx_constant, clippy::type_complexity, clippy::unreadable_literal))]",
```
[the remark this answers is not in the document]
sdroege: We have to switch to 1.31 anyway already, @GuillaumeGomez has that on his list already. Major parts of the ecosystem already don't compile with pre-1.31 anymore, including things like `cc`.

For GStreamer I already switched to 1.31 because of that.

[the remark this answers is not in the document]
EPashkin: Thanks for info, then I remove feature here too.

PR #688, opened 2019-01-05
Replace all try! with ?

kornelski:
Done with https://users.rust-lang.org/t/automatically-replace-all-try-with/16873


EPashkin:
@kornelski Thanks for fix and method.

@GuillaumeGomez @sdroege @tmiasko Merging this not cause too much problem with your changes?

GuillaumeGomez:
It might cause conflicts with my current PR but besides that it doesn't seem to change any return type so that shouldn't be too dramatic...

GuillaumeGomez:
My PR is already big enough to avoid such changes. I prefer PR to focus on one thing. It's already hard enough to review. :p

sdroege:
> @GuillaumeGomez @sdroege @tmiasko Merging this not cause too much problem with your changes?

We'll see but I can update my changes, no worries :)

PR #690, opened 2019-01-10
[WIP] Always generate IsA<T> bounds for classes even if they have no known …

sdroege:
…subclasses here

They might have subclasses in another crate!

----

TODO
* [ ] Properly detect final types (they have no public class or instance struct) and don't do it for them
* [ ] Generate typed `None` constants like `const NoneWidget: Option<Widget> = None;` for convenience

See https://github.com/gtk-rs/gir/pull/689#discussion_r246656283 for details.


sdroege:
Closing this here as it needs to be done as part of #689 to have things compile

PR #667, opened 2018-11-18
[WIP] Start of user callbacks generation

GuillaumeGomez:
Fixes https://github.com/gtk-rs/gir/issues/618.

Still heavily under development.

PR #694, opened 2019-01-20
Implement configuration/detection for final types and use that inform…

sdroege:
…ation

Final types are those that can't have any further subclasses, and as
such we don't need to generate IsA<_> bounds or NONE_XXX constants and
also don't need to generate a trait for them.

This replaces the previous "trait" configuration with "final_type",
which has a slightly wider meaning.

### DIFF
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -9,7 +9,7 @@ use analysis::out_parameters::use_function_return_for_result;
 use analysis::rust_type::{bounds_rust_type, rust_type};
 use consts::TYPE_PARAMETERS_START;
 use env::Env;
-use library::{Function, Fundamental, Nullable, ParameterDirection, Type, TypeId};
+use library::{Class, Function, Fundamental, Nullable, ParameterDirection, Type, TypeId};
 use traits::IntoString;
 
 #[derive(Clone, Eq, Debug, PartialEq)]
@@ -177,8 +177,20 @@ impl Bounds {
             Type::Fundamental(Fundamental::Filename) => Some(AsRef(None)),
             Type::Fundamental(Fundamental::OsString)=> Some(AsRef(None)),
             Type::Fundamental(Fundamental::Utf8) if *nullable => Some(Into(Some('_'), None)),
-            Type::Class(..) if !*nullable => Some(IsA(None)),
-            Type::Class(..) => Some(Into(Some('_'), Some(Box::new(IsA(None))))),
+            Type::Class(Class { final_type, ..}) if !*nullable => {
+                if final_type {
+                    None
+                } else {
+                    Some(IsA(None))
+                }
+            },
+            Type::Class(Class { final_type, ..}) => {
+                if final_type {
+                    Some(Into(Some('_'), None))
+                } else {
+                    Some(Into(Some('_'), Some(Box::new(IsA(None)))))
+                }
+            },
             Type::Interface(..) if !*nullable => Some(IsA(None)),
             Type::Interface(..) => Some(Into(Some('_'), Some(Box::new(IsA(None))))),
             Type::List(_) | Type::SList(_) | Type::CArray(_) => None,
--- a/src/analysis/object.rs
+++ b/src/analysis/object.rs
@@ -20,6 +20,7 @@ pub struct Info {
     pub get_type: String,
     pub is_interface: bool,
     pub supertypes: Vec<general::StatusedTypeId>,
+    pub final_type: bool,
     pub generate_trait: bool,
     pub trait_name: String,
     pub has_constructors: bool,
@@ -54,22 +55,6 @@ impl Deref for Info {
     }
 }
 
-pub fn has_known_subtypes(env: &Env, class_tid: library::TypeId) -> bool {
-    for child_tid in env.class_hierarchy.subtypes(class_tid) {
-        let child_name = child_tid.full_name(&env.library);
-        let status = env.config
-            .objects
-            .get(&child_name)
-            .map(|o| o.status)
-            .unwrap_or_default();
-        if status.normal() {
-            return true;
-        }
-    }
-
-    false
-}
-
 pub fn class(env: &Env, obj: &GObject, deps: &[library::TypeId]) -> Option<Info> {
     info!("Analyzing class {}", obj.name);
     let full_name = obj.name.clone();
@@ -97,20 +82,11 @@ pub fn class(env: &Env, obj: &GObject, deps: &[library::TypeId]) -> Option<Info>
 
     let supertypes = supertypes::analyze(env, class_tid, &mut imports);
 
-    let mut generate_trait = obj.generate_trait;
+    let final_type = klass.final_type;
     let trait_name = obj.trait_name
         .as_ref()
         .cloned()
         .unwrap_or_else(|| format!("{}Ext", name));
-
-    // Sanity check the user's configuration. It's unlikely that not generating
-    // a trait is wanted if there are subtypes in this very crate
-    if !generate_trait && has_known_subtypes(env, class_tid) {
-        error!(
-            "Not generating trait for {} although subtypes exist",
-            full_name
-        );
-    }
 
     let mut trampolines =
         trampolines::Trampolines::with_capacity(klass.signals.len() + klass.properties.len());
@@ -120,7 +96,7 @@ pub fn class(env: &Env, obj: &GObject, deps: &[library::TypeId]) -> Option<Info>
         env,
         &klass.functions,
         class_tid,
-        generate_trait,
+        !final_type,
         obj,
         &mut imports,
         Some(&mut signatures),
@@ -145,7 +121,7 @@ pub fn class(env: &Env, obj: &GObject, deps: &[library::TypeId]) -> Option<Info>
         env,
         &klass.signals,
         class_tid,
-        generate_trait,
+        !final_type,
         &mut trampolines,
         obj,
         &mut imports,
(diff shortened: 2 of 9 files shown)


src/analysis/bounds.rs
```diff
@@ -177,6 +177,9 @@ impl Bounds {
             Type::Fundamental(Fundamental::Filename) => Some(AsRef(None)),
             Type::Fundamental(Fundamental::OsString)=> Some(AsRef(None)),
             Type::Fundamental(Fundamental::Utf8) if *nullable => Some(Into(Some('_'), None)),
+            // FIXME: Here we need to access analysis::Object::final_type to know if we have a
+            // final type or not. How?
+            // Final types should not use IsA<_>
```
sdroege: @EPashkin do you have an idea how to do this here? This also seems a bit problematic as it requires analysis for the type to be done already, but this function here for the bounds could be called before (or not?).

EPashkin: We can add final_type to library::Class and set it in post-processing.

EPashkin:
IMHO main problem with `final_type` that we need set it in all "child" crates to works properly.

src/analysis/bounds.rs
sdroege: Postprocessing happens after parsing the config and .girs, and before analysis? So that I can basically do the same there as I'm currently doing in analysis? That would work

src/config/gobjects.rs
```diff
@@ -234,10 +234,9 @@ fn parse_object(
         .lookup("cfg_condition")
         .and_then(|v| v.as_str())
         .map(|s| s.to_owned());
-    let generate_trait = toml_object
-        .lookup("trait")
-        .and_then(|v| v.as_bool())
-        .unwrap_or(true);
+    let final_type = toml_object
+        .lookup("final_type")
+        .and_then(|v| v.as_bool());
```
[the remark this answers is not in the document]
sdroege: Agreed

sdroege:
> IMHO main problem with `final_type` that we need set it in all "child" crates to works properly.

All child crates need to add the types to the manual section of their .toml already anyway, and for some types more configuration than that is already necessary. Check the `Gir_Gst.toml` for examples. So I think that's fine.

And if a child crate forgets to put this configuration there it won't generate broken bindings. It will only crate functions that are not as convenient to use as they could otherwise.

sdroege:
Ok, implemented it like that

src/analysis/object.rs
```diff
@@ -181,15 +161,11 @@ pub fn class(env: &Env, obj: &GObject, deps: &[library::TypeId]) -> Option<Info>
...
     // There's no point in generating a trait if there are no signals, methods, properties
     // and child properties: it would be empty
-    if generate_trait && !has_signals && !has_methods && properties.is_empty()
-        && child_properties.is_empty()
-    {
-        generate_trait = false;
-    }
+    //
+    // There's also no point in generating a trait for final types: there are no possible subtypes
+    let generate_trait = !final_type && (has_signals || has_methods || !properties.is_empty() || !child_properties.is_empty());
```
EPashkin: `generate_trait` was already defined in beginning with big comment

sdroege: Yes, and here it's overridden now that we know whether signals/methods/properties are to be generated or not. Would you prefer if I call the first one `would_generate_trait` instead, to make the difference clearer that the second one is a definite one while the first just an intention (if we actually generate signals/methods/properties)?

src/config/gobjects.rs
```diff
@@ -236,8 +237,11 @@ fn parse_object(
         .map(|s| s.to_owned());
     let generate_trait = toml_object
         .lookup("trait")
+        .and_then(|v| v.as_bool());
+    let final_type = toml_object
+        .lookup("final_type")
         .and_then(|v| v.as_bool())
-        .unwrap_or(true);
+        .or(generate_trait.map(|t| !t));
```
EPashkin: Clippy suggest `.or_else` and some other nits

sdroege: What other nits?

I thought about using `or_else()` here but it seemed more effort than inverting a boolean would justify. Your decision if I should change it or not :)

src/library_postprocessing.rs
```diff
@@ -373,4 +376,52 @@ impl Library {
...
+                        };
+
+                        let full_name = tid.full_name(self);
+                        let obj = config.objects.get(&*full_name);
+
+                        if let Some(GObject { final_type: Some(final_type), ..}) = obj {
+                            // The config might also be used to override a type that is wrongly
+                            // detected as final type otherwise
+                            if *final_type {
+                                actions.push(tid);
```
EPashkin: IMHO `final_types` is better that just `actions`

src/library_postprocessing.rs
```diff
@@ -373,4 +376,52 @@ impl Library {
...
+                            }
+                        } else if klass.c_class_type.is_none() {
+                            actions.push(tid);
```
EPashkin: This case IMHO needed comment too

sdroege: There's a comment a few lines above that is explaining the whole code, I would just repeat the same thing here otherwise :)

src/analysis/object.rs
```diff
@@ -97,21 +82,16 @@ pub fn class(env: &Env, obj: &GObject, deps: &[library::TypeId]) -> Option<Info>
 
     let supertypes = supertypes::analyze(env, class_tid, &mut imports);
 
-    let mut generate_trait = obj.generate_trait;
+    let final_type = klass.final_type;
+    // We would like to generate a trait if this is not a final type as there can be
+    // subtypes, however this will be overridden again below if there is nothing that
```
sdroege: `will be overridden again below` is what I meant btw

src/library_postprocessing.rs
sdroege: At the top of the function

src/analysis/object.rs
EPashkin: Sorry missed that, maybe remove this and use `!final_type` 3 middle call instead ?

src/library_postprocessing.rs
EPashkin: Ok, misunderstand that too.

src/analysis/object.rs
sdroege: Ok, fine with me :) But then the comment makes no sense and there's no explanation

src/config/gobjects.rs
EPashkin: You right, but then need adding allow for `clippy::or_fun_call`.
Other warnings (strange that only first match/Class listed IMHO it same for both) :
```
warning: redundant pattern matching, consider using `is_some()`
   --> src\config\gobjects.rs:288:12
    |
288 |       if let Some(_) = generate_trait {
    |  _____-      ^^^^^^^
289 | |         warn!("`trait` configuration is deprecated and replaced by `final_type` for object {}", name);
290 | |     }
    | |_____- help: try this: `if generate_trait.is_some()`
    |
    = note: #[warn(clippy::redundant_pattern_matching)] on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_pattern_matching

warning: you seem to be trying to use match for destructuring a single pattern. Consider using `if let`
   --> src\library_postprocessing.rs:392:17
    |
392 | /                 match *type_ {
393 | |                     Type::Class(ref klass) => {
394 | |                         let tid = TypeId {
395 | |                             ns_id: ns_id as u16,
...   |
413 | |                     _ => (),
414 | |                 }
    | |_________________^
    |
    = note: #[warn(clippy::single_match)] on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#single_match
help: try this
    |
392 |                 if let Type::Class(ref klass) = *type_ {
393 |     let tid = TypeId {
394 |         ns_id: ns_id as u16,
395 |         id: id as u32,
396 |     };
397 | 
```

sdroege: Ok, changing :)

sdroege: Pushed

EPashkin:
@sdroege Thanks, can you change `match *self.type_mut(tid) {` to if-let too?
Note: I don't prefer not merge this before #667 so it may need conflict fixing.

EPashkin:
`if let` can have `else` 😉 , I agree that will be safer in case function extension.

sdroege:
> `if let` can have `else` wink , I agree that will be safer in case function extension.

Indeed, changed :) Sorry, was a long day!

EPashkin:
@sdroege please check generated gio,
`Cancellable` now used without `IsA`, `NONE_CANCELLABLE` not generated.
`Cancelable` has no methods, while have GCancellableClass.

sdroege:
>  `Cancelable` has no methods, while have GCancellableClass.

Why does it have no methods? `g_cancellable_cancel()`, etc? :) It's not a final type in every regard

sdroege:
> Why does it have no methods? `g_cancellable_cancel()`, etc? :) It's not a final type in every regard

Ah, because it's marked as final type in the toml file. So that needs a change in the GIO toml to not mark it as final type (because it is not).

sdroege:
I'll send a PR for that once it's merged. We ideally get regens for everything once this merged, so maybe do that together with @GuillaumeGomez's #667

EPashkin:
@sdroege You right, we just make wrong decision long ago and added "trait = false".

EPashkin:
Hm, there many final types in gio, maybe better merge this before #667 and add separate PRs while its checkable.
@GuillaumeGomez You allow merge this PR?
This will cause single conflict in `use`, but may repeated on many commits on rebase.

GuillaumeGomez:
Yep, it's fine. I'll just fix conflicts on my PR.

EPashkin:
Then lets merge

PR #698, opened 2019-01-22
Check for known subtypes when deciding if a type is a final type

sdroege:
If there's a subtype we don't consider it a final type too. The class
struct might be private but inside the library there might be subtypes.

Subtypes in other libraries are nonetheless not possible if the class
struct is private.

----

I forgot to add that back yesterday. I don't think this changes any generated code except for `gstreamer-pbutils`

### DIFF
--- a/src/library_postprocessing.rs
+++ b/src/library_postprocessing.rs
@@ -377,9 +377,22 @@ impl Library {
         }
     }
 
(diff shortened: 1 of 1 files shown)


EPashkin:
@sdroege Thanks.
In gtk-rs only affected `gdk_pixbuf::PixbufAnimation` in GtkImage

PR #699, opened 2019-01-22
Fix invalid gboolean generation

GuillaumeGomez:
When generating glib, I had a "little" issue because of this one.

cc @EPashkin

### DIFF
--- a/src/analysis/ffi_type.rs
+++ b/src/analysis/ffi_type.rs
@@ -2,6 +2,7 @@ use analysis::c_type::{implements_c_type, rustify_pointers};
 use analysis::rust_type::{Result, TypeError};
 use env::Env;
 use library::*;
+use nameutil::get_crate_name;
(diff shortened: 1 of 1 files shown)


sdroege:
:+1:

EPashkin:
@GuillaumeGomez Thanks.
I merge this but IMHO all `idle_*` and `timeout_` in glib need be manual as return `SourceId`

GuillaumeGomez:
We still have issues with `g_child_watch_add*` because of aliases. I'm ending the patch then here we go!

sdroege:
All those source/watch related things have to be manually written (and are already).

sdroege:
> All those source/watch related things have to be manually written (and are already).

Among other things, they only require `Send` for the closure, can be `FnMut()` (and are already), return a `SourceId` and the closure returns a `Continue`.

PR #702, opened 2019-01-22
handle aliases better

GuillaumeGomez:
cc @EPashkin @sdroege

### DIFF
--- a/src/library.rs
+++ b/src/library.rs
@@ -322,6 +322,10 @@ impl TypeId {
     pub fn tid_os_string() -> TypeId {
         TypeId { ns_id: 0, id: 33 }
     }
+
+    pub fn is_fundamental_type(&self, env: &Env) -> bool {
+        env.library.type_(*self).is_fundamental_type(env)
+    }
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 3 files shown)


GuillaumeGomez:
What names do you suggest then? :)

EPashkin:
Also `get_chunk` that actually  adds chunk IMHO bad name.

EPashkin:
`get_type_id`->`is_fundamental`
and second is tough, maybe just `get`->`add` and comment about returning bool

GuillaumeGomez:
Renamed and updated a bit.

EPashkin:
For `add_chunk_for_type` called `env.type_` internally, then call `env.type_` again on result IMHO just need bool

sdroege:
Seems fine to me but maybe this should be part of the `is_fundamental()` method on the types directly? It is lying currently for alias types :)

GuillaumeGomez:
Well, it's an alias so it's lying that much.

EPashkin:
Maybe instead first function better add `is_fundamental_type` to library?

EPashkin:
and remove type returning from `add_chunk_for_type` and use `is_fundamental_type`, this IMHO make code cleaner

GuillaumeGomez:
That will be better this way indeed. Let's go for that!

GuillaumeGomez:
Updated. I decided to create a function in `library` directly in case we'll need it again in the future.

sdroege:
Still good for me :)

src/library.rs
```diff
@@ -779,6 +783,15 @@ impl Type {
         }
     }
 
+    /// If the type is an Alias containing a fundamental, it'll return true (whereas
+    /// `is_fundamental` won't).
+    pub fn is_fundamental_type(&self, env: &Env) -> bool {
```
EPashkin: Seems old `is_fundamental` used only internally so we can rename it to `is_fundamental_internal" (maybe make not public)

EPashkin: But IMHO it good as is.

GuillaumeGomez:
Updated.

EPashkin:
All seems ok, waiting for apveyor just in case

EPashkin:
@GuillaumeGomez Thanks again

PR #703, opened 2019-01-22
Remove references from TypeId usage

EPashkin:
.. with clippy help

cc @GuillaumeGomez, @sdroege

### DIFF
--- a/src/library_postprocessing.rs
+++ b/src/library_postprocessing.rs
@@ -377,10 +377,10 @@ impl Library {
         }
     }
 
-    fn has_subtypes(&self, parent_tid: &TypeId) -> bool {
+    fn has_subtypes(&self, parent_tid: TypeId) -> bool {
         for (tid, _) in self.types() {
             if let Type::Class(ref class) = *self.type_(tid) {
-                if class.parent.as_ref() == Some(parent_tid) {
+                if class.parent == Some(parent_tid) {
                     return true;
                 }
             }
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
Oh indeed, thanks a lot!

src/library_postprocessing.rs
```diff
@@ -377,10 +377,10 @@ impl Library {
         }
     }
 
-    fn has_subtypes(&self, parent_tid: &TypeId) -> bool {
+    fn has_subtypes(&self, parent_tid: TypeId) -> bool {
         for (tid, _) in self.types() {
             if let Type::Class(ref class) = *self.type_(tid) {
-                if class.parent.as_ref() == Some(parent_tid) {
+                if class.parent == Some(parent_tid) {
```
sdroege: I forgot to add a TODO comment here. I think we want to move the class_hierarchy generation into postprocessing at some point, and then directly make use of the result of it here. Otherwise we have a `O(n²)` loop here and in the class hierarchy code.

sdroege:
:+1:

src/library_postprocessing.rs
EPashkin: Added FIXME

sdroege: Thanks :) All good then

EPashkin: Sourceview postprocessing take only 10msec from 1.6s total 😉

GuillaumeGomez: Oh wow. That's **A LOT**!

PR #706, opened 2019-01-26
Few improvements on user callback generation

### LINKED ISSUE #704: Fix invalid user callbacks generation
```
<slomo> you pass a Box<Option<P>> as user_data, so a *mut Option<P>. but you then cast it inside the trampoline to a &Box<Option<P>>
<slomo> that's one level of indirection too much inside the trampoline
<slomo> it should be a &Option<P>
<slomo> oh looked at the wrong function. nevermind, but that is a problem elsewhere ;)
<slomo> this one is a different case of the same general problem
<slomo> you pass an &Option<P> as user_data but get out of the user_data a *mut P
<slomo> the Option<_> got lost along the way
```

GuillaumeGomez:
Fixes #704.

This currently fixes two things:

* Remove the invalid indirection (`Box::<Option<X>>.as_ptr()` -> `&Box<Option<X>>` is not longer a thing)
* Added the missing `Option` <s>(yep, everything is an `Option` by default)</s> (it's now correctly handled everywhere)
* No more `GString` parameters: they're replaced with `&str` now
* Fix invalid Send/Sync bounds (by default, I was giving both of them :p)
* Fix invalid nullable "status" from destroy callbacks.

I intend to fix the other issues on this PR as well so please don't merge! :)

### DIFF
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -867,6 +867,7 @@ fn analyze_callback(
                     0
                 },
                 destroy_index: 0,
+                nullable: par.nullable,
             }, match par.destroy_index {
                 Some(destroy_index) => Some(c_parameters[destroy_index].1),
                 None => None,
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -388,8 +388,9 @@ fn analyze_function(
         match env.config.objects.get(&*full_name) {
             Some(obj) => {
                 match env.library.type_(type_tid) {
+                    // FIXME: Maybe check for interface and record here as well?
                     library::Type::Class(_) => obj.concurrency,
-                    _ => library::Concurrency::SendSync,
+                    _ => library::Concurrency::None,
                 }
             }
             None => library::Concurrency::SendSync,
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -318,6 +318,14 @@ fn analyze_callbacks(
                                    func_name);
                         *commented = true;
                     }
+                    // We check if the user trampoline is there. If so, we change the destroy
+                    // nullable value if needed.
+                    for call in callbacks.iter() {
+                        if call.destroy_index == pos {
+                            callback.nullable = call.nullable;
+                            break
+                        }
+                    }
                     destroys.push(callback);
                     to_remove.push(pos);
                     continue;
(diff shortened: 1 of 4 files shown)


src/codegen/function_body_chunk.rs
```diff
@@ -356,7 +356,7 @@ impl Builder {
                         if is_destroy || trampoline.scope.is_async() {
                             format!("Box_::from_raw({} as *mut _)", func)
                         } else if trampoline.scope.is_call() {
-                            format!("{} as *const _ as usize as *mut {}",
+                            format!("{} as *const _ as usize as *mut Option<{}>",
```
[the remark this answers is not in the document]
GuillaumeGomez: Simpler that way. Everything is an `Option`. :)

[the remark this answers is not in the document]
sdroege: On one side, yes. On the other side you need to unwrap the `Option<_>`, which adds code (quite a bit due to potentially panicking) and also involves runtime costs :)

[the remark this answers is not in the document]
GuillaumeGomez: And of course you're right. I'll need to add another field to know if the callback is nullable. I'll do it next.

GuillaumeGomez:
I think this is ready! If you want to look at the result, take a look at https://github.com/gtk-rs/gtk/pull/760

sdroege:
> I think this is ready! If you want to look at the result, take a look at [gtk-rs/gtk#760](https://github.com/gtk-rs/gtk/pull/760)

Great :) I'll review tomorrow or maybe later

GuillaumeGomez:
I fixed the nullable bug spotted on gtk PR (which has been updated too).

src/analysis/functions.rs
```diff
@@ -389,7 +397,7 @@ fn analyze_function(
             Some(obj) => {
                 match env.library.type_(type_tid) {
                     library::Type::Class(_) => obj.concurrency,
-                    _ => library::Concurrency::SendSync,
+                    _ => library::Concurrency::None,
```
EPashkin: Maybe interface and record need be here too,
please add comment for checking later (at first glance I don't find any)

GuillaumeGomez: Good point! I add the comment.

sdroege: Both are needed here, yes.

sdroege: I mean, add them and not just a FIXME comment :P

sdroege:
:+1:

GuillaumeGomez:
Updated.

sdroege:
All good for me

EPashkin:
Then lets merge, @GuillaumeGomez Thanks

GuillaumeGomez:
I'll send the regens this evening.

PR #710, opened 2019-01-28
Create fewer boxes for closures

sdroege:
See the individual commits for explanations.

Needs some manual code updated too, which I'll do after the regens are all through.

### DIFF
--- a/src/chunk/chunk.rs
+++ b/src/chunk/chunk.rs
@@ -56,10 +56,6 @@ pub enum Chunk {
         in_trait: bool,
     },
     Name(String),
-    BoxFn {
-        name: Option<String>,
-        typ: String,
-    },
     ExternCFunc {
         name: String,
         parameters: Vec<Param>,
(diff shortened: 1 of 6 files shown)


GuillaumeGomez:
:+1:

sdroege:
Let's get this in together with the callback generation fixes. It's small enough and we can regenerate all things at once.

EPashkin:
@sdroege Thanks, I merge it.
Agree, that result better included in closures regen.
Note: don't forget to fix signal::destroy_closure

EPashkin:
listbox_model example now fails on close edit dialog

sdroege:
>  Note: don't forget to fix signal::destroy_closure

Ah that sucks. We need to make that one generic too. Let me fix all that, sorry

GuillaumeGomez:
I wait. But I hope you can feel the pressure now. :stuck_out_tongue:

EPashkin:
@sdroege Thanks, now listbox_model works fine.

Issue #713, opened 2019-01-29
Into<Option<_>> considered harmful


sdroege:
For closures this requires type annotations everywhere to make the compiler happy and we should really remove it there to make those functions easy to use for the user. Adding a `Some()` around the closure is easy, annotating everything correctly not so much.

IMHO we should get rid of this in general also for other parameter types but that probably needs some wider discussion first. Apart from annotation mess (but less likely) it also makes it much harder for users to discover that they can also pass `None` (common question by people) and the type signature is harder to read.

sdroege:
CC @EPashkin @GuillaumeGomez

GuillaumeGomez:
I'll remove them for user callbacks. Until rust compiler becomes more intelligent at least. :)

sdroege:
https://github.com/gtk-rs/gir/pull/714#issuecomment-458612568 has one example of what happens if we do the `Into<Option<_>>` stuff btw

sdroege:
Oh! The signature for the closures is wrong!

```
R: for<'r> Fn(&'r glib::Object) -> Widget + 'static
```

that should work too. All reference parameters would need to get that lifetime.

sdroege:
So, question is whether we need the `for<'r>`, we need to drop the `Into<Option<_>>`, or we need to do both. Can you try that @EPashkin, which of the combinations would work? I'm currently not a computer where I can test that :)

EPashkin:
Ok, I check today

EPashkin:
```
fn bind_model<'a, P: IsA<gio::ListModel> + 'a, Q: Into<Option<&'a P>>, R: for<'r> Fn(&'r glib::Object)
-> Widget + 'static, S: Into<Option<R>>>(&self, model: Q, create_widget_func: S);
```
builds fine: but problem still same as for current code.
listbox_model cases
clone! - no parameter type - E0282 (type annotations needed) on macro
no clone - no type - E0282 on parameter
no clone - type - ok
no clone - some - no type - E0282 on parameter
no clone - some - type - E0282 on bind_model

EPashkin:
```
fn bind_model<'a, P: IsA<gio::ListModel> + 'a, Q: Into<Option<&'a P>>,
R: Fn(&glib::Object) -> Widget + 'static>(&self, model: Q, create_widget_func: Option<R>) {
```
without one create_widget_func.into():
clone! - no parameter type - E0282 (type annotations needed) on macro
no clone - some - no type - E0282 on parameter
[...]

PR #714, opened 2019-01-29
Want some more user-callbacks?

### LINKED ISSUE #712: Some callbacks are misgenerated
See https://github.com/gtk-rs/gtk/pull/770#issuecomment-458495560 and comments on the specific code with explanations right below.

@GuillaumeGomez

GuillaumeGomez:
Fixes #712 

cc @sdroege @EPashkin

### DIFF
--- a/src/codegen/function_body_chunk.rs
+++ b/src/codegen/function_body_chunk.rs
@@ -466,20 +466,9 @@ impl Builder {
                 body.push(Chunk::Custom("};".to_owned()));
             }
             if trampoline.ret.c_type != "void" {
-                use ::analysis::conversion_type::ConversionType;
+                use codegen::trampoline_to_glib::TrampolineToGlib;
 
-                body.push(Chunk::Custom(match ConversionType::of(env, trampoline.ret.typ) {
-                    ConversionType::Direct => "res".to_owned(),
-                    ConversionType::Scalar | ConversionType::Pointer => {
-                        match rust_type(env, trampoline.ret.typ).unwrap().as_str() {
-                            "GString" |
-                            "File" => "res.to_glib_full()".to_owned(),
-                            _ => "res.to_glib()".to_owned(),
-                        }
-                    }
-                    ConversionType::Borrow => panic!("cannot return borrowed type..."),
-                    ConversionType::Unknown => "res".to_owned(),
-                }));
+                body.push(Chunk::Custom(format!("res{}", trampoline.ret.trampoline_to_glib(env))));
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/codegen/function_body_chunk.rs
```diff
@@ -474,7 +474,13 @@ impl Builder {
                         match rust_type(env, trampoline.ret.typ).unwrap().as_str() {
                             "GString" |
                             "File" => "res.to_glib_full()".to_owned(),
-                            _ => "res.to_glib()".to_owned(),
+                            _ => {
+                                if env.type_(trampoline.ret.typ).is_fundamental_type(env) {
```
sdroege: Are enums, flags and other scalar types also considered "fundamental"?

sdroege:
Does this cause any changes in the gtk-rs crates?

GuillaumeGomez:
It makes the code work but that's pretty much all. No changes on the `to_glib`/`to_glib_full`.

sdroege:
Seems good to me then apart from the comment above :)

src/codegen/function_body_chunk.rs
GuillaumeGomez: No.

sdroege: Then this is wrong and you need to match more on the actual type (what is returned by `env.type_()`).

There should be a function already for figuring this out.

sdroege: `impl TrampolineToGlib for library::Parameter` maybe, or code related to that

GuillaumeGomez: Ok!

GuillaumeGomez:
Updated.

sdroege:
Heh, so much simpler. Looks good to me then as long as it doesn't change any other code in the gtk-rs crates apart from the broken ones :)

EPashkin:
Confirmed that it only fixed gtk in gtk-rs crates.
@GuillaumeGomez, @sdroege Many thanks to both of you.

@sdroege strange, but listbox_model is broken again.
```
error[E0282]: type annotations needed
   --> src\bin\listbox_model.rs:30:28
    |
30  |       (@param $x:ident) => ( $x );
    |                              ^^ consider giving this closure parameter a type
...
83  |       listbox.bind_model(&model, clone!(window_weak => move |item| {
    |  ________________________________-
84  | |         let box_ = gtk::ListBoxRow::new();
85  | |         let item = item.downcast_ref::<RowData>().expect("Row data is of wrong type");
86  | |
...   |
162 | |         box_
163 | |     }));
    | |______- in this macro invocation
    |
    = note: type must be known at this point
```

sdroege:
> @sdroege strange, but listbox_model is broken again.

Yes, that's https://github.com/gtk-rs/gir/issues/713

PR #715, opened 2019-01-29
Remove into bound for user callbacks

### LINKED ISSUE #713: Into<Option<_>> considered harmful

GuillaumeGomez:
Fixes #713.

cc @EPashkin @sdroege

### DIFF
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -21,6 +21,8 @@ pub enum BoundType {
     AsRef(Option<char>),
     // lifetime (if none, not a reference, like for primitive types), wrapper type
     Into(Option<char>, Option<Box<BoundType>>),
+    // Needed for user callbacks: we don't want to wrap with `Into`.
+    Nullable,
 }
 
 impl BoundType {
@@ -42,6 +44,7 @@ impl BoundType {
     fn with_lifetime(ty_: BoundType, lifetime: char) -> BoundType {
         match ty_ {
             BoundType::NoWrapper => unreachable!(),
+            BoundType::Nullable => unreachable!(),
             BoundType::IsA(_) => BoundType::IsA(Some(lifetime)),
             BoundType::AsRef(_) => BoundType::AsRef(Some(lifetime)),
             BoundType::Into(_, x) => BoundType::Into(Some(lifetime), x),
@@ -183,30 +186,10 @@ impl Bounds {
                     )
                 }
                 if need_is_into_check {
-                    if let Some(x) = if let Some(ref mut last) = self.used.last_mut() {
+                    if let Some(ref mut last) = self.used.last_mut() {
                         if last.bound_type.is_into() {
-                            let mut new_one = (*last).clone();
-                            new_one.alias = self.unused.pop_front().expect("no available bound");
-                            new_one.type_str = last.alias.to_string();
-                            new_one.parameter_name = last.parameter_name.clone();
-                            // When we create a new bound for a callback which can be NULL,
-                            // we need to generate two new bounds instead of just one. This flag
-                            // allows us to know it so we can prevent its "generation" in the
-                            // codegen part (we don't need the `Into<>` part in a few parts of the
-                            // code).
-                            new_one.callback_modified = true;
-
-                            last.bound_type = BoundType::NoWrapper;
-                            last.parameter_name = String::new();
-
-                            Some(new_one)
-                        } else {
-                            None
+                            last.bound_type = BoundType::Nullable;
                         }
-                    } else {
-                        None
-                    } {
-                        self.used.push(x);
                     }
                 }
             }
@@ -348,7 +331,7 @@ impl Bounds {
         use self::BoundType::*;
         for used in &self.used {
             match used.bound_type {
-                NoWrapper => (),
+                NoWrapper | Nullable => (),
                 IsA(_) => imports.add("glib::object::IsA", None),
                 AsRef(_) => imports.add_used_type(&used.type_str, None),
                 Into(_, Some(ref x)) => {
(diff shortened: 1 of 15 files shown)


sdroege:
Yeah that's too much. It should instead be `P: Option<Fn(i32) ...>`

GuillaumeGomez:
Okay, `Option` is now back!

GuillaumeGomez:
cc @EPashkin @sdroege

EPashkin:
Thanks, but this case also have problem: old case has was good with just type, your need Some and type

GuillaumeGomez:
Do you have an example so I can improve my code? :)

EPashkin:
I don't have, tried playground for simplified cases all not good 😢

GuillaumeGomez:
So what do you want me to change in here?

sdroege:
Don't generate any `Option<_>` in any way for callbacks, and print a warning for each such function that manual implementations might be necessary because of that.

GuillaumeGomez:
I'm almost done here. Need to find out why some `Into` bounds are removed and then we're good.

GuillaumeGomez:
Ok, here we go!

sdroege:
>  fn set_forward_page_func(&self, page_func: Option<&( dyn Fn(i32) -> i32 + 'static)>) {

Why the parenthesis and the space before the `dyn`? That should both not be needed :)

Also this one should take a `Box<dyn Fn ...>`, it's not `scope call`. This is going to crash in interesting ways :)

`scope notified` should use boxed trait objects, `scope call` should use `Option<&mut dyn FnMut...>`

----

Once that is fixed, can you paste one example for `scope notified` and another for `scope call`?

sdroege:
Just to be clear, the parameter for `scope notified` must be `Option<Box<dyn Fn...>>`. You only box that box a second time inside the function for making it 1 pointers big instead of 2.

For `scope call` you don't have any box anywhere, but the parameter is `func: Option<&mut dyn Fn...>` and you then have to pass `&mut func as *mut _` to the C function and in the trampoline you'll get out a `*mut *mut dyn Fn` (note the double indirection: this is the same as the double boxing in the other case but without heap allocation!).

GuillaumeGomez:
Sure!

GuillaumeGomez:
notified:

<details>

```rust
fn set_forward_page_func(&self, page_func: Option<Box<dyn Fn(i32) -> i32 + 'static>>) {
    let page_func = page_func.into();
    let page_func_data: Box_<Option<Box<dyn Fn(i32) -> i32 + 'static>>> = Box::new(page_func);
    unsafe extern "C" fn page_func_func(current_page: libc::c_int, data: glib_ffi::gpointer) -> \
libc::c_int {
        let callback: &Option<Box<dyn Fn(i32) -> i32 + 'static>> = &*(data as *mut _);
        let res = if let Some(ref callback) = *callback {
            callback(current_page)
        } else {
            panic!("cannot get closure...")
        };
        res
    }
    let page_func = if page_func_data.is_some() { Some(page_func_func::<> as _) } else { None };
    unsafe extern "C" fn destroy_func(data: glib_ffi::gpointer) {
        let _callback: Box_<Option<Box<dyn Fn(i32) -> i32 + 'static>>> = Box_::from_raw(data as \
*mut _);
    }
    let destroy_call3 = Some(destroy_func::<> as _);
    let super_callback0: Box_<Option<Box<dyn Fn(i32) -> i32 + 'static>>> = page_func_data;
    unsafe {
        ffi::gtk_assistant_set_forward_page_func(self.as_ref().to_glib_none().0, page_func, Box:\
:into_raw(super_callback0) as *mut _, destroy_call3);
    }
}
```
</details>

call:

<details>

```rust
fn invalidate_maybe_recurse(&self, region: &cairo::Region, child_func: Option<&mut dyn (FnMut(&W\
indow) -> bool)>) {
    let child_func_data: Option<&mut dyn (FnMut(&Window) -> bool)> = child_func;
    unsafe extern "C" fn child_func_func(window: *mut ffi::GdkWindow, user_data: glib_ffi::gpoin\
ter) -> glib_ffi::gboolean {
        let window = from_glib_borrow(window);
        let callback: *mut Option<&mut dyn (FnMut(&Window) -> bool)> = user_data as *const _ as \
usize as *mut Option<&mut dyn (FnMut(&Window) -> bool)>;
        let res = if let Some(ref mut callback) = *callback {
            callback(&window)
        } else {
            panic!("cannot get closure...")
        };
        res.to_glib()
    }
    let child_func = Some(child_func_func::<> as _);
    let super_callback0: &Option<&mut dyn (FnMut(&Window) -> bool)> = &child_func_data;
    unsafe {
        ffi::gdk_window_invalidate_maybe_recurse(self.as_ref().to_glib_none().0, region.to_glib_\
none().0, child_func, super_callback0 as *const _ as usize as *mut _);
    }
}
```

sdroege:
> page_func_func::<>

I didn't know this even compiles :) But fine with keeping it if it doesn't break the compiler.

Looks good to me otherwise.

GuillaumeGomez:
Me neither! I'll clean that up!

GuillaumeGomez:
Updated!

sdroege:
All good for me then

EPashkin:
So part ` && *trampoline.nullable` makes no difference?
This is strange, and if it true, then IMHO better removed, even it it was needed before

GuillaumeGomez:
Fine by me!

GuillaumeGomez:
Updated!

EPashkin:
@GuillaumeGomez  Thanks, lets merge this.

GuillaumeGomez:
:tada: I'll send the PRs this evening (except for gtk which is handled through @sdroege's PR).

PR #719, opened 2019-02-08
Fixed deriving Copy trait for struct with function with varargs

### LINKED ISSUE #716: un-copyable struct fields generated with #[derive(Copy)]
Similar to #534, I'm running into structs that `#[derive(Copy)]` but are un-copyable, this time because of a generated `_trucated_record_marker`, causing the build to fail.

Log:

```
error[E0204]: the trait `Copy` may not be implemented for this type
    --> libdazzle-sys/src/lib.rs:1186:10
     |
1186 | #[derive(Copy, Clone)]
     |          ^^^^
...
1201 |     _truncated_record_marker: c_void,
     |     -------------------------------- this field does not implement `Copy`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0204`.
error: Could not compile `libdazzle-sys`.

```

Struct:

```rust
#[repr(C)]
#[derive(Copy, Clone)]
pub struct DzlPreferencesInterface {
    pub parent_interface: gobject::GTypeInterface,
    pub set_page: Option<unsafe extern "C" fn(*mut DzlPreferences, *const c_char, *mut glib::GHashTable)>,
    pub add_page: Option<unsafe extern "C" fn(*mut DzlPreferences, *const c_char, *const c_char, c_int)>,
    pub add_group: Option<unsafe extern "C" fn(*mut DzlPreferences, *const c_char, *const c_char, *const c_char, c_int)>,
    pub add_list_group: Option<unsafe extern "C" fn(*mut DzlPreferences, *const c_char, *const c_char, *const c_char, gtk::GtkSelectionMode, c_int)>,
    pub add_radio: Option<unsafe extern "C" fn(*mut DzlPreferences, *const c_char, *const c_char, *const c_char, *const c_char, *const c_char, *const c_char, *const c_char, *const c_char, *const c_char, c_int) -> c_uint>,
    pub add_font_button: Option<unsafe extern "C" fn(*mut DzlPreferences, *const c_char, *const c_char, *const c_char, *const c_char, *const c_char, *const c_char, c_int) -> c_uint>,
    pub add_switch: Option<unsafe extern "C" fn(*mut DzlPreferences, *const c_char, *const c_char, *const c_char, *const c_char, *const c_char, *const c_char, *const c_char, *const c_char, *const c_char, c_int) -> c_uint>,
    pub add_spin_button: Option<unsafe extern "C" fn(*mut DzlPreferences, *const c_char, *const c_char, *const c_char, *const c_char, *const c_char, *const c_char, *const c_char, *const c_char, c_int) -> c_uint>,
    pub add_file_chooser: Option<unsafe extern "C" fn(*mut DzlPreferences, *const c_char, *const c_char, *const c_char, *const c_char, *const c_char, *const c_char, *const c_char, gtk::GtkFileChooserAction, *const c_char, c_int) -> c_uint>,
    pub add_custom: Option<unsafe extern "C" fn(*mut DzlPreferences, *const c_char, *const c_char, *mut gtk::GtkWidget, *const c_char, c_int) -> c_uint>,
    pub remove_id: Option<unsafe extern "C" fn(*mut DzlPreferences, c_uint) -> gboolean>,
    pub get_widget: Option<unsafe extern "C" fn(*mut DzlPreferences, c_uint) -> *mut gtk::GtkWidget>,
    // /*Unimplemented*/Option<unsafe extern "C" fn(*mut DzlPreferences, *const c_char, *const c_char, *mut gtk::GtkWidget, /*Unimplemented*/va_list) -> c_uint>
}

```

EPashkin:
Fix #716

Now gir understand that this field - function is incomplete
```
Option<unsafe extern "C" fn(*mut DzlPreferences, *const c_char, *const c_char,
  *mut gtk::GtkWidget, /*Unimplemented*/va_list) -> c_uint>
```

Has no effect on gtk-rs

cc @GuillaumeGomez, @sdroege, @BrainBlasted


sdroege:
What does it generate now? The `va_list` part should simply be a `...` like [here](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs-sys/blob/e73bc4d9cdc02520e3346361ed5da1f7b126fb50/gstreamer-sys/src/lib.rs#L4166)

sdroege:
It's impossible to define such functions in Rust right now, but you can at least name them and call them :)

EPashkin:
@sdroege This slightly different types (at minimum now)
Your `gst_caps_new_full` has parameter with gir's type `Fundamental::VarArgs`
```
          <parameter name="..." transfer-ownership="none">
            <doc xml:space="preserve">additional structures to add</doc>
            <varargs/>
          </parameter>
```
Field `add_table_row_va` of record `DzlPreferencesInterface` is callback with
```
            <parameter name="args" transfer-ownership="none">
              <type name="va_list" c:type="va_list"/>
            </parameter>
```

src/analysis/types.rs
```diff
@@ -13,12 +13,12 @@ pub trait IsPtr {
 impl IsPtr for Field {
     fn is_ptr(&self) -> bool {
         if let Some(ref c_type) = self.c_type {
-            return c_type.contains('*')
+            return c_type.contains('*');
```
GuillaumeGomez: No need for `return` in here.

GuillaumeGomez:
Besides the nit, all good for me! :)

src/analysis/types.rs
EPashkin: Thanks, I don't notice that many of unneeded early returns here

EPashkin:
@sdroege If you think that varags and va_list must generated same code, then I add other PR

sdroege:
> @sdroege If you think that varags and va_list must generated same code, then I add other PR

That's exactly what `va_list` is, yes. They should be treated the same and it's not clear why gobject-introspection does not do that :)

For the record, https://github.com/rust-lang/rust/issues/44930 is the RFC for being able to define such varargs functions from Rust.

EPashkin:
Seems is not same as this is 2 different function,
```
int vprintf ( const char * format, va_list arg );
int printf ( const char * format, ... );
```

sdroege:
Oh indeed, I misread. Sorry!
`...` is the "unexpanded" form of `va_list`. So yes, these are different. What we have here is basically [this](https://doc.rust-lang.org/std/ffi/struct.VaList.html)

EPashkin:
Then I merge this.

PR #724, opened 2019-02-22
Remove Into trait generation

### LINKED ISSUE #722: Remove `Into` trait
It leads to very annoying situations like:

```Rust
let scroll = gtk::ScrolledWindow::new(None::<&gtk::Adjustment>, None::<&gtk::Adjustment>);
```

GuillaumeGomez:
Fixes #722 

It doesn't fix the whole `None::<Whatever>` issue but I guess it's a first step? The goal here being to go back to more explicit code.

cc @sdroege @EPashkin

### DIFF
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -19,32 +19,13 @@ pub enum BoundType {
     IsA(Option<char>),
     // lifetime <- shouldn't be used but just in case...
     AsRef(Option<char>),
-    // lifetime (if none, not a reference, like for primitive types), wrapper type
-    Into(Option<char>, Option<Box<BoundType>>),
 }
 
 impl BoundType {
-    pub fn is_into(&self) -> bool {
-        match *self {
-            BoundType::Into(_, _) => true,
-            _ => false,
-        }
-    }
-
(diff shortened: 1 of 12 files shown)


sdroege:
Need to look at generated code but a big :+1: for the intention here at least :) Thanks for working on this

GuillaumeGomez:
That makes me very sad that we reached rust's limit so quickly...

GuillaumeGomez:
You can take a look at the result [here](https://github.com/gtk-rs/gtk/pull/791).

sdroege:
In #722 @EPashkin wrote
> if we want remove it, then instead complete removal like in #724,
> we better remove it only for non-final objects and maybe for function too (if there any),
> in other cases `Into` works fine.

The point of this PR here is not to fix the problem with `IsA`, it's completely unrelated. It makes the compiler errors slightly less confusing but that's about it.

As @GuillaumeGomez wrote in the initial comment here: "The goal here being to go back to more explicit code."

EPashkin:
As @sdroege also for this solution, then lets merge this.
@GuillaumeGomez Thanks

PR #728, opened 2019-02-23
Fixup

GuillaumeGomez:
Follow up of #724.

I still have an for the generation of `.map()`: I don't get how the rules work here. Why in some cases we do generate it and in other cases we don't?

It leads to errors such as:

```
error[E0283]: type annotations required: cannot resolve `gdk_pixbuf::Pixbuf: std::convert::AsRef<_>`
  --> src/auto/image.rs:78:84
   |
78 |             Widget::from_glib_none(ffi::gtk_image_new_from_pixbuf(pixbuf.map(|p| p.as_ref()).to_glib_none().0)).unsafe_cast()
   |                                                                                    ^^^^^^

error: aborting due to previous error
```

### DIFF
--- a/src/analysis/function_parameters.rs
+++ b/src/analysis/function_parameters.rs
@@ -6,10 +6,11 @@ use env::Env;
 use library::{self, TypeId, ParameterScope};
 use nameutil;
 use super::conversion_type::ConversionType;
-use super::rust_type::rust_type;
-use super::ref_mode::RefMode;
 use super::out_parameters::can_as_return;
 use super::override_string_type::override_string_type_parameter;
+use super::rust_type::rust_type;
+use super::ref_mode::RefMode;
+use traits::MaybeRef;
 use traits::IntoString;
 
 //TODO: remove unused fields
@@ -248,8 +249,6 @@ pub fn analyze(
         let data_param_name = "user_data";
         let callback_param_name = "callback";
 
-        let mut nullable_into = false;
-
         if add_rust_parameter {
             let rust_par = RustParameter {
                 name: name.clone(),
@@ -262,6 +261,25 @@ pub fn analyze(
             ind_rust = None;
         }
 
+        let mut trans_nullable = false;
+        let type_ = env.type_(par.typ);
+        let to_glib_extra = if par.instance_parameter || !*nullable ||
+                               (!type_.is_interface() && !type_.is_class()) {
+            String::new()
+        } else {
+            trans_nullable = *nullable;
+            match type_.maybe_ref() {
+                Some(&library::Class { final_type, ..}) => {
+                    if final_type {
+                        String::new()
+                    } else {
+                        ".as_ref()".to_owned()
+                    }
+                }
+                None => String::new(),
+            }
+        };
+
         let transformation_type = match ConversionType::of(env, typ) {
             ConversionType::Direct => TransformationType::ToGlibDirect { name },
             ConversionType::Scalar => TransformationType::ToGlibScalar {
@@ -273,11 +291,11 @@ pub fn analyze(
                 instance_parameter: par.instance_parameter,
                 transfer,
                 ref_mode,
-                to_glib_extra: String::new(),
+                to_glib_extra,
                 explicit_target_type: String::new(),
                 pointer_cast: String::new(),
                 in_trait,
-                nullable: nullable_into,
+                nullable: trans_nullable,
             },
             ConversionType::Borrow => TransformationType::ToGlibBorrow,
             ConversionType::Unknown => TransformationType::ToGlibUnknown { name },
--- a/src/analysis/function_parameters.rs
+++ b/src/analysis/function_parameters.rs
@@ -10,7 +10,6 @@ use super::out_parameters::can_as_return;
 use super::override_string_type::override_string_type_parameter;
 use super::rust_type::rust_type;
 use super::ref_mode::RefMode;
-use traits::MaybeRef;
 use traits::IntoString;
 
 //TODO: remove unused fields
@@ -268,15 +267,10 @@ pub fn analyze(
             String::new()
         } else {
             trans_nullable = *nullable;
-            match type_.maybe_ref() {
-                Some(&library::Class { final_type, ..}) => {
-                    if final_type {
-                        String::new()
-                    } else {
-                        ".as_ref()".to_owned()
-                    }
-                }
-                None => String::new(),
+            if !type_.is_final_type() {
+                ".as_ref()".to_owned()
+            } else {
+                String::new()
             }
         };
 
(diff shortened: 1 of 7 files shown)


src/analysis/function_parameters.rs
```diff
@@ -262,6 +260,15 @@ pub fn analyze(
             ind_rust = None;
         }
 
+        let mut trans_nullable = false;
+        let to_glib_extra = if par.instance_parameter || !*nullable ||
+                               (!env.type_(par.typ).is_interface() && !env.type_(par.typ).is_class()) {
+            String::new()
+        } else {
+            trans_nullable = true;
+            ".as_ref()".to_owned()
```
EPashkin: Seems you need check if parameter type is final,
Ex. "gdk_pixbuf::Pixbuf" is final and its use don't have `IsA<xx>` so `as_ref` don't work.
Also please get type in separate variable before checking `.is_interface()` and `.is_class()`

GuillaumeGomez:
Better thanks @EPashkin's help but still not there:

```rust
pub fn set_icon_from_gicon<P: IsA<gio::Icon>>(&self, gicon: Option<&P>, size: IconSize) {
    unsafe {
        ffi::gtk_tooltip_set_icon_from_gicon(self.to_glib_none().0, gicon.as_ref().to_glib_none().0, size.to_glib());
    }
}
```

Which gives:

```
error[E0308]: mismatched types
  --> src/auto/tooltip.rs:38:73
   |
38 |             ffi::gtk_tooltip_set_icon_from_gicon(self.to_glib_none().0, gicon.as_ref().to_glib_none().0, size.to_glib());
   |                                                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `gio_ffi::GIcon`, found associated type
   |
   = note: expected type `*mut gio_ffi::GIcon`
              found type `*mut <P as glib::ObjectType>::GlibType`
```

EPashkin:
Here map needed as icon not final type

GuillaumeGomez:
Fixed!

EPashkin:
@GuillaumeGomez Please update https://github.com/gtk-rs/gtk/pull/791

GuillaumeGomez:
I was waiting for this one to get merged but let's do it!

GuillaumeGomez:
Done.

EPashkin:
Gio broken
```diff
-    pub fn launch_default_for_uri_async_future<'a, P: IsA<AppLaunchContext> + Clone + 'static, Q: Into<Option<&'a P>>>(uri: &str, context: Q) -> Box_<futures_core::Future<Item = (), Error = Error>> {
+    pub fn launch_default_for_uri_async_future<P: IsA<AppLaunchContext> + Clone + 'static>(uri: &str, context: Option<&P>) -> Box_<futures_core::Future<Item = (), Error = Error>> {
         use GioFuture;
         use fragile::Fragile;
 
         let uri = String::from(uri);
-        let context = context.into();
-        let context = context.map(ToOwned::to_owned);
+        let context = context.clone();
         GioFuture::new(&(), move |_obj, send| {
             let cancellable = Cancellable::new();
             let send = Fragile::new(send);
             Self::launch_default_for_uri_async(
                  &uri,
-                 context.as_ref().map(::std::borrow::Borrow::borrow),
+                 &context,
                  Some(&cancellable),
```
```
error[E0308]: mismatched types
   --> D:/eap/rust/0/gio\src\auto\app_info.rs:120:18
    |
120 |                  &context,
    |                  ^^^^^^^^ expected enum `std::option::Option`, found reference
    |
    = note: expected type `std::option::Option<&_>`
               found type `&std::option::Option<&P>`
help: you can convert from `&Option<T>` to `Option<&T>` using `.as_ref()`
    |
120 |                  &context.as_ref(),
    |                  ^^^^^^^^^^^^^^^^^
help: try using a variant of the expected type
    |
120 |                  Some(&context),
    |                  ^^^^^^^^^^^^^^
help: consider removing the borrow
    |
120 |                  context,
    |                  ^^^^^^^
```

GuillaumeGomez:
I never think about checking gio... Fixing it!

GuillaumeGomez:
We also have some issues like:

```
-    fn launch_uris<'a, P: IsA<AppLaunchContext> + 'a, Q: Into<Option<&'a P>>>(&self, uris: &[&str], context: Q) -> Result<(), Error>;
+    fn launch_uris<P: IsA<AppLaunchContext>>(&self, uris: &[&str], context: &P) -> Result<(), Error>;
```

GuillaumeGomez:
The second error has been fixed. Trying to figure out what's going on for the first.

GuillaumeGomez:
The second issue is fixed as well. Now we'll need to check if there are no surprises left.

src/codegen/function.rs
```diff
@@ -406,7 +408,10 @@ pub fn body_chunk_futures(env: &Env, analysis: &analysis::functions::Info) -> St
         } else {
             let c_par = &analysis.parameters.c_parameters[par.ind_c];
 
-            if c_par.ref_mode != RefMode::None {
+            if *c_par.nullable {
+                try!(writeln!(body, "         {}.as_ref().map(::std::borrow::Borrow::borrow),",
```
EPashkin: Please use '\t' here if it really needed

EPashkin: and change next lines too

src/analysis/function_parameters.rs
```diff
@@ -262,6 +261,25 @@ pub fn analyze(
...
+        let mut trans_nullable = false;
+        let type_ = env.type_(par.typ);
+        let to_glib_extra = if par.instance_parameter || !*nullable ||
+                               (!type_.is_interface() && !type_.is_class()) {
+            String::new()
+        } else {
+            trans_nullable = *nullable;
+            match type_.maybe_ref() {
+                Some(&library::Class { final_type, ..}) => {
+                    if final_type {
```
EPashkin: IMHO final_type better moved to library and return false for all other types,
bad that this don't help optimize this condition as we need set `trans_nullable` too

EPashkin:
Also can you fix this too, if it don't cause problems
```diff
+++ b/src/auto/calendar.rs
@@ -205,7 +205,7 @@ impl<O: IsA<Calendar>> CalendarExt for O {
         let func_data: Box_<P> = Box::new(func);
         unsafe extern "C" fn func_func<P: Fn(&Calendar, u32, u32, u32) -> String + 'static>(calendar: *mut ffi::GtkCalendar, year: libc::c_uint, month: libc::c_uint, day: libc::c_uint, user_data: glib_ffi::gpointer) -> *mut libc::c_char {
             let calendar = from_glib_borrow(calendar);
-            let callback: &P = &*(user_data as *mut _);
+            let callback: &(P) = &*(user_data as *mut _);
```

EPashkin:
.. or add comment with TODO about make it smarter somewhere near
https://github.com/gtk-rs/gir/blob/master/src/codegen/function_body_chunk.rs#L325
or https://github.com/gtk-rs/gir/blob/master/src/codegen/function_body_chunk.rs#L386

src/codegen/function.rs
GuillaumeGomez: I just took back the old code but fine with me!

src/analysis/function_parameters.rs
GuillaumeGomez: Fine by me!

EPashkin:
@GuillaumeGomez Thanks

PR #733, opened 2019-02-28
Allow to overwrite returned type

### LINKED ISSUE #729: Allow to overwrite returned type
We had this issue in gtk: https://github.com/gtk-rs/gtk/pull/789. Might be nice to be able to instead of ignoring the function because of an invalid returned type, to allow to provide another one instead.

GuillaumeGomez:
Fixes #729.

cc @sdroege @EPashkin

### DIFF
--- a/src/analysis/return_value.rs
+++ b/src/analysis/return_value.rs
@@ -23,13 +23,13 @@ pub fn analyze(
     used_types: &mut Vec<String>,
     imports: &mut Imports,
 ) -> Info {
-    let typ = match configured_functions.iter()
-                                              .filter_map(|f| f.ret.type_name.as_ref())
-                                              .map(|typ| env.library.find_type(0, typ))
-                                              .next() {
-        Some(Some(typ)) => typ,
-        _ => override_string_type_return(env, func.ret.typ, configured_functions),
-    };
+    let typ = configured_functions.iter()
+                                  .filter_map(|f| f.ret.type_name.as_ref())
+                                  .next()
+                                  .and_then(|typ| env.library.find_type(0, typ))
+                                  .unwrap_or_else(|| override_string_type_return(env,
+                                                                                 func.ret.typ,
+                                                                                 configured_functions));
     let mut parameter = if typ == Default::default() {
         None
     } else {
--- a/src/analysis/return_value.rs
+++ b/src/analysis/return_value.rs
@@ -23,14 +23,12 @@ pub fn analyze(
     used_types: &mut Vec<String>,
     imports: &mut Imports,
 ) -> Info {
-    let typ = match match configured_functions.iter()
+    let typ = match configured_functions.iter()
                                               .filter_map(|f| f.ret.type_name.as_ref())
+                                              .map(|typ| env.library.find_type(0, typ))
                                               .next() {
-        Some(ref typ) => env.library.find_type(0, typ),
-        None => None
-    } {
-        Some(typ) => typ,
-        None => override_string_type_return(env, func.ret.typ, configured_functions),
+        Some(Some(typ)) => typ,
+        _ => override_string_type_return(env, func.ret.typ, configured_functions),
     };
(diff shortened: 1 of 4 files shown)


GuillaumeGomez:
The error is:

```
$ if [ "$TRAVIS_RUST_VERSION" == "nightly" ]; then rustup component add clippy; cargo clippy --release; fi
error: component 'clippy' for target 'x86_64-unknown-linux-gnu' is unavailable for download
error: 'cargo-clippy' is not installed for the toolchain 'nightly-x86_64-unknown-linux-gnu'
To install, run `rustup component add clippy`
The command "if [ "$TRAVIS_RUST_VERSION" == "nightly" ]; then rustup component add clippy; cargo clippy --release; fi" exited with 1.
```

sdroege:
Use stable clippy?

GuillaumeGomez:
I didn't change it and I won't (at least no in this PR).

GuillaumeGomez:
Oh crap, forgot to add documentation though!

src/analysis/return_value.rs
```diff
@@ -23,7 +23,15 @@ pub fn analyze(
     used_types: &mut Vec<String>,
     imports: &mut Imports,
 ) -> Info {
-    let typ = override_string_type_return(env, func.ret.typ, configured_functions);
+    let typ = match match configured_functions.iter()
+                                              .filter_map(|f| f.ret.type_name.as_ref())
+                                              .next() {
+        Some(ref typ) => env.library.find_type(0, typ),
```
EPashkin: IMHO better use `.map(|typ| env.library.find_type(0, typ))` instead

src/config/functions.rs
```diff
@@ -117,11 +119,15 @@ impl Return {
                 }
             }
         };
+        let type_name = v.lookup("type")
+            .and_then(|v| v.as_str())
+            .map(|s| s.to_owned());
```
EPashkin: Please add check allowing only one for "type" and "string_type"

src/analysis/return_value.rs
GuillaumeGomez: It is indeed!

GuillaumeGomez:
Updated.

src/analysis/return_value.rs
```diff
@@ -23,7 +23,13 @@ pub fn analyze(
     used_types: &mut Vec<String>,
     imports: &mut Imports,
 ) -> Info {
-    let typ = override_string_type_return(env, func.ret.typ, configured_functions);
+    let typ = match configured_functions.iter()
+                                              .filter_map(|f| f.ret.type_name.as_ref())
+                                              .map(|typ| env.library.find_type(0, typ))
+                                              .next() {
+        Some(Some(typ)) => typ,
```
EPashkin: To remove double some, please use .and_then, and maybe `if let` instead match

GuillaumeGomez: I don't see how `.and_then` would improve the situation. Another `.filter_map` could though but then we'd look to the next arg and I don't want that so I think the only improvement possible is to use `if let` as you suggested.

EPashkin: I thought about something like this:
```
    let typ = configured_functions.iter()
        .filter_map(|f| f.ret.type_name.as_ref())
        .next()
        .and_then(|type_name| env.library.find_type(0, type_name))
        .unwrap_or_else(|| override_string_type_return(env, func.ret.typ, configured_functions));
```

GuillaumeGomez: Oh indeed! Didn't think about it this way...

EPashkin:
Rustfmt not agree with your formating and with my too 😉 
Anyway, please merge PR yourself when CI ended, I will be off for hour-two.

GuillaumeGomez:
`rustfmt` is the enemy of nice code! (we should enforce it at some point too...)

PR #734, opened 2019-03-03
Recheck crate names usage

### LINKED ISSUE #731: crate_name vs package_name problem
Currently we have packages on https://crates.io with names like "gtk-sys",
but it include in lib.rs file as `extern crate glib_sys`,
this cause some obscurity for me especially in "codegen/sys/".
I propose fixate these name as next terms:
- `package_name` - name with `-` as in crates.io,
- `crate_name` - name with underscores as in code,

and change Gir codebase accordingly.

IMHO this better done before related issue #685

EPashkin:
Fix #731

No changes in current gtk-rs

cc @GuillaumeGomez, @sdroege

### DIFF
--- a/src/nameutil.rs
+++ b/src/nameutil.rs
@@ -33,6 +33,7 @@ pub fn file_name_sys(name: &str) -> String {
     path.to_str().unwrap().into()
 }
 
+/// Crate name with undescores for `use` statement
 pub fn crate_name(name: &str) -> String {
     let name = name.to_snake();
     if name.starts_with("g_") {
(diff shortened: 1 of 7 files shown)


GuillaumeGomez:
👍

src/nameutil.rs
```diff
@@ -42,6 +42,20 @@ pub fn crate_name(name: &str) -> String {
     }
 }
 
+pub fn exported_crate_name(crate_name: &str) -> String {
```
sdroege: Maybe add some docs to these functions here to explain what they are actually returning and in which context that returned string is used

sdroege:
Looks good to me otherwise

src/nameutil.rs
EPashkin: Updated

PR #739, opened 2019-03-06
[ffi] Support alignment on objects

### LINKED ISSUE #736: Add support for defining alignment of FFI structs
This would mean adding `#[repr(align(XXX))]` to the struct definitions, e.g.

```rust
#[repr(C)]
#[repr(align(16))]
struct Foo {
    x: u8,
}
```

Currently this would have to be done via configuration, but ideally there would be a gobject-introspection annotation for that.

@EPashkin Can we do per-type configuration in -sys mode right now?

sfanxiang:
Closes #736.

### DIFF
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -330,11 +330,15 @@ fn generate_unions(w: &mut Write, env: &Env, unions: &[&Union]) -> Result<()> {
             continue;
         }
         let full_name = format!("{}.{}", env.namespaces.main().name, union.name);
-        if !env.type_status_sys(&full_name).need_generate() {
+        let config = env.config.objects.get(&full_name);
+
+        if let Some(false) = config.map(|c| c.status.need_generate()) {
             continue;
         }
+
+        let align = config.and_then(|c| c.align);
         let fields = fields::from_union(env, union);
-        try!(generate_from_fields(w, &fields));
+        try!(generate_from_fields(w, &fields, align));
     }
     Ok(())
 }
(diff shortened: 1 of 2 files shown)


src/codegen/sys/lib_.rs
```diff
@@ -333,8 +333,11 @@ fn generate_unions(w: &mut Write, env: &Env, unions: &[&Union]) -> Result<()> {
         if !env.type_status_sys(&full_name).need_generate() {
             continue;
         }
+        let config = env.config.objects.get(&full_name);
+        let align = config.and_then(|c| c.align);
```
EPashkin: Can you move `config` acquisition before status check and use it instead `env.type_status_sys`,
maybe with some new function in `Env`
```
pub fn unwrap_status_sys(status: Option<GStatus>) -> GStatus {
    status.unwrap_or(GStatus::Generate)
}
```

src/config/gobjects.rs
```diff
@@ -287,6 +290,17 @@ fn parse_object(
                 .collect()
         })
         .unwrap_or_else(|_| Vec::new());
+    let align = toml_object
+        .lookup("align")
+        .and_then(|v| v.as_integer())
+        .and_then(|v| {
+            if v.count_ones() != 1 || v > i64::from(u32::max_value()) || v < 0 {
+                warn!("align configuration must be a power of two of type u32, found {}", v);
```
EPashkin: Please use "\`align\` configuration ..." like in https://github.com/gtk-rs/gir/blob/master/src/config/error.rs#L62

EPashkin:
Thanks.

src/codegen/sys/lib_.rs
EPashkin: .. or in global namespace with code near `GStatus` definition

EPashkin: Sorry, forgot about `.map(|o| o.status)`
```
pub fn unwrap_status_sys(object: Option<GObject>) -> GStatus {
    if let Some(object) = object {
       object.status
    } else {
        GStatus::Generate
    }
}
```

sfanxiang: Other code in this file does it in a way that makes it easier to spot what's going on without increasing LOC:

```rust
if let Some(false) = config.map(|c| c.status.need_generate()) {
    continue;
}
```

I actually prefer this one. Let me know if you still want `unwrap_status_sys()`.

EPashkin: This is better solution and `Env::type_status_sys` maybe unneeded now. Thanks.

EPashkin: It was almost year ago :wink:
https://github.com/gtk-rs/gir/commit/032ec3fcc3cea62166c7a1cd941ff38946a18115#diff-0c37b0c87c3c91119e579feee853b5d0R187

EPashkin:
Thanks

PR #742, opened 2019-03-13
Handle nullable parameters for callbacks as well

### LINKED ISSUE #741: Nullable parameters aren't handled in "user callbacks"
You can see it from https://github.com/gtk-rs/gtk/issues/797.

GuillaumeGomez:
Fixes #741.

You can see the result here: https://github.com/gtk-rs/gtk/pull/798

cc @EPashkin @sdroege

### DIFF
--- a/src/analysis/rust_type.rs
+++ b/src/analysis/rust_type.rs
@@ -51,6 +51,11 @@ pub fn rust_type(env: &Env, type_id: library::TypeId) -> Result {
                    library::Concurrency::None)
 }
 
+pub fn rust_type_nullable(env: &Env, type_id: library::TypeId, nullable: Nullable) -> Result {
+    rust_type_full(env, type_id, nullable, RefMode::None, ParameterScope::None,
+                   library::Concurrency::None)
+}
+
 pub fn rust_type_with_scope(
     env: &Env,
     type_id: library::TypeId,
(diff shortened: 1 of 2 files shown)


src/analysis/rust_type.rs
```diff
@@ -230,12 +235,19 @@ pub fn rust_type_full(
...
                 }
-                match rust_type(env, p.typ) {
+                match rust_type_nullable(env, p.typ, p.nullable) {
                     Ok(x) => {
                         let is_fundamental = p.typ.is_fundamental_type(env);
+                        let y = rust_type(env, p.typ).unwrap_or_else(|_| String::new());
                         s.push(format!("{}{}",
                                        if is_fundamental { "" } else { "&" },
-                                       if x != "GString" { x } else { "&str".to_owned() }));
+                                       if y != "GString" { x } else {
```
EPashkin: Maybe just `!ends_with("GString")` ?

src/analysis/rust_type.rs
```diff
@@ -230,12 +235,19 @@ pub fn rust_type_full(
...
+                                           if *p.nullable {
```
EPashkin: Please `else if` without `{` between

EPashkin:
@GuillaumeGomez Thanks, looks good but there some small nits

src/analysis/rust_type.rs
GuillaumeGomez: I don't think this is a good idea: if we ever have a type that for any reason ends with "GString" (because why not?), well, debugging will be complicated. :)

src/analysis/rust_type.rs
GuillaumeGomez: Good point!

GuillaumeGomez:
Updated.

src/analysis/rust_type.rs
```diff
@@ -250,12 +262,21 @@ pub fn rust_type_full(
...
-                                 if x != "GString" { &x } else { "String" },
-                                 concurrency),
+            let mut ret = match rust_type_nullable(env, f.ret.typ, f.ret.nullable) {
+                Ok(x) => {
+                    let y = rust_type(env, f.ret.typ).unwrap_or_else(|_| String::new());
+                    format!("{}({}) -> {}{}",
+                            closure_kind,
+                            s.join(", "),
+                            if y != "GString" { &x } else {
+                                if *f.ret.nullable {
```
EPashkin: Here `else if` too please

GuillaumeGomez: Woups, fixed as well!

sdroege:
The generated code looks good to me. Thanks for working on this :)

GuillaumeGomez:
I made the error, I fix it. Sounds reasonable. :)

EPashkin:
Thanks

GuillaumeGomez:
I'll send the regen PRs tomorrow.

PR #743, opened 2019-03-23
Fix doc generation

GuillaumeGomez:
cc @EPashkin


EPashkin:
Hm, I thought that "gir PR incoming I assume!" means you don't have time for this and added  #744 too.

EPashkin:
Your has lesser addition, but IMHO better use `is_totally_deprecated` as crate can have `deprecate_by_min_version = false`.
Whose PR will be main and changed?

GuillaumeGomez:
Ah crap. No, it litteraly meant that I was about to write a PR for it. ;)

GuillaumeGomez:
Well, yours is more complete so I'll close this one.

Issue #748, opened 2019-04-24
Cannot Generate Bindings From XML with <source-position> element


BrainBlasted:
Gir files can include the position of documentation in elements as of https://gitlab.gnome.org/GNOME/gobject-introspection/commit/925c3b875861e83d3d6808439b61e80fe5ee8b48

PR #750, opened 2019-04-24
parser: ignore source-position

BrainBlasted:
G-i now generates a source-position element, but for our
purposes we don't need it yet.

This allows us to generate bindings for libraries that
include it, closing https://github.com/gtk-rs/gir/issues/748


GuillaumeGomez:
Thanks! Looks good to me.

cc @EPashkin

EPashkin:
@BrainBlasted Thanks

Issue #727, opened 2019-02-23
gir does not convert dashes to spaces in tests


BrainBlasted:
```
error: crate name using dashes are not valid in `extern crate` statements
 --> libdazzle-sys/tests/abi.rs:5:14
  |
5 | extern crate libdazzle-sys;
  |              ^^^^^^^^^^^^^ dash-separated idents are not valid
help: if the original crate name uses dashes you need to use underscores in the code
  |
5 | extern crate libdazzle_sys;
  |                       ^

error: expected one of `::`, `;`, or `as`, found `-`
  --> libdazzle-sys/tests/abi.rs:14:14
   |
14 | use libdazzle-sys::*;
   |              ^ expected one of `::`, `;`, or `as` here

error: aborting due to 2 previous errors

```

BrainBlasted:
The fix for this was undone, problem occurs on master @ 58cffd4 again: https://github.com/gtk-rs/gir/commit/39d85b2f00c043f7af4414fc2838e4e5faa5e53f

EPashkin:
@BrainBlasted Thanks for report problem again
Can you show/link your sys's Cargo.toml ?

BrainBlasted:
https://paste.rs/OHq.toml

EPashkin:
Thanks, as I suspected, you don't have `lib` section so `package.name` used, #751 must fix that
```
[lib]
name = "libdazzle_sys"
```

PR #751, opened 2019-04-25
Force replacing '-' in crate name for `use`

EPashkin:
Refix #727 

cc @BrainBlasted


BrainBlasted:
This works, thank you.

PR #757, opened 2019-05-10
Generate builder for widgets

antoyo:
The usage of the generated code looks like:
```rust
let button = LockButtonBuilder::new()
    .text_lock("Verrouiller")
    .text_unlock("Déverrouiller")
    .build();
```
This creates a `gtk::LockButton` and set the properties `text_lock` and `text_unlock` at the widget construction.

The generated code looks like:
```rust
#[cfg(any(feature = "builders", feature = "dox"))]
pub struct LockButtonBuilder {
    #[cfg(any(feature = "v3_14", feature = "dox"))]
    text_lock: Option<String>,
    #[cfg(any(feature = "v3_14", feature = "dox"))]
    text_unlock: Option<String>,
    // ...
}

#[cfg(any(feature = "builders", feature = "dox"))]
impl LockButtonBuilder {
    pub fn new() -> Self {
        Self {
            #[cfg(any(feature = "v3_14", feature = "dox"))]
            text_lock: None,
            #[cfg(any(feature = "v3_14", feature = "dox"))]
            text_unlock: None,
            #[cfg(any(feature = "v3_14", feature = "dox"))]
            tooltip_lock: None,
            #[cfg(any(feature = "v3_14", feature = "dox"))]
            tooltip_not_authorized: None,
            #[cfg(any(feature = "v3_14", feature = "dox"))]
            tooltip_unlock: None,
        }
    }
    pub fn build(self) -> LockButton {
        let mut n_properties = 0;
        let mut property_names: Vec<CString> = vec![];
        let mut names = vec![];
        let mut values = vec![];
        #[cfg(any(feature = "v3_14", feature = "dox"))]
        {
        if let Some(text_lock) = self.text_lock {
            property_names.push(CString::new("text-lock").unwrap());
            names.push(property_names[property_names.len() - 1].as_ptr());
            let property = text_lock.to_value();
            values.push(property.into_raw());
            n_properties += 1;
        }
        }
        #[cfg(any(feature = "v3_14", feature = "dox"))]
        {
        if let Some(text_unlock) = self.text_unlock {
            property_names.push(CString::new("text-unlock").unwrap());
            names.push(property_names[property_names.len() - 1].as_ptr());
            let property = text_unlock.to_value();
            values.push(property.into_raw());
            n_properties += 1;
        }
        }
        // ...
        unsafe {
            crate::Object::from_glib_none(gobject_sys::g_object_new_with_properties(
                LockButton::static_type().to_glib(), n_properties, names.as_mut_ptr(), values.as_ptr())
            as *mut _).downcast().expect("downcast")
        }
    }
    #[cfg(any(feature = "v3_14", feature = "dox"))]
    pub fn text_lock(mut self, text_lock: &str) -> Self {
        self.text_lock = Some(text_lock.to_string());
        self
    }
    #[cfg(any(feature = "v3_14", feature = "dox"))]
    pub fn text_unlock(mut self, text_unlock: &str) -> Self {
        self.text_unlock = Some(text_unlock.to_string());
        self
    }
    // ...
}
```

### DIFF
--- a/src/codegen/object.rs
+++ b/src/codegen/object.rs
@@ -113,7 +113,7 @@ pub fn generate(
     }
 
     // TODO: include parent properties.
-    if !analysis.builder_properties.is_empty() {
+    if env.config.objects.get(&format!("{}Builder", analysis.base.full_name)).is_some() && !analysis.builder_properties.is_empty() {
         generate_builder(w, env, analysis)?;
     }
 
@@ -201,7 +201,9 @@ pub struct {}Builder {{", analysis.name)?;
                 let name = nameutil::mangle_keywords(nameutil::signal_to_snake(&property.name));
                 version_condition(w, env, property.version, false, 1)?;
                 writeln!(w, "    {}: Option<{}>,", name, type_string)?;
-                let prefix = version_condition_string(env, property.version, false, 1).unwrap_or_default();
+                let prefix = version_condition_string(env, property.version, false, 1)
+                    .map(|version| format!("{}\n", version))
+                    .unwrap_or_default();
                 methods.push(format!("\n{prefix}    pub fn {name}(mut self, {name}: {param_type}) -> Self {{
         self.{name} = Some({name}{conversion});
         self
--- a/src/codegen/object.rs
+++ b/src/codegen/object.rs
@@ -112,76 +112,9 @@ pub fn generate(
         try!(writeln!(w));
     }
 
-    // TODO: include parent and writable properties.
+    // TODO: include parent properties.
     if !analysis.builder_properties.is_empty() {
-        let mut methods = vec![];
-        let mut properties = vec![];
-        writeln!(w, "#[cfg(any(feature = \"builders\", feature = \"dox\"))]
-pub struct {}Builder {{", analysis.name)?;
-        for property in &analysis.builder_properties {
-            match rust_type(env, property.typ) {
-                Ok(type_string) => {
-                    let type_string =
-                        match type_string.as_str() {
-                            "GString" => "String",
-                            "Vec<GString>" => "Vec<String>",
-                            typ => typ,
-                        };
-                    let (param_type, conversion) =
-                        match type_string {
-                            "String" => ("&str", ".to_string()"),
-                            typ => (typ, ""),
-                        };
-                    let name = nameutil::mangle_keywords(nameutil::signal_to_snake(&property.name));
-                    version_condition(w, env, property.version, false, 1)?;
-                    writeln!(w, "    {}: Option<{}>,", name, type_string)?;
-                    let prefix = version_condition_string(env, property.version, false, 1).unwrap_or_default();
-                    methods.push(format!("\n{prefix}    pub fn {name}(mut self, {name}: {param_type}) -> Self {{
-        self.{name} = Some({name}{conversion});
-        self
-    }}", prefix=prefix, param_type=param_type, name=name, conversion=conversion));
-                    properties.push(property);
-                },
-                Err(_) => writeln!(w, "    //{}: /*Unknown type*/,", property.name)?,
-            }
-        }
-        writeln!(w, "}}\n
-
-#[cfg(any(feature = \"builders\", feature = \"dox\"))]
-impl {}Builder {{
-    pub fn new() -> Self {{
-        Self {{", analysis.name)?;
-        for property in &properties {
-            version_condition(w, env, property.version, false, 3)?;
-            let name = nameutil::mangle_keywords(nameutil::signal_to_snake(&property.name));
-            writeln!(w, "            {}: None,", name)?;
-        }
-        writeln!(w, "        }}
-    }}
-
-    pub fn build(self) -> {} {{
-        let mut properties: Vec<(&str, &dyn ToValue)> = vec![];", analysis.name)?;
-        for property in &properties {
-            let name = nameutil::mangle_keywords(nameutil::signal_to_snake(&property.name));
-            version_condition(w, env, property.version, false, 2)?;
-            if property.version.is_some() {
-                writeln!(w, "        {{")?;
-            }
-            writeln!(w,
-"        if let Some(ref {property}) = self.{property} {{
-            properties.push((\"{}\", {property}));
-        }}", property = name)?;
-            if property.version.is_some() {
-                writeln!(w, "        }}")?;
-            }
-        }
-        writeln!(w,
-"        crate::Object::new({}::static_type(), &properties).expect(\"object new\").downcast().expect(\"downcast\")
-    }}", analysis.name)?;
-        for method in methods {
-            writeln!(w, "{}", method)?;
-        }
-        writeln!(w, "}}")?;
+        generate_builder(w, env, analysis)?;
     }
 
     match analysis.concurrency {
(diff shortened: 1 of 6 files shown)


GuillaumeGomez:
If all fields require the `v3_14`, then it would be nicer to just set the feature to `v3_14` on the struct directly instead.

src/codegen/object.rs
```diff
@@ -109,6 +111,96 @@ pub fn generate(
         try!(writeln!(w));
     }
 
+    if analysis.properties.iter().any(|property| property.construct || property.construct_only) {
+        let mut methods = vec![];
+        let mut properties = vec![];
+        writeln!(w, "#[cfg(any(feature = \"builders\", feature = \"dox\"))]")?;
```
GuillaumeGomez: You have the right to add empty line so your code can "breath" a bit. ;)

src/codegen/object.rs
```diff
@@ -109,6 +111,96 @@ pub fn generate(
...
+            let name = nameutil::mangle_keywords(nameutil::signal_to_snake(&property.name));
+            writeln!(w, "            {}: None,", name)?;
+        }
+        writeln!(w, "        }}")?;
+        writeln!(w, "    }}")?;
+        writeln!(w, "    pub fn build(self) -> {} {{", analysis.name)?;
+        writeln!(w, "        let mut n_properties = 0;")?;
+        writeln!(w, "        let mut property_names: Vec<CString> = vec![];")?;
+        writeln!(w, "        let mut names = vec![];")?;
+        writeln!(w, "        let mut values = vec![];")?;
```
GuillaumeGomez: One `writeln!` is enough I think. :)

src/codegen/object.rs
```diff
@@ -109,6 +111,96 @@ pub fn generate(
...
+                writeln!(w, "        #[cfg(any(feature = \"{}\", feature = \"dox\"))]", version.to_feature())?;
+                writeln!(w, "        {{")?;
+            }
+            writeln!(w, "        if let Some({property}) = self.{property} {{", property=name)?;
+            writeln!(w, "            property_names.push(CString::new(\"{}\").unwrap());", property.name)?;
+            writeln!(w, "            names.push(property_names[property_names.len() - 1].as_ptr());")?;
+            writeln!(w, "            let property = {}.to_value();", name)?;
+            writeln!(w, "            values.push(property.into_raw());")?;
+            writeln!(w, "            n_properties += 1;")?;
+            writeln!(w, "        }}")?;
```
GuillaumeGomez: One `writeln!` to rule them all!

GuillaumeGomez:
Looks nice, thanks! Few things to update though. ;)

src/codegen/object.rs
```diff
@@ -109,6 +111,96 @@ pub fn generate(
...
+            if property.version.is_some() {
+                writeln!(w, "        }}")?;
+            }
+        }
+        writeln!(w, "        unsafe {{")?;
+        writeln!(w, "            crate::Object::from_glib_none(gobject_sys::g_object_new_with_properties(")?;
```
sdroege: Why so complicated? :) `glib::Object::new()` can do exactly this here for you with only safe code

src/codegen/object.rs
```diff
@@ -109,6 +111,96 @@ pub fn generate(
         try!(writeln!(w));
     }
 
+    if analysis.properties.iter().any(|property| property.construct || property.construct_only) {
```
sdroege: So it currently only generates a `Builder` if there's any construct property. But wouldn't it also make sense for some types to generate it otherwise and provide access to the non-construct, writable properties via the builder?

Another problem is that for some types it is not enough to call `g_object_new()` with their properties to create a valid instance, but the actual `foo_new()` function of the specific type has to be called.

And yet another problem I see is that object construction could fail like this if some of the properties *must* be provided during construction and the default values are not enough. This would end up as a `g_critical` and then we're in UB country. `g_object_new()` can't ever fail unfortunately, it must always return an instance... but that instance might be unusable because of this.

There are the `GInitable` and `GAsyncInitable` interfaces in `gio` for that very reason, but they're not widely used.

I think what I'm saying here is that the generation of the builders for now should probably be opt-in per type.

sdroege:
Looks good generally, thanks :)

src/codegen/object.rs
antoyo: Can you give examples of types where `g_object_new()` is not enough?

You're right, I forgot the properties without default values. For that, I thought I could just add the required properties in the `new()` method.

sdroege: > Can you give examples of types where g_object_new() is not enough?

`GstGhostPad` is one I remember currently. Basically everything where the `foo_new()` functions do more than just calling `g_object_new()`.

> You're right, I forgot the properties without default values. For that, I thought I could just add the required properties in the new() method.

All properties have default values (unless they're object/boxed types, in which case the default is always `None`). But those default values might not be enough for constructing a valid, usable object.

src/codegen/object.rs
```diff
@@ -109,6 +111,89 @@ pub fn generate(
...
+                match bounds_rust_type(env, property.typ) {
+                    Ok(type_string) => {
+                        let attribute_type =
+                            match type_string.as_str() {
+                                "str" => "String",
+                                _ => &type_string,
+                            };
+                        let name = nameutil::mangle_keywords(nameutil::signal_to_snake(&property.name));
+                        if let Some(version) = property.version {
+                            writeln!(w, "    #[cfg(any(feature = \"{}\", feature = \"dox\"))]", version.to_feature())?;
```
EPashkin: IMHO `version_condition` can be used here

src/codegen/object.rs
```diff
@@ -109,6 +111,89 @@ pub fn generate(
...
+                        }
+                        writeln!(w, "    {}: Option<{}>,", name, attribute_type)?;
+                        let prefix =
+                            if let Some(version) = property.version {
+                                format!("    #[cfg(any(feature = \"{}\", feature = \"dox\"))]\n", version.to_feature())
```
EPashkin: `version_condition_string` ?

src/codegen/object.rs
```diff
@@ -305,4 +390,17 @@ pub fn generate_reexports(
         }
         traits.push(format!("\tpub use super::{};", analysis.trait_name));
     }
+
+    let mut generate_builder = false;
+    for property in &analysis.properties {
+        if (!property.is_get && property.construct) || property.construct_only {
+            generate_builder = true;
+            break;
```
EPashkin: This not different condition from that used in `generate`?
Maybe better define separate function here?

EPashkin:
Can you move builder code generation to function too

src/codegen/object.rs
EPashkin: This function can return iterator filtered by `(!property.is_get && property.construct) || property.construct_only`,
and it can be used in all 3 places

src/codegen/object.rs
```diff
@@ -256,6 +179,77 @@ impl {}Builder {{
...
+                    };
+                let (param_type, conversion) =
+                    match type_string {
+                        "String" => ("&str", ".to_string()"),
+                        typ => (typ, ""),
+                    };
+                let name = nameutil::mangle_keywords(nameutil::signal_to_snake(&property.name));
+                version_condition(w, env, property.version, false, 1)?;
+                writeln!(w, "    {}: Option<{}>,", name, type_string)?;
+                let prefix = version_condition_string(env, property.version, false, 1).unwrap_or_default();
```
EPashkin: As both version string is same you can get it only one time.

src/codegen/object.rs
```diff
@@ -171,6 +179,77 @@ pub fn generate(
...
+    writeln!(w, "}}\n
+
+#[cfg(any(feature = \"builders\", feature = \"dox\"))]
+impl {}Builder {{
+    pub fn new() -> Self {{
+        Self {{", analysis.name)?;
+    for property in &properties {
+        version_condition(w, env, property.version, false, 3)?;
+        let name = nameutil::mangle_keywords(nameutil::signal_to_snake(&property.name));
+        writeln!(w, "            {}: None,", name)?;
```
EPashkin: Oh, missed that completely. Please use `/t` in format string instead four spaces

EPashkin: Actually `\t`

EPashkin: @antoyo, sorry, ignore talk about \t

EPashkin:
Maybe better add `generate_builder: bool` to `GObject` instead?
Anyway Gir.toml format changes required change of `README.md` too or we will forget about it

EPashkin:
Also config checking IMHO better do on analysis phase

EPashkin:
@antoyo Thanks.
I don't like way of configure, but lets be it.

GuillaumeGomez:
@EPashkin Why merging if you didn't agree with it? We could have discussed about it you know. :) (we still can discuss about it)

EPashkin:
I don't want stall #759
Current version don't generate anything new without direct order,
so we can postpone discussion without any problem.

GuillaumeGomez:
Sure but then let's not forget about it!

EPashkin:
@antoyo There problem with parent properties, even if builder not generated many `uses` added from parent properties.

PR #759, opened 2019-05-11
Rust 2018 and code formatting

heftig:
- Port the code to Rust 2018. The output is untouched and still Rust 2015.
- Format all the code! Used rustfmt nightly with `merge_imports = true`. This does not conflict with stable rustfmt.


sdroege:
:+1: after fixing the merge conflicts. Thanks!

heftig:
Actually, found a few more things to do... one moment.

tests/sys/atk-sys/Cargo.toml
```diff
@@ -30,3 +30,4 @@ build = "build.rs"
 links = "atk"
 name = "atk-sys"
 version = "0.2.0"
+edition = "2018"
```
EPashkin: Only in root Cargo.toml please

EPashkin:
@GuillaumeGomez What you think about this PR?


@heftig Thanks.
I only see 2 problems: wrongly added `edition = "2018"` to old generated code,
and mismatched usage of `use`: sometimes there 2 or more sections, separated by empty lines.
Maybe better combine and reformat it?
I don't sure, wherever it is better put "external uses" in separate section or not.

PS.
Seems first problem is actually not problem as travis passed.

heftig:
> mismatched usage of `use`: sometimes there 2 or more sections, separated by empty lines.
> Maybe better combine and reformat it?
> I don't sure, wherever it is better put "external uses" in separate section or not.

Whatever you prefer. rustfmt doesn't (yet) reorder between sections. In the "Manually clean up imports" commit I reordered some manually to have external imports first, then internal ones, because this scheme was already in use in many files.



> PS.
> Seems first problem is actually not problem as travis passed.

I just removed the `edition` from the sub-projects, as requested. Should I undo this again?

EPashkin:
Thanks,  I prefer `edition` only in one file, so no undo needed.
This PR is changing a lot. can you make only one `use` section too?

heftig:
Assuming AppVeyor passes, thigh should be done, now.

GuillaumeGomez:
A few PRs on gir are pending and rebasing them on this one would be a nightmare. Once things are settled down, it'll make sense but beforehand it'd very complicated.

sdroege:
Can we get this in then? :)

GuillaumeGomez:
Now that my PR has been merged, yes for sure. :)

heftig:
Does it need another rebase? I can do that once I'm on the train.

EPashkin:
@heftig Yes, please.
And fix or remove checks in apveyor.

heftig:
Added two new commits to deal with new warnings on master.
- `GStatus`: Rename `Generate_Builders`
- Reduce type complexity for `group_by_user_data`

appveyor.yml
```diff
@@ -12,12 +12,16 @@ install:
...
   - SET LIBGIT2_SYS_USE_PKG_CONFIG=1
+  - IF "%RUST%" == "stable" rustup component add rustfmt clippy
   - rustc -Vv
   - cargo -Vv
   - pacman --noconfirm -S mingw-w64-%ARCH%-gtk3 mingw-w64-%ARCH%-libgit2 mingw-w64-%ARCH%-libsecret
 
 build_script:
   - git clone -q https://github.com/gtk-rs/gir-files tests/gir-files
+  - IF "%RUST%" == "stable" cargo clippy --all --release
+  - IF "%RUST%" == "stable" cargo check --all --release
```
EPashkin: apveyor check only stables, so `if` don't needed,
You sure that `cargo check` can show something new after `cargo clippy` ?
Why is `--all` used, gir don't have workspaces.

.travis.yml
```diff
@@ -9,10 +9,10 @@ addons:
...
     - libssh2-1-dev
 script:
-  - if [ "$TRAVIS_RUST_VERSION" == "stable" ]; then
-    rustup component add clippy;
-    cargo clippy --release;
-    fi
+  - test "$TRAVIS_RUST_VERSION" == "nightly" || rustup component add rustfmt clippy
+  - test "$TRAVIS_RUST_VERSION" == "nightly" || cargo clippy --all --release
+  - test "$TRAVIS_RUST_VERSION" == "nightly" || cargo check --all --release
+  - test "$TRAVIS_RUST_VERSION" == "nightly" || cargo fmt --all -- --check
```
EPashkin: We use clippy on "stable" because is broke too many times on nightly.

heftig: Yes, this runs on every version *but* `nightly`. I expressed it this way to ensure the list has a zero exit status when being skipped without suppressing an error from cargo.

appveyor.yml
heftig: No, I'm not sure. `--all` is from copying the invocations from somewhere else.

.travis.yml
EPashkin: Oh, I missed this.
Why not combine clippy and check calls too?
Or there some cases where cargo check give new error/warning which clippy not give?

EPashkin: Sorry, not checked changes before.
All good now.

EPashkin:
cc @GuillaumeGomez, @sdroege just in case

sdroege:
:+1:

sdroege:
Next step: all our crates :)

EPashkin:
@heftig Thanks

Issue #760, opened 2019-05-11
Use disguised="1" in gir files


EPashkin:
Before we don't use disguised flag because it was in many wrong places
```
        // Currently g-ir-scanner is too eager to mark all typedef to non-complete
        // types as disguised, so we limit this behaviour to the two most common cases.
        // https://gitlab.gnome.org/GNOME/gobject-introspection/issues/101
        let disguised = c_type == "GdkAtom" || c_type == "GIConv";
```
Seems problem fixed now and we can start of using this flag.

I created this PR as linking place for all related PRs,
please don't merge its until we decide that all is good.

cc @GuillaumeGomez, @sdroege

GuillaumeGomez:
Good finding. That's great if we can fix this bug!

EPashkin:
Last only sourceview with StyleExt removed and pangocairo with change usage of pango::*,
I will add it after we decide that to do.

EPashkin:
Not sys changes because new detection for final types:
disguised type is final if it don't have children

EPashkin:
@GuillaumeGomez, @sdroege, @antoyo 
All related PR is done.
Builders added to gtk, gio, for others I don't see good candidates.
Half PR passed CI (as always gtk has one error in apveyor).

EPashkin:
All related PR merged, so I close this

PR #762, opened 2019-05-11
Use disguised in final types check

EPashkin:
Part of https://github.com/gtk-rs/gir/issues/760

Removes traits from more types

Instead `klass.c_class_type` is `klass.type_struct` now used because it allow to get type from library.

cc @GuillaumeGomez, @sdroege

### DIFF
--- a/src/codegen/sys/tests.rs
+++ b/src/codegen/sys/tests.rs
@@ -3,7 +3,7 @@ use crate::{
     codegen::general,
     env::Env,
     file_saver::save_to_file,
-    library::{Bitfield, Enumeration, Type, MAIN_NAMESPACE},
+    library::{self, Bitfield, Enumeration, Namespace, Type, MAIN_NAMESPACE},
 };
 use log::info;
 use std::{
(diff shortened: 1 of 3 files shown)


src/codegen/sys/tests.rs
```diff
@@ -99,6 +81,27 @@ fn prepare_ctypes(env: &Env) -> Vec<CType> {
...
+    let full_name = format!("{}.{}", ns.name, t.get_name());
+    if env.type_status_sys(&full_name).ignored() {
+        return None;
+    }
+    let name = match t.get_glib_name() {
+        None => return None,
+        Some(name) => name,
+    };
+    if is_name_made_up(name) {
+                    return None;
```
GuillaumeGomez: Weird indent.

src/codegen/sys/tests.rs
```diff
@@ -64,33 +64,15 @@ fn prepare_ctypes(env: &Env) -> Vec<CType> {
...
-                }
-                let cfg_condition = env.config.objects.get(&full_name).and_then(|obj| {
-                    obj.cfg_condition.clone()
-                });
-                Some(CType {
-                    name: name.to_owned(),
-                    cfg_condition,
-                })
-            },
+                => prepare_ctype(env, ns, t),
```
GuillaumeGomez: Why going to next line?

GuillaumeGomez:
Seems good to me. But I might have miss something so let's wait for @sdroege review. :)

sdroege:
The results look fine to me

EPashkin:
IMHO we need decide about edition-2018,
then this PR and GuillaumeGomez's will need be rebased and reformated anyway.

sdroege:
IMHO we should move everything to 2018. gir, its output and the manual code in the other crates.

GuillaumeGomez:
Yes, but not right now. Let's first merge all pending PRs so we have a "clean" state to start working on.

EPashkin:
Problem that #759 already added and make loot changes,
Asking to make these changes again IMHO is too much.

GuillaumeGomez:
#759 is literally just a re-run of `cargo fmt` (unless I'm missing something?). Therefore, it's way easier to update rather than updating the pending PRs.

sdroege:
> #759 is literally just a re-run of `cargo fmt` (unless I'm missing something?). Therefore, it's way easier to update rather than updating the pending PRs.

It also updates to 2018.

GuillaumeGomez:
Ah. Then we have a problem because like I already said, updating pending PRs would be a nightmare. So we have the current choice:

 * Update to 2018 and rewriting the pending PRs (let's be honest, rebasing won't work)
 * Merge the pending PRs and rebase/rework this work.

I obviously prefer the second option but what do you think @EPashkin, @sdroege?

EPashkin:
I for first.

sdroege:
Either way seems fine to me. It shouldn't be too hard to update the 2018 PR. One part is "cargo fmt", the other is manually updating everything for 2018. The second part should be easy to merge, the first part can be automatically redone.

sdroege:
We should just get the 2018 PR in really soon. Let's put tomorrow early afternoon as deadline, every other PR that is not ready before then will have to be rebased on top of that.

PR #764, opened 2019-05-14
Fix builders

EPashkin:
Part of #763
Makes minor fixes after #757

cc @antoyo

PR #767, opened 2019-05-15
Builder setters accept references

### LINKED ISSUE #763: Builder generation configuration
#757 added possibility to generate object builder if it added to Gir.toml
```toml
generate_builders = [
    "Gtk.WindowBuilder",
]
```
I don't like this configuration method because it add new items to `GObjects`,
and its names theoretically can conflict with real objects.

I prefer this way
```
[[object]]
name = "Gtk.Window"
status = "generate"
generate_builder = true
```

@GuillaumeGomez, @sdroege , @antoyo  What you think about this?

PS. ~Current version also cause warning due misnaming~
```
warning: variant `Generate_Builders` should have an upper camel case name
  --> src\config\gobjects.rs:23:5
   |
23 |     Generate_Builders,
   |     ^^^^^^^^^^^^^^^^^ help: convert the identifier to upper camel case: `GenerateBuilders`
   |
   = note: #[warn(non_camel_case_types)] on by default
```

PPS. Not related to this issue, but GStatus.Comment also not used

EPashkin:
Close #763
Makes minor fixes after #757, #764

```diff
+++ b/src/auto/about_dialog.rs
@@ -505,8 +505,8 @@ impl AboutDialogBuilder {
         self
     }
 
-    pub fn logo(mut self, logo: gdk_pixbuf::Pixbuf) -> Self {
-        self.logo = Some(logo);
+    pub fn logo(mut self, logo: &gdk_pixbuf::Pixbuf) -> Self {
+        self.logo = Some(logo.clone());
         self
     }
```
Only changes classes parameters, vector functions remain unchanged
```rust
    pub fn artists(mut self, artists: Vec<String>) -> Self {
        self.artists = Some(artists);
        self
    }
```

cc @GuillaumeGomez , @sdroege , @antoyo

### DIFF
--- a/src/codegen/object.rs
+++ b/src/codegen/object.rs
@@ -1,10 +1,14 @@
 use super::{child_properties, function, general, properties, signal, trait_impls, trampoline};
 use crate::{
-    analysis::{self, rust_type::rust_type},
+    analysis::{
+        self,
+        rust_type::{rust_type, rust_type_full},
+    },
     case::CaseExt,
     codegen::general::{version_condition, version_condition_string},
     env::Env,
     library, nameutil,
+    traits::IntoString,
 };
 use std::io::{Result, Write};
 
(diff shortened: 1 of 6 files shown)


src/codegen/object.rs
```diff
@@ -178,10 +181,24 @@ fn generate_builder(w: &mut dyn Write, env: &Env, analysis: &analysis::object::I
...
+                    library::Concurrency::None,
+                )
+                .unwrap(); //Must be Ok if rust_type is Ok
+                let (param_type_override, conversion) = match &param_type[..] {
+                    "&str" => ("", ".to_string()"),
+                    "&[&str]" => ("Vec<String>", ""),
+                    typ if typ.starts_with('&') => ("", ".clone()"),
+                    _ => ("", ""),
                 };
+                if !param_type_override.is_empty() {
```
antoyo: I'd rather use `None` instead of the empty string and do:
```
if let Some(typ) = param_type_override {
```

EPashkin: You right.
Done

GuillaumeGomez:
:+1:

EPashkin:
Added object configuration `generate_builder = true`.
`builders` array now support both builder name and object name
```toml
builders = [
    "Gtk.AboutDialog",
    "Gtk.WindowBuilder",
]
```

GuillaumeGomez:
Awesome!

PR #768, opened 2019-05-15
Ignore the <attribute> element in any other XML element

sdroege:
It is added by newer versions of gobject-introspection in various
places.


EPashkin:
Thanks

PR #775, opened 2019-05-20
Don't bring trait ObjectType into scope

### LINKED ISSUE #772: Collision between generated type and glib helper type
Since updating gir and glib on my project ([bindings for libostree](fkrull/ostree-rs)), it has started importing `glib::object::ObjectType` in at least one generated file. Unfortunately, libostree has a type `OSTree.ObjectType` which also gets imported into the same generated file. As you can probably imagine, [this breaks rather noisily](https://gitlab.com/fkrull/ostree-rs/-/jobs/214727451).

From my scan of the readme, I couldn't see an obvious way to rename an entire type. Even if there was, that's a solution I'd like to avoid. I didn't try setting `ObjectType` to manual either, but thinking about it, I guess that wouldn't work because the problem is in a file referencing the type.

EPashkin:
Closes #772

@GuillaumeGomez, @sdroege  Seems this fix is enough,
or you prefer some general solution like special function for this?

cc @fkrull

### DIFF
--- a/src/analysis/properties.rs
+++ b/src/analysis/properties.rs
@@ -227,7 +227,8 @@ fn analyze_property(
     };
 
     if !generate_trait && (writable || readable || notifiable) {
-        imports.add("glib::object::ObjectType", prop_version);
+        //Don't bring trait into scope to resolve conflict with OSTree::ObjectType
+        imports.add("glib::object::ObjectType as _", prop_version);
     }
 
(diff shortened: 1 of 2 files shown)


sdroege:
Seems fine to me

GuillaumeGomez:
Sounds good to me as well.

PR #776, opened 2019-05-21
Rename trait instead hiding, because your minimum version 1.31 don't …

EPashkin:
…support this

Part of #772

cc @GuillaumeGomez, @sdroege, @fkrull

### DIFF
--- a/src/analysis/properties.rs
+++ b/src/analysis/properties.rs
@@ -227,8 +227,8 @@ fn analyze_property(
     };
 
     if !generate_trait && (writable || readable || notifiable) {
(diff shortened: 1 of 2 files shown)


EPashkin:
Or "glib::object::ObjectType as ObjectType_" same as we do with box "use std::boxed::Box as Box_" ?

sdroege:
See https://github.com/gtk-rs/cairo/pull/260/files#r285871201 but works for me either way

GuillaumeGomez:
I'd prefer "glib::object::ObjectType as ObjectType_". Just in case.

EPashkin:
Ok, let it be `glib::object::ObjectType as ObjectType_`

PR #779, opened 2019-05-21
Fix gconstpointer fixup

fkrull:
Ran into this: I have an _equals and _hash that take gconstpointer and I had to change the *mut to *const in the generated code to make it compile.

This change feels like it should have unintended side effects, but also if you're dealing with gconstpointer, *const seems always correct.


EPashkin:
@fkrull This change broke glib::DateTime because `Shared` types implement only `ToGlibPtr<*mut _>`,
can you show bad object definition`s "glib_wrapper!" and generated `hash` ?

fkrull:
@EPashkin Here's [the entire offending file](https://gitlab.com/fkrull/ostree-rs/blob/master/src/auto/collection_ref.rs). This is generated by this modified branch, but the only difference is the `*const` instead of `*mut`. I can dig out the error message, but I think it was just the opposite of what you described: that the type *doesn't* implement `ToGlibPtr<*mut _>`.

(The compiler helpfully suggested the `*const` implementation to me so I may have jumped to conclusions there.)

EPashkin:
Yes, Boxed types don't implement `ToGlibPtr<*mut _>`, so only boxed requires "*const" in `fixup_special_functions`.
Bad that detection logic for it is not compact.
You can set flag like `is_boxed` in https://github.com/gtk-rs/gir/blob/master/src/analysis/record.rs#L89-L91 and pass it all way down.

fkrull:
@EPashkin Ok, I've added a check on that flag which should fix glib.

EPashkin:
Can you fix pointer for `RecordType::AutoBoxed` too? This is second branch in https://github.com/gtk-rs/gir/blob/master/src/analysis/record.rs#L89-L91

fkrull:
@EPashkin Thanks for the hints. I hope that should be it now? glib generates as before and my tricky type got fixed up even without having to specify `use_boxed_functions` explicitly.

src/analysis/record.rs
```diff
@@ -80,23 +80,21 @@ pub fn new(env: &Env, obj: &GObject) -> Option<Info> {
         None => return None,
     };
 
-    let use_boxed_functions = obj.use_boxed_functions;
+    let use_boxed_functions =
+        obj.use_boxed_functions || RecordType::of(&record) == RecordType::AutoBoxed;
```
EPashkin: Please, rename variable to "is_boxed".

EPashkin:
@fkrull Good to hear that config not needed in your case.

EPashkin:
Thanks, I waiting CI just in case, then will do squash+merge

Issue #777, opened 2019-05-21
Mix of Option<&T> and &Option<T>


sdroege:
We currently use `Option<&T>` for function parameters and `&Option<T>` for callbacks (signal handlers, closures, etc).

This is rather inconsistent and makes it sometimes a bit tricky to pass arguments from a closure to another function, see e.g. https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/issues/203 .

`Option<&T>` is the more idiomatic approach (and more convenient to use!) and we should switch that all over. I'll take a look at that over the next days.

PR #780, opened 2019-05-22
Use Option<&T> instead of &Option<T> for signal callback parameters

sdroege:
See https://github.com/gtk-rs/gir/issues/777

Next step is to do the same for callbacks passed to functions. That apparently does not even use the same machinery to find the correct type mappings between Rust and FFI types.

@GuillaumeGomez Any hint where to look and why you didn't use the `trampoline_from_glib` function?

### DIFF
--- a/src/analysis/rust_type.rs
+++ b/src/analysis/rust_type.rs
@@ -274,9 +274,17 @@ pub fn rust_type_full(
                         let y = rust_type(env, p.typ).unwrap_or_else(|_| String::new());
                         s.push(format!(
                             "{}{}",
-                            if is_fundamental { "" } else { "&" },
+                            if is_fundamental || *p.nullable {
+                                ""
+                            } else {
+                                "&"
+                            },
                             if y != "GString" {
-                                x
+                                if !is_fundamental && *p.nullable {
+                                    x.replace("Option<", "Option<&")
+                                } else {
+                                    x
+                                }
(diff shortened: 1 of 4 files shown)


GuillaumeGomez:
Certainly because I didn't know it existed?

sdroege:
Ok, so I might be able to move it over to those functions. Can you tell me where all that type translation magic is that you added here? I would've expected to find a special case somewhere that converts from `Option<T>` to `&Option<T>` like there is for signal trampolines, but couldn't find it.

sdroege:
Found it, I guess. No special case, but by-passing the type transformation we do elsewhere ;)

sdroege:
This should be for all the autogenerated code. Once this is in, things need to be regenerated and manual code needs to be updated.

EPashkin:
@sdroege Thanks.
Affected gtk, gio, examples

sdroege:
Let's get https://github.com/gtk-rs/gtk/pull/812 merged first? That one is already big enough

EPashkin:
It's up to you.  That PR ready for merge.

sdroege:
Unless you want to regen with this change and then merge, just get it in now and I'll care for it later :)

EPashkin:
Sorry, I don't want to update 812.

sdroege:
No problem :)

PR #781, opened 2019-05-26
Allow disguised types to use ref/unref

moggiesir:
Since the c:type for a disguised record is already a pointer to the record, `define_shared_type` should use the underlying record name instead of the c:type.

### DIFF
--- a/src/codegen/record.rs
+++ b/src/codegen/record.rs
@@ -32,11 +32,13 @@ pub fn generate(w: &mut dyn Write, env: &Env, analysis: &analysis::record::Info)
         analysis.specials.get(&Type::Ref),
         analysis.specials.get(&Type::Unref),
     ) {
+        
+        let c_type = if type_.disguised { "_".to_string() + &type_.c_type } else { type_.c_type.to_string() };
         general::define_shared_type(
             w,
             env,
             &analysis.name,
-            &type_.c_type,
+            &c_type,
             ref_fn,
             unref_fn,
             &analysis.glib_get_type,


EPashkin:
Thanks for PR, but its really needed?

We have only one shared disguised record in gtk-rs and change broke it.
```diff
+++ b/src/auto/coverage.rs
@@ -10,7 +10,7 @@ use std::ptr;
 
 glib_wrapper! {
     #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
-    pub struct Coverage(Shared<pango_sys::PangoCoverage>);
+    pub struct Coverage(Shared<pango_sys::_PangoCoverage>);
 
     match fn {
```
Its happened because your `ref` and `unref` defined on outer ctype not on inner.
```rust
    pub fn pango_coverage_ref(coverage: *mut PangoCoverage) -> *mut PangoCoverage;
    pub fn pango_coverage_unref(coverage: *mut PangoCoverage);
```
Maybe this because current disguised definition,
```rust
#[repr(C)]
pub struct _PangoCoverage(c_void);

pub type PangoCoverage = *mut _PangoCoverage;
```
that differs from C's
```
typedef struct _PangoCoverage PangoCoverage;
```

So, @moggiesir, please show related code fragments for your case and error it produced.
Maybe we need change `disguised definition` in sys crate.

moggiesir:
Ugh, according to https://gitlab.gnome.org/GNOME/gobject-introspection/issues/101 `disguised` has two meanings, and I didn't see the second one until now.

I'm working on getting https://github.com/gtk-rs/javascriptcore-rs to be more complete (and generated from .gir files). JSGlobalContextRef is a disguised type according to the first meaning and generates: `type JSGlobalContextRef = *mut _JSGlobalContextRef;` where _JSGlobalContextRef is an opaque struct. That all matches the C nicely.

IMO, the pango sys definition is actually incorrect, but may not be worth fixing. I'm happy to either:

1. drop this PR and just manually wrap JSGlobalContextRef
2. change the gir generation code for shared disguised records so that it looks at the type that the ref/unref functions take.

EPashkin:
@moggiesir Seems currently better do first case.
I don't have enough information to decide:
while WebKit code contains JSGlobalContextRef, it not present in latest Ubuntu's javascript-core .gir files.
Maybe @sdroege or @antoyo have more info on this?

antoyo:
The `JavaScriptCore.gir` file is missing many things.

moggiesir:
Here's my WIP changes to the gir file for JSC: https://github.com/moggiesir/gir-files/commit/605f744fa59f7a219ee4bd2ac41aa56ec24e3076

AFAICT, the problem with the gir file is that before webkit2 v2_22, webkit2 directly referenced Apple's non-glib JavaScriptCore library and never added that to the gir files. Starting with 2.22, added a glib-based wrapper for JavaScriptCore and they deprecated anything that referred to the apple types: https://webkitgtk.org/2018/09/03/webkitgtk2.22.0-released.html

It would be easiest to just drop rust support for the deprecated methods, but I think that maintaining compatibility for a while might be nice, which means adding some additional types to the gir file.

PR #782, opened 2019-05-26
Declare parameter to be mut if necessary

sfanxiang:
Diff (gtk4):
```diff
diff --git a/src/auto/entry.rs b/src/auto/entry.rs
--- a/src/auto/entry.rs
+++ b/src/auto/entry.rs
@@ -647,7 +647,7 @@ impl<O: IsA<Entry>> EntryExt for O {
         }
     }
 
-    fn set_tabs(&self, tabs: Option<&mut pango::TabArray>) {
+    fn set_tabs(&self, mut tabs: Option<&mut pango::TabArray>) {
         unsafe {
             gtk_sys::gtk_entry_set_tabs(self.as_ref().to_glib_none().0, tabs.to_glib_none_mut().0);
         }
diff --git a/src/auto/text.rs b/src/auto/text.rs
--- a/src/auto/text.rs
+++ b/src/auto/text.rs
@@ -180,7 +180,7 @@ impl Text {
         }
     }
 
-    pub fn set_tabs(&self, tabs: Option<&mut pango::TabArray>) {
+    pub fn set_tabs(&self, mut tabs: Option<&mut pango::TabArray>) {
         unsafe {
             gtk_sys::gtk_text_set_tabs(self.to_glib_none().0, tabs.to_glib_none_mut().0);
         }
```

cc @EPashkin @sdroege

### DIFF
--- a/src/codegen/function.rs
+++ b/src/codegen/function.rs
@@ -59,7 +59,7 @@ pub fn generate(
         }
         Visibility::Hidden => return Ok(()),
     }
-    let declaration = declaration(env, analysis);
+    let declaration = declaration(env, analysis, only_declaration);
     let suffix = if only_declaration { ";" } else { " {" };
 
     writeln!(w)?;
@@ -88,7 +88,7 @@ pub fn generate(
     }
 
     if analysis.async_future.is_some() {
-        let declaration = declaration_futures(env, analysis);
+        let declaration = declaration_futures(env, analysis, only_declaration);
         let suffix = if only_declaration { ";" } else { " {" };
 
         writeln!(w)?;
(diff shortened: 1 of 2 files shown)


EPashkin:
@sfanxiang IMHO both functions not change `tabs`, you tried use it as `const` like in
https://github.com/gtk-rs/gtk/blob/master/Gir.toml#L663

sfanxiang:
@EPashkin I think you're right. But should this still be fixed?

EPashkin:
Not sure about this:
1) Using `const*PangoTabArray` instead mutable pointer is better fixed in upstream,
2) Almost all actual mutable parameters are `out` and we move them to return value without using `Option<&mut XXX>`,
3) Rare `inout`s don't supported by gir and need be manual.

sfanxiang:
OK, closing this. set_tabs() in gtk3 needs to be fixed then.

EPashkin:
@sfanxiang Your right, thanks for reporting problem
I created PR

PR #796, opened 2019-06-20
Add warning to gobject-sys's build script

EPashkin:
Part of https://github.com/gtk-rs/sys/pull/141

cc @GuillaumeGomez, @sdroege


GuillaumeGomez:
I'm not a big fan considering we'll need to update this code as soon as the problem is fixed (if it's ever fixed haha). But meanwhile, it does the job so good enough for me.

Thanks @EPashkin !

EPashkin:
Yes, only if "problem is fixed" 😉

EPashkin:
@sdroege Thanks for snippet, I merge my PR, so you can update yours

Issue #797, opened 2019-06-22
Allow define function trait for ignored functions in doc mode


EPashkin:
Ex. manual function `SocketListenerExtManual::accept_async` don't get help with `--features embed-lgpl-docs` because gio/vendor.md contains
```
<!-- trait SocketListenerExt::fn accept_async -->
This is the asynchronous version of `SocketListenerExt::accept`.
```

EPashkin:
Seems this blocked by https://github.com/GuillaumeGomez/rustdoc-stripper/issues/53 anyway.

PR #798, opened 2019-06-22
Add overriding for function trait for manual implemented functions

EPashkin:
Fix https://github.com/gtk-rs/gir/issues/797

@GuillaumeGomez, not sure that it right way, but it works

test for gio:
```diff
+++ b/Gir.toml
@@ -664,11 +664,13 @@ manual_traits = ["SocketListenerExtManual"]
     name = "accept_socket_async"
     # finish function misses nullable return annotation
     ignore = true
+    doc_trait_name = "SocketListenerExtManual"
 
     [[object.function]]
     name = "accept_async"
     # finish function misses nullable return annotation
     ignore = true
+    doc_trait_name = "SocketListenerExtManual"
 
 [[object]]
 name = "Gio.Subprocess"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,6 +185,9 @@ generate_builder = true
     doc_hidden = true
     # disable length_of autodetection
     disable_length_detect = true
+    # write function docs to trait other than default "xxxExt",
+    # also works in [object.signal] and [object.property]
+    doc_trait_name = "SocketListenerExtManual"
         # override for parameter
         [[object.function.parameter]]
         # filter by name
--- a/src/codegen/doc/mod.rs
+++ b/src/codegen/doc/mod.rs
@@ -155,6 +155,12 @@ fn create_object_doc(w: &mut dyn Write, env: &Env, info: &analysis::object::Info
     let signals: &[Signal];
     let properties: &[Property];
 
+    let obj = env
+        .config
+        .objects
+        .get(&info.full_name)
+        .expect("Object not found");
+
     match *env.library.type_(info.type_id) {
         Type::Class(ref cl) => {
             doc = cl.doc.as_ref();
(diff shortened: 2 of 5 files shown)


EPashkin:
Signal need this too 😢

EPashkin:
Properties too, it also generated wrong function_name:
```
<!-- trait CellRendererPixbufExt::fn set_property_icon-name -->
The name of the themed icon to display.
This property only has an effect if not overridden by "stock_id"
or "pixbuf" properties.
```

EPashkin:
I hope that this final version,
@GuillaumeGomez  I will process other crates tomorrow, if you don't want do it yourself

EPashkin:
Checked all gtk-rs crates: PR covered all cases.
@GuillaumeGomez, @sdroege plase, check code.
Also maybe you have better solution than `doc_trait_name`?

GuillaumeGomez:
Isn't it possible to get if the function is in a manual trait directly instead of adding this new key?

EPashkin:
Theoretically we can scan all manual files and store all traits and its functions, then auto-detect its docs,
but at the moment I don't want do parse yourself and use syn too.

GuillaumeGomez:
No need to have a super efficient thing. A little python script which will run through `src` folder and store all "trait *ExtManual" things should be enough no?

EPashkin:
We have many traits like EditableSignals in gtk, so trait name can be any,
and it require parsing anyway.
Also I don't understand how "little python script" helps with "Isn't it possible to get if the function is in a manual trait directly instead of adding this new key?"

GuillaumeGomez:
> Also I don't understand how "little python script" helps with "Isn't it possible to get if the function is in a manual trait directly instead of adding this new key?"

Nothing, sorry. I thought you were talking about finding missing declared manual traits in the gir files... What I meant initially was: "why not using `manual_trait = "SomeExtManual"` when generating docs?

EPashkin:
> I meant initially was: "why not using manual_trait = "SomeExtManual" when generating docs instead of adding a new key?

Mostly because it array `manual_traits = ["DialogExtManual"]`,
also I don't think that all ignored need moved to it.

GuillaumeGomez:
It seems weird to me that it's not possible to just base ourselves on this... What do you think of this @sdroege ?

EPashkin:
Ok, I still think that we can't change trait for all ignored functions,
so I merge this as is.

PR #800, opened 2019-06-23
Fix doc generation

GuillaumeGomez:
cc @EPashkin @sdroege 

Result of the changes: https://github.com/gtk-rs/lgpl-docs/pull/62

For now I only ran the new gir fix on `atk`, `gdk` and `gtk`.


GuillaumeGomez:
Actually it's already fixed in #798

PR #806, opened 2019-07-06
Use new mem::uninitialized API

### LINKED ISSUE #770: Use MaybeUninit instead of mem::unitialized
The latter causes UB very easily, and e.g. the following already is UB from what I understand

```rust
let mut x = mem::uninitialized();
foo_sys::blabla(&mut x);
```

https://doc.rust-lang.org/std/mem/union.MaybeUninit.html

`MaybeUninit` is just being stabilized now: https://github.com/rust-lang/rust/pull/60445

GuillaumeGomez:
Fixes #770.

cc @sdroege @EPashkin

### DIFF
--- a/src/codegen/function_body_chunk.rs
+++ b/src/codegen/function_body_chunk.rs
@@ -38,6 +38,15 @@ enum OutMemMode {
     NullMutPtr,
 }
 
+impl OutMemMode {
+    fn is_uninitialized(&self) -> bool {
+        match *self {
+            OutMemMode::Uninitialized => true,
+            _ => false,
+        }
+    }
+}
+
 #[derive(Clone, Default)]
 struct ReturnValue {
     pub ret: return_value::Info,
@@ -102,7 +111,10 @@ impl Builder {
     pub fn out_parameter(&mut self, env: &Env, parameter: &AnalysisCParameter) -> &mut Builder {
         let mem_mode = c_type_mem_mode(env, parameter);
         self.parameters.push(Parameter::Out {
-            parameter: parameter_ffi_call_out::Parameter::new(parameter),
+            parameter: parameter_ffi_call_out::Parameter::new(
+                parameter,
+                mem_mode.is_uninitialized(),
+            ),
             mem_mode,
         });
         self.outs_as_return = true;
@@ -121,9 +133,11 @@ impl Builder {
     pub fn generate(&self, env: &Env, bounds: String, bounds_names: String) -> Chunk {
         let mut body = Vec::new();
 
-        if self.outs_as_return {
-            self.write_out_variables(&mut body);
-        }
+        let mut uninitialized_vars = if self.outs_as_return {
+            self.write_out_variables(&mut body)
+        } else {
+            Vec::new()
+        };
 
         let mut group_by_user_data = FuncParameters::new();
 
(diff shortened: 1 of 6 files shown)


GuillaumeGomez:
An example of the result: https://github.com/gtk-rs/pango/pull/155

sdroege:
See comments there, invalid usage of `MaybeUninit` :)

GuillaumeGomez:
More like lazy usage but yes. ;) I'll update the PR later on.

sdroege:
Not lazy, it's undefined behaviour just like with `mem::uninitialized` now. It's wrong :)

GuillaumeGomez:
I wanted to mean that my implementation was lazy. :)

sdroege:
Oh that's possible :) Changing the code generator to use `MaybeUninit` correctly is probably quite some work

GuillaumeGomez:
This is **finally** done. T_T

src/codegen/function_body_chunk.rs
```diff
@@ -739,7 +793,11 @@ impl Builder {
             .output_params
             .iter()
             .filter(|param| param.direction == ParameterDirection::Out && param.name != "error")
-            .map(|param| (param, type_mem_mode(env, param)))
+            .map(|param| {
+                let kind = type_mem_mode(env, param);
+                ;
+                (param, kind)
```
EPashkin: Unneeded debug change?

GuillaumeGomez: Indeed, good catch.

EPashkin:
Thanks, look good for me, it still need `cargo fmt` run

src/codegen/translate_from_glib.rs
```diff
@@ -81,22 +81,27 @@ impl TranslateFromGlib for analysis::return_value::Info {
 }
 
 fn from_glib_xxx(transfer: library::Transfer, array_length: Option<&String>) -> (String, String) {
+    let good_print = |name: &str| {
+        if name == "error" {
+            format!(", {} as usize", name)
+        } else {
+            format!(", {{ let {0} = {0}.assume_init(); {0} as usize }})", name)
```
EPashkin: Seems this can be just `format!(", {}.assume_init() as usize)", name)` without "error" branch

sdroege:
Can you regenerate the pango PR so we can also check the results? :)

src/codegen/translate_from_glib.rs
EPashkin: @GuillaumeGomez If you don't want this one-liner, please, merge this PR yourself.

EPashkin: Oh, error branch still needed.
Can you tell me which function?

GuillaumeGomez: I updated it to a one-liner. However, I don't understand your last comment. What error?

EPashkin: I meant branch `if name == "error"`, I suspect that it never reached:
note that it not ends with ')', 
and I updated it to produce bad code, but gtk-rs crates still builds fine after regen.

GuillaumeGomez: I'll remove this check and see the result.

GuillaumeGomez:
@EPashkin You were right: the condition was completely useless so I removed it.

EPashkin:
@GuillaumeGomez Thanks

GuillaumeGomez:
I'll send the PRs soon then!

PR #812, opened 2019-07-11
Better handling of uninitialized primitives

### LINKED ISSUE #809: Not use MaybeUninit for primitive types
It's a bit overkill to use it on primitives considering we could just initialize them like `let x = 0;`. Less code for the same result.

GuillaumeGomez:
Fixes #809.

cc @EPashkin @sdroege

### DIFF
--- a/src/library.rs
+++ b/src/library.rs
@@ -250,6 +250,32 @@ impl Fundamental {
             _ => true,
         }
     }
+
+    pub fn get_initialization(self) -> Option<&'static str> {
+        match self {
+            Fundamental::Boolean
+            | Fundamental::Int8
+            | Fundamental::UInt8
+            | Fundamental::Int16
+            | Fundamental::UInt16
+            | Fundamental::Int32
+            | Fundamental::UInt32
+            | Fundamental::Int64
+            | Fundamental::UInt64
+            | Fundamental::Char
+            | Fundamental::UChar
+            | Fundamental::Short
+            | Fundamental::UShort
+            | Fundamental::Int
+            | Fundamental::UInt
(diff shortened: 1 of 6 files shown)


src/library.rs
```diff
@@ -250,6 +250,32 @@ impl Fundamental {
...
+            | Fundamental::Char
+            | Fundamental::UChar
+            | Fundamental::Short
+            | Fundamental::UShort
+            | Fundamental::Int
+            | Fundamental::UInt
+            | Fundamental::Long
+            | Fundamental::ULong
+            | Fundamental::Size
+            | Fundamental::SSize => Some("0"),
```
sdroege: `let b: bool = 0` is not going to work :)

GuillaumeGomez: Is there any boolean declared anywhere?

sdroege:
I'm not entirely convinced that this is the right thing to do btw, I only mentioned it on IRC because changing everything manually is quite some work. For autogenerated code, using `MaybeUninit` has no additional cost to humans and is minimally faster for computers.

src/library.rs
sdroege: 15 lines above or so in the same block of code

EPashkin: At minimum no booleans in gtk-rs

EPashkin:
Sorry, @GuillaumeGomez, I also think that `MaybeUninit` is better for this case, at minimum in generated code.

PS. If you still insist, then this PR breaks atk and clippy found unneeded `clone` for `&'static str`.

GuillaumeGomez:
No, @sdroege convinced me so I'll just close it. :)

PR #814, opened 2019-07-17
Prevent invalid function parameter order on async callback by not enforcing it

### LINKED ISSUE #813: Get the signature of the "finish" methods from the gir file
The signature of the "finish" methods (to finish an asynchronous call) is usually the same, and gir is based on that standard. However, sometimes, that standard is not respected, for example in libgoa: https://github.com/nemequ/vala-girs/blob/eef9b2f23eccce46af487e23888268cbfd6c6284/gir-1.0/Goa-1.0.gir#L698

In this situation, the generated code cannot compile because the parameters are swapped. Is it possible to generate this kind of code by following the declaration in the gir files?

GuillaumeGomez:
Fixes #813.

Can you check it @Boiethios please?

### DIFF
--- a/src/codegen/function_body_chunk.rs
+++ b/src/codegen/function_body_chunk.rs
@@ -718,13 +718,20 @@ impl Builder {
                 type_: "*mut _".to_string(),
             });
         }
-        finish_args.push(Chunk::Name("res".to_string()));
+        let mut found_async_result = false;
         finish_args.extend(
             trampoline
                 .output_params
                 .iter()
-                .filter(|param| param.direction == ParameterDirection::Out)
+                .filter(|param| {
+                    param.direction == ParameterDirection::Out
+                        || param.typ.full_name(&env.library) == "Gio.AsyncResult"
+                })
                 .map(|param| {
+                    if param.typ.full_name(&env.library) == "Gio.AsyncResult" {
+                        found_async_result = true;
+                        return Chunk::Name("res".to_string());
+                    }
                     let kind = type_mem_mode(env, param);
                     let mut par: parameter_ffi_call_out::Parameter = param.into();
                     if kind.is_uninitialized() {
@@ -733,6 +740,10 @@ impl Builder {
                     }
                     Chunk::FfiCallOutParameter { par }
                 }),
+        );
+        assert!(
+            found_async_result,
+            "The check *wasn't* performed in analysis part: Guillaume was wrong!"
         );
         let index_to_ignore = find_index_to_ignore(&trampoline.output_params);
         let mut result: Vec<_> = trampoline
@@ -802,7 +813,7 @@ impl Builder {
             });
         body.extend(output_vars);
 
-        let ret_name = if trampoline.ffi_ret.is_some() || !uninitialized_vars.is_empty() {
+        let ret_name = if trampoline.ffi_ret.is_some() {
             "ret"
         } else {
             "_"


EPashkin:
@GuillaumeGomez PR broke gio, now many errors like this:
```
error[E0061]: this function takes 2 parameters but 1 parameter was supplied
   --> D:/eap/rust/0/gio\src\auto\app_info.rs:144:21
    |
144 |             let _ = gio_sys::g_app_info_launch_default_for_uri_finish(&mut error);
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected 2 parameters

```

src/codegen/function_body_chunk.rs
```diff
@@ -718,13 +718,15 @@ impl Builder {
                 type_: "*mut _".to_string(),
             });
         }
-        finish_args.push(Chunk::Name("res".to_string()));
         finish_args.extend(
             trampoline
                 .output_params
                 .iter()
-                .filter(|param| param.direction == ParameterDirection::Out)
+                .filter(|param| param.direction == ParameterDirection::Out || param.name == "res")
```
sdroege: Don't ever work with the names of parameters :) Use the C type (`GAsyncResult`) or the Rust type `gio::AsyncResult`) here instead.

Async finish functions must have this, and exactly once.

GuillaumeGomez: Good idea!

GuillaumeGomez:
Updated.

src/codegen/function_body_chunk.rs
```diff
@@ -718,13 +718,16 @@ impl Builder {
...
         finish_args.extend(
             trampoline
                 .output_params
                 .iter()
-                .filter(|param| param.direction == ParameterDirection::Out)
+                .filter(|param| param.direction == ParameterDirection::Out ||
+                                param.typ.full_name(&env.library) == "Gio.AsyncResult")
                 .map(|param| {
+                    if param.typ.full_name(&env.library) == "Gio.AsyncResult" {
+                        return Chunk::Name("res".to_string());
```
sdroege: Should panic/error out if there is no such parameter. Otherwise looks like this should work.

GuillaumeGomez: At this point, this is already way too late. I *think* it's already checked beforehand.

GuillaumeGomez: Good point! If we ever panic because of it, a fix will be required. I'll add that in the assert message. :D

GuillaumeGomez:
Added the assert.

src/codegen/function_body_chunk.rs
```diff
@@ -734,6 +741,10 @@ impl Builder {
                     Chunk::FfiCallOutParameter { par }
                 }),
         );
+        assert!(
+            found_async_result == true,
```
sdroege: Without the `== true`. This is a bool :P

Boiethios:
That works well. The generated code for GOA compiles correctly :ok_hand:

Boiethios:
There is just a small thing: the return is not used, and that triggers a warning:

```none
warning: unused variable: `ret`
   --> src/auto/account.rs:246:17
    |
246 |             let ret = goa_sys::goa_account_call_ensure_credentials_finish(_source_object as *mut _, out_expires_in.as_mut_ptr(), res, &mut error);
    |                 ^^^ help: consider prefixing with an underscore: `_ret`
    |
    = note: #[warn(unused_variables)] on by default
```

sdroege:
That's an actual bug I guess. The `out_expires_in` is only valid if `true` was returned. And if `false` was returned we need to handle the error.

sdroege:
Sure

GuillaumeGomez:
@Boiethios Try again? :)

EPashkin:
@GuillaumeGomez Thanks, please merge yourself when @Boiethios accepts

Boiethios:
That seems to be ok. Here is the generated code:

    fn call_ensure_credentials<P: IsA<gio::Cancellable>, Q: FnOnce(Result<i32, Error>) + Send + 'static>(&self, cancellable: Option<&P>, callback: Q) {
        let user_data: Box<Q> = Box::new(callback);
        unsafe extern "C" fn call_ensure_credentials_trampoline<Q: FnOnce(Result<i32, Error>) + Send + 'static>(_source_object: *mut gobject_sys::GObject, res: *mut gio_sys::GAsyncResult, user_data: glib_sys::gpointer) {
            let mut out_expires_in = mem::MaybeUninit::uninit();
            let _ = goa_sys::goa_account_call_ensure_credentials_finish(_source_object as *mut _, out_expires_in.as_mut_ptr(), res, &mut error);
            let out_expires_in = out_expires_in.assume_init();
            let result = if error.is_null() { Ok(out_expires_in) } else { Err(from_glib_full(error)) };
            let callback: Box<Q> = Box::from_raw(user_data as *mut _);
            callback(result);
        }
        let callback = call_ensure_credentials_trampoline::<Q>;
        unsafe {
            goa_sys::goa_account_call_ensure_credentials(self.as_ref().to_glib_none().0, cancellable.map(|p| p.as_ref()).to_glib_none().0, Some(callback), Box::into_raw(user_data) as *mut _);
        }
    }

If I understand correctly the code, the return value is discarded since it only indicates if `error` has been set.

GuillaumeGomez:
Ok, then I merge!

Issue #815, opened 2019-07-20
Specific implementation for glib::Pid


GuillaumeGomez:
For the moment, we can't do much since it's an alias over `i32`. Implementing `FromGlib` trait in glib as a first step then handle it directly into `gir` would fix the issue.

Cogitri:
Adding to this, It appears that this doesn't work for at least vte-rs:

```
error[E0308]: mismatched types
    --> /home/rasmus/Projects/vte-rs/src/auto/terminal.rs:1160:79
     |
1160 |             vte_sys::vte_terminal_watch_child(self.as_ref().to_glib_none().0, child_pid);
     |                                                                               ^^^^^^^^^ expected i32, found struct `glib::Pid`
     |
     = note: expected type `i32`
                found type `glib::Pid`
```

For the code: https://github.com/antoyo/vte-rs/pull/4

EPashkin:
It not because it need be `child.pid_to_glib()` too ?

Cogitri:
Huh, if I do `child_pid.pid_to_glib()` I get:

```
error[E0599]: no method named `pid_to_glib` found for type `glib::Pid` in the current scope
    --> /home/rasmus/Projects/vte-rs/src/auto/terminal.rs:1160:89
     |
1160 |             vte_sys::vte_terminal_watch_child(self.as_ref().to_glib_none().0, child_pid.pid_to_glib());
```

Seems like Rust can't decide if it's a i32 or glib::Pid :P

GuillaumeGomez:
Not `pid_to_glib` but `to_glib`.

EPashkin:
@Cogitri sorry, I meant just "to_glib"

Cogitri:
Doesn't work either:
```
error[E0599]: no method named `to_glib` found for type `glib::Pid` in the current scope
    --> src/auto/terminal.rs:1160:89
     |
1160 |             vte_sys::vte_terminal_watch_child(self.as_ref().to_glib_none().0, child_pid.to_glib());
     |                                                                                         ^^^^^^^

```

EPashkin:
Very strange, this line was added 2 month ago  https://github.com/gtk-rs/glib/blob/master/src/source.rs#L56

Cogitri:
Ah, I guess it hasn't made into glib 0.8.2 yet then.

GuillaumeGomez:
Ah indeed. Next release. :)

Cogitri:
Okie, thanks

[...]

PR #816, opened 2019-07-20
Allow nullable callbacks

GuillaumeGomez:
An example of generation: https://github.com/gtk-rs/gio/pull/230

Note that the gio PR needs https://github.com/gtk-rs/gir/issues/815

cc @EPashkin @sdroege

### DIFF
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -224,6 +224,21 @@ fn fixup_special_functions(
     }
 }
 
+fn find_callback_bound_to_destructor(
+    callbacks: &[Trampoline],
+    destroy: &mut Trampoline,
+    destroy_index: usize,
+) -> bool {
+    for call in callbacks {
+        if call.destroy_index == destroy_index {
+            destroy.nullable = call.nullable;
+            destroy.bound_name = call.bound_name.clone();
+            return true;
+        }
+    }
+    false
+}
+
 fn analyze_callbacks(
     env: &Env,
     func: &library::Function,
@@ -262,6 +277,7 @@ fn analyze_callbacks(
             Some(n) => &n,
             None => &func.name,
         };
+        let mut destructors_to_update = Vec::new();
         for pos in 0..parameters.c_parameters.len() {
             // If it is a user data parameter, we ignore it.
             if cross_user_data_check.values().any(|p| *p == pos) || user_data_indexes.contains(&pos)
@@ -345,17 +361,15 @@ fn analyze_callbacks(
                         warn_main!(
                             type_tid,
                             "`{}`: no user data point to the destroy callback",
-                            func_name
+                            func_name,
                         );
                         *commented = true;
                     }
                     // We check if the user trampoline is there. If so, we change the destroy
                     // nullable value if needed.
-                    for call in callbacks.iter() {
-                        if call.destroy_index == pos {
-                            callback.nullable = call.nullable;
-                            break;
-                        }
+                    if !find_callback_bound_to_destructor(&callbacks, &mut callback, pos) {
+                        // Maybe the linked callback is after so we store it just in case...
+                        destructors_to_update.push((pos, destroys.len()));
                     }
                     destroys.push(callback);
                     to_remove.push(pos);
@@ -372,6 +386,19 @@ fn analyze_callbacks(
                     par.scope,
                 )
                 .is_err();
+            }
+        }
+        for (destroy_index, pos_in_destroys) in destructors_to_update {
+            if !find_callback_bound_to_destructor(
+                &callbacks,
+                &mut destroys[pos_in_destroys],
+                destroy_index,
+            ) {
+                warn_main!(
+                    type_tid,
+                    "`{}`: destructor without linked callback",
+                    func_name
+                );
             }
         }
     }
@@ -692,7 +719,7 @@ fn analyze_function(
     if !commented {
         if !destroys.is_empty() || !callbacks.is_empty() {
             if callbacks.iter().any(|c| !c.scope.is_call()) {
-                imports.add("std::boxed::Box as Box_", None);
+                imports.add("std::boxed::Box as Box_");
             }
         }
         if !commented {
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -691,16 +691,8 @@ fn analyze_function(
 
     if !commented {
         if !destroys.is_empty() || !callbacks.is_empty() {
-            if callbacks.iter().any(|c| c.scope.is_async() && *c.nullable) {
-                warn_main!(
-                    type_tid,
-                    "{}: gir cannot generate nullable async callback...",
-                    func.c_identifier.as_ref().unwrap_or(&func.name)
-                );
-                commented = true;
-            }
-            if !commented && callbacks.iter().any(|c| !c.scope.is_call()) {
-                imports.add("std::boxed::Box as Box_");
+            if callbacks.iter().any(|c| !c.scope.is_call()) {
+                imports.add("std::boxed::Box as Box_", None);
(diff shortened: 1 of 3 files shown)


src/codegen/function_body_chunk.rs
```diff
@@ -244,12 +244,13 @@ impl Builder {
                     &bounds,
                     &bounds_names,
                     false,
+                    &self.callbacks[callback_pos].bound_name,
```
EPashkin: Why not `&trampoline.bound_name,` in this for?

GuillaumeGomez: I need to check if it's the same. Thanks for the lead!

EPashkin:
"glib::Pid" can be added with this diff, but not sure that it best solution
```diff
+++ b/src/codegen/function_body_chunk.rs
@@ -1307,6 +1307,13 @@ fn add_chunk_for_type(
             )));
             true
         }
+        library::Type::Alias(_) if ty_name == "glib::Pid" => {
+            body.push(Chunk::Custom(format!(
+                "let {0} = glib::Pid({0});",
+                par.name
+            )));
+            true
+        },
         library::Type::Alias(ref x) => add_chunk_for_type(env, x.typ, par, body, ty_name, nullable),
         ref x =>
```

GuillaumeGomez:
I don't want it like this. I'd prefer use a trait instead.

EPashkin:
Agree, also this diff not fix generation of "glib::spawn_async"

src/codegen/function_body_chunk.rs
GuillaumeGomez: I confirm, it's the same!

GuillaumeGomez: Actually only for trampoline. Destroy callbacks have this issue. We need to update their bounds.

EPashkin:
Seems `cargo fmt` is needed

GuillaumeGomez:
I need to update the PR in any case.

GuillaumeGomez:
Updated.

EPashkin:
Not sure that it better than current version, Ex. in gdk.
```diff
+/src/auto/functions.rs
@@74,7 +574,7 @@ pub fn threads_add_idle_full<P: Fn() -> bool + Send + Sync + 'static>(
unsafe extern "C" fn notify_func<P: Fn() -> bool + Send + Sync + 'static>(
        data: glib_sys::gpointer,
    ) {
-        let _callback: Box_<P> = Box_::from_raw(data as *mut _);
+        let _callback: Box_<Fn() + 'static> = Box_::from_raw(data as *mut _);
    }
    let destroy_call3 = Some(notify_func::<P> as _);
    let super_callback0: Box_<P> = function_data;
@@ -630,7 +630,7 @@ pub fn threads_add_timeout_full<P: Fn() -> bool + Send + Sync + 'static>(
    unsafe extern "C" fn notify_func<P: Fn() -> bool + Send + Sync + 'static>(
        data: glib_sys::gpointer,
    ) {
-        let _callback: Box_<P> = Box_::from_raw(data as *mut _);
+        let _callback: Box_<Fn() + 'static> = Box_::from_raw(data as *mut _);
    }
    let destroy_call4 = Some(notify_func::<P> as _);
    let super_callback0: Box_<P> = function_data;
@@ -687,7 +687,7 @@ pub fn threads_add_timeout_seconds_full<P: Fn() -> bool + Send + Sync + 'static>
    unsafe extern "C" fn notify_func<P: Fn() -> bool + Send + Sync + 'static>(
        data: glib_sys::gpointer,
    ) {
-        let _callback: Box_<P> = Box_::from_raw(data as *mut _);
+        let _callback: Box_<Fn() + 'static> = Box_::from_raw(data as *mut _);
```
in gtk:
```diff
+++ b/src/auto/cell_layout.rs
@@ -159,9 +159,7 @@ impl<O: IsA<CellLayout>> CellLayoutExt for O {
             None
         };
         unsafe extern "C" fn destroy_func<P: IsA<CellRenderer>>(data: glib_sys::gpointer) {
-            let _callback: Box_<
-                Option<Box<dyn Fn(&CellLayout, &CellRenderer, &TreeModel, &TreeIter) + 'static>>,
-            > = Box_::from_raw(data as *mut _);
+            let _callback: Box_<Q> = Box_::from_raw(data as *mut _);
         }
         let destroy_call4 = Some(destroy_func::<P> as _);
         let super_callback0: Box_<
```
```

**Comment by sdroege** [comment]
Yes this does not look right. Will comment in more detail later.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

**Comment by sdroege** [comment]
> ```diff
> +/src/auto/functions.rs
> @@74,7 +574,7 @@ pub fn threads_add_idle_full<P: Fn() -> bool + Send + Sync + 'static>(
> unsafe extern "C" fn notify_func<P: Fn() -> bool + Send + Sync + 'static>(
>         data: glib_sys::gpointer,
>     ) {
> -        let _callback: Box_<P> = Box_::from_raw(data as *mut _);
> +        let _callback: Box_<Fn() + 'static> = Box_::from_raw(data as *mut _);
>     }
> ```

This changes things partially from static to dynamic dispatch, and will crash. We don't create a trait object, but the box is then interpreted as one here. And let's keep static dispatch please.



> ```diff
> +++ b/src/auto/cell_layout.rs
> @@ -159,9 +159,7 @@ impl<O: IsA<CellLayout>> CellLayoutExt for O {
>              None
>          };
>          unsafe extern "C" fn destroy_func<P: IsA<CellRenderer>>(data: glib_sys::gpointer) {
> -            let _callback: Box_<
> -                Option<Box<dyn Fn(&CellLayout, &CellRenderer, &TreeModel, &TreeIter) + 'static>>,
> -            > = Box_::from_raw(data as *mut _);
> +            let _callback: Box_<Q> = Box_::from_raw(data as *mut _);
>          }
> ```

This does the exact opposite. Created as trait object but then not interpreted as one.

Something's not right with this change :)

**Comment by GuillaumeGomez** [comment]
Indeed. It seems like I only made *half* the changes required...

**Comment by sdroege** [comment]
> Indeed. It seems like I only made _half_ the changes required...

It should not have any effect on already generated code though, so you also changed things that do not have to be changed :)

**Comment by GuillaumeGomez** [comment]
No, some changes had to be done, notably on the destructors. The current code couldn't handle more than one correctly.

**Comment by GuillaumeGomez** [comment]
Should be ready now. Updated the associated PR as well.

File: src/analysis/functions.rs [hunk 2]
```diff
@@ -224,6 +224,21 @@ fn fixup_special_functions(
     }
 }
 
+fn find_callback_bound_to_destructor(
+    callbacks: &[Trampoline],
+    destroy: &mut Trampoline,
+    destroy_index: usize,
+) -> bool {
+    for call in callbacks.iter() {
```
EPashkin: Seems `.iter()` not needed

EPashkin: Maybe change to `let Some(call) = callbacks.iter().find(...) {...}` but not sure that it better

EPashkin:
@GuillaumeGomez Thanks, I have only small nit

sdroege:
See comment in the gio PR. Otherwise the generated code looks fine

src/analysis/functions.rs
GuillaumeGomez: I prefer the `for` loop syntax so I just removed the `.iter()` call.

EPashkin: I too 😉

EPashkin:
@GuillaumeGomez Thanks, now wait for CI

GuillaumeGomez:
CI is green! :)

EPashkin:
@GuillaumeGomez Thanks again

PR #817, opened 2019-07-22
Handle glib::Pid conversions specifically

GuillaumeGomez:
cc @EPashkin @sdroege


GuillaumeGomez:
I updated https://github.com/gtk-rs/gio/pull/230 so you can see this change too.

sdroege:
Generated code looks good to me

EPashkin:
@GuillaumeGomez Thanks, I merge this as is. but, please, don't forget to fix it in `glib::spawn_async` too

GuillaumeGomez:
@EPashkin I'll create an issue just in case.

Issue #818, opened 2019-07-23
Allow to skip the build.rs when building the doc


Boiethios:
A published sys crate cannot have its doc on [docs.rs](docs.rs) (unless the linked library is installed in the machine that does the build) because the build script returns an error while searching for the library, and thus the doc build breaks.

There should be a feature "docs.rs", so that the build.rs could be:

```rust
#[cfg(feature = "docs-rs")]
fn main() {} // Skip the script when the doc is building

#[cfg(not(feature = "docs-rs"))]
fn main() {
    // The generated code for the build
}
```

And in the Cargo.toml:

```toml
[features]
docs-rs = []

[package.metadata.docs.rs]
features = [ "docs.rs" ]
```

I've made a POC, and it works well: https://docs.rs/goa-sys

GuillaumeGomez:
We should allow gir users to be able to generate it but I don't want to enforce it. I'll implement it.

PR #820, opened 2019-07-23
Handle doc generation for docs.rs

GuillaumeGomez:
Fixes https://github.com/gtk-rs/gir/issues/818.

I'm pretty sure I'm forgetting something here though... Well it's late, I'll see that tomorrow...

cc @EPashkin @sdroege

### DIFF
--- a/src/codegen/sys/build.rs
+++ b/src/codegen/sys/build.rs
@@ -26,14 +26,24 @@ fn generate_build_script(w: &mut dyn Write, env: &Env) -> Result<()> {
     write!(
         w,
         "{}",
-        r##"extern crate pkg_config;
+        r##"#[cfg(not(feature = "docs-rs"))]
+extern crate pkg_config;
 
+#[cfg(not(feature = "docs-rs"))]
 use pkg_config::{Config, Error};
+#[cfg(not(feature = "docs-rs"))]
 use std::env;
+#[cfg(not(feature = "docs-rs"))]
 use std::io::prelude::*;
+#[cfg(not(feature = "docs-rs"))]
 use std::io;
+#[cfg(not(feature = "docs-rs"))]
 use std::process;
 
+#[cfg(feature = "docs-rs")]
+fn main() {} // prevent linking libraries to avoid documentation failure
+
+#[cfg(not(feature = "docs-rs"))]
 fn main() {
     if let Err(s) = find() {
         let _ = writeln!(io::stderr(), "{}", s);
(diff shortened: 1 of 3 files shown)


src/codegen/sys/build.rs
```diff
@@ -26,21 +26,32 @@ fn generate_build_script(w: &mut dyn Write, env: &Env) -> Result<()> {
...
 use std::env;
+#[cfg(not(feature = "docs-rs"))]
 use std::io::prelude::*;
+#[cfg(not(feature = "docs-rs"))]
 use std::io;
+#[cfg(not(feature = "docs-rs"))]
 use std::process;
 
+#[cfg(feature = "docs-rs")]
+fn main() {} // prevent linking libraries to prevent documentation failure
```
EPashkin: Maybe rephrase this to remove double "prevent"

EPashkin:
@GuillaumeGomez Thanks, LGFM
IMHO only possible addition is updating "README.md"

GuillaumeGomez:
No, I just remembered! I need to add in the features the highest version and "dox".

GuillaumeGomez:
Ok, I think it's ready now.

src/codegen/sys/cargo_toml.rs
```diff
@@ -92,19 +92,42 @@ fn fill_in(root: &mut Table, env: &Env) {
...
+        let mut versions = env
             .namespaces
             .main()
             .versions
             .iter()
-            .filter(|&&v| v > env.config.min_cfg_version);
-        versions.fold(None::<Version>, |prev, &version| {
+            .filter(|&&v| v > env.config.min_cfg_version)
+            .collect::<Vec<_>>();
+        let max_version = versions.iter().max().map(|x| *x);
```
EPashkin: Why not get last element in sorted array?

EPashkin:
It is great idea to add highest version, thanks

src/codegen/sys/cargo_toml.rs
GuillaumeGomez: Good point!

sdroege:
Shouldn't we by default just add all features apart from `purge-lgpl-docs`? That's the right thing for almost every crate.

GuillaumeGomez:
The `lgpl-*` features aren't part of sys crates and should be provided through the new option. Also, I'm not a big fan of adding all features. For instance, versions "contain" lower versions so it'd just be noise.

GuillaumeGomez:
Updated.

EPashkin:
@GuillaumeGomez Thanks

PR #825, opened 2019-08-08
Migrate property getters to new signature for `Value::get`

fengalin:
... in consequence of https://github.com/gtk-rs/glib/pull/513

### DIFF
--- a/src/codegen/property_body.rs
+++ b/src/codegen/property_body.rs
@@ -138,7 +138,10 @@ impl Builder {
         } else {
             ".unwrap()"
         };
-        body.push(Chunk::Custom(format!("value.get(){}", unwrap)));
+        body.push(Chunk::Custom(format!(
+            "value.get().expect(\"Return Value for property `{}` getter\"){}",
+            self.name, unwrap,
+        )));
 
         let unsafe_ = Chunk::Unsafe(body);
 
--- a/src/codegen/signal.rs
+++ b/src/codegen/signal.rs
@@ -138,11 +138,19 @@ pub fn generate(
             )?;
 
             if trampoline.ret.typ != Default::default() {
-                if trampoline.ret.nullable == library::Nullable(true) {
-                    writeln!(w, "{}res.unwrap().get()", tabs(indent + 1),)?;
+                let unwrap = if trampoline.ret.nullable == library::Nullable(true) {
+                    ""
                 } else {
-                    writeln!(w, "{}res.unwrap().get().unwrap()", tabs(indent + 1),)?;
-                }
+                    ".unwrap()"
+                };
+
+                writeln!(
+                    w,
+                    "{}res.unwrap().get().expect(\"Return Value for `{}`\"){}",
+                    tabs(indent + 1),
+                    emit_name,
+                    unwrap,
+                )?;
             }
             writeln!(w, "{}}}", tabs(indent))?;
         }


src/codegen/property_body.rs
```diff
@@ -138,7 +138,10 @@ impl Builder {
         } else {
             ".unwrap()"
         };
-        body.push(Chunk::Custom(format!("value.get(){}", unwrap)));
+        body.push(Chunk::Custom(format!(
+            "value.get().expect(\"Return Value for property getter\"){}",
```
sdroege: Maybe include the property name here and why this fails

fengalin: Ok for name. However, it though the [error would be displayed in the panic message](https://doc.rust-lang.org/std/result/enum.Result.html#method.expect), so we only needed to write the context in the `expect` message.

sdroege: true

fengalin: ... but unfortunately, `expect` seems to be using the `Debug` version, not the `Display` one: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7da532d4cd23ff02e61fb7085df8bab8

So we would get (example from `gio`):

```
thread 'main' panicked at 'Return Value for property `from-charset` getter: GetError { actual: guint, requested: gboolean }'
```

We loose the `TypeMismatch` part even though it is kind of obvious from the fields. So I guess the `enum` version could be applicable finally:

``` rust
pub enum GetError {
    TypeMismatch { actual: Type, requested: Type },
}
```

sdroege: :+1:

sdroege: OTOH, how does it work for `TryFromIntError` then? That should have the same problem you have here.

fengalin: Not much details: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=0c0d17053f263ea0bb6b2c6cb2340847

sdroege: Let's keep consistent with that then, I guess

sdroege: I.e. it's not our problem but something that should be solved in `expect()` and `unwrap()`

fengalin: All right!

sdroege:
Is this all ready from your side then? :) I'd give it another review later

fengalin:
> Is this all ready from your side then?

Not yet, I'm checking the other crates now :)

fengalin:
It should be good to go now. So far, I checked:

- `gtk-rs/gtk`: generation & unit tests with features `subclassing,futures` (which also pulls `atk`, `gdk`, `gdk-pixbuf` & `gio`)
- `gstreamer-rs/gstreamer`: generation & unit tests with features `subclassing,futures,serde`.

fengalin:
Guys, when you get time, could you review & merge this one please? This will allow me to regen the other crates with the actual `gir` version.

src/codegen/property_body.rs
```diff
@@ -138,7 +138,10 @@ impl Builder {
         } else {
             ".unwrap()"
         };
-        body.push(Chunk::Custom(format!("value.get(){}", unwrap)));
+        body.push(Chunk::Custom(format!(
+            "value.get().expect(\"Return Value for property `{}` getter\"){}",
```
EPashkin: This error not tell about actual reason.
IMHO just `unwrap` give better error.

fengalin: Yes its does... somewhat. See this discussion: https://github.com/gtk-rs/gir/pull/825#pullrequestreview-272474063 (you will need to click "Show resolved").

EPashkin: Sorry, I missed that discussion.

EPashkin:
@fengalin Thanks, I will merge then

fengalin:
Thanks @EPashkin

Issue #829, opened 2019-08-17
Gir think that array is optional in futures


EPashkin:
Ex. parameter `uris` converted with `map`
```
    #[cfg(any(feature = "futures", feature = "dox"))]
    #[cfg(any(feature = "v2_60", feature = "dox"))]
    fn launch_uris_async_future<P: IsA<AppLaunchContext> + Clone + 'static>(
        &self,
        uris: &[&str],
        context: Option<&P>,
    ) -> Box_<dyn future::Future<Output = Result<(), Error>> + std::marker::Unpin> {
        use fragile::Fragile;
        use GioFuture;

        let uris = uris.map(ToOwned::to_owned);
        let context = context.map(ToOwned::to_owned);
        GioFuture::new(self, move |obj, send| {
            let cancellable = Cancellable::new();
            let send = Fragile::new(send);
            obj.launch_uris_async(
                uris.as_ref().map(::std::borrow::Borrow::borrow),
                context.as_ref().map(::std::borrow::Borrow::borrow),
                Some(&cancellable),
                move |res| {
                    let _ = send.into_inner().send(res);
                },
            );

            cancellable
        })
    }
```

cc @sdroege, @GuillaumeGomez

EPashkin:
Write this function tough for me (lifetime problem)
so I add "no_future" flag to gir to possibility to disable future generation,
and this function can be added separately as manual.

sdroege:
The non-futures version of this function is equally broken. The `uris` have to stay alive until `finish()` function is called, which is not guaranteed in either case currently.

We need to convert to some owned array of owned strings before calling `launch_uris_async()` and ensure that this stays alive long enough.

EPashkin:
I tried that but can't solve problem with lifetimes,
as this function is blocker I propose to ignore it, until we can find manual or generated solution

sdroege:
Not sure what problem you ran into, this is basically like `write_async` :) But yes, for now ignore the async and futures variant until someone has to time to add a manual implementation.

Please create an issue in gio so we don't forget it though.

PR #830, opened 2019-08-18
Add flag to disable future generation

EPashkin:
Part of https://github.com/gtk-rs/gir/issues/829

cc @GuillaumeGomez , @sdroege


sdroege:
While this would be useful to have, I'd rather if we fixed the code generation bug in the function you noticed.

Issue #827, opened 2019-08-14
Generate set_property with IsA


EPashkin:
See https://github.com/gtk-rs/gtk/pull/875

`Bounds` need to have function that return bound for property, maybe with type other that `Bound`

EPashkin:
All PRs merged, I close this

PR #838, opened 2019-09-07
Use IsA for property setters for non-final objects

EPashkin:
Part of https://github.com/gtk-rs/gir/issues/827

cc @GuillaumeGomez , @sdroege

### DIFF
--- a/src/analysis/properties.rs
+++ b/src/analysis/properties.rs
@@ -1,5 +1,6 @@
 use crate::{
     analysis::{
+        bounds::PropertyBound,
         imports::Imports,
         ref_mode::RefMode,
         rust_type::*,
@@ -25,6 +26,7 @@ pub struct Property {
     pub nullable: library::Nullable,
     pub get_out_ref_mode: RefMode,
     pub set_in_ref_mode: RefMode,
+    pub set_bound: Option<PropertyBound>,
     pub version: Option<Version>,
     pub deprecated_version: Option<Version>,
 }
@@ -189,6 +191,7 @@ fn analyze_property(
             nullable,
             get_out_ref_mode,
             set_in_ref_mode,
+            set_bound: None,
             version: prop_version,
             deprecated_version: prop.deprecated_version,
         })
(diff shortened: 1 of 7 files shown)


src/analysis/properties.rs
```diff
@@ -200,9 +203,21 @@ fn analyze_property(
...
         }
+        let set_bound = PropertyBound::get(env, prop.typ);
         if type_string.is_ok() {
             imports.add("gobject_sys");
             imports.add("glib::Value");
+            if set_bound.is_some() {
+                imports.add("glib::object::IsA");
+                imports.add("glib::value::SetValueOptional");
+                if !*nullable {
+                    //TODO: support nonnulable setter if found any
```
EPashkin: No nonnulable setters in gtk-rs

src/codegen/general.rs
```diff
@@ -514,16 +514,13 @@ pub fn declare_default_from_new(
...
-            }
-            _ => es.push(c),
-        })
-        .count();
+    for c in s.chars() {
+        match c {
+            '\"' | '\\' => es.push('\\'),
+            _ => (),
+        }
+        es.push(c)
```
EPashkin: nightly clippy strart nags about this and other `.count()`

src/analysis/properties.rs
sdroege: What does this mean? :)

src/codegen/general.rs
sdroege: What exactly does it complain about, can you paste?

EPashkin:
```
warning: this call to `map()` won't have an effect on the call to `count()`
   --> src\codegen\general.rs:517:13
    |
517 |       let _ = s
    |  _____________^
518 | |         .chars()
519 | |         .map(|c| match c {
520 | |             '\"' | '\\' => {
...   |
525 | |         })
526 | |         .count();
    | |________________^
    |
    = note: `#[warn(clippy::suspicious_map)]` on by default
    = help: make sure you did not confuse `map` with `filter`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#suspicious_map
```

src/analysis/properties.rs
EPashkin: I thought that maybe exists some property that can't be NULL,
then generator support this case too (with `param: &T` instead of `param: Option<&T>`).
But check shows that no that properties in gtk-rs crates, so I just add warning and ToDo.

src/codegen/general.rs
sdroege: Maybe you want `for_each()` here instead?

src/analysis/properties.rs
sdroege: Sounds fine to me

src/codegen/general.rs
EPashkin: IMHO just `for` is better, no closure created etc.

sdroege: For this kind of code, I agree, yes.

The closure is optimized away btw, but with a for loop it's simply more readable IMHO :)

src/nameutil.rs
```diff
@@ -116,9 +115,8 @@ lazy_static! {
             "where", "while", "yield",
         ]
         .iter()
-        .map(|k| map.insert(*k, format!("{}_", k)))
-        .count();
-        map
+        .map(|k| (*k, format!("{}_", k)))
+        .collect()
```
EPashkin: Seems this case can be just collected without temporary HashMap

PR #842, opened 2019-09-13
Correctly detect not generated builders

### LINKED ISSUE #841: not_bound options shows generated builders whereas it shouldn't
For example, on gtk, running with `not_bound` option shows `Gtk.AboutDialogBuilder` as being not generated whereas it is.

GuillaumeGomez:
Fixes #841.

cc @EPashkin @sdroege


EPashkin:
@GuillaumeGomez Thanks

Issue #845, opened 2019-09-18
Invalid cargo key for docs.rs


GuillaumeGomez:
I badly added it. It should be `[package.metadata.docs.rs]` instead of `["package.metadata.docs.rs"]`.

PR #846, opened 2019-09-18
Docs rs

### LINKED ISSUE #844: Handle dox feature like non-sys crates?
Currently, non-sys crate's `dox` feature includes the highest version and other features. It'd be nice if the sys crates did the same so we don't have to include more than the `dox` feature when building on `docs.rs` ([example](https://github.com/gtk-rs/sys/blob/master/gdk-sys/Cargo.toml#L53)).

GuillaumeGomez:
Fixes #844 
Part of #845 

cc @EPashkin @sdroege

### DIFF
--- a/src/codegen/sys/cargo_toml.rs
+++ b/src/codegen/sys/cargo_toml.rs
@@ -92,33 +92,25 @@ fn fill_in(root: &mut Table, env: &Env) {
 
     {
         let features = upsert_table(root, "features");
-        let mut versions = env
+        let versions = env
             .namespaces
             .main()
             .versions
             .iter()
-            .filter(|&&v| v > env.config.min_cfg_version)
-            .collect::<Vec<_>>();
-        versions.sort_unstable();
-        let max_version = versions.last().copied();
-        versions
-            .into_iter()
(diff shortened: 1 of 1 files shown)


src/codegen/sys/cargo_toml.rs
```diff
@@ -92,33 +92,29 @@ fn fill_in(root: &mut Table, env: &Env) {
...
-        versions
-            .into_iter()
-            .fold(None::<Version>, |prev, &version| {
-                let prev_array: Vec<Value> =
-                    prev.iter().map(|v| Value::String(v.to_feature())).collect();
-                features.insert(version.to_feature(), Value::Array(prev_array));
-                Some(version)
-            });
+            .filter(|&&v| v > env.config.min_cfg_version);
+        versions.fold(None::<Version>, |prev, &version| {
```
EPashkin: `versions` seems not need now

EPashkin: Sorry, missed that it fill features.

EPashkin:
@GuillaumeGomez Thanks

PR #848, opened 2019-10-04
Use Box<dyn Error> instead of Box<Error>

sdroege:
And also update dependencies to the latest versions.


EPashkin:
@sdroege Thanks

PR #852, opened 2019-10-19
Remove some duplicate clone calls

alex179ohm:
Hi guys, I just started learning the code and I found some duplicate clone calls.
Maybe this pr can help

### DIFF
--- a/src/analysis/class_builder.rs
+++ b/src/analysis/class_builder.rs
@@ -85,8 +85,6 @@ fn analyze_property(
     configured_properties: &[&config::properties::Property],
     imports: &mut Imports,
 ) -> Option<Property> {
-    //let name = prop.name.clone();
-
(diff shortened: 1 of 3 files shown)


src/analysis/class_builder.rs
```diff
@@ -85,7 +85,7 @@ fn analyze_property(
     configured_properties: &[&config::properties::Property],
     imports: &mut Imports,
 ) -> Option<Property> {
-    let name = prop.name.clone();
+    //let name = prop.name.clone();
```
EPashkin: Seems this line can be removed too

EPashkin:
@alex179ohm Thanks

src/analysis/class_builder.rs
alex179ohm: Ouch, Tanx :)

PR #855, opened 2019-10-20
Remove dependencies

antoyo:
I'll update this text with the compilation time for every change.
It turns out `git2-sys` is very slow to compile (+1 minute).
Maybe we can re-enable the features of `env_logger` since it does not help much.

## Before
Clean build: 4m 15s
Rebuild:     1m 30s

## After removing docopt
Clean build: 2m 26s
Rebuild:     1m 20s

## After disable features of env_logger
Clean build: 2m 23s
Rebuild:     1m 18s

## After removing features of regex
Clean build: 2m 11s
Rebuild:     1m 17s

## After removing git2
Clean build: 1m 28s
Rebuild:     55.24s

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,30 +1,45 @@
+use std::{cell::RefCell, str::FromStr};
+use std::env;
+use std::process;
+
+use getopts::Options;
 use hprof::Profiler;
 use libgir::{self as gir, Config, Library, WorkMode};
-use std::{cell::RefCell, str::FromStr};
 
-static USAGE: &str = "
-Usage: gir [options] [<library> <version>]
-       gir (-h | --help)
+fn print_usage(program: &str, opts: Options) {
 
-Options:
-    -h, --help              Show this message.
-    -c CONFIG               Config file path (default: Gir.toml)
-    -d GIRSPATH             Directory for girs
-    -m MODE                 Work mode: doc, normal, sys or not_bound
-    -o PATH                 Target path
-    --doc-target-path PATH  Doc target path
-    -b, --make-backup       Make backup before generating
-    -s, --stats             Show statistics
-";
+    let brief = format!("Usage: {program} [options] [<library> <version>]
+       {program} (-h | --help)", program = program);
+    print!("{}", opts.usage(&brief));
+}
 
 fn build_config() -> Result<Config, String> {
-    let args = match docopt::Docopt::new(USAGE).and_then(|dopt| dopt.parse()) {
-        Ok(args) => args,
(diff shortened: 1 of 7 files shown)


src/main.rs
```diff
@@ -34,15 +61,18 @@ fn build_config() -> Result<Config, String> {
...
-        args.get_str("<library>"),
-        args.get_str("<version>"),
-        args.get_str("-o"),
-        args.get_str("--doc-target-path"),
-        args.get_bool("-b"),
-        args.get_bool("-s"),
+        matches.opt_str("d").as_ref().map(|string| string.as_str()),
+        matches.free.get(0).as_ref().map(|string| string.as_str()),
+        matches.free.get(1).as_ref().map(|string| string.as_str()),
+        matches.opt_str("o").as_ref().map(|string| string.as_str()),
```
EPashkin: Maybe better add function for `.as_ref().map(|string| string.as_str())`?

EPashkin:
@antoyo Thanks, faster is better.

@GuillaumeGomez, @sdroege What you think?

GuillaumeGomez:
I'm the one who told him to do it. :stuck_out_tongue:

EPashkin:
By way: as I remember rust don't use libgit2 but call git from command line, maybe we can do same.

src/main.rs
```diff
@@ -1,30 +1,67 @@
...
+    let program = args[0].clone();
+
+    let mut options = Options::new();
+    options.optopt(
+        "c",
+        "config",
+        "Config file path (default: Gir.toml)",
+        "CONFIG",
+    );
+    options.optflag("h", "help", "Show this message");
```
sdroege: Compared to docopt or clap this is quite ugly :)

sdroege:
>  It turns out `git2-sys` is very slow to compile (+1 minute).

I think it's much faster if you actually have libgit2 installed on your system :)

But yeah, works for me generally. I still think `getops` is rather ugly compared to alternatives but go for it if @EPashkin is fine with it too

EPashkin:
@antoyo What is time with CLI git?

src/main.rs
GuillaumeGomez: Yep but it's just options and quite a small amount of code so who cares. :)

antoyo:
@EPashkin: I think it's a bit faster than with the previous attempt which was using `libgit2` as a shared library.

antoyo:
This is ready to be merged.
The other crates to remove to improve compile time will require much more work: regex and toml.

EPashkin:
@antoyo Thanks

Issue #849, opened 2019-10-05
Various builders are generated with parameter-less new() function but no Default impl


sdroege:
See e.g. in gio

```
warning: you should consider deriving a `Default` implementation for `auto::list_store::ListStoreBuilder`
  --> src/auto/list_store.rs:37:5
   |
37 | /     pub fn new() -> Self {
38 | |         Self {
39 | |             #[cfg(any(feature = "v2_44", feature = "dox"))]
40 | |             item_type: None,
41 | |         }
42 | |     }
   | |_____^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#new_without_default
help: try this
   |
31 | #[derive(Default)]
   |
```

bilelmoussaoui:
This one is no longer an issue

PR #857, opened 2019-10-21
#849 Add derive Default on Objects with parameters-less new functions

alex179ohm:
has been tested with the gtk updated master brach.
it changes from this previous generated code:
```rust
pub struct AboutDialogBuilder {
     artists: Option<Vec<String>>,
     authors: Option<Vec<String>>,
....
}
impl Default for AboutDialogBuilder {
      fn default() -> Self {
            Self::new()
      }
}
```
to:
```rust
#[derive(Default)]
pub struct AboutDialogBuilder {
     artists: Option<Vec<String>>,
     authors: Option<Vec<String>>,
....
}
```
maybe the `generate_default_for_new` function could be removed definitely, but is still used in `src/codegen/record.rs` in the `generate` function.

### DIFF
--- a/src/codegen/object.rs
+++ b/src/codegen/object.rs
@@ -151,11 +151,8 @@ pub fn generate(
 fn generate_builder(w: &mut dyn Write, env: &Env, analysis: &analysis::object::Info) -> Result<()> {
     let mut methods = vec![];
     let mut properties = vec![];
-    if analysis.functions.iter().any(|f| {
-        !f.visibility.hidden() && f.name == "new" && f.parameters.rust_parameters.is_empty()
-    }) {
-        writeln!(w, "#[derive(Default)]")?;
-    }
+    general::declare_default_from_new(w, env, &analysis.name, &analysis.functions)?;
+    writeln!(w)?;
     writeln!(w, "pub struct {}Builder {{", analysis.name)?;
     for property in &analysis.builder_properties {
         match rust_type(env, property.typ) {
--- a/src/codegen/object.rs
+++ b/src/codegen/object.rs
@@ -70,7 +70,7 @@ pub fn generate(
 
         writeln!(w, "}}")?;
 
-        general::declare_default_from_new(w, env, &analysis.name, &analysis.functions)?;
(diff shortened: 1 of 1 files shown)


EPashkin:
@alex179ohm Thanks.

@antoyo You agree with this change?

antoyo:
@EPashkin: I'm not sure. Can we get a branch of `gtk-rs` generated with this?

sdroege:
@EPashkin @antoyo @GuillaumeGomez Any reason why this wasn't merged yet?

GuillaumeGomez:
No idea.

antoyo:
Because some [stuff](https://github.com/gtk-rs/gtk/pull/905/files#diff-d96e500340aab6b0a9a9ed5c8853d7f4L46-L51) is removed in the generated code while it should not.

sdroege:
I see, that's a problem indeed :) Thanks for writing a comment about that here. @alex179ohm can you fix that?

EPashkin:
Added https://github.com/gtk-rs/gtk/pull/919 with only adding #[derive(Default)] IMHO it is enough,
will add gir PR if CI will be happy

EPashkin:
Seems @alex179ohm missed that we wasn't have `impl Default for xxBuilder` at all, so only needed add one line to builder generator (I missed that too :wink: ).
But deriving `Default` and maybe `Clone` for all builders is IMHO good idea.
@sdroege, @GuillaumeGomez, @antoyo What you think about it?

EPashkin:
Also we can replace https://github.com/gtk-rs/gir/blob/master/src/codegen/object.rs#L222-L234
with call of "xxxBulder::default()"`

GuillaumeGomez:
For me it sounds good but I didn't work at all on builders so I prefer to let @sdroege and @antoyo answer on this.

antoyo:
@EPashkin: Generating both derive is good.
Did you make a new PR?

EPashkin:
@antoyo Yes, I added it, will merge after CI, then update gtk's

EPashkin:
Closing in favor of #876

Issue #735, opened 2019-03-04
Adds unused glib::GString imports


sdroege:
E.g. here
```
warning: unused import: `glib::GString`
  --> gstreamer-editing-services/src/auto/track_element.rs:14:5
   |
14 | use glib::GString;
   |     ^^^^^^^^^^^^^
   |
   = note: #[warn(unused_imports)] on by default

warning: unused import: `glib::GString`
  --> gstreamer-gl/src/auto/glsl_stage.rs:11:5
   |
11 | use glib::GString;
   |     ^^^^^^^^^^^^^
   |
   = note: #[warn(unused_imports)] on by default
```

Code is [here](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/blob/631eee13da3322bb1d6d254c09e9e8a8941f093e/gstreamer-editing-services/src/auto/track_element.rs) and [here](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/blob/631eee13da3322bb1d6d254c09e9e8a8941f093e/gstreamer-gl/src/auto/glsl_stage.rs).

CC @philn

sdroege:
It's probably from the `&[&str]` parameters?

philn:
Most likely :(

PR #858, opened 2019-10-21
Fix invalid import add

GuillaumeGomez:
Fixes https://github.com/gtk-rs/gtk/issues/872
Fixes https://github.com/gtk-rs/gir/issues/735

cc @sdroege @EPashkin

### DIFF
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -977,10 +977,8 @@ fn analyze_callback(
             }
         }
         if let Ok(s) = used_rust_type(env, func.ret.typ, false) {
-            if s != "GString" {
+            if !s.ends_with("GString") {
                 imports_to_add.push(s);
-            } else {
-                imports_to_add.push("String".to_owned());
             }
         }
         let user_data_index = par.user_data_index.unwrap_or_else(|| 0);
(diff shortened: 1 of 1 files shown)


sdroege:
That's covering all of https://github.com/gtk-rs/gir/issues/735 ?

GuillaumeGomez:
I didn't check on gstreamer. I doubt it'd work as is though. I can extend it if you want.

sdroege:
Please do, would be nice to get rid of all those unused `GString` imports

EPashkin:
@GuillaumeGomez You don't know which function added this import in calendar?
I don't see any function with string :(
I worry that we don't add import "String" in some cases (I not found that case in gtk-rs).

GuillaumeGomez:
Ok, extended it a bit. Doesn't have warning on `gstreamer-editing-services` anymore.

@EPashkin I'm removing the imports from the callbacks only so that should be pretty accurate.

EPashkin:
@GuillaumeGomez Now I understand that, thanks.

EPashkin:
Last change broke gio's g_settings_schema_source_list_schemas

GuillaumeGomez:
Interesting! I'll take a look.

GuillaumeGomez:
It now works on gio as well. :)

EPashkin:
@GuillaumeGomez Yes, not only needed changes. Thanks.

EPashkin:
* now only needed changes

EPashkin:
Lets merge

PR #860, opened 2019-10-22
Extend gpointer to void*

### LINKED ISSUE #856: `user_data` for callbacks being `<type name="gpointer" c:type="void*"/>` makes generation fail
Trying to generate high-level bindings for https://github.com/AravisProject/aravis/blob/master/src/arvcamera.h#L46 appears to fail due to `void*` being used instead of `gpointer`. Given the error raised, I assume this is due to https://github.com/gtk-rs/gir/blob/master/src/analysis/functions.rs#L904 .

Running `sed -i 's#void\*#gpointer#' Aravis-0.6.gir` works as a temporary workaround for this issue.

GuillaumeGomez:
Fixes #856.

Waiting for @Ekleog to confirm it works in his case.

### DIFF
--- a/src/analysis/ffi_type.rs
+++ b/src/analysis/ffi_type.rs
@@ -1,6 +1,7 @@
 use crate::{
     analysis::{
         c_type::{implements_c_type, rustify_pointers},
+        is_gpointer,
         rust_type::{Result, TypeError},
     },
     env::Env,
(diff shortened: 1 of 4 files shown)


EPashkin:
This PR comment many function in gio

GuillaumeGomez:
That explains why gio compiled without issues... Not good.

GuillaumeGomez:
No more commented issues.

EPashkin:
@GuillaumeGomez Yes, now no unneeded changes in gtk-rs, Thanks

Ekleog:
@GuillaumeGomez Can confirm it works, thanks!

PR #863, opened 2019-10-23
Correctly generate glib::Error import

### LINKED ISSUE #862: `glib::Error` gets `use`d as `use Error` instead of `use glib::Error` when in the `manual` array
Generated code looks like https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/blob/master/gstreamer-editing-services/src/auto/asset.rs#L24 ; with `GLib.Error` in the `manual` array of the `Gir.toml`.

This can be worked around with `use glib::Error;` at the root of the crate (as is currently done by most if not all gtk-rs crates), but without this it'd fail to compile.

I think the `use` statement should be `use glib::Error` directly, to avoid this issue :)

GuillaumeGomez:
Fixes #862.

cc @EPashkin @sdroege @Ekleog

### DIFF
--- a/src/analysis/imports.rs
+++ b/src/analysis/imports.rs
@@ -58,7 +58,7 @@ impl Imports {
     /// Declares that name should be available through its last path component.
     ///
     /// For example, if name is `X::Y::Z` then it will be available as `Z`.
-    /// Uses defaults
+    /// Uses defaults.
     pub fn add(&mut self, name: &str) {
         if let Some(ref defined) = self.defined {
             if name == defined {
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -323,8 +323,8 @@ fn find_error_type(env: &Env, function: &Function) -> String {
         .iter()
         .find(|param| param.direction == ParameterDirection::Out && param.name == "error")
         .expect("error type");
-    if let Type::Record(ref record) = *env.type_(error_param.typ) {
-        return record.name.clone();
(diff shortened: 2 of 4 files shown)


src/analysis/imports.rs
```diff
@@ -175,7 +175,10 @@ impl Imports {
     ///
     /// Returns `None` if name matches crate name exactly. Otherwise returns
     /// name with crate name prefix stripped or full name if there was no match.
-    fn strip_crate_name<'a>(&self, name: &'a str) -> Option<&'a str> {
+    fn strip_crate_name<'a>(&self, mut name: &'a str) -> Option<&'a str> {
+        if name == "Error" && self.crate_name != "glib" {
+            name = "glib::Error";
```
sdroege: This does not seem correct. Why does it even end up here without the `glib::` prefix?

GuillaumeGomez: This is actually an excellent question. Since it's part of `manual` imports everytime, it \*should\* not be needed. This fix is therefore incomplete. I need to find what's going on.

GuillaumeGomez:
I now extended the fix. Way cleaner, better, faster, etc. :)

GuillaumeGomez:
And now with fixed formatting.

GuillaumeGomez:
ping @EPashkin @sdroege :)

EPashkin:
@GuillaumeGomez Thanks, I will wait for @sdroege just in case

sdroege:
:+1:

Issue #819, opened 2019-07-23
Full implementation for glib::Pid


GuillaumeGomez:
As stated in this [comment](https://github.com/gtk-rs/gir/pull/817#issuecomment-513892295).

PR #864, opened 2019-10-23
Add missing from_glib conversion for Pid

GuillaumeGomez:
Fixes https://github.com/gtk-rs/gir/issues/819

cc @EPashkin @sdroege


EPashkin:
@GuillaumeGomez Thanks, glib now compiled

PR #866, opened 2019-10-28
Provide the full path to the Inhibit type for signals

sdroege:
Otherwise glib::signal::Inhibit needs to be re-exported from the crate root.

CC @EPashkin


EPashkin:
@sdroege Thanks

PR #869, opened 2019-11-11
Generate pinned box futures and use the stabilized futures

sdroege:
@GuillaumeGomez This depends on https://github.com/gtk-rs/glib/pull/531

When regenerating, all other crates need their `futures` feature removed and the dependency on the `futures` crate


sdroege:
@EPashkin Please review and merge, then we update everything :)

EPashkin:
@sdroege Thanks

PR #870, opened 2019-11-15
Migrate to `tempfile` from deprecated `temdir`

silwol:
The `tempdir` crate was deprecated by its authors, and `tempfile` is
recommended as a replacement.

I assume this is the first time that a dependency needs to be removed
from `Cargo.toml` files, because I found no functionality to remove the
entry. I created an `unset` function analogous to the `set_string`
function found in the `cargo_toml.rs` file.

Rationale: I do packaging work for rust crates in Debian, and due to
it's deprecation, we decided to not package the `tempdir` crate at all,
but to submit patches to the projects that still use it instead.

### DIFF
--- a/src/codegen/sys/cargo_toml.rs
+++ b/src/codegen/sys/cargo_toml.rs
@@ -87,7 +87,8 @@ fn fill_in(root: &mut Table, env: &Env) {
     {
         let dev_deps = upsert_table(root, "dev-dependencies");
         set_string(dev_deps, "shell-words", "0.1.0");
-        set_string(dev_deps, "tempdir", "0.3");
+        set_string(dev_deps, "tempfile", "3");
+        unset(dev_deps, "tempdir");
     }
 
     {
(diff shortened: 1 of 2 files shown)


src/codegen/sys/cargo_toml.rs
```diff
@@ -151,6 +152,10 @@ fn set_string<S: Into<String>>(table: &mut Table, name: &str, new_value: S) {
     table.insert(name.into(), Value::String(new_value.into()));
 }
 
+fn unset(table: &mut Table, name: &str) {
+    table.remove(name.into());
```
EPashkin: Seems `into` not needed: https://travis-ci.org/gtk-rs/gir/jobs/612236198#L663
```
warning: identical conversion

   --> src/codegen/sys/cargo_toml.rs:156:18

    |

156 |     table.remove(name.into());

    |                  ^^^^^^^^^^^ help: consider removing `.into()`: `name`

    |

    = note: `#[warn(clippy::identity_conversion)]` on by default

    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#identity_conversion
```

silwol:
You're right. I had simply taken the set_value function as a basis without considering the fact that `remove` takes the key by ref in contrast to `insert`.

EPashkin:
@silwol Thanks. If @GuillaumeGomez and @sdroege is not against this change I will merge it today and add PR to sys

EPashkin:
@GuillaumeGomez IMHO not, as we don't add lock file to git in sys crates

GuillaumeGomez:
Ah indeed, my bad! I thought the change was done on the gir's `Cargo.lock` file. Then all good for me!

PR #873, opened 2019-11-19
Rename

GuillaumeGomez:
cc @EPashkin @sdroege

### DIFF
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -514,6 +514,11 @@ fn analyze_function(
         .filter_map(|f| f.version)
         .min()
         .or(func.version);
+    let name = configured_functions
+        .iter()
+        .filter_map(|f| f.rename.clone())
+        .next()
+        .unwrap_or_else(|| name);
     let version = env.config.filter_version(version);
     let deprecated_version = func.deprecated_version;
     let cfg_condition = configured_functions
--- a/src/codegen/doc/mod.rs
+++ b/src/codegen/doc/mod.rs
@@ -435,10 +435,7 @@ where
     if st.name != fn_name {
         st.name = fn_name.to_owned();
     }
-    let ty = TypeStruct {
-        parent,
-        ..st
-    };
+    let ty = TypeStruct { parent, ..st };
     let self_name: Option<String> = fn_
         .parameters()
         .iter()
--- a/src/codegen/doc/mod.rs
+++ b/src/codegen/doc/mod.rs
@@ -254,8 +254,10 @@ fn create_object_doc(w: &mut dyn Write, env: &Env, info: &analysis::object::Info
     };
 
     for function in functions {
+        let configured_functions = obj.functions.matched(&function.name);
         let ty = if has_trait && function.parameters.iter().any(|p| p.instance_parameter) {
-            let configured_functions = obj.functions.matched(&function.name);
+            // We use "original_name" here to be sure to get the correct object since the "name"
+            // field could have been renamed.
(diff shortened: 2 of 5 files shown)


EPashkin:
@GuillaumeGomez Thanks

src/analysis/functions.rs
```diff
@@ -514,6 +514,11 @@ fn analyze_function(
         .filter_map(|f| f.version)
         .min()
         .or(func.version);
+    let name = configured_functions
+        .iter()
+        .filter_map(|f| f.rename.clone())
```
sdroege: Shouldn't you filter here first to make sure `f.name == name`? Or does that happen elsewhere already?

src/config/ident.rs
```diff
@@ -9,6 +10,15 @@ pub enum Ident {
     Pattern(Regex),
 }
 
+impl fmt::Display for Ident {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match *self {
+            Ident::Name(ref name) => write!(f, "{}", name),
+            Ident::Pattern(_) => write!(f, "Regex"),
```
sdroege: Should we maybe store the regex pattern string here so that it can be printed? Seems useful

src/analysis/functions.rs
GuillaumeGomez: Indeed!

src/config/ident.rs
GuillaumeGomez: I can add a todo if you want? For now, it's not used so...

sdroege: Sure

src/analysis/functions.rs
sdroege: Does the `filter_map` just above also miss that? Maybe the filtering is actually happening somewhere else already and that's why it works at all? Check the code in more detail please :)

GuillaumeGomez: I can confirm that it's not good. Fixing it. :)

EPashkin: configured_functions is filtered by function name by caller,
only problem is it pattern that applied to many functions,
and IMHO its not problem.
I more worry about docs.

GuillaumeGomez: That's a good point.

GuillaumeGomez: I'll need to check for the doc generation. For now, it's very certainly broken in this case.

src/codegen/doc/mod.rs
```diff
@@ -254,8 +254,10 @@ fn create_object_doc(w: &mut dyn Write, env: &Env, info: &analysis::object::Info
     };
 
     for function in functions {
+        let configured_functions = obj.functions.matched(&function.name);
```
EPashkin: IMHO function name here replaced already, so no one will be found here

GuillaumeGomez: You fell in the same trap as I did. :)

Here is a `library::Function` list, whereas I make the change inside the `analysis::functions::Info` type. So at this point, `function.name` is still correct, which is why I need to get the `rename` thing [here](https://github.com/gtk-rs/gir/pull/873/files#diff-30816be45ddaf9346e6d262a214fb1b4R273).

EPashkin: You right :wink:
Then only need some formating

GuillaumeGomez:
Formatting fixed too now.

EPashkin:
@GuillaumeGomez Thanks

PR #875, opened 2019-11-20
Generate GIO futures code a bit more simple and without requiring all…

sdroege:
… crates to directly use the fragile crate

See https://github.com/gtk-rs/gio/issues/257


sdroege:
This and the gio PR will fail CI until everything's regenerated.

EPashkin:
Ok, then I merge as is.
@sdroege Thanks

PR #878, opened 2019-11-29
generate imports in the same order as rustfmt

GuillaumeGomez:
This is in order to reduce the noise before running `cargo fmt`.

cc @EPashkin @sdroege


EPashkin:
@GuillaumeGomez Thanks. Please merge it when CI pass

PR #881, opened 2019-12-15
Fix fields with fixed size array type

### LINKED ISSUE #880: Misgenerates -sys structs with fixed-size, non-pointer arrays
The .gir file changed like this with the latest version of `gobject-introspection`.

```diff
       <field name="kv_data" writable="1">
-        <doc xml:space="preserve">key validity data</doc>
-        <array zero-terminated="0" c:type="guint8" fixed-size="2">
+        <doc xml:space="preserve"
+             filename="gstmikey.h"
+             line="512">key validity data</doc>
+        <array zero-terminated="0" fixed-size="2">
           <type name="guint8" c:type="guint8*"/>
         </array>
       </field>
```

This causes the generated struct to change like this
```diff
-    pub kv_data: [*mut u8; 2],
+    pub kv_data: *mut [*mut u8; 2],
```

The old version was correct. The C header looks like this

```c
  guint8  *kv_data[2];
```

@EPashkin Any ideas how we can handle this correctly again?

EPashkin:
Fix #880 

cc @sdroege , @GuillaumeGomez

### DIFF
--- a/src/library_postprocessing.rs
+++ b/src/library_postprocessing.rs
@@ -342,7 +342,10 @@ impl Library {
                             }
                             if let Type::FixedArray(_, _, Some(ref c_type)) = *field_type {
                                 // fixed-size Arrays can only have inner c_type
-                                actions.push((tid, fid, Action::SetCType(c_type.clone())));
+                                // HACK: field c_type used only in sys mode for pointer checking
+                                // so any string without * will work
+                                let array_c_type = "fixed_array".to_owned();
+                                actions.push((tid, fid, Action::SetCType(array_c_type)));
                                 continue;
                             }
                             error!("Field `{}::{}` is missing c:type", name, &field.name);


EPashkin:
May be better make `create_fixed_array_c_type_from_inner` return some fixed string like "fixed_array" as c_type actually not used in normal mode and only used for detecting pointers in sys.
Current version can go wrong way for double pointers "gpointer**" but no such changes in gtk_rs.

src/library_postprocessing.rs
```diff
@@ -503,3 +504,14 @@ impl Library {
         }
     }
 }
+
+fn create_fixed_array_c_type_from_inner(inner_c_type: &str) -> String {
+    let mut c_type = inner_c_type.trim_end();
+    // removing last *
+    if let Some(pos) = c_type.rfind("*") {
```
GuillaumeGomez: You're sure this is a good idea?

GuillaumeGomez:
We'd better make the code handle cases of double pointers as well otherwise it'll blow at some point.

EPashkin:
@GuillaumeGomez Then simplest solution just return "fixed_array"

GuillaumeGomez:
If that works and is simpler, then let's go for it!

src/library_postprocessing.rs
EPashkin: Not fuly. But before in .gir files in "array" tag was element type while in inner "type" - array type

EPashkin:
"fixed_array" works: in both sys and normal mode no unneeded changes.
@sdroege are you not against this hack?

GuillaumeGomez:
He said that as long as it didn't change anything else, it was ok for him so whenever you're ready to merge @EPashkin !

EPashkin:
Ok, then I use "hack"

GuillaumeGomez:
:+1:

sdroege:
Now that the release is out, is this solution good for now or should we work on a cleaner/better/more versatile solution?

EPashkin:
IMHO current solution is good until someone report that it not works.
I.E. I almost sure that we will have problem with field with type pointer to fixed size arrays,
but are there any in wildlife ?

sdroege:
You mean if we have a `gchar *foo[2]` in C for example, which would have to map to a `[*const c_char; 2]` or similar? There are at least the padding arrays but those we handle just fine (i.e. `gpointer padding[16]`).

EPashkin:
No I mean rust `*const [*const c_char; 2]`, i.e. field (or argument) is pointer.

sdroege:
There's no such thing in C, so we're safe. That fixed-length array there would just be another pointer indirection.

EPashkin:
https://www.geeksforgeeks.org/pointer-array-array-pointer/ ?

sdroege:
Yes you can do that but from a compiler point of view it's the same as just another pointer indirection. I think current C compilers don't even warn you if the arrays you're using are of the wrong size.

Also I doubt the gobject-introspection C parser can actually handle that. I'm not aware of any GLib-based API to do that.

EPashkin:
@sdroege Thanks for explanation, so we really safe

PR #883, opened 2019-12-17
Add new nullable_return_is_error configuration for return types

sdroege:
This allows converting an Option<T> into an Result<T, glib::BoolError>,
which in case of Option signalling error allows for more ergonomic error
handling code.

----

CC @EPashkin @GuillaumeGomez

### DIFF
--- a/src/codegen/return_value.rs
+++ b/src/codegen/return_value.rs
@@ -43,13 +43,29 @@ impl ToReturnValue for library::Parameter {
 impl ToReturnValue for analysis::return_value::Info {
     fn to_return_value(&self, env: &Env, is_trampoline: bool) -> String {
         match self.parameter {
-            Some(ref par) => par.to_return_value(env, is_trampoline),
+            Some(ref par) => {
+                let name = par.to_return_value(env, is_trampoline);
+                if self.nullable_return_is_error.is_some() && name.starts_with(" -> Option<") {
+                    // Change ` -> Option<T>` to ` -> Result<T, glib::BoolError>`
+                    format!(
+                        " -> Result<{}, {}BoolError>",
+                        &name[11..(name.len() - 1)],
+                        if env.namespaces.glib_ns_id == namespaces::MAIN {
+                            ""
+                        } else {
+                            "glib::"
+                        }
+                    )
+                } else {
+                    name
+                }
+            }
             None => String::new(),
         }
     }
 }
 
-pub fn out_parameter_as_return_parts(
+fn out_parameter_as_return_parts(
     analysis: &analysis::functions::Info,
     is_glib_crate: bool,
 ) -> (&'static str, &'static str) {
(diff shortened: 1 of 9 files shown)


src/codegen/return_value.rs
```diff
@@ -43,13 +43,29 @@ impl ToReturnValue for library::Parameter {
 impl ToReturnValue for analysis::return_value::Info {
     fn to_return_value(&self, env: &Env, is_trampoline: bool) -> String {
         match self.parameter {
-            Some(ref par) => par.to_return_value(env, is_trampoline),
+            Some(ref par) => {
+                let name = par.to_return_value(env, is_trampoline);
+                if self.nullable_return_is_error.is_some() && name.starts_with(" -> Option<") {
+                    let mut name = name.replace("Option<", "Result<");
```
GuillaumeGomez: I'm definitely not a big fan of this! `Option<Option<` would be bad. Please change it to:

```rust
let parts = name.splitn(2, "<").collect();
if parts.len() > 1 && parts[0] == "Option" {
    format!("Result<{}", parts[1])
} else {
    parts.join('<')
}
```

src/codegen/return_value.rs
```diff
@@ -43,13 +43,29 @@ impl ToReturnValue for library::Parameter {
 impl ToReturnValue for analysis::return_value::Info {
     fn to_return_value(&self, env: &Env, is_trampoline: bool) -> String {
         match self.parameter {
-            Some(ref par) => par.to_return_value(env, is_trampoline),
+            Some(ref par) => {
+                let name = par.to_return_value(env, is_trampoline);
```
GuillaumeGomez: Also, shouldn't it be done inside this method instead?

src/codegen/return_value.rs
EPashkin: Or as we already checked for " -> Option<" we can replace it to " -> Result<"

src/codegen/return_value.rs
sdroege: Inside that function is more generic and also used in other contexts than return value of normal functions, that's why I've put it outside

src/codegen/return_value.rs
GuillaumeGomez: That'd be better indeed!

sdroege: @EPashkin you mean simply doing `format!(" -> Result<{}", name[11..]")` or similar?

EPashkin: yes, or change replace text

src/codegen/return_value.rs
```diff
@@ -79,9 +95,25 @@ pub fn out_parameter_as_return_parts(
...
+                        ("Result<(", "), BoolError>")
+                    } else {
+                        ("Result<(", "), glib::BoolError>")
+                    }
+                } else {
+                    ("Option<(", ")>")
+                }
             } else {
-                ("Option<", ">")
+                if analysis.ret.nullable_return_is_error.is_some() {
```
EPashkin: Clippy says that it better be `} else if analysis.."`

EPashkin:
Also new warning than IMHO better ignored by "#[allow(clippy::large_enum_variant))".
```
warning: large size difference between variants
  --> src\chunk\chunk.rs:23:5
   |
23 | /     FfiCallConversion {
24 | |         ret: return_value::Info,
25 | |         array_length_name: Option<String>,
26 | |         call: Box<Chunk>,
27 | |     },
   | |_____^
   |
   = note: `#[warn(clippy::large_enum_variant)]` on by default
help: consider boxing the large fields to reduce the total size of the enum
  --> src\chunk\chunk.rs:23:5
   |
23 | /     FfiCallConversion {
24 | |         ret: return_value::Info,
25 | |         array_length_name: Option<String>,
26 | |         call: Box<Chunk>,
27 | |     },
   | |_____^
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
```

sdroege:
Fixed, and also some other clippy warnings. Thanks for the review

EPashkin:
@sdroege Big thanks

PR #886, opened 2019-12-30
Use once_cell::sync::Lazy instead of lazy_static

sdroege:
This works without requiring any macros.

----

@EPashkin @GuillaumeGomez any opinions?

### DIFF
--- a/src/codegen/sys/functions.rs
+++ b/src/codegen/sys/functions.rs
@@ -12,7 +12,7 @@ use std::io::{Result, Write};
 //used as glib:get-type in GLib-2.0.gir
 const INTERN: &str = "intern";
 
-static DEFAULT_OBJ: Lazy<GObject> = Lazy::new(|| Default::default());
+static DEFAULT_OBJ: Lazy<GObject> = Lazy::new(Default::default);
 
 pub fn generate_records_funcs(
     w: &mut dyn Write,
--- a/src/codegen/sys/functions.rs
+++ b/src/codegen/sys/functions.rs
@@ -6,15 +6,13 @@ use crate::{
     library, nameutil,
     traits::*,
 };
-use lazy_static::lazy_static;
+use once_cell::sync::Lazy;
 use std::io::{Result, Write};
 
 //used as glib:get-type in GLib-2.0.gir
 const INTERN: &str = "intern";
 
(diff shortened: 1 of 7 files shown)


EPashkin:
@sdroege big thanks 👍 
I wait for @GuillaumeGomez approval just in case.
Note: travis fails on glib compiling than fixed by adding "once_cell" to it.

src/codegen/sys/functions.rs
```diff
@@ -6,15 +6,13 @@ use crate::{
...
+use once_cell::sync::Lazy;
 use std::io::{Result, Write};
 
 //used as glib:get-type in GLib-2.0.gir
 const INTERN: &str = "intern";
 
-lazy_static! {
-    static ref DEFAULT_OBJ: GObject = Default::default();
-}
+static DEFAULT_OBJ: Lazy<GObject> = Lazy::new(|| Default::default());
```
EPashkin:
```
  --> src\codegen\sys\functions.rs:15:47
   |
15 | static DEFAULT_OBJ: Lazy<GObject> = Lazy::new(|| Default::default());
   |                                               ^^^^^^^^^^^^^^^^^^^^^ help: remove closure as shown: `Default::default`
   |
   = note: `#[warn(clippy::redundant_closure)]` on by default
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_closure

```

GuillaumeGomez:
The code looks good to me. However I'm wondering: what's the advantage of using `once_cell` over `lazy_static`?

sdroege:
> what's the advantage of using `once_cell` over `lazy_static`?

It does not require macros and is faster to compile, doesn't cause a few types to be generated via a macro at each usage, etc. while otherwise having exactly the same behaviour.

There's also an RFC now to get this into std, so switching later and dropping the dependency would be easy: https://github.com/rust-lang/rfcs/pull/2788

GuillaumeGomez:
Awesome! You'll have to update all crates to switch the dependencies but otherwise all good. Thanks!

src/codegen/sys/functions.rs
EPashkin: @sdroege I fixed this in your repo

PR #888, opened 2020-02-04
Add support for PtrArray

majorz:
This is the consecutive PR after including the PtrArray support in glib. Tested with latest NetworkManager introspection definitions.

Related to https://github.com/gtk-rs/glib/pull/559

### DIFF
--- a/src/analysis/conversion_type.rs
+++ b/src/analysis/conversion_type.rs
@@ -75,6 +75,7 @@ impl ConversionType {
             FixedArray(..) => ConversionType::Pointer,
             List(_) => ConversionType::Pointer,
             SList(_) => ConversionType::Pointer,
+            PtrArray(_) => ConversionType::Pointer,
             Function(super::library::Function { ref name, .. }) if name == "AsyncReadyCallback" => {
                 ConversionType::Direct
             }
--- a/src/analysis/ref_mode.rs
+++ b/src/analysis/ref_mode.rs
@@ -41,6 +41,7 @@ impl RefMode {
             | Interface(..)
             | List(..)
             | SList(..)
+            | PtrArray(..)
             | CArray(..) => {
                 if direction == library::ParameterDirection::In {
                     RefMode::ByRef
--- a/src/analysis/rust_type.rs
+++ b/src/analysis/rust_type.rs
@@ -187,7 +187,7 @@ pub fn rust_type_full(
                 Ok(name)
             }
         }
-        List(inner_tid) | SList(inner_tid) | CArray(inner_tid)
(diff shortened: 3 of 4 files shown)


sdroege:
Can you show some generated code from this, maybe in some PR elsewhere? :) The changes here look simple enough, but I'm curious to what that translates in practice.

EPashkin:
@sdroege changes in atk:
```diff
--- a/src/auto/relation.rs
+++ b/src/auto/relation.rs
@@ -47,7 +47,7 @@ pub trait RelationExt: 'static {
 
     fn get_relation_type(&self) -> RelationType;
 
-    //fn get_target(&self) -> /*Unknown conversion*//*Unimplemented*/PtrArray TypeId { ns_id: 1, id: 9 };
+    fn get_target(&self) -> Vec<Object>;
 
     fn remove_target<P: IsA<Object>>(&self, target: &P) -> bool;
 
@@ -81,9 +81,13 @@ impl<O: IsA<Relation>> RelationExt for O {
         }
     }
 
-    //fn get_target(&self) -> /*Unknown conversion*//*Unimplemented*/PtrArray TypeId { ns_id: 1, id: 9 } {
-    //    unsafe { TODO: call atk_sys:atk_relation_get_target() }
-    //}
+    fn get_target(&self) -> Vec<Object> {
+        unsafe {
+            FromGlibPtrContainer::from_glib_none(atk_sys::atk_relation_get_target(
+                self.as_ref().to_glib_none().0,
+            ))
+        }
+    }
 
     fn remove_target<P: IsA<Object>>(&self, target: &P) -> bool {
         unsafe {
```

majorz:
Here is an example with an out parameter: https://github.com/balena-io-modules/libnm-rs/blob/master/src/connection.rs#L699-L708

majorz:
And this is an example with input parameter and return value: https://github.com/balena-io-modules/libnm-rs/blob/master/src/device.rs#L105

EPashkin:
Seems only atk affected from gtk-rs and examples builds fine for me

majorz:
And some example code from me that uses those (I am still short with examples): https://github.com/balena-io-modules/libnm-rs/blob/master/examples/connections.rs

The `get_connections` method + the `filter_connections` method from above.

majorz:
Yeah, from gtk-rs I did not see much PtrArray usage. I think I saw some in gstreamer, but in general it looks like only NetworkManager makes a heavy usage of it.

sdroege:
> I think I saw some in gstreamer

It should also make more usage of it instead of `GList` in many places, but API-stability... :)

Looks all good to me, thanks a lot!

majorz:
Excellent, thanks guys!

PR #890, opened 2020-03-01
Run rustfmt on generated code

EPashkin:
@GuillaumeGomez , @sdroege 

Not fully sure that it good idea: as on windows it runs on gtk for 27sec instead of 0.5sec
and single `cargo fmt` runs for 2 sec,
so I hope that on linux it runs faster.

Also, I am completely against using rustfmt as a library now.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -175,6 +175,11 @@ fn do_main() -> Result<(), String> {
         gir::codegen_generate(&env);
     }
 
+    if !env.config.disable_format && env.config.work_mode.is_generate_rust_files() {
+        let _watcher = statistics.enter("Formatting");
+        gir::fmt::format(&env.config.target_path);
+    }
+
     drop(watcher_total);
     statistics.end_frame();
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -48,6 +48,7 @@ fn build_config() -> Result<Config, String> {
     options.optopt("p", "doc-target-path", "Doc target path", "PATH");
     options.optflag("b", "make-backup", "Make backup before generating");
     options.optflag("s", "stats", "Show statistics");
+    options.optflag("", "disable_format", "Disable formatting generated code");
 
     let matches = match options.parse(&args[1..]) {
         Ok(matches) => matches,
@@ -80,6 +81,7 @@ fn build_config() -> Result<Config, String> {
         matches.opt_str("doc-target-path").as_str_ref(),
         matches.opt_present("b"),
         matches.opt_present("s"),
+        matches.opt_present("disable_format"),
     )
 }
 
@@ -103,6 +105,7 @@ fn do_main() -> Result<(), String> {
         Ok(cfg) => cfg,
         Err(err) => return Err(err),
     };
+    cfg.check_disable_format();
 
     let statistics = Profiler::new("Gir");
     statistics.start_frame();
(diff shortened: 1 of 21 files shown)


GuillaumeGomez:
I'm fine with it but it might be "unexpected" for some people, no?

EPashkin:
IMHO if we do this option "opt-in" it will be used by no one

sdroege:
Maybe it's so slow because your serialize everything on formatting and do it file by file? `cargo fmt` probably does everything in parallel.

Might be looking into what `cargo fmt` actually does.

EPashkin:
IMHO it most reason why it so slow is windows (slower spawn)+antivirus,
I have same problem when `git` was called for all files to get version 😉 .

`Cargo fmt` uses rustfmt as library without cost of spawning at minimum.

Can you time this version on gtk with and without "--disable_format" option?

sdroege:
4.4s without parameters, 1.51s with `--disable_format`.

Why don't we call rustfmt as a library if the spawning is so expensive? Or why don't we spawn rustfmt only once for all files at the end, i.e. just call `cargo fmt` in the end?

EPashkin:
I don't like to add big dependency, that also still have "-nightly" suffix.
I like idea about just running  "cargo fmt"

EPashkin:
Use `cargo fmt` now

EPashkin:
```
Timing information for Gir:                                                                                                                     Total - 1 * 5.0s = 5.0s (100.0%)                                                                                                                Loading - 1 * 2.7s = 2.7s (53.7%)                                                                                                             Preprocessing - 1 * 0ns = 0ns (0.0%)                                                                                                          Update library by config - 1 * 0ns = 0ns (0.0%)                                                                                               Postprocessing - 1 * 14.6ms = 14.6ms (0.3%)                                                                                                   Resolving type ids - 1 * 0ns = 0ns (0.0%)                                                                                                     Checking versions - 1 * 0ns = 0ns (0.0%)                                                                                                      Namespace/symbol/class analysis - 1 * 6.9ms = 6.9ms (0.1%)                                                                                    Analyzing - 1 * 150.0ms = 150.0ms (3.0%)                                                                                                      Generating - 1 * 229.8ms = 229.8ms (4.6%)                                                                                                     Formatting - 1 * 1.9s = 1.9s (38.4%)   
```

EPashkin:
So speed problem is "solved", IMHO now only left is to decide: opt-in, opt-out (current) or this feature not needed.

sdroege:
> I don't like to add big dependency, that also still have "-nightly" suffix.

It's using the compiler infrastructure, so always has to be nightly :)

> IMHO now only left is to decide: opt-in, opt-out (current) or this feature not needed.

opt-out seems fine to me

src/main.rs
```diff
@@ -80,6 +81,7 @@ fn build_config() -> Result<Config, String> {
         matches.opt_str("doc-target-path").as_str_ref(),
         matches.opt_present("b"),
         matches.opt_present("s"),
+        matches.opt_present("disable_format"),
```
sdroege:
```suggestion
        matches.opt_present("disable-format"),
```

EPashkin:
Then I will merge when CI pass.

sdroege:
Now we can remove it from the scripts :)

PR #891, opened 2020-03-12
Allow to override use-return-for-result in function configuration.

de-vri-es:
This PR allows the configuration file to override `use-return-for-result` per function. There are many functions in practice that return a `boolean` or `guint` where the return value is a real value and doesn't just indicate if an error occurred.

You could argue that libraries shouldn't do that, or that the GIR format should allow a tag to distinguish this. However, I think we also have to be able to deal with the current situation, and I figured this is the most logical work-around.

Feedback welcome :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,10 @@ generate_builder = true
         # convert Option return types to Result<T, glib::BoolError> with
         # the given error message on failure
         nullable_return_is_error = "Function failed doing what it is supposed to do"
+        # always include the return value of throwing functions in the returned Result<...>,
+        # without this option bool and guint return values are assumed to indicate success or error,
+        # and are not included in the returned Result<...>
+        use_return_for_result = true
         # change string type. Variants: "utf8", "filename", "os_string"
         string_type = "os_string"
         # overwrite type
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -6,6 +6,7 @@ use crate::{
         out_parameters::use_function_return_for_result,
         rust_type::{bounds_rust_type, rust_type, rust_type_with_scope},
     },
+    config,
     consts::TYPE_PARAMETERS_START,
     env::Env,
     library::{
@@ -81,6 +82,7 @@ impl Bounds {
         par: &CParameter,
         r#async: bool,
         concurrency: Concurrency,
+        configured_functions: &[&config::functions::Function],
     ) -> (Option<String>, Option<CallbackInfo>) {
         let type_name = bounds_rust_type(env, par.typ);
         if (r#async && async_param_to_remove(&par.name)) || type_name.is_err() {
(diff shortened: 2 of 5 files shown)


sdroege:
Can you show an example for such function?

de-vri-es:
One example is [`arv_camera_gv_get_packet_size()`](https://github.com/AravisProject/aravis/blob/fff18fb9763125d0460bd54e62af77e9333b5260/src/arvcamera.c#L2463) from the Aravis project. There are many more in Aravis.

There is also atleast one in glib itself: [`g_key_file_has_key`](https://developer.gnome.org/glib/2.62/glib-Key-value-file-parser.html#g-key-file-has-key) . While the error reporting part of the documentation says it is deprecated, the documentation of that function itself doesn't mention that.

sdroege:
This is not really GObject-Introspection compatible API unfortunately. If your function takes a `GError` then there needs to be a return value that signifies failure, and any other "return values" would have to be returned via an out parameter.

This probably also does not work in various other bindings without manual intervention.

Can you show some generated code for this btw?

de-vri-es:
You don't need the return value to detect an error, you can simply look if the error pointer has been set. In fact, the generated code currently ignores the return value if it is `boolean` or `guint` (without this PR). Other return types already work without this patch. 

You can see an example here of code generated with this PR: https://github.com/robohouse-delft/aravis-rs/blob/c310df778d75cec90b83bafcd56b535094fd6ce8/aravis/src/auto/camera.rs#L1277-L1290

sdroege:
Yeah looks good to me (the generated code), what do you think @EPashkin ? :)

I think it would be good to merge this feature, but you'll probably also have problems with this API in other bindings.

de-vri-es:
I agree. Ideally, the GIR annotations would convey this information. I may try to fly that idea with `glib`, but even if they agree it will take a while before that is really usable.

Note that I couldn't find anything on treating `guint` as error indicator. Only about booleans. From https://developer.gnome.org/glib/2.62/glib-Error-Reporting.html#gerror-rules:
> By convention, if you return a boolean value indicating success then TRUE means success and FALSE means failure. Avoid creating functions which have a boolean return value and a GError parameter, but where the boolean does something other than signal whether the GError is set. Among other problems, it requires C callers to allocate a temporary error. Instead, provide a "gboolean *" out parameter. There are functions in GLib itself such as g_key_file_has_key() that are deprecated because of this. If FALSE is returned, the error must be set to a non-NULL value. One exception to this is that in situations that are already considered to be undefined behaviour (such as when a g_return_val_if_fail() check fails), the error need not be set. Instead of checking separately whether the error is set, callers should ensure that they do not provoke undefined behaviour, then assume that the error will be set on failure.

It doesn't mention `guint` anywhere. So I was somewhat surprised that the int values were also thrown away, not just the boolean values. Is this some undocumented de-facto standard, or a bug in `gtk-rs/gir` (or did I miss something)?

EPashkin:
@sdroege There no changes in gtk-rs crates as this flag was not added to config.
@de-vri-es Thanks

de-vri-es:
Thank you both for the quick reviews :)

EPashkin:
Functions with `u32`
`gtk_builder_add_from_resource` and other function in `Builder`, seems return value has no meaning for us.
`g_credentials_get_unix_user` in gio - this IMHO need fix by added flag.
`gtk_source_search_context_replace_all` also need this fix.

EPashkin:
Added PR for sourceview,
for gio IMHO better combine with https://github.com/gtk-rs/gtk/issues/973#issuecomment-598373364

sdroege:
@EPashkin can you create issues or PRs for this so we don't forget? Thanks :)

EPashkin:
Added PR for gio too, without updating prelude for subclasses

PR #893, opened 2020-04-02
Check gir file

GuillaumeGomez:
The first commit is in order to provide a way to check gir files for GNOME so they don't have to wait for us to report bugs. Once this PR is merged, I'll create a Dockerfile and send it to https://gitlab.gnome.org/GNOME/gtk/tree/master/.gitlab-ci/

The second commit is "funnier": when testing the option, I decided to use the dl.sh script to get the latest (and buggy) gir files for gtk4. And when I ran it (expecting a failure), I got:

```
Can't open file "../gir-files/Gsk-4.0.gir": Too many open files (os error 24)
```

After a little search, and a nice run using strace:

```bash
strace -e trace=open,close,read,write,connect,accept ./target/release/gir --check-gir-file ../gir-files/Gtk-4.0.gir
```

I discovered that gir was opening thousands of file. At this point, it seemed pretty obvious that the problem was a circular dependency. Therefore, I added a small check in order to prevent it to happen.

cc @EPashkin @sdroege

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -118,7 +118,7 @@ fn run_check(check_gir_file: &str) -> Result<(), String> {
         Some(p) => p,
         None => return Err(format!("Failed to get parent directory from `{}`", check_gir_file)),
     };
-    return library.read_file(&parent, &lib_name);
+    return library.read_file(&parent, &mut vec![lib_name.to_owned()]);
 }
 
 fn do_main() -> Result<(), String> {
@@ -147,7 +147,7 @@ fn do_main() -> Result<(), String> {
         let _watcher = statistics.enter("Loading");
 
         library = Library::new(&cfg.library_name);
-        library.read_file(&cfg.girs_dir, &cfg.library_full_name())?;
+        library.read_file(&cfg.girs_dir, &mut vec![cfg.library_full_name()])?;
     }
 
     {
--- a/src/main.rs
+++ b/src/main.rs
@@ -26,7 +26,12 @@ impl<S: AsRef<str>> OptionStr for Option<S> {
     }
 }
 
-fn build_config() -> Result<(Option<Config>, Option<String>), String> {
+enum RunKind {
+    Config(Config),
+    CheckGirFile(String),
+}
+
+fn build_config() -> Result<RunKind, String> {
     let args: Vec<_> = env::args().collect();
     let program = args[0].clone();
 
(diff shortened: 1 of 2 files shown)


src/parser.rs
```diff
@@ -17,17 +17,17 @@ pub fn is_empty_c_type(c_type: &str) -> bool {
 }
 
 impl Library {
-    pub fn read_file(&mut self, dir: &Path, lib: &str) -> Result<(), String> {
-        let file_name = make_file_name(dir, lib);
+    pub fn read_file(&mut self, dir: &Path, libs: &mut Vec<String>) -> Result<(), String> {
+        let file_name = make_file_name(dir, &libs[libs.len() - 1]);
```
EPashkin, line 21: Maybe add `assert!(!libs.is_empty())`?` before this?

EPashkin: Or check and return error.

src/main.rs
```diff
@@ -25,7 +26,7 @@ impl<S: AsRef<str>> OptionStr for Option<S> {
     }
 }
 
-fn build_config() -> Result<Config, String> {
+fn build_config() -> Result<(Option<Config>, Option<String>), String> {
```
EPashkin, line 29: Maybe better use local enum { Config(Config), CheckGirFile(String) }  ?

src/parser.rs
GuillaumeGomez: It cannot be empty since the call stack starts with `read_file` with one element, the only other place where we push/pop is the read_repository function. We add one element, call `read_file` then once it has returned, it pops what it just added. The array always has at least one element.

src/main.rs
GuillaumeGomez: Great idea!

GuillaumeGomez:
Updated!

EPashkin:
Thanks, Look good for me
Waiting for @sdroege just in case

sdroege:
@GuillaumeGomez What is the cycle with gsk-4.0? What are all the gir files involved? This should be reported to the GTK people

sdroege:
Looks good to me otherwise. There are two potential problems though:
1. Some things we only check at later stages, like things related to async functions
2. If we advertise this as a "validate gir files" thingy, we need to take PRs from others that make the parser more correct. Currently we are more strict than needed in a few places, I can't remember details though. We'll find out.

GuillaumeGomez:
The cycle was `gtk4 -> gsk4 -> gtk4`. The goal here is mostly to have a start, we can extend this check later on.

sdroege:
> The cycle was `gtk4 -> gsk4 -> gtk4`.

Did you report that already? Please provide link to the issue so I can subscribe to it :)

GuillaumeGomez:
I didn't. Since I'll update enforce it into their CI, I just assumed they'll be forced to fix it very soon. I just need this PR to get merged first. :3

cc @EPashkin ;)

GuillaumeGomez:
Thanks!

PR #894, opened 2020-04-03
git: check return value from 'git' command

gdesmott:
Command::output() succeeds as long as the command was actually called.
It's up to the user to check if the command was properly executed or
not.

Fix returning Some("") instead of None when calling repo_hash() with an
invalid git repo path.

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -8,7 +8,11 @@ pub fn repo_hash<P: AsRef<Path>>(path: P) -> Option<String> {
         None => vec![],
     };
     args.extend(&["rev-parse", "--short", "HEAD"]);
-    let hash = String::from_utf8(Command::new("git").args(&args).output().ok()?.stdout).ok()?;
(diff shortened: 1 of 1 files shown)


src/git.rs
```diff
@@ -8,7 +8,11 @@ pub fn repo_hash<P: AsRef<Path>>(path: P) -> Option<String> {
         None => vec![],
     };
     args.extend(&["rev-parse", "--short", "HEAD"]);
-    let hash = String::from_utf8(Command::new("git").args(&args).output().ok()?.stdout).ok()?;
+    let output = Command::new("git").args(&args).output().ok()?;
+    if !&output.status.success() {
```
EPashkin, line 12: & really needed here?

EPashkin:
Thanks for fix.

src/git.rs
gdesmott: nope, typo :)

PR #896, opened 2020-04-09
Remove all transmute calls in generated code

GuillaumeGomez:
Regen sample: https://github.com/gtk-rs/gtk/pull/983

cc @EPashkin @sdroege

### DIFF
--- a/src/analysis/properties.rs
+++ b/src/analysis/properties.rs
@@ -293,7 +293,6 @@ fn analyze_property(
             }
             imports.add("glib::signal::connect_raw");
             imports.add("glib::signal::SignalHandlerId");
-            imports.add("std::mem::transmute");
             imports.add("std::boxed::Box as Box_");
             imports.add("glib_sys");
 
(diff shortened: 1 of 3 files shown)


sdroege:
:+1:

EPashkin:
@GuillaumeGomez Thanks.

GuillaumeGomez:
Time to regen everything then!

sdroege:
This seems to be wrong: https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/451/diffs#note_462758

sdroege:
I think this happens here because there are more casts than needed and it happens to cast something wrong. We should probably replace the second `_` with `gobject_sys::GCallback` in the generated code.

GuillaumeGomez:
Please open an issue so we can take a look.

sdroege:
See https://github.com/gtk-rs/gir/issues/901

PR #900, opened 2020-04-12
Simplify chunk::Assert* and fix missing init check for async_future methods

### LINKED ISSUE #898: Some init asserts are missing in the generated code
As showed [here](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/424#note_462625), some asserts are not generated whereas they should.

GuillaumeGomez:
Fixes #898 

cc @sdroege @EPashkin

### DIFF
--- a/src/analysis/safety_assertion_mode.rs
+++ b/src/analysis/safety_assertion_mode.rs
@@ -37,4 +37,11 @@ impl SafetyAssertionMode {
 
         InMainThread
     }
+
+    pub fn is_none(&self) -> bool {
+        match *self {
+            SafetyAssertionMode::None => true,
+            _ => false,
+        }
+    }
 }
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -63,6 +63,7 @@ pub struct AsyncFuture {
     pub name: String,
     pub success_parameters: String,
     pub error_parameters: String,
+    pub assertion: SafetyAssertionMode,
 }
 
 #[derive(Debug)]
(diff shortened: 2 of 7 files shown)


src/analysis/safety_assertion_mode.rs
```diff
@@ -13,6 +14,16 @@ impl Default for SafetyAssertionMode {
     }
 }
 
+impl fmt::Display for SafetyAssertionMode {
```
EPashkin, line 17: @sdroege Use Display for code snippets is acceptable for you?

sdroege:
@GuillaumeGomez This is also for async functions, not only for async_future functions or not?

src/analysis/safety_assertion_mode.rs
sdroege: No, but your decision :)

EPashkin: @GuillaumeGomez  can you move this match as function to writer?
IMHO just to "mod.rs" will do.

GuillaumeGomez: Ok!

GuillaumeGomez:
The async functions work as expected.

sdroege:
> The async functions work as expected.

The futures functions don't need assertions because they just call the async functions anyway.

GuillaumeGomez:
But the script doesn't care about it. Also, I'm not sure if it'd be a very nice idea to not check functions ending with `_future`. But your call.

sdroege:
> But the script doesn't care about it. Also, I'm not sure if it'd be a very nice idea to not check functions ending with `_future`. But your call.

Could add the `skip` version of the macro to these functions

GuillaumeGomez:
Fine by me!

GuillaumeGomez:
Updated!

EPashkin:
@GuillaumeGomez Thanks, waiting CI

Issue #901, opened 2020-04-13
Signal handler trampoline casts can cause crashes


sdroege:
See https://github.com/gtk-rs/gir/pull/896#issuecomment-612642197 and following.

Unclear what exactly happens there. In this specific code, no cast at all is needed. My guess is that the two wildcard casts are happening to convert it to something else in the middle than what is needed.

sdroege:
Another case (`gstreamer-app/src/app_sink.rs`):

```rust
            connect_raw(
                self.as_ptr() as *mut _,
                b"new-sample\0".as_ptr() as *const _,
                Some(*(&new_sample_trampoline::<F> as *const _ as *const _)),
                Box_::into_raw(f),
            )
```

For testing purposes I've added a lot of debug output:

```rust
        unsafe {
            let f: Box_<F> = Box_::new(f);
            let t = new_sample_trampoline::<F> as *const unsafe extern "C" fn(*mut gst_app_sys::GstAppSink, *mut libc::c_void) -> i32;
            dbg!(t);
            let t = t as *const unsafe extern "C" fn();
            dbg!(t);
            dbg!(*t);
            let t = new_sample_trampoline::<F> as unsafe extern "C" fn(*mut gst_app_sys::GstAppSink, *mut libc::c_void) -> i32;
            dbg!(t);
            let mut t = std::mem::transmute(t);
            dbg!(t);
            let u = &new_sample_trampoline::<F> as *const _ as *const _;
            dbg!(u);
            t = *u;
            dbg!(t);
            connect_raw(
                self.as_ptr() as *mut _,
                b"new-sample\0".as_ptr() as *const _,
                Some(t),
                Box_::into_raw(f),
            )
        }
```

This prints

```
[/home/slomo/Projects/rust/gstreamer-rs/gstreamer-app/src/app_sink.rs:295] t = 0x00005603710fe920
[/home/slomo/Projects/rust/gstreamer-rs/gstreamer-app/src/app_sink.rs:297] t = 0x00005603710fe920
[/home/slomo/Projects/rust/gstreamer-rs/gstreamer-app/src/app_sink.rs:298] *t = 0x247c894858ec8348
[/home/slomo/Projects/rust/gstreamer-rs/gstreamer-app/src/app_sink.rs:300] t = 0x00005603710fe920
[/home/slomo/Projects/rust/gstreamer-rs/gstreamer-app/src/app_sink.rs:302] t = 0x00005603710fe920
[...]

PR #902, opened 2020-04-13
Revert "Remove all transmute calls in generated code"

EPashkin:
Part of https://github.com/gtk-rs/gir/issues/901

This reverts commit 5ca26d9e8400e8fa904f23c9dc8597241a7ba1d0.

### DIFF
--- a/src/writer/to_code.rs
+++ b/src/writer/to_code.rs
@@ -137,7 +137,7 @@ impl ToCode for Chunk {
                 );
                 let self_str = if in_trait { "Self, " } else { "" };
                 let s2 = format!(
-                    "\tSome(transmute({}::<{}F> as usize)), Box_::into_raw(f))",
+                    "\tSome(transmute::<_, unsafe extern \"C\" fn()>({}::<{}F> as *const ())), Box_::into_raw(f))",
                     trampoline, self_str
                 );
                 vec![s1, s2]
--- a/src/writer/to_code.rs
+++ b/src/writer/to_code.rs
@@ -137,7 +137,7 @@ impl ToCode for Chunk {
                 );
                 let self_str = if in_trait { "Self, " } else { "" };
                 let s2 = format!(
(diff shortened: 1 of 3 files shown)


src/writer/to_code.rs
```diff
@@ -137,7 +137,7 @@ impl ToCode for Chunk {
                 );
                 let self_str = if in_trait { "Self, " } else { "" };
                 let s2 = format!(
-                    "\tSome(*(&{}::<{}F> as *const _ as *const _)), Box_::into_raw(f))",
+                    "\tSome(transmute({}::<{}F> as usize)), Box_::into_raw(f))",
```
sdroege, line 140: For completeness, can we do

```suggestion
                    "\tSome(transmute::<_, unsafe extern "C" fn()>::({}::<{}F> as *const ())), Box_::into_raw(f))",
```

That's less likely to fall apart and closer to the unsafe code guidelines.

EPashkin: Ok, but without second `::`

sdroege: Indeed, updated comment

EPashkin:
```
Some(transmute::<_, unsafe extern "C" fn()>(
                    activate_trampoline::<Self, F> as *const (),
                )),
```

sdroege: Looks correct

EPashkin: Only problem of this change that we need change manual code again instead just reverting commit :(

sdroege: Need to do that anyway, there were some other changes in the manual code that should be kept

sdroege: @EPashkin Can you update?

EPashkin: Oops, forgot to push.
Done

sdroege:
Thanks, let's get this in then and then fix up everything else :)

PR #903, opened 2020-04-13
Revert "Remove all transmute calls in generated code"

GuillaumeGomez:
Unfortunately, it wasn't a successful experiment... A solution is still possible without `transmute`:

```rust
fn foo() {
    println!("bla");
}

fn bar(i: u32) {
    println!("bar");
}

fn main() {
    unsafe {
        let b = bar as fn(u32);
        let b = &b;
        let b = b as *const fn(u32) as *const fn();
        (*b)();
    }
}
```

But `transmute` seems to be the simplest solution. So let's go back to it.

Reverts gtk-rs/gir#896

cc @EPashkin @sdroege

### DIFF
--- a/src/analysis/properties.rs
+++ b/src/analysis/properties.rs
@@ -293,6 +293,7 @@ fn analyze_property(
             }
             imports.add("glib::signal::connect_raw");
             imports.add("glib::signal::SignalHandlerId");
+            imports.add("std::mem::transmute");
             imports.add("std::boxed::Box as Box_");
             imports.add("glib_sys");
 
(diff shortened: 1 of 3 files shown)


sdroege:
See #902

EPashkin:
@GuillaumeGomez I close this as my PR merged

GuillaumeGomez:
Oh my bad, didn't see it. Thanks for doing it! I'll copy/paste my initial comment into yours.

EPashkin:
Better copy to issue

PR #904, opened 2020-04-15
Add a check to avoid unforeseen failures

GuillaumeGomez:
cc @EPashkin @sdroege


GuillaumeGomez:
CI passed. Anyone? :)

sdroege:
Sure, why not. Adding more checks to existing jobs and piling stuff up instead of creating separate jobs feels wrong to me, but that's apparently how travis is supposed to be used.

GuillaumeGomez:
Merging then!

PR #907, opened 2020-04-22
Show more not generated types

GuillaumeGomez:
This is a bit tricky since we currently filter out based on names. However, enums can have names ending with "Type" for example, filtering them out. I wonder if classes or other types may be filtered out for the same reason? I couldn't find one but I think it's a possibility.

cc @EPashkin @sdroege

### DIFF
--- a/src/library.rs
+++ b/src/library.rs
@@ -837,6 +837,20 @@ impl Type {
             _ => true,
         }
     }
+
+    pub fn is_enumeration(&self) -> bool {
+        match *self {
+            Type::Enumeration(_) => true,
+            _ => false,
+        }
+    }
+
+    pub fn is_bitfield(&self) -> bool {
+        match *self {
+            Type::Bitfield(_) => true,
+            _ => false,
+        }
+    }
 }
 
 macro_rules! impl_maybe_ref {
@@ -996,6 +1010,7 @@ impl Library {
             "Accessible",
             "Iface",
             "Type",
+            "Interface",
         ];
         let namespace_name = self.namespaces[MAIN_NAMESPACE as usize].name.clone();
         let mut parents = HashSet::new();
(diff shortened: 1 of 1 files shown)


src/library.rs
```diff
@@ -1006,7 +1013,7 @@ impl Library {
                 let full_name = format!("{}.{}", namespace_name, name);
```
sdroege, line 1013: The list of names above should probably also contain `Interface`. But for what reason is `Type` included there?

src/library.rs
```diff
@@ -1006,7 +1013,7 @@ impl Library {
...
                 let mut check_methods = true;
 
-                if !not_allowed_ending.iter().any(|s| name.ends_with(s)) {
+                if !not_allowed_ending.iter().any(|s| name.ends_with(s)) || x.is_enumeration() {
```
sdroege, line 1016: Also flags I guess?

GuillaumeGomez: flags?

sdroege: Aka `Bitfield`, the thing we use the `bitflags` crate for

GuillaumeGomez: Oh I see! Added it. :)

src/library.rs
GuillaumeGomez: I don't know actually. On `atk`, all "*Type" are enumerations at least. Maybe it's not the case on other crates?

EPashkin: Agree about `Interface`

GuillaumeGomez: I added it.

EPashkin:
@GuillaumeGomez Thanks. Can you make manual squash and merge yourself?

src/library.rs
sdroege: So should we remove `Type` here?

sdroege: Yes? If not, why not? :)

GuillaumeGomez:
Sure!

PR #911, opened 2020-05-01
Generate duplicated enum values in the -sys crate

sdroege:
We generate them only as constants so overlaps are not really a problem
and like this we a) represent the C header more accurately and b) don't
randomly skip values based on their order. Duplicated values could be
aliases or old names for the same thing.

### DIFF
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -7,11 +7,9 @@ use crate::{
     library::*,
     nameutil::*,
     traits::*,
-    version::Version,
 };
 use log::info;
 use std::{
-    collections::HashMap,
     fs,
     io::{Result, Write},
 };
@@ -286,7 +284,6 @@ fn generate_enums(w: &mut dyn Write, env: &Env, items: &[&Enumeration]) -> Resul
         if let Some(false) = config.map(|c| c.status.need_generate()) {
             continue;
         }
-        let mut vals: HashMap<String, (String, Option<Version>)> = HashMap::new();
         writeln!(w, "pub type {} = c_int;", item.c_type)?;
         for member in &item.members {
             let member_config = config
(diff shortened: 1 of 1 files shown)


sdroege:
Proper fix for https://gitlab.freedesktop.org/gstreamer/gstreamer-rs-sys/-/commit/efe335fe2966a3a077de5d7d1d1f9a56056ac395 , which disappeared because a new member with the same value was added.

sdroege:
CC @EPashkin @GuillaumeGomez

GuillaumeGomez:
:+1:

EPashkin:
I thought that we will get problems with gdk and its "2BUTTON", but sys seems normal as in `GTK_2BUTTON" here.
But IMHO then we need remove "alias" from sys completely or make it act as ignore Ex. for `GDK_EVENT_LAST`,
but it still needed for normal mode.

sdroege:
I don't understand what your suggestion is. What should we do? :)

EPashkin:
Or alias completely from sys (and fix all other warnings).
Or add/restore `if alias { continue }`

sdroege:
What exactly is alias, how are they marked in the .gir file?

EPashkin:
https://github.com/gtk-rs/sys/blob/master/conf/gir-gdk.toml#L22

EPashkin:
It was analog `ignore` for enums, may be currently we possibly need just add `ignore` for enum and remove this, I not sure.
Anyway in this PR IMHO restore "alias" is enough.

sdroege:
I agree, let's add back alias then and in another commit I rename it to ignore?

EPashkin:
Better only restore alias. Renaming need more work (as it affect normal crate too) and better done in separate PR.

sdroege:
@EPashkin Done :)

PR #915, opened 2020-05-05
Add options.split_build_rs

EPashkin:
Part of https://github.com/gtk-rs/sys/pull/167#issuecomment-623858337

cc @sdroege , @GuillaumeGomez , @kinnison

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,11 @@ target_path = "."
 #Path where lib.rs generated (defaults to <target_path>/src)
 #auto_path = "src"
 work_mode = "sys"
+#If true then build.rs will be splitted into 2 parts:
+# always generated build_version.rs,
+# and build.rs that generated only if not exists.
+# Defaults to false
+split_build_rs = false
 ```
 
 You can mark some functions that has suffix `_utf8` on Windows:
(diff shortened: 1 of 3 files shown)


kinnison:
Can you show an example of how this would look in use?  I'm not as familiar with the Gir codebase.

EPashkin:
https://github.com/gtk-rs/sys/pull/168

sdroege:
This is great but can we also have it the other way around? :)

In `gstreamer-sys` I'm currently editing the generated `build.rs` for the version part specifically to add missing versions for making user's lives a bit easier: if no new API was added between GStreamer 1.14 and 1.16 in one library there would still be a `v1_16` feature so that users can simply select `v1_16` for all libraries without having to check for each of them if the feature actually exists. Or maybe for that I instead want a configuration option `add_empty_major_versions` to automate it?

EPashkin:
@sdroege Yes IMHO you want `add_empty_major_versions` as currently we just check all item in namespace to get versions and your mode need change of this process.

EPashkin:
Maybe it better be `include_versions_by_step="0.2"`

EPashkin:
@sdroege if you want, I can add a `include_version_by_step` today.

EPashkin:
.. or `add_empty_major_versions` with same functionality, but it will be buildin

EPashkin:
.. only problem, how to detect max version if its not present in .gir

sdroege:
That would have to be configured too, I guess. Would be great if you could add that :)

EPashkin:
Added issue https://github.com/gtk-rs/gir/issues/916

PR #917, opened 2020-05-05
Add static lifetime to generated build_version.rs

EPashkin:
Fix errors in #915

PR #918, opened 2020-05-05
tests: use split_build_rs in gdk gir test

kinnison:
The reason #917 was needed (that #915 passed CI) was that the new feature @EPashkin added wasn't actually tested by any of the tests.  This patch causes `gdk-sys` in the tests to require the split build option.  This way bugs in this code in the future should be caught.

### DIFF
--- a/tests/sys/gir-gdk.toml
+++ b/tests/sys/gir-gdk.toml
@@ -5,6 +5,7 @@ library = "Gdk"
 version = "3.0"
 min_cfg_version = "3.4"
 target_path = "./gdk-sys"
+split_build_rs = true
 external_libraries = [
    "GLib",
    "GObject",


kinnison:
Aah I missed that the gtk tests also failed, give me a moment.

kinnison:
@EPashkin The failures here were due to gdk-sys having the bad version run and merged.  If you fix gdk-sys then this test would pass.

EPashkin:
After fixing can you make final test by removing `'static` again?

kinnison:
I tested locally by removing `'static` and confirming that `tests/sys/test.sh` fails - is that enough or do you want me to try and come up with a test which can stay in the code?

EPashkin:
> The failures here were due to gdk-sys having the bad version run and merged. If you fix gdk-sys then this test would pass.

Then what test this steps if it not depend of current gir or sys changed by it?

kinnison:
It tests that *GTK*'s gir when regenerated still builds.  GTK depends on gdk-sys which had merged a badly generated build script which is not regenerated by this test.

EPashkin:
You right.

EPashkin:
Thanks.
Last nagging: by current log "sys_build (Windows)" doing same as end of tests/sys/test.sh, maybe better remove it?

kinnison:
The tests there are based on the behaviour of the old travis and appveyor -- appveyor also built explicitly with v3_20 whereas travis didn't.

EPashkin:
Sorry, it not same, it tests big version too.
So I restart CI here and merge when it pass

EPashkin:
@kinnison Thanks for help with CI, github's is new for me.

kinnison:
You're very welcome.  It's a bit odd initially but I quite like it once you get your head around it.  It's wordier than travis or appveyor, but more flexible and clever.

Issue #661, opened 2018-11-15
Use quote crate instead of string interpolation to generate code


meh:

sdroege:
@meh what's the status of this? :)

sdroege:
@meh any updates?

bilelmoussaoui:
As per #919, this doesn't sound like a viable approach. Maybe the issue should be closed?

sdroege:
IIRC we use quote in a few places. Let's just do that where it's useful.

PR #919, opened 2020-05-05
WIP: Use quote crate for generating sys bindings

sdroege:
CC @EPashkin @GuillaumeGomez @gdesmott 

This is only an experiment so far. It generates the enums correctly but I'm not sure how to
* Add newlines in the output of the quote crate (between each enum we had an empty line before)
* Output comments (the `// Enums` at the top)

I also didn't check if the version condition part works, it probably needs some more refactoring.

Ideally I'd like to move the sys generation fully over the quote, and then we can check how to make use of that for the safe bindings code generation.

Opinions?

### DIFF
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -275,17 +275,23 @@ fn generate_constants(w: &mut dyn Write, env: &Env, constants: &[Constant]) -> R
 }
 
 fn generate_enums(w: &mut dyn Write, env: &Env, items: &[&Enumeration]) -> Result<()> {
-    if !items.is_empty() {
-        writeln!(w, "// Enums")?;
+    use proc_macro2::{Ident, Literal, Span};
+    use quote::quote;
+
+    if items.is_empty() {
+        return Ok(());
     }
-    for item in items {
+
+    let enums = items.iter().map(|item| {
         let full_name = format!("{}.{}", env.namespaces.main().name, item.name);
         let config = env.config.objects.get(&full_name);
         if let Some(false) = config.map(|c| c.status.need_generate()) {
-            continue;
+            return quote! {};
         }
-        writeln!(w, "pub type {} = c_int;", item.c_type)?;
-        for member in &item.members {
+
+        let c_type = Ident::new(&item.c_type, Span::call_site());
+
+        let members = item.members.iter().map(|member| {
             let member_config = config
                 .as_ref()
                 .map(|c| c.members.matched(&member.name))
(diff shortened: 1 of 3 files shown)


src/codegen/sys/lib_.rs
```diff
@@ -294,18 +300,40 @@ fn generate_enums(w: &mut dyn Write, env: &Env, items: &[&Enumeration]) -> Resul
...
-                w,
-                "pub const {}: {} = {};",
-                member.c_identifier, item.c_type, member.value,
-            )?;
+            let c_identifier = Ident::new(&member.c_identifier, Span::call_site());
+            let value = Literal::i32_unsuffixed(member.value.parse().unwrap());
+
+            let version_condition = match version {
+                Some(v) if v > env.config.min_cfg_version => {
+                    let v = v.to_cfg();
```
EPashkin, line 311: Need different function, Ex. to_cfg_tokens(), as string not works in this case
```diff
-#[cfg(any(feature = "v3_18", feature = "dox"))]
+#[cfg(any("feature = \"v3_18\"", feature = "dox"))]
 pub const GDK_TOUCHPAD_SWIPE: GdkEventType = 41;
-#[cfg(any(feature = "v3_18", feature = "dox"))]
+#[cfg(any("feature = \"v3_18\"", feature = "dox"))]
 pub const GDK_TOUCHPAD_PINCH: GdkEventType = 42;
```

EPashkin:
Part of https://github.com/gtk-rs/gir/issues/661

EPashkin:
I also have no idea how to add empty line or comments :(

EPashkin:
Found solution:
```rust
    let c1 = Punct::new('/', Spacing::Joint);
    let c2 = Punct::new('/', Spacing::Alone);
    let nl = Punct::new('\n', Spacing::Alone);
...
        quote! {
            pub type #c_type = c_int;
            #(#members)*
            #nl#nl
        }
...
    let tokens = quote! {
        #c1#c2 Enums #nl
        #(#enums)*
    };

    writeln!(w, "{}", tokens)?;
```

src/codegen/sys/lib_.rs
sdroege: Ah the string escaping :) Will take care of it once @GuillaumeGomez also agreed that this is the way forward.

In the end each of the functions would return a `TokenStream` and we would only write things out in the very end btw, instead of doing it in each function.

sdroege:
The newline/comments part is not very beautiful, but that will do for now :) We can move that into helper functions to make it less ugly. Thanks for finding that!

src/codegen/sys/lib_.rs
```diff
@@ -294,18 +300,40 @@ fn generate_enums(w: &mut dyn Write, env: &Env, items: &[&Enumeration]) -> Resul
...
             }
 
-            version_condition(w, env, version, false, 0)?;
-            writeln!(
-                w,
-                "pub const {}: {} = {};",
-                member.c_identifier, item.c_type, member.value,
-            )?;
+            let c_identifier = Ident::new(&member.c_identifier, Span::call_site());
+            let value = Literal::i32_unsuffixed(member.value.parse().unwrap());
```
EPashkin, line 307: IMHO we get problems with `i32`

EPashkin:
For the record, I’m not sure if use `quote` for generation will give us any benefit.

src/codegen/sys/lib_.rs
sdroege: Why / when?

sdroege:
> For the record, I’m not sure if use `quote` for generation will give us any benefit.

For this specific part of the code not so much, but for example for the futures code in the non-sys code generation, the signal connection/handler code, etc. it would make it considerably easier to read and maintain compared to the existing code generator

src/codegen/sys/lib_.rs
EPashkin: I not remember where, but there was constant that not fill into i32, only in u32.
But it really not problem as we declare type as c_int.
Sorry, my mistake

EPashkin:
I hope too.

sdroege:
@GuillaumeGomez What's your opinion?

GuillaumeGomez:
I find the code a bit harder to read but overall, I think it'll improve it.

sdroege:
While writing the code is much more convenient, the generated code looks quite awful unless whitespace is inserted manually everywhere... when trying to use the `quote` crate in another code generator I'm writing this made it not a good choice.

There's no interest in fixing/improving this in the `quote` crate (see https://github.com/dtolnay/quote/issues/154), so let's do something different then...

PR #920, opened 2020-05-05
Add options.extra_versions and lib_version_overrides configurations

### LINKED ISSUE #916: Support adding versions in sys mode
https://github.com/gtk-rs/gir/pull/915#issuecomment-623887761
> In gstreamer-sys I'm currently editing the generated build.rs for the version part specifically to add missing versions for making user's lives a bit easier: if no new API was added between GStreamer 1.14 and 1.16 in one library there would still be a v1_16 feature so that users can simply select v1_16 for all libraries without having to check for each of them if the feature actually exists. Or maybe for that I instead want a configuration option add_empty_major_versions to automate it?

or https://github.com/gtk-rs/gir/pull/915#issuecomment-623890033
> Maybe it better be include_versions_by_step="0.2"

in both cases we seems have problem if part not have elements with maximum version.

cc @sdroege , @GuillaumeGomez

EPashkin:
Fix #916 

@sdroege Please test this.

I also found that previous version lost "v1_36_7" in pango and "v2_44" in glib in build.rs, and only added it to Cargo.toml

cc @GuillaumeGomez

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -17,15 +17,6 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         toolchain: [stable]
-        include:
-          - os: windows-latest
-            bits: 32
-            arch: i686
-            toolchain: stable-i686-pc-windows-gnu
-          - os: windows-latest
-            bits: 64
-            arch: x86_64
-            toolchain: stable-x86_64-pc-windows-gnu
 
     defaults:
       run:
@@ -42,15 +33,6 @@ jobs:
           profile: minimal
           components: rustfmt, clippy
         id: toolchain
-      - name: Install MSYS2
-        uses: numworks/setup-msys2@v1
-        if: matrix.os == 'windows-latest'
-      - name: Set up MSYS2 path and other windows env
-        run: |
-          echo "::add-path::$RUNNER_TEMP\\msys\\msys64\\usr\\bin"
-          echo "::add-path::$RUNNER_TEMP\\msys\\msys64\\mingw${{ matrix.bits }}\\bin"
-          echo "::set-env name=LIBGIT2_SYS_USE_PKG_CONFIG::1"
-        if: matrix.os == 'windows-latest'
       - name: Cache cargo registry
         uses: actions/cache@v1
         with:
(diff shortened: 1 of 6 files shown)


sdroege:
That works for me, now there's only one diff left: `v1_12 = ["v1_10", "gstreamer-sys/v1_12"]` is what I need, for example, but `v1_12 = ["v1_10"]` is what I get. For one of the dependencies at least the version needs to be passed through.

Any ideas how to make that last part  better? Then I won't have any manual changes in the sys crates anymore.

EPashkin:
This is tough.
Sys crate really needed this dependence?
In gtk all this dependences seems moved to normal crates.

EPashkin:
We can add 
```
feature_dependencies=[
"gstreamer-sys",
]
```
and pass same versions or 
```
[[feature_dependencies]]
version = "1.12"
dependency = "gstreamer-sys"
dependency_version= "1.12"
```
maybe with combining last two params

sdroege:
```
[[feature_dependencies]]
version = "1.12"
dependency = "gstreamer-sys/v1_12"
```

Seems good, especially if that can be given multiple times (so you can depend on `gstreamer-sys/v1_12` and `gstreamer-base-sys/v1_12`, for example).

And for simplification `feature_dependencies_with_same_version` like what you suggested.

EPashkin:
@sdroege You have cases with different versions?
IMHO having both mode will make code to complex.

If we threat `feature_dependencies.dependency` as simply string, then we can write "gstreamer-sys/v1_12, gstreamer-base-sys/v1_12" to it.

sdroege:
> @sdroege You have cases with different versions?

I don't, but I could imagine such cases to happen. For example it could happen that GTK 3.30 uses a type from GLib 2.66 in one of the function signatures.

> If we threat `feature_dependencies.dependency` as simply string, then we can write "gstreamer-sys/v1_12, gstreamer-base-sys/v1_12" to it

Having only that (and not the same version thing) would also work for me, just requires some more configuration but that's acceptable :)

EPashkin:
@sdroege Added `feature_dependencies`

sdroege:
Thanks, I'll try that later. From the documentation this looks exactly like what I need :)

Issue #839, opened 2019-09-13
Problem with generating enums/flags where the `get_type()` function was added later


sdroege:
E.g. https://gitlab.freedesktop.org/gstreamer/gstreamer-rs-sys/blob/master/gstreamer-video-sys/src/lib.rs#L2332

This `get_type()` function was added in 1.16, the corresponding flags type exists since much longer.

There are two problems here
a) In GStreamer this function needs to be marked as `Since: 1.16`
b) We need to generate the `#[cfg()]` thing for it in the -sys crate
c) We need to figure out how to handle that conditional compilation in the non-sys crate

Thoughts?

EPashkin:
"a" and "b" IMHO fixing by changing GstVideo-1.0.gir,
for "c" IMHO if there any value that custom `gtk_align_get_type` can return without broke all,
then simpler just add manual function to sys crate with inverse condition,
so no changes will needed in non-syses

EPashkin:
Alternative also needed fixes for "a" and "b" anyway.
We can compare versions for "get_type" and "enum" and if get_type added later not generate "StaticType"  and `"*Value"*` traits for enum.

sdroege:
> "a" and "b" IMHO fixing by changing GstVideo-1.0.gir,

Probably, but would we even generate the b) for the -sys crate?

> for "c" IMHO if there any value that custom `gtk_align_get_type` can return without broke all,
> then simpler just add manual function to sys crate with inverse condition,
> so no changes will needed in non-syses

We can't implement our own `get_type` function. We'd have to put the whole `StaticType` and GValuer traits impls into `#[cfg]`

EPashkin:
About "b": IMHO get_type is normal function so "[cfg]" will be generated with right .gir file.

You right that instead not generating we can add "[cfg]" to trait "StaticType" etc.

"Breaking change" alternative just increase enum version to match get_type's.

EPashkin:
About "b" if you meant "inverse function" then seems we can't generate it 😢

sdroege:
> "Breaking change" alternative just increase enum version to match get_type's.

That's not an option for GStreamer unfortunately :)

sdroege:
Ok, so we have a plan I guess.

PR #928, opened 2020-06-09
Correctly get function configuration and make use of it for get_type(…

sdroege:
…) functions in sys mode

We shouldn't look for the full get_type function name but for the name
without the namespace and type name prefix.

Once we have that, make use of the ignore and version configuration in
addition to the general #[cfg] handling. This allows us to ignore
get_type() functions and override their version, for example if the
get_type() function was added after the type like in Pango.

First part of https://github.com/gtk-rs/gir/issues/839

----

CC @EPashkin @GuillaumeGomez 

With this we can configure in sys mode as follows, for example

```toml
[[object]]
name = "Pango.AttrIterator"
status = "generate"
    [[object.function]]
    name = "get_type"
    version = "1.44"
```

### DIFF
--- a/src/codegen/sys/functions.rs
+++ b/src/codegen/sys/functions.rs
@@ -23,11 +23,13 @@ pub fn generate_records_funcs(
     for record in records {
         let name = format!("{}.{}", env.config.library_name, record.name);
         let obj = env.config.objects.get(&name).unwrap_or(&DEFAULT_OBJ);
+        let version = std::cmp::max(record.version, obj.version);
         let glib_get_type = record.glib_get_type.as_ref().unwrap_or(&intern_str);
         generate_object_funcs(
             w,
             env,
             obj,
+            version,
             &record.c_type,
             glib_get_type,
             &record.functions,
@@ -45,10 +47,12 @@ pub fn generate_classes_funcs(
     for klass in classes {
         let name = format!("{}.{}", env.config.library_name, klass.name);
         let obj = env.config.objects.get(&name).unwrap_or(&DEFAULT_OBJ);
+        let version = std::cmp::max(klass.version, obj.version);
         generate_object_funcs(
             w,
             env,
             obj,
+            version,
             &klass.c_type,
             &klass.glib_get_type,
             &klass.functions,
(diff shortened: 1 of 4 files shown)


src/codegen/sys/functions.rs
```diff
@@ -183,9 +183,23 @@ fn generate_object_funcs(
...
     }
     if write_get_type {
-        let configured_functions = obj.functions.matched(&glib_get_type);
-        generate_cfg_configure(w, &configured_functions, false)?;
-        writeln!(w, "    pub fn {}() -> GType;", glib_get_type)?;
+        use crate::case::CaseExt;
+
+        let type_name = c_type.to_snake();
+        assert!(glib_get_type.starts_with(&type_name));
+        let function_name = &glib_get_type[(type_name.len() + 1)..];
```
sdroege, line 190: Do we have a better way for doing the above 3 steps?

sdroege: Could go via the symbol-prefixes of the namespace and then the symbol-prefix of the type... doing so now, needs more parsing :)

sdroege:
Will fix the failures tomorrow.

src/codegen/sys/functions.rs
```diff
@@ -183,9 +208,64 @@ fn generate_object_funcs(
...
+                    if symbol_prefix.starts_with(prefix) {
+                        Some(String::from(&symbol_prefix[(prefix.len() + 1)..]))
+                    } else {
+                        None
+                    }
+                })
+                .next()
+                .unwrap_or_else(|| {
+                    panic!(
+                        "Can't guess symbol prefix for type {} with get_type() function {}",
```
EPashkin, line 236: Can you make this message slightly different from previous ?

EPashkin:
@sdroege Thanks.

src/codegen/sys/functions.rs
sdroege: Actually that makes no sense at all, let me fix that properly :)

sdroege: Done, all working a bit different now

sdroege:
@EPashkin If you want we can remove the parser changes commit as it's unneeded, but maybe we can keep it in case someone needs this information later?

EPashkin:
@sdroege Its good as is, thanks. At maximum, we can make class's "symbol-prefix" optional too, just in case.

sdroege:
Ok, then let's do the next part. Non-sys :) Thanks for merging

sdroege:
@EPashkin I would store the `get_type` version in the `analysis::Record`/`Enum`/etc structs next to the function name and get the correct version number during analysis (which then also means that for auto boxed types for example the minimum version would be the version of the `get_type`). Does that sound ok to you? Then inside codegen I would only check that version and create code conditional on that.

EPashkin:
Yes, just store it in Record/Enum/etc.

sdroege:
Forgot to mention, this PR alone already adds lots of version markers to `get_type` functions everywhere. Shows how very wrong things were before, even before this pango update...

EPashkin:
And because of this first thing we needed is sys regen ;)

GuillaumeGomez:
That's a great add, thanks! :)

PR #929, opened 2020-06-11
functions: rename constructors

### LINKED ISSUE #905: new_from_XXX() constructors should be generated as from_XXX()
The idiomatic way for Rust constructors is to be named `from_XXX()` instead of `new_from_XXX()`, see discussion on https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/459#note_470007

gdesmott:
The idiomatic way for Rust constructors is to be named from_XXX()
instead of new_from_XXX() and with_XXX() instead of new_with_XXX().

Fix #905

### DIFF
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -459,6 +459,14 @@ fn analyze_callbacks(
     }
 }
 
+fn rename_constructor(name: String) -> String {
+    if name.starts_with("new_from") || name.starts_with("new_with") {
+        name[4..].to_string()
+    } else {
+        name
+    }
+}
+
 fn analyze_function(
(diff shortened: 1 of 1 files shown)


sdroege:
:+1:

sdroege:
Need to regenerate everything after this. Also seems like gtk fails for unrelated reasons currently, which seems problematic. I'll investigate later.

EPashkin:
@GuillaumeGomez Thanks

EPashkin:
Ops, sorry
@gdesmott  Thank for renaming

GuillaumeGomez:
You're welcome. :p

PR #930, opened 2020-06-11
Handle version configuration on get_type functions in non-sys mode

sdroege:
And only make use of the functions conditionally if they're newer than
the type itself.

This only applies to records (boxed/shared), enums and flags. GObject
types always need a get_type function and won't work without, so there
will never be the case that the function is added later.

Fixes https://github.com/gtk-rs/gir/issues/839

----

This should be all for being able to update pango now. Generates for example the following:

```rust
#[cfg(any(feature = "v1_44", feature = "dox"))]
glib_wrapper! {
    pub struct AttrIterator(Boxed<pango_sys::PangoAttrIterator>);

    match fn {
        copy => |ptr| pango_sys::pango_attr_iterator_copy(mut_override(ptr)),
        free => |ptr| pango_sys::pango_attr_iterator_destroy(ptr),
        get_type => || pango_sys::pango_attr_iterator_get_type(),
    }
}

#[cfg(not(any(feature = "v1_44", feature = "dox")))]
glib_wrapper! {
    pub struct AttrIterator(Boxed<pango_sys::PangoAttrIterator>);

    match fn {
        copy => |ptr| pango_sys::pango_attr_iterator_copy(mut_override(ptr)),
        free => |ptr| pango_sys::pango_attr_iterator_destroy(ptr),
    }
}
```

Once merged we need to add the correct version configuration everywhere and then regenerate things.

### DIFF
--- a/src/analysis/record.rs
+++ b/src/analysis/record.rs
@@ -8,6 +8,7 @@ use crate::{
     library,
     nameutil::*,
     traits::*,
+    version::Version,
 };
 use log::info;
 use std::ops::Deref;
@@ -15,7 +16,7 @@ use std::ops::Deref;
 #[derive(Debug, Default)]
 pub struct Info {
     pub base: InfoBase,
-    pub glib_get_type: Option<String>,
+    pub glib_get_type: Option<(String, Option<Version>)>,
     pub use_boxed_functions: bool,
     pub derives: Derives,
     pub init_function_expression: Option<String>,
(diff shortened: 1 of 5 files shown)


sdroege:
CC @EPashkin @GuillaumeGomez

sdroege:
Oh I should've tried to compile the generated code first. This of course does not work because the `#[cfg]` is only added to the first item and not every item that comes out of the macro.

Suggestions?

sdroege:
Solved that one, that happened because both were enabled if `dox` was enabled.

src/analysis/record.rs
```diff
@@ -153,6 +154,48 @@ pub fn new(env: &Env, obj: &GObject) -> Option<Info> {
...
+                // As the get_type function only has a version if explicitly configured let's just
+                // panic here. It's easy enough for the user to move the version configuration from
+                // the function to the type.
+                panic!(
+                    "Have to use get_type function for {} but version is higher than for the type ({:?} > {:?})",
+                    full_name, get_type_version, version
+                );
+            }
+        } else {
+            error!("Missing memory management functions for {}", full_name);
```
sdroege, line 194: This here also prevents the dreadful panic that would happen inside codegen otherwise for these types and instead it reports a proper error.

EPashkin:
@sdroege Thanks, looks workable

PR #931, opened 2020-06-12
Generate version conditions for functions in sys mode no lower than t…

sdroege:
…he type's version

This causes the correct version to be added to gdk_gl_error_quark() for
example.

----

CC @EPashkin @GuillaumeGomez


GuillaumeGomez:
:+1:

EPashkin:
@sdroege Thanks

PR #932, opened 2020-06-14
Give preference to configured function versions over versions automat…

sdroege:
…ically inferred in sys mode

The automatically inferred versions might be wrong in various cases, and
if a version is specifically configured then it's going to be the case
for a reason.

gobject-introspection sometimes moves a free-standing function into a
type that was added later, like in the case of GstAudioBuffer (added in
1.16) and gst_audio_buffer_clip() (existed since 1.0).

----

CC @EPashkin @GuillaumeGomez


EPashkin:
@sdroege Thanks.
Seems gio still not updated: https://github.com/gtk-rs/gir/pull/932/checks?check_run_id=770599211#step:21:171

sdroege:
@EPashkin Indeed, that's on my list for today or tomorrow :)

PR #933, opened 2020-06-16
Allow to ignore builders

GuillaumeGomez:
cc @EPashkin @sdroege

### DIFF
--- a/src/analysis/class_builder.rs
+++ b/src/analysis/class_builder.rs
@@ -19,7 +19,7 @@ pub fn analyze(
     obj: &GObject,
     imports: &mut Imports,
 ) -> Vec<Property> {
-    if !obj.generate_builder {
+    if !obj.generate_builder || obj.ignore_builder {
         return Vec::new();
     }
 
--- a/src/config/config.rs
+++ b/src/config/config.rs
@@ -185,7 +185,6 @@ impl Config {
             .map(|t| gobjects::parse_toml(t, concurrency, generate_display_trait))
             .unwrap_or_default();
         gobjects::parse_status_shorthands(&mut objects, &toml, concurrency, generate_display_trait);
-        gobjects::parse_builders(&mut objects, &toml);
 
         let external_libraries = read_external_libraries(&toml)?;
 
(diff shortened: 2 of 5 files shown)


EPashkin:
IMHO better just add `options.ignored_builders`.
I don't like to add it to objects list directly.
@sdroege what you think about this?

sdroege:
Needs to be added to the `README.md`

sdroege:
>  @sdroege what you think about this?

See comments in the gtk PR, I think these are not actually needed and another change in gir would be needed instead.

But for this specific configuration, I would add it to the objects directly instead. It's a configuration for the object and not a global configuration.

sdroege:
On the other hand, `builders` is a global list and not per object... so this would be consistent at least. I don't like `builders` to be global either, but I also don't use it in any of my crates

GuillaumeGomez:
@EPashkin: so @sdroege suggested to remove the global builder list and instead add a `generate_builder` key to objects. What do you think about it? If we agree on this, I'll update this PR and make the change as well (and on the other repositories too).

sdroege:
@GuillaumeGomez And to make the required changes smaller, maybe there could be a global option to create builders by default or to not create them. And then have a configuration to override that default per object?

GuillaumeGomez:
Sure.

EPashkin:
@GuillaumeGomez Ok, Lets change `generate_builder: Option<bool>` in `Gobject`,
and remove global list in future

GuillaumeGomez:
Ok, I'll update this PR then.

GuillaumeGomez:
I removed the global "generate_builders", I also took into account the "abstract" property. So if we all agree on this, I'll add documentation and update the repositories using "generate_builders" (only gtk as far as I can tell).

What do you think?

EPashkin:
@GuillaumeGomez Thanks, all looks good

GuillaumeGomez:
Arf, I didn't update documentation, I just wanted your opinion before merging. :laughing: Well, sending another PR quickly to fix this then!

PR #937, opened 2020-06-22
Use system-deps

gdesmott:
Use system-deps instead of our custom deps handing code.

[Tested with gstreamer-sys](https://gitlab.freedesktop.org/gdesmott/gstreamer-rs-sys/pipelines/165046).

cc @sdroege

### DIFF
--- a/src/codegen/sys/cargo_toml.rs
+++ b/src/codegen/sys/cargo_toml.rs
@@ -82,7 +82,8 @@ fn fill_in(root: &mut Table, env: &Env) {
 
     {
         let build_deps = upsert_table(root, "build-dependencies");
-        set_string(build_deps, "pkg-config", "0.3.7");
+        set_string(build_deps, "system-deps", "1.3");
+        unset(build_deps, "pkg-config");
     }
 
     {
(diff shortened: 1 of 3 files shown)


sdroege:
Thanks, looks good to me. We should get this in for the release :)

@epashkin @GuillaumeGomez

GuillaumeGomez:
I don't see where it is used in `gstreamer-sys`. It isn't in the `cargo.toml` file?

sdroege:
> I don't see where it is used in `gstreamer-sys`. It isn't in the `cargo.toml` file?

@gdesmott didn't send an MR for it yet because it's not merged into gir yet

GuillaumeGomez:
I have troubles with merging something this huge just before a new release.

sdroege:
Functionally it's the same we had before by default, but it a) moves the logic into another crate so we don't have to copy it everywhere and b) allows overriding the behaviour in various ways via environment variables if the default is not good

src/codegen/sys/cargo_toml.rs
```diff
@@ -82,7 +82,7 @@ fn fill_in(root: &mut Table, env: &Env) {
 
     {
         let build_deps = upsert_table(root, "build-dependencies");
-        set_string(build_deps, "pkg-config", "0.3.7");
+        set_string(build_deps, "system-deps", "1.3");
```
EPashkin, line 85: Maybe add `unset(build_deps,  "pkg-config");` as it no more needed

EPashkin:
@gdesmott Thans, it works fine on my Windows/MSys64.

src/codegen/sys/cargo_toml.rs
gdesmott: Good point, I added that.

EPashkin:
@gdesmott Thanks again

src/codegen/sys/cargo_toml.rs
```diff
@@ -82,7 +82,8 @@ fn fill_in(root: &mut Table, env: &Env) {
...
+        unset(build_deps, "pkg-config");
```
sdroege, line 86: This breaks `gdk-sys`, which uses `pkg-config` in some custom build.rs code.

See CI failure that will happen here soon: https://github.com/gtk-rs/sys/pull/174 (`Cargo.toml` will have changes after running gir: the removed `pkg-config` dependency)

EPashkin: Добавил https://github.com/gtk-rs/gir/pull/941,

sdroege: Спасибо :)

GuillaumeGomez: Русское время!

PR #939, opened 2020-06-23
cargo_toml: set feature-versions for all versions

gdesmott:
collect_versions() returns all the versions while
env.namespaces.main().versions() is a filtered subset.

### DIFF
--- a/src/codegen/sys/cargo_toml.rs
+++ b/src/codegen/sys/cargo_toml.rs
@@ -121,10 +121,8 @@ fn fill_in(root: &mut Table, env: &Env) {
         set_string(meta, "version", env.config.min_cfg_version.to_string());
 
         let versions = upsert_table(meta, "feature-versions");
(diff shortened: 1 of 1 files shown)


EPashkin:
IMHO We need only version from main namespace, Ex. we need Gtk_3_16 but not glib_2_24 when generating gtk-sys

EPashkin:
Strange, but it show no difference in gtk-rs, @gdesmott you have sample where this change produce different result?

gdesmott:
Yes, I changed it because of https://gitlab.freedesktop.org/gstreamer/gstreamer-rs-sys/-/merge_requests/61#note_547014

EPashkin:
Ok, https://github.com/gtk-rs/gir/blob/master/src/codegen/sys/build.rs#L85 also use `collect_versions`
@gdesmott Thanks

PR #943, opened 2020-07-01
Generate docs for bitfields too

sdroege:
CC @EPashkin @GuillaumeGomez 

Then it's only not supported by rustdoc-stripper... which also mishandles enums currently.


EPashkin:
@sdroege Thanks

sdroege:
See https://github.com/GuillaumeGomez/rustdoc-stripper/issues/63 for the other part

PR #944, opened 2020-07-02
Correctly generate doc elements for bitfields

GuillaumeGomez:


src/codegen/doc/mod.rs
```diff
@@ -79,9 +79,9 @@ macro_rules! impl_function_like_type {
     };
 }
 
-impl_to_stripper_type!(Member, Variant);
+impl_to_stripper_type!(Member, Const);
 impl_to_stripper_type!(Enumeration, Enum);
-impl_to_stripper_type!(Bitfield, Type);
+impl_to_stripper_type!(Bitfield, Struct);
```
EPashkin, line 84: Just to be sure: This change really needed?

GuillaumeGomez: Yes, they are declared as struct and need to be stored as such, otherwise rustdoc-stripper doesn't "recognize" them.

GuillaumeGomez:
@EPashkin By the way, any idea what's going on with the windows builds?

EPashkin:
Seems libsecret is not installed, I don't have time to look more

EPashkin:
@GuillaumeGomez Thanks

PR #945, opened 2020-07-02
Fix variant doc generation

GuillaumeGomez:
I realized afterwards that bitfields fields and enum variants were both `Member`, so it's not good for the second one. Maybe I should use the same code for bitfields as well? What do you think @EPashkin ?

### DIFF
--- a/src/codegen/doc/mod.rs
+++ b/src/codegen/doc/mod.rs
@@ -79,7 +79,6 @@ macro_rules! impl_function_like_type {
     };
 }
 
-impl_to_stripper_type!(Member, Const);
 impl_to_stripper_type!(Enumeration, Enum);
 impl_to_stripper_type!(Bitfield, Struct);
 impl_to_stripper_type!(Record, Struct);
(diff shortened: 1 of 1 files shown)


EPashkin:
@GuillaumeGomez Yes, please make bitfield same.
Also seems first `write_item_doc` can be made to function, but not sure that it make code clearer

GuillaumeGomez:
Ok, I'll update the code quickly then!

GuillaumeGomez:
@EPashkin I just realized: what do you mean by "`write_item_doc` can be made into function"? It's already a function, no?

GuillaumeGomez:
In the meantime I applied the same change to the bitfield.

EPashkin:
Calls `write_item_doc` in enum and bitfield are same and can be made into function that accepting: `doc`, `deprecated_version`, `doc_deprecated`

GuillaumeGomez:
Oh I see. Indeed, I'm not sure it's very pertinent in here...

PR #946, opened 2020-07-03
Update dependencies

GuillaumeGomez:
I forgot that gir has a dependency on `rustdoc-stripper`. Now it's updated!

cc @EPashkin @sdroege


EPashkin:
@GuillaumeGomez Thanks

PR #953, opened 2020-07-25
Fix tests for libsecret

EPashkin:
cc @GuillaumeGomez , @sdroege


tests/sys/gir-secret.toml
```diff
@@ -11,6 +11,9 @@ external_libraries = [
 ]
 
 ignore = [
+   "Secret.MAJOR_VERSION",
+   "Secret.MINOR_VERSION",
+   "Secret.MICRO_VERSION",
```
GuillaumeGomez, line 16: Based on [semver spec](https://semver.org/), it should be called `PATCH`, but I prefer this wording. :3

EPashkin:
Ok, CI passed with some dirty hacks

PR #956, opened 2020-08-09
Update README.md

TopView:
Added what a 'gir' is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,8 @@
 # GIR
 
+"[GObject introspection (abbreviated ‘GIR’)](https://developer.gnome.org/programming-guidelines/stable/introspection.html.en) is a system which extracts APIs from C code and produces binary type libraries which can be used by non-C language bindings, and other tools, to introspect or wrap the original C libraries. It uses a system of annotations in documentation comments in the C code to expose extra information about the APIs which is not machine readable from the code itself."
+
+
 The `GIR` is used to generate both the sys level crate and a safe API crate to use the sys level (FFI) crate.
 
 ## How to use


TopView:
Hope I've got this right.

README.md
```diff
@@ -1,5 +1,8 @@
 # GIR
 
+"[GObject introspection (abbreviated ‘GIR’)](https://developer.gnome.org/programming-guidelines/stable/introspection.html.en) is a system which extracts APIs from C code and produces binary type libraries which can be used by non-C language bindings, and other tools, to introspect or wrap the original C libraries. It uses a system of annotations in documentation comments in the C code to expose extra information about the APIs which is not machine readable from the code itself."
```
sdroege, line 3: Why in quotes?

```suggestion
"[GObject introspection (abbreviated ‘GIR’)](https://developer.gnome.org/programming-guidelines/stable/introspection.html.en) is a system which extracts APIs from C code and produces binary type libraries and XML API descriptions which can be used by non-C language bindings, and other tools, to introspect or wrap the original C libraries. It uses a system of annotations in documentation comments in the C code to expose extra information about the APIs which is not machine readable from the code itself."
```

TopView: The double quotes, are because this is directly quoted from the URL I provided.  The single quotes around GIR aren't needed, just my personal preference.  Thank you for the additional text.  I suppose you might want to include it as '[and XML API descriptions]' (i.e. in square brackets) to show that it's not part of the original quote.  OR if you want, just remove the outer double quotes, as it's no longer a direct quote.  I don't know the right thing to do here.  ....Just that I spend a whole bunch of time trying to figure out what the heck GIR was, so I could even start to read this.  .... And while you're at it please clarify the term 'crate', thought I now see it's from the rust packager or something.  Thanks again!

EPashkin:
While names are same this crate is not GObjectIntrospection, but code generator that used its results.

TopView:
@EPashkin, I'm sorry, which names are you referring to?  And which crate, and which code generator?  And can you say more about what GIR refers to exactly?  Thank you for your comment, but unfortunately it left me with more questions and confusion that I started with here.  :-)

EPashkin:
This crate is code generator

PR #965, opened 2020-10-02
Specify link attribute for Windows to be happy

seungha-yang:
import library needs to be explicitly linked

### DIFF
--- a/src/codegen/sys/cargo_toml.rs
+++ b/src/codegen/sys/cargo_toml.rs
@@ -39,7 +39,7 @@ fn fill_empty(root: &mut Table, env: &Env, crate_name: &str) {
         set_string(
             package,
             "links",
-            nameutil::crate_name(&env.config.library_name),
+            nameutil::shared_libs_to_links(&env.namespaces.main().shared_libs),
         );
     }
 
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -23,6 +23,18 @@ pub fn generate(env: &Env) {
     save_to_file(&path, env.config.make_backup, |w| generate_lib(w, env));
 }
 
+fn write_link_attr(w: &mut dyn Write, shared_libs: &[String]) -> Result<()> {
+    for it in shared_libs {
+        writeln!(
+            w,
+            "#[link(name = \"{}\")]",
+            shared_lib_name_to_link_name(it)
+        )?;
+    }
+
+    Ok(())
+}
+
 fn generate_lib(w: &mut dyn Write, env: &Env) -> Result<()> {
     general::start_comments(w, &env.config)?;
     statics::begin(w)?;
@@ -60,6 +72,7 @@ fn generate_lib(w: &mut dyn Write, env: &Env) -> Result<()> {
     generate_classes_structs(w, env, &classes)?;
     generate_interfaces_structs(w, env, &interfaces)?;
 
+    write_link_attr(w, &env.namespaces.main().shared_libs)?;
     writeln!(w, "extern \"C\" {{")?;
     functions::generate_enums_funcs(w, env, &enums)?;
     functions::generate_bitfields_funcs(w, env, &bitfields)?;
--- a/src/nameutil.rs
+++ b/src/nameutil.rs
@@ -138,6 +138,17 @@ pub fn shared_lib_name_to_link_name(name: &str) -> &str {
     s
 }
 
+pub fn shared_libs_to_links(shared_libs: &[String]) -> String {
+    // https://github.com/rust-lang/cargo/issues/4533
+    // links = ["foo", "bar"] or ['foo', 'bar'] is not supported.
+    // Pick the first element
+    if let Some(shared_lib) = shared_libs.get(0) {
+        return format!("\"{}\"", shared_lib_name_to_link_name(&shared_lib));
+    }
+
+    panic!("empty list of shared library");
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -201,5 +212,26 @@ mod tests {
             shared_lib_name_to_link_name("libgdk_pixbuf-2.0.so.0"),
             "gdk_pixbuf-2.0"
         );
+    }
+
+    #[test]
+    fn shared_libs_to_links_works() {
+        let libs = vec![
+            String::from("libgobject-2.0.so.0"),
+            String::from("libglib-2.0.so.0"),
+        ];
+        // https://github.com/rust-lang/cargo/issues/4533
+        // list is not supported. Pick the first one
+        assert_eq!(shared_libs_to_links(&libs), "\"gobject-2.0\"");
+
+        let libs = vec![String::from("libgio-2.0.so.0")];
+        assert_eq!(shared_libs_to_links(&libs), "\"gio-2.0\"");
+    }
+
+    #[test]
+    #[should_panic]
+    fn shared_libs_to_links_panic() {
+        let libs = vec![];
+        shared_libs_to_links(&libs);
     }
 }
--- a/src/nameutil.rs
+++ b/src/nameutil.rs
@@ -125,6 +125,19 @@ pub fn lib_name_to_toml(name: &str) -> String {
     name.to_string().replace("-", "_").replace(".", "_")
 }
 
+pub fn shared_lib_name_to_link_name(name: &str) -> &str {
+    let mut s = name;
+
+    if s.starts_with("lib") {
+        s = &s[3..];
+    }
+
+    let offset = s.find(".so").unwrap_or_else(|| s.len());
+    s = &s[..offset];
+
+    s
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -179,5 +192,14 @@ mod tests {
     #[test]
     fn lib_name_to_toml_works() {
         assert_eq!(lib_name_to_toml("gstreamer-1.0"), "gstreamer_1_0");
+    }
+
+    #[test]
+    fn shared_lib_name_to_link_name_works() {
+        assert_eq!(shared_lib_name_to_link_name("libatk-1.0.so.0"), "atk-1.0");
+        assert_eq!(
+            shared_lib_name_to_link_name("libgdk_pixbuf-2.0.so.0"),
+            "gdk_pixbuf-2.0"
+        );
     }
 }


src/codegen/sys/lib_.rs
```diff
@@ -23,6 +23,26 @@ pub fn generate(env: &Env) {
     save_to_file(&path, env.config.make_backup, |w| generate_lib(w, env));
 }
 
+fn get_library_name(lib: &String) -> String {
+    let mut s = String::from(lib);
+
+    if s.starts_with("lib") {
+        s = s.split_off("lib".len());
```
sdroege, line 30:
```suggestion
        s = &s[3..0];
```

Would also do :)

src/codegen/sys/lib_.rs
```diff
@@ -23,6 +23,26 @@ pub fn generate(env: &Env) {
     save_to_file(&path, env.config.make_backup, |w| generate_lib(w, env));
 }
 
+fn get_library_name(lib: &String) -> String {
+    let mut s = String::from(lib);
```
sdroege, line 27: You don't need to create a `String` here, `&str` should be sufficient

src/codegen/sys/lib_.rs
```diff
@@ -23,6 +23,26 @@ pub fn generate(env: &Env) {
...
+
+    if s.starts_with("lib") {
+        s = s.split_off("lib".len());
+    }
+
+    let offset = s.find(".so").unwrap_or(s.len());
+    s.truncate(offset);
```
sdroege, line 34: Or just

```suggestion
    &s[..offset]
```

:)

src/codegen/sys/lib_.rs
```diff
@@ -23,6 +23,26 @@ pub fn generate(env: &Env) {
...
+    s
+}
+
+fn write_link_attr(w: &mut dyn Write, shared_libs: &Vec<String>) -> Result<()> {
+    for it in shared_libs.iter() {
+        writeln!(w, "#[link(name = \"{}\")]", get_library_name(it))?;
```
sdroege, line 40: I think this should also be used here: https://github.com/gtk-rs/gir/blob/a6737c72ae70ee7a17ae4ac7729905f9bb87e4e3/src/codegen/sys/cargo_toml.rs#L42

It's wrong what it does there currently.

src/codegen/sys/lib_.rs
```diff
@@ -23,6 +23,26 @@ pub fn generate(env: &Env) {
...
+        s = s.split_off("lib".len());
+    }
+
+    let offset = s.find(".so").unwrap_or(s.len());
+    s.truncate(offset);
+    s
+}
+
+fn write_link_attr(w: &mut dyn Write, shared_libs: &Vec<String>) -> Result<()> {
+    for it in shared_libs.iter() {
```
sdroege, line 39:
```suggestion
    for it in &shared_libs {
```

sdroege:
Apart from that various clippy warnings, you'll see them when clicking on the diff here or by running it locally :)

Otherwise looks good to me, thanks!

src/codegen/sys/lib_.rs
seungha-yang: I guess you meant `for it in shared_libs` :) done

src/codegen/sys/lib_.rs
seungha-yang: done

src/codegen/sys/lib_.rs
```diff
@@ -23,6 +23,18 @@ pub fn generate(env: &Env) {
     save_to_file(&path, env.config.make_backup, |w| generate_lib(w, env));
 }
 
+fn write_link_attr(w: &mut dyn Write, shared_libs: &Vec<String>) -> Result<()> {
```
sdroege, line 26:
```suggestion
fn write_link_attr(w: &mut dyn Write, shared_libs: &[String]) -> Result<()> {
```

src/nameutil.rs
```diff
@@ -125,6 +125,28 @@ pub fn lib_name_to_toml(name: &str) -> String {
...
+        s = &s[3..];
+    }
+
+    let offset = s.find(".so").unwrap_or(s.len());
+    s = &s[..offset];
+
+    s.to_string()
+}
+
+pub fn shared_libs_to_links(shared_libs: &Vec<String>) -> String {
```
sdroege, line 141:
```suggestion
pub fn shared_libs_to_links(shared_libs: &[String]) -> String {
```

src/nameutil.rs
```diff
@@ -125,6 +125,28 @@ pub fn lib_name_to_toml(name: &str) -> String {
...
+    let mut links = vec![];
+    for it in shared_libs {
+        links.push(format!("\"{}\"", shared_lib_name_to_link_name(it)));
+    }
+
+    format!("[{}]", links.join(", "))
```
sdroege, line 147: Maybe only if `!links.is_empty()`?

src/nameutil.rs
```diff
@@ -125,6 +125,28 @@ pub fn lib_name_to_toml(name: &str) -> String {
     name.to_string().replace("-", "_").replace(".", "_")
 }
 
+pub fn shared_lib_name_to_link_name(name: &str) -> String {
```
sdroege, line 128:
```suggestion
pub fn shared_lib_name_to_link_name(name: &str) -> &str {
```

Then you don't need the `to_string()` in the end

src/nameutil.rs
seungha-yang: oh, empty list, good point :) but just curios it's possible case apart from programming error. (no native library is required for binding?)

sdroege: That's a good point. Should probably directly error out then instead

seungha-yang: now it will cause panic in that case

sdroege: Not nice to panic there but it seems like there's no other error reporting at all implemented right now...

src/nameutil.rs
```diff
@@ -125,6 +125,36 @@ pub fn lib_name_to_toml(name: &str) -> String {
...
 }
 
+pub fn shared_lib_name_to_link_name(name: &str) -> &str {
+    let mut s = name;
+
+    if s.starts_with("lib") {
+        s = &s[3..];
+    }
+
+    let offset = s.find(".so").unwrap_or(s.len());
```
sdroege, line 135:
```suggestion
    let offset = s.find(".so").unwrap_or_else(|| s.len());
```

src/nameutil.rs
seungha-yang: right, that's why I just gave up there 😅

src/nameutil.rs
seungha-yang: done

sdroege (approved):

seungha-yang:
`links = ["foo", "bar"]` or `['foo', 'bar']` doesn't work... not sure why. the last commit was updated so that picking up the first library from list

sdroege:
> `links = ["foo", "bar"]` or `['foo', 'bar']` doesn't work... not sure why

Because cargo doesn't support it: https://github.com/rust-lang/cargo/issues/4533

src/nameutil.rs
```diff
@@ -125,6 +125,35 @@ pub fn lib_name_to_toml(name: &str) -> String {
...
+        links.push(format!("\"{}\"", shared_lib_name_to_link_name(it)));
+    }
+
+    if links.is_empty() {
+        panic!("empty list of shared library");
+    }
+
+    // FIXME: links = ["foo", "bar"] or ['foo', 'bar'] doesn't seem to be working
+    // https://doc.rust-lang.org/cargo/reference/manifest.html#the-links-field
+    // is not mentioning about this though, probably cargo bug?
```
sdroege, line 153: https://github.com/rust-lang/cargo/issues/4533 says this is not supported

src/codegen/sys/cargo_toml.rs
```diff
@@ -39,7 +39,9 @@ fn fill_empty(root: &mut Table, env: &Env, crate_name: &str) {
         set_string(
             package,
             "links",
-            nameutil::crate_name(&env.config.library_name),
+            // FIXME: Will pick the first library from list.
+            // links = ["foo", "bar"] doesn't work...
```
sdroege, line 43: Remove the `FIXME` comment from here, that workaround is in the function itself and not here :)

src/nameutil.rs
sdroege: Can you update the comment accordingly?

Looks good to me otherwise

sdroege (approved):

EPashkin (approved):
Checked build locally, it works

src/nameutil.rs
```diff
@@ -125,6 +125,35 @@ pub fn lib_name_to_toml(name: &str) -> String {
...
+
+    if links.is_empty() {
+        panic!("empty list of shared library");
+    }
+
+    // https://github.com/rust-lang/cargo/issues/4533
+    // links = ["foo", "bar"] or ['foo', 'bar'] is not supported.
+    // Pick the first element
+    links[0].clone()
+}
```
GuillaumeGomez, line 155: Considering that `shared_lib_name_to_link_name` always return something, why not replacing this function with:

```suggestion
pub fn shared_libs_to_links(shared_libs: &[String]) -> String {
    // https://github.com/rust-lang/cargo/issues/4533
    // links = ["foo", "bar"] or ['foo', 'bar'] is not supported.
    // Pick the first element
    for it in shared_libs {
        return format!("\"{}\"", shared_lib_name_to_link_name(it));
    }
    panic!("empty list of shared library");
}
```

?

seungha-yang: that looks much better :) thanks for the suggestion.

sdroege: Causes a clippy warning though, rightfully:

```
 src/nameutil.rs#L145

this loop never actually loops
```

sdroege: So instead of the loop, just do `if !shared_libs.is_empty() { ... }`

seungha-yang: hmm why my cargo build/test build doesn't tell warning? :) btw updated

sdroege: Run `cargo clippy` to get extra warnings/suggestions :)

sdroege (approved):

sdroege:
@GuillaumeGomez Let's get this in then?

src/nameutil.rs
```diff
@@ -125,6 +125,30 @@ pub fn lib_name_to_toml(name: &str) -> String {
...
+    s
+}
+
+pub fn shared_libs_to_links(shared_libs: &[String]) -> String {
+    // https://github.com/rust-lang/cargo/issues/4533
+    // links = ["foo", "bar"] or ['foo', 'bar'] is not supported.
+    // Pick the first element
+    if !shared_libs.is_empty() {
+        return format!("\"{}\"", shared_lib_name_to_link_name(&shared_libs[0]));
+    }
```
GuillaumeGomez, line 147: Sorry, should have suggested:

```suggestion
if let Some(shared_lib) = shared_libs.iter().next() {
    return format!("\"{}\"", shared_lib_name_to_link_name(&shared_lib));
}
```

sdroege: Or even better: if let Some(item) = vec.get(0)

No need for an iterator.

GuillaumeGomez: True!

seungha-yang: thanks for the suggestions!

GuillaumeGomez (approved):
Looks good to me, thanks!

sdroege:
@GuillaumeGomez You merge? Everybody approved it now :)

GuillaumeGomez:
Normally it's up to @EPashkin but since they already approved, I guess it's fine.

PR #970, opened 2020-10-17
Add per-crate and per-object configuration to trust return value nullability

sdroege:
This has to be used carefully as many libraries forgot to put
`nullability` annotations on return values, which then causes a panic if
an unexpected `NULL` is returned.

----

CC @GuillaumeGomez @EPashkin @sophie-h

With this you can check what kind of changes are generated when trusting the gir files, and then report any missing nullable annotations to the C libraries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,13 @@ single_version_file = true
 # which do not have an override for `generate_display_trait`
 # (defaults to "true")
 generate_display_trait = true
+# Trust the nullability information about return values. If this is disabled
+# then any pointer return type is assumed to be nullable unless there is an
+# explicit override for it.
+# This has to be used carefully as many libraries are missing nullable
+# annotations for return values, which then will cause a panic once an
+# unexpected NULL is returned.
+trust_return_value_nullability = false
 # Disable running `cargo fmt` on generated files
 # (defaults to false)
 disable_format = true
@@ -209,6 +216,9 @@ cfg_condition = "mycond"
 generate_display_trait = false
 # if you want to generate builder with name SomeClassBuilder
 generate_builder = true
+# trust return value nullability annotations for this specific type.
+# See above for details and use with care
+trust_return_value_nullability = false
     # define overrides for function
     [[object.function]]
     # filter functions from object
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -144,6 +144,7 @@ pub fn analyze<F: Borrow<library::Function>>(
 
         let mut info = analyze_function(
             env,
+            obj,
             name,
             func,
             type_tid,
@@ -469,6 +470,7 @@ fn rename_constructor(name: String) -> String {
 
 fn analyze_function(
     env: &Env,
+    obj: &config::gobjects::GObject,
     name: String,
     func: &library::Function,
     type_tid: library::TypeId,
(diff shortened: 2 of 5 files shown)


GuillaumeGomez:
I'm not sure this is a good idea to have such a setting... In any case, please put a multi-line warning (on the console when running gir with this setting) to say that it needs to be use with a lot of caution.

sophie-h:
As I said in the gtk issue, I think for libraries like gtk this feature would only be realistic on a per .c-file/class basis. Otherwise we would have to check thousands of functions before being able to use this feature.

sdroege:
On 17 October 2020 15:03:04 EEST, Guillaume Gomez <notifications@github.com> wrote:
>I'm not sure this is a good idea to have such a setting... In any case, please put a multi-line warning (on the console when running gir with this setting) to say that it needs to be use with a lot of caution.
>

I would like to use it in gstreamer where it's almost correct now and where I can take care of fixing it. I don't want a warning there :)

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

sdroege:
On 17 October 2020 15:19:53 EEST, Sophie Herold <notifications@github.com> wrote:
>As I said in the gtk issue, I think for libraries like gtk this feature would only be realistic on a per .c-file/class basis. Otherwise we would have to check thousands of functions before being able to use this feature.
>

That's doable, yes. I'll also add a per type configuration for this in addition. 

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

sdroege:
On 17 October 2020 15:03:04 EEST, Guillaume Gomez <notifications@github.com> wrote:
>I'm not sure this is a good idea to have such a setting... In any case, please put a multi-line warning (on the console when running gir with this setting) to say that it needs to be use with a lot of caution.
>

Also the goal here would be to slowly move all crates to this by fixing upstream annotations so that we don't have to worry about it anymore and simply have to review new APIs like we have to do now.

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

sdroege:
@GuillaumeGomez Also following https://gitlab.gnome.org/GNOME/gtk/-/issues/3267#note_937101 this is basically the only way forward here. We can't change gobject-introspection and have to live with how wonderfully broken reality is :)

sdroege:
@sophie-h This can now also be configured per object if you want to give it a try

sdroege:
@GuillaumeGomez This is also ready for review now

sophie-h:
I'm very happy about this. Set one object to `trust_return_value_nullability`, checked the diff in src/auto, found missing annotations in gtk code. I think this is a good way forward.

sdroege:
> I'm very happy about this. Set one object to `trust_return_value_nullability`, checked the diff in src/auto, found missing annotations in gtk code. I think this is a good way forward.

Thanks for taking care of that :) I'll go through gio one of these days unless you want to do that.

sophie-h:
> I'll go through gio one of these days unless you want to do that.

No, thanks :) Not sure if I will have a lot of time for that.

sdroege:
> No, thanks :) Not sure if I will have a lot of time for that.

See https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1708 . I'll wait until that is in a GLib release, otherwise this needs too many overrides in `Gir.toml`.

sophie-h:
Is there anything to be done here or can this be merged?

sdroege:
It needs @GuillaumeGomez or @EPashkin to review it :)

EPashkin (approved):
@sdroege Thanks, I was great idea.

Please add PR with use `matches!` in separate MR sometime (see CI clippy errors)

sdroege:
> Please add PR with use `matches!` in separate MR sometime (see CI clippy errors)

@GuillaumeGomez already did that here: https://github.com/gtk-rs/gir/pull/973

Let's get this in then? :)

sdroege:
@GuillaumeGomez Can you take a look at this? The new GLib release fixes all the remaining missing annotations, so we could have a clean version of GLib now with improved API.

GuillaumeGomez (approved):

GuillaumeGomez:
After a discussion between @sdroege and I, I finally agreed on his position so let's get this in!

sdroege:
Thanks!

Issue #969, opened 2020-10-12
Error enums: Use the error-quark correctly


bilelmoussaoui:
see the comments in https://github.com/gtk-rs/glib/pull/697#discussion_r502524870 for details

sdroege:
So what we should do here is to use the `error-quark` as a string, because that's what it is. There is not necessarily a function for this.

This can then be placed in a `OnceCell<glib::Quark>`, for example.

sdroege:
Should also revert https://github.com/gtk-rs/gir-files/pull/60 at that point

sdroege:
See https://github.com/gtk-rs/gir/pull/972

PR #972, opened 2020-10-17
Properly handle error-domain

sdroege:
Fixes https://github.com/gtk-rs/gir/issues/969

### DIFF
--- a/src/library_postprocessing.rs
+++ b/src/library_postprocessing.rs
@@ -36,6 +36,7 @@ impl Library {
         self.fix_fields();
         self.make_unrepresentable_types_opaque();
         self.mark_final_types(config);
+        self.update_error_domain_functions();
     }
 
     fn fix_gtype(&mut self) {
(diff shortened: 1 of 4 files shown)


src/library_postprocessing.rs
```diff
@@ -505,4 +506,47 @@ impl Library {
             }
         }
     }
+
+    fn update_error_domain_functions(&mut self) {
```
sdroege, line 510: This does not seem to work, investigating later

sdroege:
CC @GuillaumeGomez ready for a review. This causes no changes in gio, and the only change in glib is

```diff
--- a/src/auto/enums.rs
+++ b/src/auto/enums.rs
@@ -581,7 +581,11 @@ impl FromGlib<glib_sys::GUriError> for UriError {
 #[cfg(any(feature = "v2_66", feature = "dox"))]
 impl ErrorDomain for UriError {
     fn domain() -> Quark {
-        unsafe { from_glib(glib_sys::g_uri_error_quark()) }
+        static QUARK: once_cell::sync::Lazy<glib_sys::GQuark> =
+            once_cell::sync::Lazy::new(|| unsafe {
+                glib_sys::g_quark_from_static_string(b"g-uri-quark\0".as_ptr() as *const _)
+            });
+        from_glib(*QUARK)
     }
 
     fn code(self) -> i32 {
```

While there **is** a function for the quark, it's wrongly named. We could add a configuration for that but IMHO we should just ignore this special case for now instead of adding a configuration for that.

sdroege:
Added some more heuristics that should cover the GStreamer and GLib special cases

GuillaumeGomez (approved):

sdroege:
@GuillaumeGomez Let's get that in then, and then revert that workaround in `gir-files`? Regenerating everything is not needed yet, there are no real changes, it's just cleanup :)

src/library_postprocessing.rs
```diff
@@ -505,4 +506,129 @@ impl Library {
...
+                                .iter()
+                                .any(|c| f.c_identifier.as_ref() == Some(c))
+                        }) {
+                            error_domains.push((
+                                ns_id,
+                                enum_tid,
+                                None,
+                                func.c_identifier.as_ref().unwrap().clone(),
+                            ));
+                            continue 'next_enum;
```
GuillaumeGomez, line 556: Why not just `continue` btw?

src/library_postprocessing.rs
```diff
@@ -505,4 +506,129 @@ impl Library {
...
+                                    .iter()
+                                    .any(|c| f.c_identifier.as_ref() == Some(c))
+                            }) {
+                                error_domains.push((
+                                    ns_id,
+                                    enum_tid,
+                                    Some(domain_tid),
+                                    func.c_identifier.as_ref().unwrap().clone(),
+                                ));
+                                continue 'next_enum;
```
GuillaumeGomez, line 586: Why not just `continue`?

GuillaumeGomez: Ah nevermind.

src/library_postprocessing.rs
sdroege: Here it could be plain `continue` but I thought it would be less confusing to have it with a label in both cases so it's clear that both `continue`s do the same and that there's not another loop hidden somewhere :)

GuillaumeGomez: Fine by me!

GuillaumeGomez:
Thanks!

sdroege:
https://github.com/gtk-rs/gir-files/pull/63 reverts the workaround in `gir-files`

PR #973, opened 2020-10-17
Fix clippy warnings

### LINKED ISSUE #971: Lots of clippy warnings with latest clippy
See title, needs to be cleaned up :)

GuillaumeGomez:
Fixes #971.

cc @sdroege @EPashkin

### DIFF
--- a/src/codegen/function_body_chunk.rs
+++ b/src/codegen/function_body_chunk.rs
@@ -40,10 +40,7 @@ enum OutMemMode {
 
 impl OutMemMode {
     fn is_uninitialized(&self) -> bool {
-        match *self {
-            OutMemMode::Uninitialized => true,
-            _ => false,
-        }
+        matches!(*self, OutMemMode::Uninitialized)
     }
 }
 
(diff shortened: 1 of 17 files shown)


GuillaumeGomez:
Is it a big issue if it doesn't compile on stable? :-°

src/codegen/function_body_chunk.rs
```diff
@@ -1080,17 +1076,15 @@ impl Builder {
                 ref mem_mode,
             } = *par
             {
-                if self
-                    .transformations
-                    .iter()
-                    .any(|tr| match tr.transformation_type {
+                if self.transformations.iter().any(|tr| {
```
sdroege, line 1079: Clippy complains about this still, if you don't want to change it can you add an `#[allow(clippy::...)]` in the right place?

sdroege:
Looks good to me otherwise, no function changes and can go in after that last warning is fixed or ignored

src/codegen/function_body_chunk.rs
GuillaumeGomez: It doesn't on nightly, so...

GuillaumeGomez:
Let's get it in then!

Issue #962, opened 2020-08-30
Rename get/set_XXX_property() to get/set_XXX()


sdroege:
Came up a while ago in some issue. This should make the API nicer in 99% of the cases (and we already don't create the property functions if there are normal getters/setters), and in the few cases where there would be conflicts this can be resolved via renaming.

@EPashkin @GuillaumeGomez What do you think?

GuillaumeGomez:
It's nicer but also "harder" for people to use. Explanations: mostly functions discoverability, if the name is too different from the C one, it makes it harder to find or know it's the one to use.

So personally, I don't mind, but I think it was important to be noted.

GuillaumeGomez:
PS: I intend to use "doc_alias" in docs to allow to search for C function names directly. :3

sdroege:
> mostly functions discoverability, if the name is too different from the C one, it makes it harder to find or know it's the one to use.

These would only exist if there are no C functions, otherwise we use the C getter/setter anyway (we do already now and don't generate the `get/set_XXX_property()` then).

Other bindings are behaving exactly the same btw (without a `_property` prefix), so it would seem more consistent in that regard at least.

sdroege:
This should be done together with https://github.com/gtk-rs/gir/issues/963

That other ticket makes things follow Rust code style better, this one removes duplication from the API and makes us behave like literally all other bindings.

@bilelmoussaoui What do you think?

bilelmoussaoui:
This can be closed now

PR #977, opened 2020-10-27
Use char::TryFrom::<u32> to convert from UniChar

fengalin:
This holds if we assume the C side Unicode validity can be trusted.

Consequence of the removal of `FromGlib<u32>` for `char` and `Option<char>` in https://github.com/gtk-rs/glib/pull/718.

The following projects were updated:
- https://github.com/fengalin/atk/tree/translate-option-result
- https://github.com/fengalin/pango/tree/translate-option-result
- https://github.com/fengalin/gtk/tree/translate-option-result

With these updates, `gtk-rs/examples` builds successfully. `gstreamer-rs` doesn't seem affected.

I'll send pull requests for `atk`, `pango` and `gtk` when this is merged.

### DIFF
--- a/src/codegen/function_body_chunk.rs
+++ b/src/codegen/function_body_chunk.rs
@@ -1310,10 +1310,17 @@ fn add_chunk_for_type(
     let type_ = env.type_(typ_);
     match *type_ {
         library::Type::Fundamental(ref x) if !x.requires_conversion() => true,
-        library::Type::Fundamental(library::Fundamental::Boolean)
-        | library::Type::Fundamental(library::Fundamental::UniChar) => {
+        library::Type::Fundamental(library::Fundamental::Boolean) => {
             body.push(Chunk::Custom(format!(
                 "let {0} = from_glib({0});",
+                par.name
+            )));
+            true
+        }
+        library::Type::Fundamental(library::Fundamental::UniChar) => {
+            body.push(Chunk::Custom(format!(
+                "let {0} = std::convert::TryFrom::try_from({0})\
+                     .expect(\"conversion from an invalid Unicode value attempted\");",
                 par.name
             )));
             true
--- a/src/codegen/translate_from_glib.rs
+++ b/src/codegen/translate_from_glib.rs
@@ -23,7 +23,13 @@ impl TranslateFromGlib for Mode {
         use crate::analysis::conversion_type::ConversionType::*;
         match ConversionType::of(env, self.typ) {
             Direct => (String::new(), String::new()),
-            Scalar => ("from_glib(".into(), ")".into()),
+            Scalar => match env.library.type_(self.typ) {
+                library::Type::Fundamental(library::Fundamental::UniChar) => (
+                    "std::convert::TryFrom::try_from(".into(),
+                    ").expect(\"conversion from an invalid Unicode value attempted\")".into(),
+                ),
+                _ => ("from_glib(".into(), ")".into()),
+            },
             Pointer => {
                 let trans = from_glib_xxx(self.transfer, array_length);
                 match *env.type_(self.typ) {
--- a/README.md
+++ b/README.md
@@ -395,6 +395,23 @@ status = "generate"
     ignore = true
 ```
 
+Which will prevent gir from generating `stock_list_ids`. If you want to specify
+that a function will be manually implemented, you can use:
+
+
+```toml
+[[object]]
+name = "Gtk.Entry"
+status = "generate"
+    [[object.function]]
+    name = "get_invisible_char"
+    manual = true
+```
+
+This will prevent gir from generating `get_invisible_char` and it won't generate
+`get_property_invisible_char` which would have been generated if we had used
+"ignore = true".
+
 Note that you must not place `Gtk.*` into the `generate` array and
 additionally configure its members.
 
--- a/src/analysis/class_builder.rs
+++ b/src/analysis/class_builder.rs
@@ -63,7 +63,10 @@ fn analyze_properties(
             continue;
         }
         let configured_properties = obj.properties.matched(&prop.name);
-        if configured_properties.iter().any(|f| f.ignore) {
+        if !configured_properties
+            .iter()
+            .all(|f| f.status.need_generate())
+        {
             continue;
         }
 
--- a/src/analysis/constants.rs
+++ b/src/analysis/constants.rs
@@ -22,7 +22,10 @@ pub fn analyze<F: Borrow<library::Constant>>(
         let constant = constant.borrow();
         let configured_constants = obj.constants.matched(&constant.name);
 
-        if configured_constants.iter().any(|c| c.ignore) {
+        if !configured_constants
+            .iter()
+            .all(|c| c.status.need_generate())
+        {
             continue;
         }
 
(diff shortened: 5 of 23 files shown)


sdroege (approved):

fengalin:
@GuillaumeGomez: what do you think? I'll submit the other modifications on `gtk-rs/gtk-rs` when / if this one is merged.

src/codegen/function_body_chunk.rs
```diff
@@ -1308,14 +1308,20 @@ fn add_chunk_for_type(
...
+        library::Type::Fundamental(library::Fundamental::Boolean) => {
             body.push(Chunk::Custom(format!(
                 "let {0} = from_glib({0});",
                 par.name
             )));
             true
         }
+        library::Type::Fundamental(library::Fundamental::UniChar) => {
+            body.push(Chunk::Custom(format!(
+                "let {0} = std::convert::TryFrom::try_from({0}).unwrap();",
```
GuillaumeGomez, line 1320: Maybe using an `expect` in here instead? :)

src/codegen/translate_from_glib.rs
```diff
@@ -23,7 +23,13 @@ impl TranslateFromGlib for Mode {
         use crate::analysis::conversion_type::ConversionType::*;
         match ConversionType::of(env, self.typ) {
             Direct => (String::new(), String::new()),
-            Scalar => ("from_glib(".into(), ")".into()),
+            Scalar => match env.library.type_(self.typ) {
+                library::Type::Fundamental(library::Fundamental::UniChar) => (
+                    "std::convert::TryFrom::try_from(".into(),
+                    ").unwrap()".into(),
```
GuillaumeGomez, line 29: `expect` here too.

GuillaumeGomez:
Just nits but looks good to me otherwise. :)

src/codegen/function_body_chunk.rs
fengalin: We can, IIUIC, the `unwrap` will print the `CharTryFromError` message in this case.

GuillaumeGomez: I want proof. 😄

fengalin: :) try it yourself: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=bb2e9c3697d3567db45085c521c3d8c3

:D

GuillaumeGomez:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: CharTryFromError(())', src/main.rs:4:54
```

Sorry but this is just not good enough. For a "novice" rust user, this isn't helpful at all. Please use `expect`.

fengalin:
Hold on please! I just noticed a discrepancy with latest `gir-files`.

fengalin:
So, the problem I noticed was not introduced by this PR as I can reproduce it with current `master`, but it appears as a consequence of the changes required to `gtk/Gir.toml`. This is what I observe:

Switching `Gtk.Entry.get_invisible_char` to `ignore = true` in `gtk/Gir.toml`, `get_invisible_char` is not generated as expected, but `get_property_invisible_char` is generated when it isn't with current settings:

```
    name = "get_invisible_char"
        [object.function.return]
        nullable = true
```
Is this expected?
Should I do anything to prevent `get_property_invisible_char` from being generated?
Note that the property is declared as returning `u32` and I think we could do better by manually returning `char`. Do you agree?

src/codegen/function_body_chunk.rs
sdroege: Yeah unclear why `unwrap()` prints the `Debug` impl of the error instead of the `Display` impl. That's quite useless behaviour.

sdroege:
>  Note that the property is declared as returning `u32` and I think we could do better by manually returning `char`. Do you agree?

Yes that should be a `char`. I think there's a configuration for overriding the return type?

sdroege:
>  Should I do anything to prevent `get_property_invisible_char` from being generated?

If there's a normal getter then disable the property getter. See also https://github.com/gtk-rs/gir/issues/962

fengalin:
> If there's a normal getter then disable the property getter. See also #962

Oh, that's fine. However, can I instruct `gir` that I'll implement `get_invisible_char` manually so that it doesn't need to implement `get_property_invisible_char`? If there's nothing yet for this, I can add a commit.

sdroege:
> However, can I instruct `gir` that I'll implement `get_invisible_char` manually so that it doesn't need to implement `get_property_invisible_char`? If there's nothing yet for this, I can add a commit.

You can configure the property to be ignored, yes.

fengalin:
> You can configure the property to be ignored, yes.

The problem with that is that it also gets removed from the builders. What do you think about detecting that `get_invisible_char` is declared as `ignore = true` so that the `get_property_invisible_char` is not generated but the rest remains?

sdroege:
See the issue I linked above. I would trait properties as equivalent to getters/setters with the same name for all practical purposes apart from the notify signal.

And I think for builders the problem is simply that we don't distinguish between "manual" and "ignore" currently. You might have more luck with not ignoring the property but only generating the notify signal (and no getter/setter). There's also configuration for that.

fengalin:
Ok thanks! I'll resume this tomorrow.

fengalin:
I implemented the "manual" option for functions so that we can prevent `get_property_...` from being generated when the getter is manually generated: https://github.com/gtk-rs/gir/pull/977/commits/c656603ce8208da5f00667a62aa4646b8007a313.

Used in `gtk-rs/gtk` and non-regression checked on `gtk-rs/atk` and `gtk-rs/pango`. See last commits in this branch: https://github.com/fengalin/gtk-rs/tree/translate-option-result. I'll insert a  commit to update the `gir` submodule and update the "regenerate" commits when this is merged.

fengalin:
I also tested on `gstreamer-rs/gstreamer` `Gir.toml`. See https://gitlab.freedesktop.org/fengalin/gstreamer-rs/-/commits/translate-option-result
and updated the last commit to make sure we get the same result with "ignore = true" and "manual = true" where applicable.

sdroege (approved):
Still looks good to me, @GuillaumeGomez ?

GuillaumeGomez (approved):
Looks good to me as well!

GuillaumeGomez:
The CI failures however seem to be compilation failures.

fengalin:
> The CI failures however seem to be compilation failures.

Forgot to update tests... Sorry.

fengalin:
CI needs https://github.com/gtk-rs/gtk-rs/pull/11 which depends on this one.

fengalin:
One last update: I forgot to add a "manual" in a list so that is doesn't issue an ERROR line (doesn't change the generated code though).

fengalin:
Should be good to go this time. I changed applicable "ignore = true" to "manual = true" in `gtk-rs` and `gstreamer-rs` and regenerated with success.

GuillaumeGomez:
It's still broken, however: I guess it's expected?

GuillaumeGomez:
Ok, it seems to be expected, merging then. Thanks @fengalin !

PR #980, opened 2020-10-30
toml: use new system-deps versions syntax

gdesmott:
This will allow us to override the library name depending on the
version, see https://github.com/gdesmott/system-deps/issues/11

### DIFF
--- a/src/codegen/sys/cargo_toml.rs
+++ b/src/codegen/sys/cargo_toml.rs
@@ -82,7 +82,7 @@ fn fill_in(root: &mut Table, env: &Env) {
 
     {
         let build_deps = upsert_table(root, "build-dependencies");
-        set_string(build_deps, "system-deps", "1.3");
+        set_string(build_deps, "system-deps", "2.0");
     }
 
     {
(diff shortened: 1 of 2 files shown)


gdesmott:
@sdroege : let me know if that's ok to you and I'll merge and release the system-deps changes.

src/codegen/sys/cargo_toml.rs
```diff
@@ -82,6 +82,7 @@ fn fill_in(root: &mut Table, env: &Env) {
 
     {
         let build_deps = upsert_table(root, "build-dependencies");
+        // TODO: bump to 1.4 once released
```
sdroege, line 85: Shouldn't this become 2.0 because it breaks API?

sdroege:
Seems good to me otherwise

src/codegen/sys/cargo_toml.rs
gdesmott: I was planning to do 1.3 -> 1.4 for the API break but you're right, let's go to 2.0 to make it clearer.

gdesmott:
This should be ready now, I just released system-deps 2.0.0.

sdroege (approved):

sdroege:
@GuillaumeGomez Let's get this in?

GuillaumeGomez:
Thanks!

PR #985, opened 2020-11-06
Fix CI

GuillaumeGomez:
Follow-up of https://github.com/gtk-rs/gtk-rs/pull/18

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest]
+        os: [ubuntu-20.04]
         toolchain: [stable]
 
     defaults:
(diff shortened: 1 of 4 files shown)


sdroege:
@GuillaumeGomez What do we do about Windows? Disable that for now until someone figures out why it's completely useless? This might just need something like `pacman update` in the beginning, not sure.

GuillaumeGomez:
I'm investigating. I think we should already merge this PR and I'll continue to look on another branch since we at least fixed linux build. What do you think?

sdroege:
Works for me

ids1024:
I'm not sure what the best practices on Github Actions are, but using `pacman -Syu` in place of `pacman -S` should update the package database and install updates (just `pacman -Sy` will only do the first; not sure if that is likely to cause issues in practice).

GuillaumeGomez:
@ids1024 None of us know much about Windows so any lead is good. I'll try what you suggested. :)

ids1024:
> None of us know much about Windows so any lead is good. 

I don't know much about Windows either, but `pacman` is the Arch Linux package manger, which msys2 apparently decided to use.

PR #986, opened 2020-11-06
Try to fix Windows CI

GuillaumeGomez:
Follow-up of #985.

PR #991, opened 2020-11-15
Generate `impl` blocks for associated enum functions

MarijnS95:
Leverage existing function generation for class/record to generate associated functions on enumerations as well, instead of having to write these by hand :tada:. The MR for `gstreamer-rs` can be found [here](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/630). These changes came up during https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/581.

This relies on functions being associated with the `<enumeration>` in `.gir` files as opposed to being free (same story with class/record), see https://gitlab.freedesktop.org/gstreamer/gst-plugins-base/-/merge_requests/805 for an example where this was not the case.

Note that we have [decided](https://gitlab.freedesktop.org/gstreamer/gst-plugins-base/-/merge_requests/805#note_616786) against turning the first argument from `<parameter>` into `<instance-parameter>` in Gir files (to make these functions behave like `fn my_func(&self)` instead of `fn my_func(my_enum: Self)`). This is instead overridden during analysis, it's a bit ugly so perhaps we need to revisit this discussion. It all depends on other tooling, and how many other languages (potentially) support functions on enum instances like Rust.

- [x] Apologies for finding out about SSR in the history of this repo and performing some refactorings :grimacing:. Perhaps those are better submitted in a separate, easier-to-review PR?

### Related PRs/MRs

- gstreamer-rs: https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/630

### TODO

- [x]  Active review comment threads
  - [x] `to_string` seems quite a large task still:
    - [x] When do we use it?
    - [x] How do we generate generic (error-handling) code for it?
    - [x] Fix recursion because of `fn to_string(self)` (ie. by naming it consistently `to_str` across the board)
    - [x] Fix that one remaining unused `use glib::GString` in `webrtc/enums` (due to `&'static str` replacement)
    - [x] Decide on panicking rules:
      ```rust
      .map(|ptr| CStr::from_ptr(ptr).to_str().unwrap())
      .unwrap_or("INVALID")
      ```
      versus:
      ```rust
      .and_then(|ptr| CStr::from_ptr(ptr).to_str().ok())
      .unwrap_or("INVALID")
      ```
      Even though the latter discards a descriptive `Err` in favour of `None`.
- [x]  Adding support for flags (which should be exactly the same)
  - [x] Check if we can reduce code duplication
- [x]  Moving the quote stuff into a separate PR
  Removed from this PR, separate PR comes when rebasing doesn't conflict with these changes anymore :wink: 
- [x]  Removing the commits from #994 after that one is merged
- [x] ~Autogenerate `from_string`, which is pretty straightforward~ Not worth it: many different types use this function, all with their own semantics.
  - [ ] However manual implementations are updated to call the autogenerated version, which takes care of unsafe, asserting initialization, and glib to/from conversion. Where possible this should be applied to other manual functions as well.
- [x] Do not import `std::fmt` in flags when `generate_display_trait` is true but nothing has `to_string` (only problematic in gtk)
  - This was my mistake, copying the analysis code to add `std::fmt` when `generate_display_trait` is true, without any matching codegen. A sensible Display trait is now emitted in that case.
- [x] Rebase this branch on master as soon as edition2018 drops on gstreamer-rs :)
- [x] Rename `gst_video_chroma_*` functions to `gst_video_chroma_site_*` for autogeneration.
- [x] If there is an unknown value in a returned enum (`fn from_*(xxx) -> EnumType`), turn that into a `Result<..., BoolError>`;

### DIFF
--- a/src/analysis/enums.rs
+++ b/src/analysis/enums.rs
@@ -26,7 +26,7 @@ impl Info {
     }
 }
 
-pub fn new(env: &Env, obj: &GObject) -> Option<Info> {
+pub fn new(env: &Env, obj: &GObject, imports: &mut Imports) -> Option<Info> {
     info!("Analyzing enumeration {}", obj.name);
 
     let enumeration_tid = env.library.find_type(0, &obj.name)?;
@@ -35,8 +35,6 @@ pub fn new(env: &Env, obj: &GObject) -> Option<Info> {
 
     let name = split_namespace_name(&obj.name).1;
 
-    let mut imports = Imports::with_defined(&env.library, name);
-
     let mut functions = functions::analyze(
         env,
         &enumeration.functions,
@@ -44,13 +42,13 @@ pub fn new(env: &Env, obj: &GObject) -> Option<Info> {
         false,
         false,
         obj,
-        &mut imports,
+        imports,
         None,
         None,
     );
     let specials = special_functions::extract(&mut functions);
 
-    special_functions::analyze_imports(&specials, &mut imports);
+    special_functions::analyze_imports(&specials, imports);
 
     let (version, deprecated_version) = info_base::versions(
         env,
@@ -66,7 +64,8 @@ pub fn new(env: &Env, obj: &GObject) -> Option<Info> {
         name: name.to_owned(),
         functions,
         specials,
-        imports,
+        // TODO: Don't use!
+        imports: Imports::new(&env.library),
         version,
         deprecated_version,
         cfg_condition: obj.cfg_condition.clone(),
--- a/src/analysis/enums.rs
+++ b/src/analysis/enums.rs
@@ -1,20 +1,15 @@
-use super::{function_parameters::TransformationType, imports::Imports, info_base::InfoBase, *};
+use super::{function_parameters::TransformationType, imports::Imports, *};
 use crate::{config::gobjects::GObject, env::Env, nameutil::*, traits::*};
 
 use log::info;
-use std::ops::Deref;
 
 #[derive(Debug, Default)]
 pub struct Info {
-    pub base: InfoBase,
-}
-
-impl Deref for Info {
-    type Target = InfoBase;
-
-    fn deref(&self) -> &InfoBase {
-        &self.base
-    }
+    pub full_name: String,
+    pub type_id: library::TypeId,
+    pub name: String,
+    pub functions: Vec<functions::Info>,
+    pub specials: special_functions::Infos,
 }
 
 impl Info {
@@ -116,29 +111,13 @@ pub fn new(env: &Env, obj: &GObject, imports: &mut Imports) -> Option<Info> {
 
     special_functions::analyze_imports(&specials, imports);
 
-    let (version, deprecated_version) = info_base::versions(
-        env,
-        obj,
-        &functions,
-        enumeration.version,
-        enumeration.deprecated_version,
-    );
-
-    let base = InfoBase {
+    let info = Info {
         full_name: obj.name.clone(),
         type_id: enumeration_tid,
         name: name.to_owned(),
         functions,
         specials,
-        // TODO: Don't use!
-        imports: Imports::new(&env.library),
-        version,
-        deprecated_version,
-        cfg_condition: obj.cfg_condition.clone(),
-        concurrency: obj.concurrency,
     };
-
-    let info = Info { base };
 
     Some(info)
 }
--- a/src/analysis/enums.rs
+++ b/src/analysis/enums.rs
@@ -35,6 +35,9 @@ pub fn new(env: &Env, obj: &GObject, imports: &mut Imports) -> Option<Info> {
 
     let name = split_namespace_name(&obj.name).1;
 
+    // Mark the type as available within the enum namespace:
+    imports.add_defined(&format!("crate::{}", name));
+
     let mut functions = functions::analyze(
         env,
         &enumeration.functions,
--- a/src/analysis/enums.rs
+++ b/src/analysis/enums.rs
@@ -1,0 +1,79 @@
+use super::{imports::Imports, info_base::InfoBase, *};
+use crate::{config::gobjects::GObject, env::Env, library, nameutil::*, traits::*};
+use log::info;
+use std::ops::Deref;
+
+#[derive(Debug, Default)]
+pub struct Info {
+    pub base: InfoBase,
+}
+
+impl Deref for Info {
+    type Target = InfoBase;
+
+    fn deref(&self) -> &InfoBase {
+        &self.base
+    }
+}
+
+impl Info {
+    pub fn type_<'a>(&self, library: &'a library::Library) -> &'a library::Enumeration {
+        let type_ = library
+            .type_(self.type_id)
+            .maybe_ref()
+            .unwrap_or_else(|| panic!("{} is not an enumeration.", self.full_name));
+        type_
+    }
+}
+
+pub fn new(env: &Env, obj: &GObject) -> Option<Info> {
+    info!("Analyzing enumeration {}", obj.name);
+
+    let enumeration_tid = env.library.find_type(0, &obj.name)?;
+    let type_ = env.type_(enumeration_tid);
+    let enumeration: &library::Enumeration = type_.maybe_ref()?;
+
+    let name = split_namespace_name(&obj.name).1;
+
+    let mut imports = Imports::with_defined(&env.library, name);
+
+    let mut functions = functions::analyze(
+        env,
+        &enumeration.functions,
+        enumeration_tid,
+        false,
+        false,
+        obj,
+        &mut imports,
+        None,
+        None,
+    );
+    let specials = special_functions::extract(&mut functions);
+
+    special_functions::analyze_imports(&specials, &mut imports);
+
+    let (version, deprecated_version) = info_base::versions(
+        env,
+        obj,
+        &functions,
+        enumeration.version,
+        enumeration.deprecated_version,
+    );
+
+    let base = InfoBase {
+        full_name: obj.name.clone(),
+        type_id: enumeration_tid,
+        name: name.to_owned(),
+        functions,
+        specials,
+        imports,
+        version,
+        deprecated_version,
+        cfg_condition: obj.cfg_condition.clone(),
+        concurrency: obj.concurrency,
+    };
+
+    let info = Info { base };
+
+    Some(info)
+}
--- a/src/analysis/enums.rs
+++ b/src/analysis/enums.rs
@@ -1,5 +1,5 @@
 use super::{function_parameters::TransformationType, imports::Imports, info_base::InfoBase, *};
-use crate::{config::gobjects::GObject, env::Env, library, nameutil::*, traits::*};
+use crate::{config::gobjects::GObject, env::Env, nameutil::*, traits::*};
 
 use log::info;
 use std::ops::Deref;
@@ -112,7 +112,7 @@ pub fn new(env: &Env, obj: &GObject, imports: &mut Imports) -> Option<Info> {
         }
     }
 
-    let specials = special_functions::extract(&mut functions);
+    let specials = special_functions::extract(&mut functions, type_);
 
     special_functions::analyze_imports(&specials, imports);
 
--- a/src/analysis/enums.rs
+++ b/src/analysis/enums.rs
@@ -107,7 +107,7 @@ pub fn new(env: &Env, obj: &GObject, imports: &mut Imports) -> Option<Info> {
         }
     }
 
-    let specials = special_functions::extract(&mut functions, type_);
+    let specials = special_functions::extract(&mut functions, type_, obj);
 
     special_functions::analyze_imports(&specials, imports);
 
--- a/src/analysis/enums.rs
+++ b/src/analysis/enums.rs
@@ -29,6 +29,17 @@ impl Info {
 pub fn new(env: &Env, obj: &GObject, imports: &mut Imports) -> Option<Info> {
     info!("Analyzing enumeration {}", obj.name);
 
+    if !obj.status.need_generate() {
+        return None;
+    }
+
+    if !obj
+        .type_id
+        .map_or(false, |tid| tid.ns_id == namespaces::MAIN)
+    {
+        return None;
+    }
+
     let enumeration_tid = env.library.find_type(0, &obj.name)?;
     let type_ = env.type_(enumeration_tid);
     let enumeration: &library::Enumeration = type_.maybe_ref()?;
@@ -37,6 +48,25 @@ pub fn new(env: &Env, obj: &GObject, imports: &mut Imports) -> Option<Info> {
 
     // Mark the type as available within the enum namespace:
     imports.add_defined(&format!("crate::{}", name));
+
+    let has_get_quark = enumeration.error_domain.is_some();
+    if has_get_quark {
+        imports.add("glib::Quark");
+        imports.add("glib::error::ErrorDomain");
+    }
+
+    let has_get_type = enumeration.glib_get_type.is_some();
+    if has_get_type {
+        imports.add("glib::Type");
+        imports.add("glib::StaticType");
+        imports.add("glib::value::SetValue");
+        imports.add("glib::value::FromValue");
+        imports.add("glib::value::FromValueOptional");
+    }
+
+    if obj.generate_display_trait {
+        imports.add("std::fmt");
+    }
 
     let mut functions = functions::analyze(
         env,
--- a/src/analysis/enums.rs
+++ b/src/analysis/enums.rs
@@ -1,5 +1,6 @@
-use super::{imports::Imports, info_base::InfoBase, *};
+use super::{function_parameters::TransformationType, imports::Imports, info_base::InfoBase, *};
 use crate::{config::gobjects::GObject, env::Env, library, nameutil::*, traits::*};
+
 use log::info;
 use std::ops::Deref;
 
@@ -79,6 +80,38 @@ pub fn new(env: &Env, obj: &GObject, imports: &mut Imports) -> Option<Info> {
         None,
         None,
     );
+
+    // Gir does not currently mark the first parameter of associated enum functions -
+    // that are identical to its enum type - as instance parameter since most languages
+    // do not support this.
+    for f in &mut functions {
+        if f.parameters.c_parameters.is_empty() {
+            continue;
+        }
+
+        let first_param = &mut f.parameters.c_parameters[0];
+
+        if first_param.typ == enumeration_tid {
+            first_param.instance_parameter = true;
+
+            let t = f
+                .parameters
+                .transformations
+                .iter_mut()
+                .find(|t| t.ind_c == 0)
+                .unwrap();
+
+            if let TransformationType::ToGlibScalar { name, .. } = &mut t.transformation_type {
+                *name = "self".to_owned();
+            } else {
+                panic!(
+                    "Enumeration function instance param must be passed as scalar, not {:?}",
+                    t.transformation_type
+                );
+            }
+        }
+    }
+
     let specials = special_functions::extract(&mut functions);
 
     special_functions::analyze_imports(&specials, imports);
--- a/src/analysis/flags.rs
+++ b/src/analysis/flags.rs
@@ -101,7 +101,7 @@ pub fn new(env: &Env, obj: &GObject, imports: &mut Imports) -> Option<Info> {
         }
     }
 
-    let specials = special_functions::extract(&mut functions, type_);
+    let specials = special_functions::extract(&mut functions, type_, obj);
 
     special_functions::analyze_imports(&specials, imports);
 
--- a/src/analysis/flags.rs
+++ b/src/analysis/flags.rs
@@ -1,0 +1,117 @@
+use super::{function_parameters::TransformationType, imports::Imports, *};
+use crate::{config::gobjects::GObject, env::Env, nameutil::*, traits::*};
+
+use log::info;
+
+#[derive(Debug, Default)]
+pub struct Info {
+    pub full_name: String,
+    pub type_id: library::TypeId,
+    pub name: String,
+    pub functions: Vec<functions::Info>,
+    pub specials: special_functions::Infos,
+}
+
+impl Info {
+    pub fn type_<'a>(&self, library: &'a library::Library) -> &'a library::Bitfield {
+        let type_ = library
+            .type_(self.type_id)
+            .maybe_ref()
+            .unwrap_or_else(|| panic!("{} is not an flags.", self.full_name));
+        type_
+    }
+}
+
+pub fn new(env: &Env, obj: &GObject, imports: &mut Imports) -> Option<Info> {
+    info!("Analyzing flags {}", obj.name);
+
+    if !obj.status.need_generate() {
+        return None;
+    }
+
+    if !obj
+        .type_id
+        .map_or(false, |tid| tid.ns_id == namespaces::MAIN)
+    {
+        return None;
+    }
+
+    let flags_tid = env.library.find_type(0, &obj.name)?;
+    let type_ = env.type_(flags_tid);
+    let flags: &library::Bitfield = type_.maybe_ref()?;
+
+    let name = split_namespace_name(&obj.name).1;
+
+    // Mark the type as available within the bitfield namespace:
+    imports.add_defined(&format!("crate::{}", name));
+
+    let has_get_type = flags.glib_get_type.is_some();
+    if has_get_type {
+        imports.add("glib::Type");
+        imports.add("glib::StaticType");
+        imports.add("glib::value::SetValue");
+        imports.add("glib::value::FromValue");
+        imports.add("glib::value::FromValueOptional");
+    }
+
+    if obj.generate_display_trait {
+        imports.add("std::fmt");
+    }
+
+    let mut functions = functions::analyze(
+        env,
+        &flags.functions,
+        flags_tid,
+        false,
+        false,
+        obj,
+        imports,
+        None,
+        None,
+    );
+
+    // Gir does not currently mark the first parameter of associated bitfield functions -
+    // that are identical to its bitfield type - as instance parameter since most languages
+    // do not support this.
+    for f in &mut functions {
+        if f.parameters.c_parameters.is_empty() {
+            continue;
+        }
+
+        let first_param = &mut f.parameters.c_parameters[0];
+
+        if first_param.typ == flags_tid {
+            first_param.instance_parameter = true;
+
+            let t = f
+                .parameters
+                .transformations
+                .iter_mut()
+                .find(|t| t.ind_c == 0)
+                .unwrap();
+
+            if let TransformationType::ToGlibScalar { name, .. } = &mut t.transformation_type {
+                *name = "self".to_owned();
+            } else {
+                panic!(
+                    "Bitfield function instance param must be passed as scalar, not {:?}",
+                    t.transformation_type
+                );
+            }
+        }
+    }
+
+    let specials = special_functions::extract(&mut functions, type_);
+
+    special_functions::analyze_imports(&specials, imports);
+
+    let info = Info {
+        full_name: obj.name.clone(),
+        type_id: flags_tid,
+        name: name.to_owned(),
+        functions,
+        specials,
+    };
+
+    Some(info)
+}
--- a/src/analysis/imports.rs
+++ b/src/analysis/imports.rs
@@ -3,6 +3,7 @@ use crate::{library::Library, nameutil::crate_name, version::Version};
 use std::borrow::Cow;
 use std::cmp::Ordering;
 use std::collections::btree_map::BTreeMap;
+use std::collections::HashSet;
 use std::ops::{Deref, DerefMut};
 use std::vec::IntoIter;
 
@@ -70,10 +71,8 @@ fn compare_imports(a: &(&String, &ImportConditions), b: &(&String, &ImportCondit
 pub struct Imports {
     /// Name of the current crate.
     crate_name: String,
-    /// Name defined within current module. It doesn't need use declaration.
-    ///
-    /// NOTE: Currently we don't need to support more than one such name.
-    defined: Option<String>,
+    /// Names defined within current module. It doesn't need use declaration.
+    defined: HashSet<String>,
     defaults: ImportConditions,
     map: BTreeMap<String, ImportConditions>,
 }
@@ -82,7 +81,7 @@ impl Imports {
     pub fn new(gir: &Library) -> Imports {
         Imports {
             crate_name: make_crate_name(gir),
-            defined: None,
+            defined: HashSet::new(),
             defaults: ImportConditions::default(),
             map: BTreeMap::new(),
         }
@@ -91,7 +90,7 @@ impl Imports {
     pub fn with_defined(gir: &Library, name: &str) -> Imports {
         Imports {
             crate_name: make_crate_name(gir),
-            defined: Some(name.to_owned()),
+            defined: std::iter::once(name.to_owned()).collect(),
             defaults: ImportConditions::default(),
             map: BTreeMap::new(),
         }
@@ -127,12 +126,22 @@ impl Imports {
         // The ffi namespace is used directly, including it is a programmer error.
         assert_ne!(name, "crate::ffi");
 
-        if !name.contains("::") && name != "xlib" {
+        if (!name.contains("::") && name != "xlib") || self.defined.contains(name) {
             false
-        } else if let Some(ref defined) = self.defined {
-            !((name.starts_with("crate::") && &name[7..] == defined) || name == defined)
+        } else if let Some(name) = name.strip_prefix("crate::") {
+            !self.defined.contains(name)
         } else {
             true
+        }
+    }
+
+    /// Declares that `name` is defined in scope
+    ///
+    /// Removes existing imports from `self.map` and marks `name` as
+    /// available to counter future import "requests".
+    pub fn add_defined(&mut self, name: &str) {
+        if self.defined.insert(name.to_owned()) {
+            self.map.remove(name);
         }
     }
 
--- a/src/analysis/imports.rs
+++ b/src/analysis/imports.rs
@@ -120,11 +120,14 @@ impl Imports {
         self.defaults.clear();
     }
 
-    /// The goals of this function is to discard unwanted imports like "ffi" and "crate". It
+    /// The goals of this function is to discard unwanted imports like "crate". It
     /// also extends the checks in case you are implementing "X". For example, you don't want to
     /// import "X" or "crate::X" in this case.
     fn common_checks(&self, name: &str) -> bool {
-        if name == "crate::ffi" || (!name.contains("::") && name != "xlib") {
+        // The ffi namespace is used directly, including it is a programmer error.
+        assert_ne!(name, "crate::ffi");
+
+        if !name.contains("::") && name != "xlib" {
             false
         } else if let Some(ref defined) = self.defined {
             !((name.starts_with("crate::") && &name[7..] == defined) || name == defined)
--- a/src/analysis/mod.rs
+++ b/src/analysis/mod.rs
@@ -65,6 +65,7 @@ pub fn run(env: &mut Env) {
     }
 
     let mut enum_imports = Imports::new(&env.library);
+    enum_imports.add("glib::translate::*");
 
     let mut analyzed = 1;
     while analyzed > 0 {
--- a/src/analysis/mod.rs
+++ b/src/analysis/mod.rs
@@ -64,9 +64,6 @@ pub fn run(env: &mut Env) {
         to_analyze.push((tid, deps));
     }
 
-    let mut enum_imports = Imports::new(&env.library);
-    enum_imports.add("glib::translate::*");
-
     let mut analyzed = 1;
     while analyzed > 0 {
         analyzed = 0;
@@ -76,14 +73,12 @@ pub fn run(env: &mut Env) {
                 new_to_analyze.push((tid, deps.clone()));
                 continue;
             }
-            analyze(env, tid, deps, &mut enum_imports);
+            analyze(env, tid, deps);
             analyzed += 1;
         }
 
         to_analyze = new_to_analyze;
     }
-
-    env.analysis.enum_imports = enum_imports;
 
     if !to_analyze.is_empty() {
         error!(
@@ -93,10 +88,35 @@ pub fn run(env: &mut Env) {
         return;
     }
 
+    analyze_enums(env);
+
     analyze_constants(env);
 
     // Analyze free functions as the last step once all types are analyzed
     analyze_global_functions(env);
+}
+
+fn analyze_enums(env: &mut Env) {
+    let mut imports = Imports::new(&env.library);
+    imports.add("glib::translate::*");
+
+    for obj in env.config.objects.values() {
+        if obj.status.ignored() {
+            continue;
+        }
+        let tid = match env.library.find_type(0, &obj.name) {
+            Some(x) => x,
+            None => continue,
+        };
+
+        if let Type::Enumeration(_) = env.library.type_(tid) {
+            if let Some(info) = enums::new(env, obj, &mut imports) {
+                env.analysis.enumerations.push(info);
+            }
+        }
+    }
+
+    env.analysis.enum_imports = imports;
 }
 
 fn analyze_global_functions(env: &mut Env) {
@@ -161,7 +181,7 @@ fn analyze_constants(env: &mut Env) {
     env.analysis.constants = constants::analyze(env, &constants, obj);
 }
 
-fn analyze(env: &mut Env, tid: TypeId, deps: &[TypeId], enum_imports: &mut Imports) {
+fn analyze(env: &mut Env, tid: TypeId, deps: &[TypeId]) {
     let full_name = tid.full_name(&env.library);
     let obj = match env.config.objects.get(&*full_name) {
         Some(obj) => obj,
@@ -181,12 +201,6 @@ fn analyze(env: &mut Env, tid: TypeId, deps: &[TypeId], enum_imports: &mut Impor
         Type::Record(_) => {
             if let Some(info) = record::new(env, obj) {
                 env.analysis.records.insert(full_name, info);
-            }
-        }
-        Type::Enumeration(_) => {
-            // Cannot mutably borrow env.analysis.enum_imports here
-            if let Some(info) = enums::new(env, obj, enum_imports) {
-                env.analysis.enumerations.push(info);
             }
         }
         _ => {}
--- a/src/analysis/mod.rs
+++ b/src/analysis/mod.rs
@@ -12,6 +12,7 @@ pub mod class_builder;
 pub mod class_hierarchy;
 pub mod constants;
 pub mod conversion_type;
+pub mod enums;
 pub mod ffi_type;
 pub mod function_parameters;
 pub mod functions;
@@ -44,6 +45,7 @@ pub struct Analysis {
     pub records: BTreeMap<String, record::Info>,
     pub global_functions: Option<info_base::InfoBase>,
     pub constants: Vec<constants::Info>,
+    pub enumerations: Vec<enums::Info>,
 }
 
 pub fn run(env: &mut Env) {
@@ -171,6 +173,11 @@ fn analyze(env: &mut Env, tid: TypeId, deps: &[TypeId]) {
         Type::Record(_) => {
             if let Some(info) = record::new(env, obj) {
                 env.analysis.records.insert(full_name, info);
+            }
+        }
+        Type::Enumeration(_) => {
+            if let Some(info) = enums::new(env, obj) {
+                env.analysis.enumerations.push(info);
             }
         }
         _ => {}
--- a/src/analysis/mod.rs
+++ b/src/analysis/mod.rs
@@ -15,6 +15,7 @@ pub mod constants;
 pub mod conversion_type;
 pub mod enums;
 pub mod ffi_type;
+pub mod flags;
 pub mod function_parameters;
 pub mod functions;
 pub mod general;
@@ -48,6 +49,9 @@ pub struct Analysis {
     pub constants: Vec<constants::Info>,
     pub enumerations: Vec<enums::Info>,
     pub enum_imports: Imports,
+
+    pub flags: Vec<flags::Info>,
+    pub flags_imports: Imports,
 }
 
 pub fn run(env: &mut Env) {
@@ -90,6 +94,8 @@ pub fn run(env: &mut Env) {
 
     analyze_enums(env);
 
+    analyze_flags(env);
+
     analyze_constants(env);
 
     // Analyze free functions as the last step once all types are analyzed
@@ -117,6 +123,30 @@ fn analyze_enums(env: &mut Env) {
     }
 
     env.analysis.enum_imports = imports;
+}
+
+fn analyze_flags(env: &mut Env) {
+    let mut imports = Imports::new(&env.library);
+    imports.add("glib::translate::*");
+    imports.add("bitflags::bitflags");
+
+    for obj in env.config.objects.values() {
+        if obj.status.ignored() {
+            continue;
+        }
+        let tid = match env.library.find_type(0, &obj.name) {
+            Some(x) => x,
+            None => continue,
+        };
+
+        if let Type::Bitfield(_) = env.library.type_(tid) {
+            if let Some(info) = flags::new(env, obj, &mut imports) {
+                env.analysis.flags.push(info);
+            }
+        }
+    }
+
+    env.analysis.flags_imports = imports;
 }
 
 fn analyze_global_functions(env: &mut Env) {
--- a/src/analysis/mod.rs
+++ b/src/analysis/mod.rs
@@ -110,7 +110,6 @@ fn analyze_global_functions(env: &mut Env) {
 
     let mut imports = imports::Imports::new(&env.library);
     imports.add("glib::translate::*");
-    imports.add(&format!("crate::{}", env.main_sys_crate_name()));
 
     let functions = functions::analyze(
         env,
--- a/src/analysis/mod.rs
+++ b/src/analysis/mod.rs
@@ -2,6 +2,7 @@ use crate::{
     env::Env,
     library::{self, Type, TypeId},
 };
+use imports::Imports;
 use log::error;
 use std::collections::BTreeMap;
 
@@ -46,6 +47,7 @@ pub struct Analysis {
     pub global_functions: Option<info_base::InfoBase>,
     pub constants: Vec<constants::Info>,
     pub enumerations: Vec<enums::Info>,
+    pub enum_imports: Imports,
 }
 
 pub fn run(env: &mut Env) {
@@ -62,6 +64,8 @@ pub fn run(env: &mut Env) {
         to_analyze.push((tid, deps));
     }
 
+    let mut enum_imports = Imports::new(&env.library);
+
     let mut analyzed = 1;
     while analyzed > 0 {
         analyzed = 0;
@@ -71,12 +75,15 @@ pub fn run(env: &mut Env) {
                 new_to_analyze.push((tid, deps.clone()));
                 continue;
             }
-            analyze(env, tid, deps);
+            analyze(env, tid, deps, &mut enum_imports);
             analyzed += 1;
         }
 
         to_analyze = new_to_analyze;
     }
+
+    env.analysis.enum_imports = enum_imports;
+
     if !to_analyze.is_empty() {
         error!(
             "Not analyzed {} objects due unfinished dependencies",
@@ -153,7 +160,7 @@ fn analyze_constants(env: &mut Env) {
     env.analysis.constants = constants::analyze(env, &constants, obj);
 }
 
-fn analyze(env: &mut Env, tid: TypeId, deps: &[TypeId]) {
+fn analyze(env: &mut Env, tid: TypeId, deps: &[TypeId], enum_imports: &mut Imports) {
     let full_name = tid.full_name(&env.library);
     let obj = match env.config.objects.get(&*full_name) {
         Some(obj) => obj,
@@ -176,7 +183,8 @@ fn analyze(env: &mut Env, tid: TypeId, deps: &[TypeId]) {
             }
         }
         Type::Enumeration(_) => {
-            if let Some(info) = enums::new(env, obj) {
+            // Cannot mutably borrow env.analysis.enum_imports here
+            if let Some(info) = enums::new(env, obj, enum_imports) {
                 env.analysis.enumerations.push(info);
             }
         }
--- a/src/analysis/special_functions.rs
+++ b/src/analysis/special_functions.rs
@@ -4,6 +4,7 @@ use crate::{
         imports::Imports,
     },
     library::{Type as LibType, TypeId},
+    version::Version,
 };
 use std::{collections::BTreeMap, str::FromStr};
 
@@ -52,6 +53,7 @@ impl Type {
 pub struct Info {
     pub glib_name: String,
     pub returns_static_ref: bool,
+    pub version: Option<Version>,
 }
 
 pub type Infos = BTreeMap<Type, Info>;
@@ -144,6 +146,7 @@ pub fn extract(functions: &mut Vec<FuncInfo>, parent_type: &LibType) -> Infos {
                 Info {
                     glib_name: func.glib_name.clone(),
                     returns_static_ref,
+                    version: func.version,
                 },
             );
         }
@@ -152,12 +155,14 @@ pub fn extract(functions: &mut Vec<FuncInfo>, parent_type: &LibType) -> Infos {
     if has_copy && !has_free {
         if let Some((glib_name, pos)) = destroy {
             let ty_ = Type::from_str("destroy").unwrap();
-            update_func(&mut functions[pos], ty_, parent_type);
+            let func = &mut functions[pos];
+            update_func(func, ty_, parent_type);
             specials.insert(
                 ty_,
                 Info {
                     glib_name,
                     returns_static_ref: false,
+                    version: func.version,
                 },
             );
         }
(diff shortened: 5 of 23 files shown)


sdroege:
The exact same code would also apply to flags types, or not? Can that be unified so (mostly) the same code can cover both?

Also if you could submit all the cleanup commits as a separate PR that would speed up reviewing here. The cleanups all look fine from what I saw so far and could be merged more or less immediately :)

sdroege:
Can you also check what (if any) effects this has on https://github.com/gtk-rs/gtk-rs and submit a WIP PR?

MarijnS95:
> The exact same code would also apply to flags types, or not? Can that be unified so (mostly) the same code can cover both?

I guess you mean `bitfield`s (inconveniently called `codegen/flags.rs`)? Those should indeed share all the same semantics, I'll see if I can rework the current `analysis` to work for both and unify most of the codegen.

> Also if you could submit all the cleanup commits as a separate PR that would speed up reviewing here. The cleanups all look fine from what I saw so far and could be merged more or less immediately :)

Pushed as #986, including some commits that change generated output (ie. `f.write_str()`) but are not fully relevant to this PR either.

sdroege:
> I guess you mean `bitfield`s (inconveniently called `codegen/flags.rs`)

Yeah, `bitfield` is very confusing too because that's also a completely separate concept in C :)

sdroege:
This needs to be rebased now

sdroege:
> > Can you also check what (if any) effects this has on https://github.com/gtk-rs/gtk-rs and submit a WIP PR?
> 
> As mentioned in the edit above I or someone else will have to go through gtk and remove the same handwritten implementations that clash with generated functions, while also validating semantics of generated code (strings...).

That's OK for now, just submit a PR that does not remove the handwritten functions so we can easily compare what is generated there too :)

MarijnS95:
> That's OK for now, just submit a PR that does not remove the handwritten functions so we can easily compare what is generated there too :)

@sdroege Removed them anyway because it was only one class, as well as fixing multiple doc issues that appeared when testing with `--all-features`.

There's a temp commit here to make sure the CI builds on the fixed gtk-rs branch, once this gets closer I guess we'll merge gtk-rs, bump the commit off and make sure it succeeds one final time before merging?

MarijnS95:
@sdroege Pushed a simple yet unsustainable version of a safer, static `.to_string` implementation. We can discuss the details over at our thread on gstreamer-rs, I want to take this opportunity for something else: `quote`. Continuing onward from https://github.com/gtk-rs/gir/pull/992#discussion_r524124393: this `to_string` is my first actual block of code in gir, and using `writeln` + `format` is horrible. I pushed a commit on top that replaces this block with a `quote` macro to show what it'd look like. It should probably not be part of this PR but I don't think it's particularly bad to restart the discussion of switching to this crate.

The benefit of having `Display` on `TokenStream` makes it a drop-in replacement for localized segments of code without requiring `gir` in its entirety to be converted, a massive win IMO.

The only problem I see right now is formatting in comments and macros (`bitflags! {}`), the rest is covered by `fmt` without ever thinking about indentation, `\n` and `\t` :grin: 

Further advantages:
- Sort of syntax checking at compile-time (ie. no mismatched braces, but remember it's merely tokens);
- Highlighting in an IDE;
- Code looks _way closer_ to the generated result;
- No writeln, indentation or `Option` unpacking boilerplate;
- Support for iterable expansion.

Conversion is pretty easy, too: copy an existing block of _generated_ code (instead of tearing down `writeln!`) and substitute variable parts with `#variable_to_expand`.

src/analysis/enums.rs
```diff
@@ -117,6 +117,14 @@ pub fn new(env: &Env, obj: &GObject, imports: &mut Imports) -> Option<Info> {
         }
     }
 
+    // Enum to_string functions provide static strings and need special error handling.
+    if let Some(_to_string) = functions.iter_mut().find(|f| f.name == "to_string") {
```
sdroege, line 121: This is only valid if the C function's return value is `transfer none`

src/analysis/enums.rs
```diff
@@ -117,6 +117,14 @@ pub fn new(env: &Env, obj: &GObject, imports: &mut Imports) -> Option<Info> {
...
+        // TODO: Ignore or manual? Manual means it's available in other scopes right?
+        // to_string.status = crate::config::gobjects::GStatus::Ignore;
+        // to_string.name = "to_str";
```
sdroege, line 124: We have to rename the function because otherwise it conflicts with the owned `ToString::to_string()` and that's not so great :)

I would automatically rename it without configuration. That's not very different from the automatic renaming of e.g. `type` to `type_` because it's a keyword.

src/codegen/enums.rs
```diff
@@ -68,6 +73,56 @@ pub fn generate(env: &Env, root_path: &Path, mod_rs: &mut Vec<String>) {
...
+
+    // TODO: Can we somehow reuse function_body_chunk::Builder here?
+
+    writeln!(
+        w,
+        "\
+\t\tunsafe {{
+\t\t\tlet ptr = {}::{}(self.to_glib());
+\t\t\tif ptr.is_null() {{
+\t\t\t\treturn \"INVALID\";
```
sdroege: Just like `to_string()`, this should probably panic if it's `NULL` unless the return value is marked as `nullable` in the introspection data. If it is marked as `nullable`, then we should probably **not** generate this at all and print a warning. It will require case by case specific manual implementations.

sdroege:
> There's a temp commit here to make sure the CI builds on the fixed gtk-rs branch. Once this gets closer I guess we'll merge gtk-rs, bump the commit off and make sure it succeeds one final time before merging?

What's that commit and why? :)

sdroege:
> I pushed a commit on top that replaces this block with a `quote` macro to show what it'd look like. It should probably not be part of this PR but I don't think it's particularly bad to restart the discussion of switching to this crate.

Let's indeed do that as a separate PR afterwards. I agree that the string manipulations without quote are annoying. Last time I tried moving to quote the big problem was that it was rather annoying to a) do comments, b) format code in a reasonable way (e.g. quote didn't output empty lines between any items IIRC), c) even rustfmt was not converting the code into something reasonable looking.

If that can be solved I'm fine with moving to quote. But as said, different PR. Not too many things at once :)

src/analysis/enums.rs
```diff
@@ -0,0 +1,157 @@
...
+        enumeration.deprecated_version,
+    );
+
+    let base = InfoBase {
+        full_name: obj.name.clone(),
+        type_id: enumeration_tid,
+        name: name.to_owned(),
+        functions,
+        specials,
+        // TODO: Don't use!
```
sdroege, line 146: What does this mean? :)

src/codegen/object.rs
```diff
@@ -82,6 +82,7 @@ pub fn generate(
         } else {
             None
         },
+        None, // TODO
```
sdroege, line 85: What's to do here?

src/codegen/record.rs
```diff
@@ -114,6 +114,7 @@ pub fn generate(w: &mut dyn Write, env: &Env, analysis: &analysis::record::Info)
         &analysis.functions,
         &analysis.specials,
         None,
+        None, // TODO
```
sdroege, line 117: What's to do here?

src/analysis/enums.rs
```diff
@@ -84,6 +85,38 @@ pub fn new(env: &Env, obj: &GObject, imports: &mut Imports) -> Option<Info> {
...
+                .iter_mut()
+                .find(|t| t.ind_c == 0)
+                .unwrap();
+
+            if let TransformationType::ToGlibScalar { name, .. } = &mut t.transformation_type {
+                *name = "self".to_owned();
+            } else {
+                panic!(
+                    "Enum function instance param must be passed as scalar, not {:?}",
+                    t.transformation_type
```
sdroege, line 114: Wouldn't this panic if there was a non-instance enum function? Something that does not take `Self` as first parameter but something completely different?

sdroege:
Looks ok to me otherwise. So
  * [ ] The comments above
  * [ ] Adding support for flags (which should be exactly the same)
  * [ ] Moving the quote stuff into a separate PR
  * [ ] Removing the commits from https://github.com/gtk-rs/gir/pull/994 after that one is merged

Doesn't seem like too much missing here :)

src/codegen/enums.rs
MarijnS95: > Just like to_string(), this should probably panic if it's NULL

Like the implicit panic through an assertion in `from_glib_none` -> `CStr::from_ptr`?

> unless the return value is marked as nullable in the introspection data

Hmm:

https://github.com/gtk-rs/gir/blob/062b3d649ccc09c17564d6bc429dc682829d71f0/src/analysis/special_functions.rs#L45-L48

As mentioned before this is a hacky approach, perhaps I should pass down the surrounding type to `special_functions::extract` and make a `StaticToString` (activated on flags/enum) to generate `impl Display` that leverages `.to_str()` instead. Still like `.to_str()` itself to show up in the original `impl` block though.

> It will require case by case specific manual implementations.

Well, it was the goal to generate this since all implementations thus far are identical. Should I abandon this effort instead?

src/analysis/enums.rs
MarijnS95: Agreed. I didn't add too much upfront checking in this proof of concept, all `to_string`s (in gstreamer) have that.

src/analysis/enums.rs
MarijnS95: > We have to rename the function because otherwise it conflicts with the owned ToString::to_string() and that's not so great :)

That is exactly the issue I highlighted, creating infinite recursion and such :D

> I would automatically rename it without configuration. That's not very different from the automatic renaming of e.g. type to type_ because it's a keyword.

What do you mean with "without configuration"? The configuration parsed from `Gir.toml`? I don't think the commented out `.status =` here makes any sense, though we should at least adhere to it when it's not set to `Generate`.

src/analysis/enums.rs
MarijnS95: If it takes something different than `Self` it should not reach past `if first_param.typ == enumeration_tid {` correct?

src/codegen/record.rs
MarijnS95: Oh, pass down `analysis.version` :grimacing: , good catch!

src/codegen/object.rs
MarijnS95: Same, `analysis.version` :)

src/analysis/enums.rs
MarijnS95: `InfoBase` contains an `imports` field that "should not" be used for `enums` as they share a single import block. This is initialized to some useless default and should rather be `None` but that pollutes code elsewhere.

However, the current way of passing down `env.analysis.enum_imports` in `analysis/mod.rs` is far from ideal. I can remove that in favour of using this `InfoBase::imports` field for every enum, and merge them (including filtering on `Imports::derives`) in codegen.

MarijnS95:
> > There's a temp commit here to make sure the CI builds on the fixed gtk-rs branch. Once this gets closer I guess we'll merge gtk-rs, bump the commit off and make sure it succeeds one final time before merging?
> 
> What's that commit and why? :)

That [commit](https://github.com/gtk-rs/gir/commit/cdb578bb2159907c20cf10c1f74e7b49dc5aa7ed) switched the CI over to use my enum functions gtk-rs branch so that we can see the whole thing working and make sure the CI succeeds :)

> > I pushed a commit on top that replaces this block with a `quote` macro to show what it'd look like. It should probably not be part of this PR but I don't think it's particularly bad to restart the discussion of switching to this crate.
> 
> Let's indeed do that as a separate PR afterwards. I agree that the string manipulations without quote are annoying. Last time I tried moving to quote the big problem was that it was rather annoying to a) do comments, b) format code in a reasonable way (e.g. quote didn't output empty lines between any items IIRC), c) even rustfmt was not converting the code into something reasonable looking.

I'm glad to hear quote gets a second chance! The only issues with `rustfmt` are, as highlighted: not being able to format inside comments (duh) and inside macro calls, notably (actually, limited to) `#[cfg]` annotations inside `bitflags! {}`. It's sad but seems like we have to write quite a bit of `Punct::new(..., Spacing::Joint)` to clean those up. Same for commented out functions. I wonder if we can do a pre-formatting pass on these codeblocks before stringifying them into their respective `//` / `bitflags! {}` cages instead.

Guess it only makes sense to, for a first approach, convert (or implement new) tiny bits of `gir` in `quote` as long as they have no or little effect on the output. Over time this can be scaled up, but that's not a hard requirement at all :)

> If that can be solved I'm fine with moving to quote. But as said, different PR. Not too many things at once :)

Of course, pushing it here was merely to showcase :D

src/codegen/enums.rs
sdroege: > Like the implicit panic through an assertion in from_glib_none -> CStr::from_ptr?

For example

> Well, it was the goal to generate this since all implementations thus far are identical. Should I abandon this effort instead?

You mean the example in GStreamer that handles `Unknown` as `"UNKNOWN"`? That's not a general GLib pattern so I don't think we should automatically do that.

src/analysis/enums.rs
sdroege: Either do that, or make the comment a bit more descriptive :)

src/codegen/object.rs
sdroege: See my reply there then :)

src/analysis/enums.rs
sdroege: Indeed, sorry :)

src/analysis/enums.rs
sdroege: It's possible to rename functions in the Gir.toml. We shouldn't require the user to rename `to_string()` to `to_str()` in these cases though but automatically do that anyway, unless there actually is such a configuration (then we should follow it of course).

sdroege:
> I'm glad to hear quote gets a second chance! The only issues with `rustfmt` are, as highlighted: not being able to format inside comments (duh) and inside macro calls, notably (actually, limited to) `#[cfg]` annotations inside `bitflags! {}`. It's sad but seems like we have to write quite a bit of `Punct::new(..., Spacing::Joint)` to clean those up. Same for commented out functions. I wonder if we can do a pre-formatting pass on these codeblocks before stringifying them into their respective `//` / `bitflags! {}` cages instead.

I think that can make sense. AFAIU you can call rustfmt also as a library so that might be worth looking into.

> Guess it only makes sense to, for a first approach, convert (or implement new) tiny bits of `gir` in `quote` as long as they have no or little effect on the output. Over time this can be scaled up, but that's not a hard requirement at all :)

Agreed. But now let's first try to get this PR done and not put too much other noise in here :)

src/codegen/object.rs
MarijnS95: I see no reply there?

"There" being https://github.com/gtk-rs/gir/pull/991#discussion_r526656146, not https://github.com/gtk-rs/gir/pull/991#discussion_r527488987 (I replied bottom-to-top, hence the confusion. Sorry :grimacing:)

This has anyway been fixed in the last push: https://github.com/gtk-rs/gir/compare/37af52f724ea03a12bdc6127f027333588af308e..165999fdd25bc572606975821b0fdb19a9838cdc#diff-fae3a30287d03db00601eac005d17f1a1f41a1a9d94dbef815cc26c685e037a0L85-R85

sdroege: Oh, sorry for the confusion. Please resolve discussions that are resolved :)

src/analysis/enums.rs
MarijnS95: Would you mind checking out the situation in `analysis/mod.rs` and recommend either? I'm leaning towards using `InfoBase::imports` so that `enum_imports` isn't helplessly passed around, and `imports` doesn't end up unused for enums without proper indication.

src/codegen/object.rs
MarijnS95: Fair, perhaps now is the time to ask how you want to deal with review threads: should I mark them as resolved after fixing or notify you so that you can take another look and approve?

I personally prefer the latter (when reviewing) so that I don't have to expand all resolved side-by-side to a push-diff and doublecheck, even though the change at hand is minor enough.

sdroege: If it's something minor, just resolve it. If it's something where you're not entirely sure, keep it open :)

src/analysis/enums.rs
sdroege: Yes that seems better

src/codegen/enums.rs
MarijnS95: > You mean the example in GStreamer that handles Unknown as "UNKNOWN"? That's not a general GLib pattern so I don't think we should automatically do that.

The entire `to_str` function: name, dealing with invalid values upfront that might cause the C function to return `NULL`, and returning a static lifetime borrow.

If this is non-standard I can revert most of the code. That means the original `to_string(&self) -> GString` functions are going to pop back up. I don't really know what direction we want to go with these functions.

src/analysis/enums.rs
MarijnS95: Okay, so:

- Keep `to_string.name = "to_str";` (in the most recent push);
- Except if overridden (or disabled) by `config`;
- `impl Display`, which leverages this function, should follow suit as well (ie. call the renamed function, or not be generated at all if the config sets it to `ignore`. We can still use it if set to `manual`).

src/analysis/enums.rs
MarijnS95: Oh, it looks like `function` doesn't use `InfoBase` in its `Info` struct either, perhaps in part because of that?

src/codegen/enums.rs
sdroege: `to_str()` for cases that are not nullable and transfer none is standard. `to_string()` for not nullable and transfer full is also standard. That should cover 99% of the cases and can be autogenerated.

nullable needs special handling per case, but we can create a `to_str()` / `to_string()` function that returns an `Option` I guess but that also means that it can't be used for automatically implementing `Display` / `ToString` (which is fine). Autogenerating that should be good, and if something better can be done it can be ignored in `Gir.toml` and a manual implementation can be done.

Also non-standard is the weird GStreamer case where one of the enum variants always returns `NULL` and that definitely needs manual handling.

src/analysis/enums.rs
sdroege: Correct, and also https://github.com/gtk-rs/gir/pull/991#discussion_r527540323

src/analysis/enums.rs
MarijnS95: Pushed a quick proof of concept, merging import dicts is ugly and I haven't even gotten into the conditions. For that I should probably do away with `.extend` and instead call the `add` functions. Turns out I like the original approach after all, despite passing `enum_functions` around which can be fixed by moving enumeration (and later bitflags) handling out of `fn analyze` like `analyze_constants`/`analyze_global_functions`.

src/analysis/enums.rs
MarijnS95: All three points seem to be mostly working. Couple more questions:

-  What are the semantics behind `manual = true`? Don't generate the function, but expect that the function is available under the same signature as if it was generated (so that other code like an automated `Display` impl can still be generated), as opposed to `ignore = true`?
- Renames are already applied when running `.find(|f| f.name == "to_string")` in analysis. Should I put more effort in, ie. ` f.glib_name.ends_with("_to_string")` (or check the full string with snake_case enum name prepended), or is there a better option? Seems like we still want to apply special behaviour when encountering `to_string` on enum types?
- Shall I move this part of the code into `special_functions::extract` so that it can be more generally applied in the future? For now that means passing in the parent type (enum) to not enable "static to_string" handling on any other functions.

src/codegen/enums.rs
MarijnS95: All functions handled by this bit of code have `transfer none` return type and are nullable. They all reside in gstreamer, gtk doesn't seem to have any.  Bitfields however have only one such function (in `GLSLProfile`), the other `to_string`s are all transfer full, not nullable and get the default `to_string -> GString` implementation (`GLAPI` and `GLPlatform`).

> nullable needs special handling per case, [...]

Given the above distribution of functions in gstreamer, what do you want to do exactly? We only have non-nullable variants in bitflags, and you seem to suggest to turn all the others into `Option<&'static str>` without `Display`?

Upfront warning: In the current push all enum/bitfield instance parameters are passed by value. This means `fn to_string` (for the transfer full case) is not found by the autogenerated `f.write_str(&self.to_string())` in `Display` resulting in aforementioned infinite recursion (caught by clippy at least). Perhaps we should rename _all_ `to_string` special functions to `to_str` and be done with this possibility forever?

> Also non-standard is the weird GStreamer case where one of the enum variants always returns NULL and that definitely needs manual handling.

You mean the `return_if_false` case that gets compiled out on release? In debug calling `to_string(unknown)` returns `NULL` while in release it returns `"unknown"`.

src/analysis/enums.rs
MarijnS95: Removed it again and moved some of the items from `InfoBase` (don't need nearly all of it) to `Info`. Quite a bit cleaner now.

MarijnS95:
Let's update and extend the list:

- [ ]  The comments above
  - [ ] `to_string` seems quite a large task still:
    - [ ] When do we use it?
    - [ ] How do we generate generic (error-handling) code for it?
    - [ ] Fix recursion because of `fn to_string(self)`
    - [ ] Fix that one remaining unused `use glib::GString` in `webrtc/enums` (due to `&'static str` replacement)
- [x]  Adding support for flags (which should be exactly the same)
  - [ ] Check if we can reduce code duplication
- [x]  Moving the quote stuff into a separate PR
  Removed from this PR, separate PR comes when rebasing doesn't conflict with these changes anymore :wink: 
- [x]  Removing the commits from #994 after that one is merged
- [ ] Autogenerate `from_string`, which is pretty straightforward
- [ ] Do not import `std::fmt` when `generate_display_trait` is true but nothing has `to_string` (only problematic in gtk)

I'll rebase this branch on master as soon as edition2018 drops on gstreamer-rs :)

src/analysis/imports.rs
```diff
@@ -126,13 +125,36 @@ impl Imports {
...
+        // Always store the name without crate:: prefix
+        let name = if let Some(name_without_crate) = name.strip_prefix("crate::") {
+            // Remove crate::<name>:
+            self.map.remove(name);
+            name_without_crate
+        } else {
+            // Take extra care to remove an eventual use crate::<name>; as well:
+            self.map.remove(&format!("crate::{}", name));
+            name
+        };
```
MarijnS95, line 151: Welp edition 2018 `crate::` handling takes quite a bit more effort than expected. I don't want to make the assumption that one or the other format stored in here, but can we perhaps make it as such?

src/analysis/enums.rs
sdroege: * `manual = true` is like `ignore = true` but for documentation and other purposes (I think currently only for property deduplication if there is also a manual getter/setter function) it is assumed the function actually exists.
* For enums/flags there would still be special behaviour in case of `transfer none`, yes. That renaming should probably happen for enums/flags directly during analysis or even in pre-process before the analysis.
* Let's refactor it to a separate function once there are more users of this :)

src/analysis/imports.rs
sdroege: The problem is that currently names are stored in both variants? That's annoying and should ideally be cleaned up to only use the 2018 variant...

That should also clean up such handling elsewhere.

src/codegen/enums.rs
sdroege: > Given the above distribution of functions in gstreamer, what do you want to do exactly? We only have non-nullable variants in bitflags, and you seem to suggest to turn all the others into Option<&'static str> without Display?

All others are `nullable`, really? I would've assumed the `nullable` case to be very rare... but yes, for `nullable` we can't really autogenerate a sensible `Display` impl as it's not clear what the `NULL` case should represent.

> Perhaps we should rename all to_string special functions to to_str and be done with this possibility forever?

Or use the full function call syntax for calling it. `Type::fmt(self, f)` vs `Type::fmt(&self, f)`. That seems safer anyway and we should probably also do that in more places, also for `<Type as Trait>::function()` things.

> You mean the return_if_false case that gets compiled out on release? In debug calling to_string(unknown) returns NULL while in release it returns "unknown".

`g_return_val_if_fail()` cases are to be transformed (manually...) into panics on the Rust side and don't have to be considered for the signature, for example. We also assume that they're not compiled out for bindings but always trigger (compiling them out is not the default for releases and you have to explicitly do that, and then you'll have a lot of fun if you're not careful).

src/analysis/enums.rs
MarijnS95: Pushed a commit yesterday that renames _all_ `to_string` to `to_str`, to err on the safe side of not creating recursion with the automatically-implemented `ToString` trait. (For now that's only the case when moving `self` in flags and enums, not sure if we want to pollute the rest with this change?).

Do we still need `to_str(self) -> &'static str` to be `pub`? We've seen that formatting - which is pretty much the only use for stringifying these - is implemented pretty efficiently.

> For enums/flags there would still be special behaviour in case of transfer none, yes

I take that as an "ack" to the current approach. Can we make a more concrete plan as to what should happen in this PR? It's kinda meta now and it looks like the code and our discussion here are on the same line. But let's see the comment below on the same topic as well...

sdroege: > Do we still need to_str(self) -> &'static str to be pub? We've seen that formatting - which is pretty much the only use for stringifying these - is implemented pretty efficiently.

Yes, check the usage of those functions in GStreamer. There are many places where a `&str` is used as a `&dyn ToValue`. By going via `Display` there it would be one more annoying allocation.

> I take that as an "ack" to the current approach. Can we make a more concrete plan as to what should happen in this PR?

Yes, what you currently implemented from what I understand (I didn't check the latest code). For everything but enums/flags it would be as before in the generated code, for enums/flags with `transfer none` and non-`nullable` return value there would be the `&'static str` special case.

Note that we ignore the nullable annotations in gir currently for return values. See https://github.com/gtk-rs/gir/pull/970 and https://github.com/gtk-rs/gtk-rs/issues/163 . For GStreamer core this can also be enabled now, for gst-plugins-base not yet. So you will see various functions in gir that claim to be nullable but are actually not. For now you can fix that in the Gir.toml by configuring the return value explicitly to be `nullable = false`.

src/codegen/enums.rs
MarijnS95: > All others are `nullable`, really? I would've assumed the `nullable` case to be very rare... but yes, for `nullable` we can't really autogenerate a sensible `Display` impl as it's not clear what the `NULL` case should represent.

Yup, every `transfer none` on flags and enums is nullable currently. I guess we can assume either a panic or `"NONE"`/`"UNKNOWN"` here as is done currently in the code?

> Or use the full function call syntax for calling it. `Type::fmt(self, f)` vs `Type::fmt(&self, f)`. That seems safer anyway and we should probably also do that in more places, also for `<Type as Trait>::function()` things.

I guess that's a bit hard to do inside the `fn fmt` `impl Display for ...` itself, isn't it?

> `g_return_val_if_fail()` cases are to be transformed (manually...) into panics on the Rust side and don't have to be considered for the signature, for example. We also assume that they're not compiled out for bindings but always trigger (compiling them out is not the default for releases and you have to explicitly do that, and then you'll have a lot of fun if you're not careful).

That looks to be the only reason for `NULL` together with out-of-bounds values. Basically the code backup I have now to `return "UNKNOWN"` on `matches!(self, Unknown, __Unknown(_))` and on a return value of `NULL` should both panic?

src/analysis/imports.rs
MarijnS95: I have to investigate this more carefully but it is quite possible that both variants are used intertwined, depending on where they are called from.

After all, needing an external symbol from outside the file is going to be imported with `use crate::MySymbol;`, but when defined it's going to be defined and used as `MySymbol` (though, supposedly, also reachable through `crate::MySymbol` even though that's a confusing).

I guess it is much easier for me to solve this by simply calling `add_defined` with a `crate::` prefixed variant after which everything should be working fine.

src/codegen/enums.rs
sdroege: > Yup, every transfer none on flags and enums is nullable currently. I guess we can assume either a panic or "NONE"/"UNKNOWN" here as is done currently in the code?

Is it really `nullable` or only because the annotation is ignored? See my other comment.

Assertions/`g_return_val_if_fail()` are not considered `nullable`, and the bound checks also not as they're impossible in proper languages unlike C. You can ignore these cases and if a function is marked `nullable` just because of these cases then that should be fixed.

src/analysis/enums.rs
MarijnS95: > Yes, check the usage of those functions in GStreamer. There are many places where a `&str` is used as a `&dyn ToValue`. By going via `Display` there it would be one more annoying allocation.

Right, there's one `format!("{}", ...)` case where I removed the (now unnecessary) `to_str` call which is implicit through `Display::fmt`, the others need a string conversion for `ToValue` to work :+1: 

> Yes, what you currently implemented from what I understand (I didn't check the latest code). For everything but enums/flags it would be as before in the generated code, for enums/flags with `transfer none` and non-`nullable` return value there would be the `&'static str` special case.

They are all "nullable", but do read on:

> Note that we ignore the nullable annotations in gir currently for return values. See #970 and [gtk-rs/gtk-rs#163](https://github.com/gtk-rs/gtk-rs/issues/163) . For GStreamer core this can also be enabled now, for gst-plugins-base not yet. So you will see various functions in gir that claim to be nullable but are actually not. For now you can fix that in the Gir.toml by configuring the return value explicitly to be `nullable = false`.

I think I have pointed out that `nullable = false;` code before in `special_functions.rs`. I printed the value of `nullable` before calling that, but looking it up in `.gir` now (for example `gst_audio_format_to_string`) shows no `nullable="true"` in xml :thinking:

sdroege: > I think I have pointed out that nullable = false; code before in special_functions.rs. I printed the value of nullable before calling that, but looking it up in .gir now (for example gst_audio_format_to_string) shows no nullable="true" in xml thinking

Yeah that function is not `nullable` and that's correct. So I guess our main problem here is the ignoring of the annotation? :)

MarijnS95: I might have missed the point, but I think I see what you mean now:

- `.gir` only ever has `nullable="1"`, implying that leaving it out equals to not-nullable;
- Since the mentioned PRs those are treated as nullable as well, unless:
  - `trust_return_value_nullability` is set;
  - That particular function is marked as non-nullable in `Gir.toml`;

So there is no ignored annotation, this properly leads to the `ToString`s being nullable.

Except for one little override that we should (IMO) stick to for the static implementation as well:

https://github.com/gtk-rs/gir/blob/918f87cafa876b0ce3b593f98719763295f6c22b/src/analysis/special_functions.rs#L45-L48

sdroege: > do you intend that before or after this override?

I would put `nullable = false` overrides into `Gir.toml` for now for these functions when they are not `nullable` until we can move to `trust_return_value_nullability = true`.

Does that answer your question?

Is there any `to_string()` function that is really nullable currently?

MarijnS95: > I would put `nullable = false` overrides into `Gir.toml` for now for these functions when they are not `nullable` until we can move to `trust_return_value_nullability = true`.
> 
> Does that answer your question?

Yeah, makes sense given the below:
 
> Is there any `to_string()` function that is really nullable currently?

Why don't we throw some fun XPath at that?

```
xmlstarlet sel -t -v "//_:function[@name=\"to_string\" and _:return-value[@nullable=1]]/@c:identifier" gir-files/*.gir
gst_glsl_profile_to_string
gst_glsl_version_to_string
```

According to the documentation returns the `the name for @profile (/@version) or %NULL on error`.

So we can't - like the current code in `special_functions.rs` asumes - treat all `to_string` as non-nullable.

MarijnS95: So, now that we established the fact that not all `to_string`s can be assumed to be non-nullable, I need to add a check and manually set `nullable = false` for the specific functions that we want to accept (until that trust property can land). Then, which of the checks should I drop from the alternate `to_str` static function?

```rust
if matches!(self, Self::Unknown | Self::__Unknown(_)) {
    // Should be a panic or not?
    return "UNKNOWN";
}
unsafe {
    ffi::gst_video_field_order_to_string(self.to_glib())
        .as_ref()
        .map(|ptr| CStr::from_ptr(ptr).to_str().unwrap())
        // Should be .expect("{type_name}::to_string() should never return NULL!")?
        .unwrap_or("INVALID")
}
```

sdroege: Thanks for the research, that all makes sense :)

> Then, which of the checks should I drop from the alternate to_str static function?

* `Self::Unknown`: This is a special case of those one or two GStreamer functions and should be handled in a manual implementation for those. No point in having this in gir
* `Self::__Unknown(_)`: Pass it through to the function. It can only correctly come from C, or the user used non-public API wrongly and it would panic afterwards then
* `.unwrap_or("INVALID")`: Make this an `expect()`

MarijnS95: > * `Self::Unknown`: This is a special case of those one or two GStreamer functions and should be handled in a manual implementation for those. No point in having this in gir

That means `Unknown` might return `"UNKNOWN"` or panic when `g_return_val_if_fail` is compiled in...

> * `Self::__Unknown(_)`: Pass it through to the function. It can only correctly come from C, or the user used non-public API wrongly and it would panic afterwards then

Right, I think I advocated for this one before when Rust and GStreamer could possibly be out of sync. 

> * `.unwrap_or("INVALID")`: Make this an `expect()`

Same for `.unwrap_or("NONE")` in `flags.rs` I guess?

sdroege: > That means Unknown might return "UNKNOWN" or panic when g_return_val_if_fail is compiled in...

Well, it means that we have to implement this function manually in GStreamer because it has so strange behaviour :)

> Same for .unwrap_or("NONE") in flags.rs I guess?

Yes. All these are special cases in GStreamer that should just get a manual implementation

MarijnS95: > Well, it means that we have to implement this function manually in GStreamer because it has so strange behaviour :)

Yeah, I totally quoted the part where you said to implement it manually, replying that the autogenerator would mess this up :facepalm: 

The 3 affected functions with an `unknown` field have been set to `manual = true` in gstreamer. Pushed another hacky implementation here that unifies the `to_str` function between flags and bitfields again, which I think is close to final form. The only problem: this "needs" another `impl Type {}` block right below. I'd rather generate it inside the one and only autogenerated one, ie. by calling a special function from `codegen/{flags,enums}.rs` or stuffing this path into `function::generate` (eww). Any suggestions?

sdroege: You could have a generic function for generating those functions that assumes to be called from inside the `impl Type { ... }` block. But having two `impl Type { ... }` blocks is also not really a problem IMHO. Whatever is easier really :)

MarijnS95: The most annoying thing is writing extra code that checks if the impl block is supposed to be empty... And if it has just one function that is this `to_str` it's empty too. That's just ugly. Even then the blocks are currently emitted right below eachother.

Moved the function into `src/codegen/function.rs` and assume it is called inside the impl again. There's just that `if "to_str"` remaining in enums and flags codegen now.

sdroege: Makes sense.

Also let me know when I should give the whole thing some time for another review :)

MarijnS95: Please, it's time to resolve all conversations and give this a fresh review.

Let's see if we can work out the final few points in the list above, perhaps generalize `if func_analysis.name == "to_str" && has_static_tostr` and then this should, in theory, be almost ready :)

sdroege:
>  If there is an unknown value in a returned `enum (fn from_*(xxx) -> EnumType)`, turn that into a `Result<..., BoolError>;`

What is this one about?

src/analysis/enums.rs
```diff
@@ -0,0 +1,143 @@
...
+pub struct Info {
+    pub full_name: String,
+    pub type_id: library::TypeId,
+    pub name: String,
+    pub functions: Vec<functions::Info>,
+    pub specials: special_functions::Infos,
+}
+
+impl Info {
+    //TODO: add test in tests/ for panic
```
sdroege, line 22: That seems unneeded

src/analysis/flags.rs
```diff
@@ -0,0 +1,137 @@
...
+pub struct Info {
+    pub full_name: String,
+    pub type_id: library::TypeId,
+    pub name: String,
+    pub functions: Vec<functions::Info>,
+    pub specials: special_functions::Infos,
+}
+
+impl Info {
+    //TODO: add test in tests/ for panic
```
sdroege, line 22: And this also

src/analysis/enums.rs
```diff
@@ -0,0 +1,143 @@
...
+                    t.transformation_type
+                );
+            }
+        }
+    }
+
+    // Enum to_string functions provide static strings and need special error handling.
+    if let Some(to_string) = functions
+        .iter_mut()
+        .find(|f| f.name == "to_string" && f.status.need_generate())
```
sdroege, line 120: One thing we might want to consider here also is `get_name()` functions with the correct signature. That seems to be a common pattern too for enums (and flags? not sure)

src/analysis/enums.rs
```diff
@@ -0,0 +1,143 @@
...
+    {
+        if let Some(ret_param) = &to_string.ret.parameter {
```
sdroege, line 122: Maybe we also want to check if:
 - the function has no other arguments
 - the function actually returns a string

And otherwise handle it not as a "special" function

src/analysis/special_functions.rs
```diff
@@ -44,6 +46,11 @@ fn update_func(func: &mut FuncInfo, type_: Type) -> bool {
     }
     // I assume `to_string` functions never return `NULL`
     if type_ == Type::ToString {
+        // Hack to make sure the names never, ever clash.
+        // TODO: Should perhaps merge ToStr and ToString, then store in
+        // specials whether this implementation returns pointers to 'static
+        // strings (which is possible but not guaranteed with `transfer none`)
```
sdroege, line 52: That sounds like a good idea. That way we could also store the actual function name (`get_name()` / `to_string()` and whatever else we might find at later time).

src/codegen/function.rs
```diff
@@ -456,3 +456,36 @@ pub fn body_chunk_futures(
...
+    version_condition(w, env, function.version, false, 1)?;
+
+    let vis = match function.visibility {
+        Visibility::Public => "pub ",
+        _ => "",
+    };
+
+    writeln!(w, "\t{}fn {}(self) -> &'static str {{", vis, function.name)?;
+
+    // TODO: Can we somehow reuse function_body_chunk::Builder here?
```
sdroege, line 475: If you feel bored, probably. We ideally want to get rid of all that code generation abstraction as it makes things more complicated than necessary (thus also the experiment with quote)

sdroege:
Looks almost good to go to me

src/analysis/enums.rs
MarijnS95: Copypaste from record.rs, will remove.

MarijnS95:
> > If there is an unknown value in a returned `enum (fn from_*(xxx) -> EnumType)`, turn that into a `Result<..., BoolError>;`
> 
> What is this one about?

There are 3 `from_iso` functions that used to return `Result<..., glib::BoolError>` on `__Unknown(_)` (_not_ `Unknown`), but this is not dealt with by autogeneration.

Since `__Unknown` seems to be standard, whereas `Unkown` is not, perhaps this is worth taking into the autogenerator?
Not sure what the preconditions are though: Return type must be `Self`, `from_` in the name, takes a single argument...

src/codegen/function.rs
MarijnS95: Not really, unfortunately.

`quote` might be a good starting point to refactor certain chunks of the code. Not going to be me though, I'd love to introduce it in smaller places to show how it simplifies things, but a full-on refactor is something for someone more invested in `gir` ;)

src/analysis/special_functions.rs
MarijnS95: It would beak `impl FromStr for Type` though, but then again we should probably insert such info in the `specials` `Map`.

Would make most of the code more generic though, which is my main goal here. Let's see what I can come up with.

src/codegen/function.rs
sdroege: If you feel like it, submit a PR with some minor parts moved over to quote after this big one here is in :)

src/analysis/enums.rs
MarijnS95:
```
xmlstarlet sel -t -v "//_:bitfield/_:function[contains(@name, \"get_name\")]/@c:identifier" gir-files/*.gir
gst_message_type_get_name
gst_stream_type_get_name
```
On bitfield, and these on enumerations:

```
xmlstarlet sel -t -v "//_:enumeration/_:function[contains(@name, \"get_name\")]/@c:identifier" gir-files/*.gir
gst_debug_level_get_name
gst_event_type_get_name
gst_format_get_name
gst_pad_mode_get_name
gst_query_type_get_name
gst_state_change_get_namegst_install_plugins_return_get_namegst_player_color_balance_type_get_name
gst_player_error_get_name
gst_player_state_get_name
```
(not sure what xmlstarlet is doing to the newlines...)

Afaik they're all marked nullable though...

sdroege:
> There are 3 `from_iso` functions that used to return `Result<..., glib::BoolError>` on `__Unknown(_)` (_not_ `Unknown`), but this is not dealt with by autogeneration.

Those can go away, I'm not sure what I thought when adding them.

src/codegen/function.rs
MarijnS95: That has always been the plan since removing it... But I converted parts of this PR that have massively changed since then :grimacing:

src/analysis/enums.rs
sdroege: None of those should be `nullable` (in the annotations and .gir file), but yes it would need overrides in `Gir.toml` for now.

MarijnS95: `get_name` support is added and refactored quite a bit, but someone invested in the C code should check and add `nullable = false` in `Gir.toml` for every affected function. I've disabled the aforementioned `ret.nullable = false` assumption in `special_functions` for enumerations/bitfields.

src/analysis/enums.rs
MarijnS95: Done that for all of `to_string` and `get_name`,

src/analysis/special_functions.rs
MarijnS95: Implemented like that. Perhaps we should rename `ToString` to `Display` so that it's clear this is about the trait, that can be implemented by either `to_string` (renamed to `to_str` to prevent `ToString::to_string` clashes) or `get_name`.

src/analysis/enums.rs
MarijnS95: Marking this as resolved. I've pushed more proof-of-concept code to clean up the implementation, but it's messy because of mixed assumptions on nullability. This thread dragged on for too long, let's spawn new review threads on fresh lines.

src/analysis/special_functions.rs
sdroege: Yes naming it `Display` would make a lot more sense. Feel free to rename that

src/analysis/enums.rs
sdroege: You mean now everything is simply goes through the normal nullable code? That's good, this special casing was a bit problematic :)

If you give me a list of functions where you're unsure, I can tell you if they're nullable or not, or if they need special handling (-> manual implementation). Would that help?

sdroege:
What's the status from your side now? Basically ready apart from updating gtk-rs/gstreamer-rs completely and in need of another review, or do you have some open issues currently?

src/analysis/enums.rs
MarijnS95: > You mean now everything is simply goes through the normal nullable code? That's good, this special casing was a bit problematic :)

Nope, just enumerations and bitfields; with prior code `to_string` on `GLSLVersion`/`GLSLProfile` would otherwise be generated as non-nullable despite containing explicit `nullable="1"` in GstGL.gir. `GLAPI`/`GLPlatform` return non-NULL and have now gotten an explicit override as well.

> If you give me a list of functions where you're unsure, I can tell you if they're nullable or not, or if they need special handling (-> manual implementation). Would that help?

The aforementioned `_get_name` functions extracted with `xmlstarlet`. I'm not unsure, it's just a time-consuming process to find and/or override `[object.function.return] nullable = false`. Perhaps we should wait until `trust_return_value_nullability` can be enabled? Otherwise, here's the full list, including everything outside of flags/enums as well. Note that it only helps if you set up a commit adding/overriding `nullable = false` in Gir directly :grimacing: 

```console
test@gstreamer-rs $ xmlstarlet sel -t -v "//_:function[contains(@name, \"get_name\")]/@c:identifier" gir-files/*.gir
gst_debug_level_get_name
gst_element_state_change_return_get_name
gst_element_state_get_name
gst_event_type_get_name
gst_format_get_name
gst_message_type_get_name
gst_pad_link_get_name
gst_pad_mode_get_name
gst_query_type_get_name
gst_state_change_get_name
gst_stream_type_get_name
gst_debug_level_get_name
gst_event_type_get_name
gst_flow_get_name
gst_format_get_name
gst_message_type_get_name
gst_pad_mode_get_name
gst_query_type_get_name
gst_state_change_get_name
gst_stream_type_get_namegst_install_plugins_return_get_name
gst_install_plugins_return_get_namegst_player_color_balance_type_get_name
gst_player_error_get_name
gst_player_state_get_name
gst_player_color_balance_type_get_name
gst_player_error_get_name
gst_player_state_get_name
```

MarijnS95:
@sdroege This is developed on gstreamer-rs which I try to keep in sync (the PR), and gtk-rs is checked and updated every so often.

This code is ready from my side. The only thing I'm not completely okay with is the `generate_static_to_str` call and surrounding `if` in enums and flags, that could be more generic. As well as the bit of code adding the `CStr` import for it in `special_functions`, that doesn't belong there. Suggestions are welcome :)

Other than that all boxes above have been checked except reducing code duplication (done for the most part) and (nice to have) check if there are no C calls in manual code that are better delegated to these newly generated enum functions (see what I did to `from_string`).

src/analysis/enums.rs
sdroege: > Nope, just enumerations and bitfields; with prior code to_string on GLSLVersion/GLSLProfile would otherwise be generated as non-nullable despite containing explicit nullable="1" in GstGL.gir.

Yeah those are indeed properly nullable. Annoying :)

> Otherwise, here's the full list, including everything outside of flags/enums as well. Note that it only helps if you set up a commit adding/overriding nullable = false in Gir directly 

If you put that list into the GStreamer MR as a discussion (so we don't forget), I'll add overrides (as needed) for all of them to the `Gir.toml`

sdroege:
> Found one problem: `check_init_asserts` wants to see the assertions in `from_str`, even though these are delegated to `from_string` for exactly the cleanup described in the previous paragraph. How do we want to solve that, just add it regardless (while still benefiting from glib translation provided by the autogenerated function)?

Add the assertion to the lowest function that is calling the C code, and all callers can have `skip_init_assert!` instead

I'll try give this another round of review later

src/codegen/function.rs
```diff
@@ -456,3 +459,41 @@ pub fn body_chunk_futures(
...
+
+    let visibility = match function.visibility {
+        Visibility::Public => "pub ",
+        _ => "",
+    };
+
+    writeln!(
+        w,
+        "\
+\t{visibility}fn {rust_fn_name}(self) -> &'static str {{
```
GuillaumeGomez, line 481: The `\t` everywhere are making the reading super complicated... We'll need to improve that later on.

src/codegen/function.rs
```diff
@@ -456,3 +459,41 @@ pub fn body_chunk_futures(
...
+\t\tunsafe {{
+\t\t\tCStr::from_ptr(
+\t\t\t\t{ns}::{glib_fn_name}(self.to_glib())
+\t\t\t\t\t.as_ref()
+\t\t\t\t\t.expect(\"{glib_fn_name} returned NULL\"),
+\t\t\t)
+\t\t\t.to_str()
+\t\t\t.unwrap()
```
GuillaumeGomez, line 489: Please never use `unwrap()`, always `expect()`.

GuillaumeGomez (changes requested):
The `.unwrap()` needs to be replaced but the rest seems good. From just the code, it seems pretty obvious that we missed things but we'll fixed them when they'll appear. Thanks for working on this!

src/codegen/function.rs
MarijnS95: @GuillaumeGomez I've simply adopted the standards used elsewhere in `gir`, but this is one of the larger predefined codeblocks making it more pronounced. This is why I've been advocating for `quote`, for which I have a PR in the works (submitted after this one) that replaces most of the code blocks in this PR.

src/codegen/function.rs
MarijnS95: Here I've adopted what [was already there](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/blob/8b9ef8b10961ba2a9dad933c79219423253247ef/gstreamer-audio/src/audio_format.rs#L127-129) in manual code, will update both.

src/codegen/function.rs
sdroege: I think it's fine as is. It's bad but consistent with code elsewhere.

Once this is in we should incrementally start moving to a better code generation mechanism.

src/codegen/function.rs
GuillaumeGomez: Yes please! :)

MarijnS95:
> From just the code, it seems pretty obvious that we missed things but we'll fixed them when they'll appear.

@GuillaumeGomez  Anything in particular? From my POV there are 2 eyesores that could use some feedback:

- `generate_static_to_str` could perhaps be called from `function::generate`
- `returns_static_ref` should not be determined by `special_functions`. After all this has little to do with the trait impls (not used in `trait_impl.rs`) and only shared between flags and enums. Any suggestions where to deal with this instead? In the end it relies on the same checks `special_functions::update_func` makes for determining if it's a valid function to generate `impl Display` for.

GuillaumeGomez:
Sorry, I should have been more explicit. What I tried to say is that the code changes look good, but considering how big it is, it's very likely that we missed bugs, which is totally expected and not a problem. That's the basic life of code projects. :)

src/codegen/function.rs
MarijnS95: Done :)

src/codegen/function.rs
MarijnS95: @sdroege Yeah I think the only consistent, _readable_ solution is quote :grimacing: 

I'll rebase and refactor most of these blocks when this PR is in :)

MarijnS95:
@GuillaumeGomez Right, this patchset has gotten quite large and I keep sneaking in patches which I don't want to pick to master, fix conflicts and submit them, then rebase this and submit them again.

That said there are 5 generic patches at the beginning of this PR I could submit separately, would that help anything?

MarijnS95:
However, the nice thing with a code generator and a fixed set of inputs is that we, for the time being, can exactly validate that the output it produces is correct. It's quite unlikely this is thrown in the wild and breaking on random occasions, though the generated code might be :grimacing:

src/codegen/function.rs
GuillaumeGomez: `quote` didn't help much iirc. I think the best way is just plain string formatting but in much bigger steps instead of all the small ones we currently have.

MarijnS95: I made quite a case for it in https://github.com/gtk-rs/gir/pull/992#discussion_r524124393, and it's only getting better than that :)

GuillaumeGomez:
Actually it's removing version information from items, which isn't great in my opinion. Explanations! :)

If you put the cfg on the file directly:

```rust
#[cfg(feature = "v3_22)]
mod some_file;
```

Then, you won't see the condition on the items in `some_file`. So for example, if you declare `Foo` in `some_file.rs`, to see that you require the `v3_22` feature to use it, you'll have to go on the parent module page. We're losing information here and I don't think it's a good idea.

MarijnS95:
> Explanations!

It's a bit of a de-sync between three repos, but here goes: https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/630#note_715517

Basically the sight of:

```rust
#[cfg(any(feature = "v1_10", feature = "dox"))]
#[cfg_attr(feature = "dox", doc(cfg(feature = "v1_10")))]
impl StreamType {
    #[cfg(any(feature = "v1_10", feature = "dox"))]
    #[cfg_attr(feature = "dox", doc(cfg(feature = "v1_10")))]
    pub fn get_name(self) -> &'static str {
```
bothered me, so the second-last commit removes it from the function... Then the last commit also removes it from all impls etc if the type itself (and hence the surrounding `mod`) already has it.

But.... As we discussed on IRC, this removes useful information from the docs, even though this is more importantly lacking from structs emitted by `glib_wrapper`.

I've reverted the changes from all three repos. Someone should make an issue that covers the following:
- rustdoc should show `doc(cfg())` constraints on a module on top of all documentation pages generated from that module;
- `doc(cfg())` should be added to `glib_wrapper`;
- Perhaps reduce current `#[cfg]` + `#[cfg_attr(... doc(cfg(...)))]` to just the `doc(cfg())` part, if the surrounding scope (impl, module) already has a `#[cfg]` for that version.

The second and third could perhaps be dropped (and the patches to omit them entirely reinstated) once part 1 is solved.

src/analysis/mod.rs
```diff
@@ -83,12 +92,65 @@ pub fn run(env: &mut Env) {
...
+        }
+    }
+
+    env.analysis.enum_imports = imports;
+}
+
+fn analyze_flags(env: &mut Env) {
+    let mut imports = Imports::new(&env.library);
+    imports.add("glib::translate::*");
+    imports.add(&format!("crate::{}", env.main_sys_crate_name()));
```
sdroege, line 132: This is actually not added to the generated code but filtered out again later, isn't it?

src/analysis/mod.rs
```diff
@@ -83,12 +92,65 @@ pub fn run(env: &mut Env) {
...
     analyze_constants(env);
 
     // Analyze free functions as the last step once all types are analyzed
     analyze_global_functions(env);
 }
 
+fn analyze_enums(env: &mut Env) {
+    let mut imports = Imports::new(&env.library);
+    imports.add("glib::translate::*");
+    imports.add(&format!("crate::{}", env.main_sys_crate_name()));
```
sdroege, line 108: Also this

src/analysis/special_functions.rs
```diff
@@ -28,74 +32,151 @@ impl FromStr for Type {
...
+            return false;
+        }
+
+        if func.name == "to_string" {
+            // Rename to to_str to make sure it doesn't clash with ToString::to_string
+            func.name = "to_str".to_owned();
+
+            // As to not change old code behaviour, assume non-nullability outside
+            // enums and flags only. Function inside enums and flags have been
+            // appropriately marked in Gir.
```
sdroege, line 88: This one should get linked to from https://github.com/gtk-rs/gtk-rs/issues/163

Also can you change it to only do that if the `trust_nullability` configuration is not set? If it is set then we better just trust it as configured by the user :)

sdroege (commented):
Looks good to me otherwise, I'll go through the generated code again some time later today. Probably during the evening.

src/codegen/function.rs
```diff
@@ -456,3 +456,39 @@ pub fn body_chunk_futures(
...
+
+    let visibility = match function.visibility {
+        Visibility::Public => "pub ",
+        _ => "",
+    };
+
+    writeln!(
+        w,
+        "\
+\t{visibility}fn {rust_fn_name}(self) -> &'static str {{
```
sdroege, line 476: See https://github.com/rust-lang/rust/pull/42417

sdroege: I.e. should be `fn foo<'a>(self) -> &'a str` instead

sdroege:
Went through the gstreamer-rs MR and there are a few things to consider for gir in addition to the above.

src/analysis/mod.rs
MarijnS95: You're right, all code was changed for 2018 to use `ffi::something` directly but all these calls were still left in the other analysis/codegen files. I've removed all these `.add` calls, as well as turn the `name == "crate::ffi"` check in `common_checks` into an assertion such that any future mistake like this is more easily caught.

MarijnS95: Note that this also removes the following from gtk-rs:

```diff
diff --git a/glib/src/gobject/auto/binding.rs b/glib/src/gobject/auto/binding.rs
index 1765250fd..d219724b3 100644
--- a/glib/src/gobject/auto/binding.rs
+++ b/glib/src/gobject/auto/binding.rs
@@ -2,7 +2,6 @@
 // from gir-files (https://github.com/gtk-rs/gir-files)
 // DO NOT EDIT
 
-use crate::gobject_ffi;
 use crate::translate::*;
 use crate::BindingFlags;
 use std::fmt;
diff --git a/glib/src/gobject/auto/flags.rs b/glib/src/gobject/auto/flags.rs
index e1fbed625..670c941e8 100644
--- a/glib/src/gobject/auto/flags.rs
+++ b/glib/src/gobject/auto/flags.rs
@@ -2,7 +2,6 @@
 // from gir-files (https://github.com/gtk-rs/gir-files)
 // DO NOT EDIT
 
-use crate::gobject_ffi;
 use crate::translate::*;
 use crate::value::FromValue;
 use crate::value::FromValueOptional;
```

src/analysis/special_functions.rs
MarijnS95: This comment has set up the link right now, is that enough?

I'll add a comment and see if I won't forget linking the exact line with a permalink, but that only makes sense once the commits are in master and are not going to be rebased anymore ;)

MarijnS95: Added `!trust_nullability` to the clause here, good idea but it changes nothing to neither gstreamer nor gtk :)

src/analysis/mod.rs
sdroege: Nice :)

sdroege (approved):

sdroege:
@GuillaumeGomez On you now, this one is "1 change requested" from you :)

sdroege:
Actually pending a decision on https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/630#note_719829 from my side

src/analysis/special_functions.rs
```diff
@@ -28,40 +34,91 @@ impl FromStr for Type {
...
+            }
+        }
+
+        !*ret.nullable
+    } else {
+        false
+    }
+}
+
+fn update_func(func: &mut FuncInfo, type_: Type) -> bool {
```
sdroege, line 98: `type_` is not used anymore :)

MarijnS95: Huh, it's passed into `visibility(type_) below:

```rust
fn update_func(func: &mut FuncInfo, type_: Type) -> bool {
    if func.visibility != Visibility::Comment {
        func.visibility = visibility(type_);
    }
    true
}
```

sdroege: Ah sorry, was confused by the diff :)

sdroege (approved):
Still looks good to me. GStreamer generated code is ready too

GuillaumeGomez (approved):

GuillaumeGomez:
No more `.unwrap()` in the generated code so I'm happy.

PR #992, opened 2020-11-16
Simplify `match` blocks into `?` operator using SSR

MarijnS95:
Extracted from #991

While addressing minor code inconsistency I stumbled upon a commit from Heftig using `rerast` for structural refactoring, and it turns out Rust-analyzer supports SSR as well. This made it possible to easily do a bulk of conversions to use `?` instead of manual `match` blocks. See the relevant commits for the rules used.

One minor deficiency is the inability to repeat labels: it is impossible to specify `Some($a) => $a`, meaning that a manual check is necessary to see if the right hand side wasn't doing any extra conversion. These had to be addressed by hand, but weren't many.

### DIFF
--- a/src/codegen/enums.rs
+++ b/src/codegen/enums.rs
@@ -156,10 +156,10 @@ fn generate_enum(
     }
     writeln!(
         w,
-        "{}",
-        "    #[doc(hidden)]
+        "\
+    #[doc(hidden)]
     __Unknown(i32),
-}
+}}
 "
     )?;
 
@@ -208,13 +208,13 @@ impl ToGlib for {name} {{
             enum_.name, member.name, sys_crate_name, member.c_name
         )?;
     }
-    writeln!(w, "\t\t\t{}::__Unknown(value) => value", enum_.name)?;
+    writeln!(w, "\t\t\t{}::__Unknown(value) => value,", enum_.name)?;
     writeln!(
         w,
-        "{}",
-        "        }
-    }
-}
+        "\
+        }}
+    }}
+}}
 "
     )?;
 
@@ -248,10 +248,10 @@ impl FromGlib<{sys_crate_name}::{ffi_name}> for {name} {{
     writeln!(w, "\t\t\tvalue => {}::__Unknown(value),", enum_.name)?;
     writeln!(
         w,
-        "{}",
-        "        }
-    }
-}
+        "\
+        }}
+    }}
+}}
 "
     )?;
 
@@ -319,10 +319,10 @@ impl FromGlib<{sys_crate_name}::{ffi_name}> for {name} {{
 
         writeln!(
             w,
-            "{}",
-            "        }
-    }
-}
+            "\
+        }}
+    }}
+}}
 "
         )?;
     }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,185 +2,189 @@
 # It is not intended for manual editing.
 [[package]]
 name = "aho-corasick"
-version = "0.7.13"
+version = "0.7.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7404febffaa47dac81aa44dba71523c9d069b1bdc50a77db41195149e17f68e5"
 dependencies = [
- "memchr 2.3.3 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr",
 ]
 
 [[package]]
 name = "autocfg"
-version = "1.0.0"
+version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cdb031dd78e28731d87d56cc8ffef4a8f36ca26c38fe2de700543e627f8a464a"
 
 [[package]]
 name = "bitflags"
 version = "1.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cf1de2fe8c75bc145a2f577add951f8134889b4795d47466a54a5c846d691693"
 
 [[package]]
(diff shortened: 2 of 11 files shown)


src/codegen/enums.rs
```diff
@@ -168,9 +168,10 @@ fn generate_enum(
         version_condition(w, env, enum_.version, false, 0)?;
         writeln!(
             w,
-            "impl fmt::Display for {0} {{\n\
-             \tfn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {{\n\
-             \t\twrite!(f, \"{0}::{{}}\", match *self {{",
+            "impl fmt::Display for {0} {{
+\tfn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {{
+\t\twrite!(f, \"{0}::{{}}\", match *self {{
+",
```
sdroege, line 174: Why? It looks nicer to have it aligned properly and the output is the same

sdroege (commented):

src/codegen/enums.rs
MarijnS95: I don't find `\n\` particularly readable, but then again that allows to have leading spaces as indentation.

Would this work? (Notice the `"\`):

```rust
            "\
impl fmt::Display for {0} {{
\tfn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {{
\t\twrite!(f, \"{0}::{{}}\", match *self {{
",
```

That also allows `\t` to be substituted with `    `.

sdroege: That would also work for me (also with four leading spaces).

Just it would be nice to have the code in the string consistently indented, otherwise it's confusing to read.

MarijnS95: > also with four leading spaces

That would resurface the need for trailing `\n\` again. I'll remove these changed from the PR as they're merely extra noise.

sdroege: Ok :) Maybe let's handle that separately then, @GuillaumeGomez probably also has opinions on this topic

MarijnS95: It appears both ways are used, one with `\n\` and proper indentation to follow the rest of the code, others with mismatched indentation on the first line and the rest justified all the way to the left. Kept the changes that turn:
```rust
    writeln!(
        w,
        "{}",
        "    #[doc(hidden)]
    __Unknown(i32),
}
"
```
into:
```rust
    writeln!(
        w,
        "\
    #[doc(hidden)]
    __Unknown(i32),
}}
"
```
As per your alignment request.

GuillaumeGomez: I don't mind to have `\n`. Please keep the indent using whitespaces. :)

MarijnS95: I don't know how much we should care about formatting (at least indentation) of generated output (seeing a PR to switch to `quote` earlier this year being closed) as the resulting files are pulled through rustfmt anyway.

sdroege: Not much, what matters is that it's readable inside the `gir` code. The autogenerated code itself will be cleaned up anyway, as you say.

MarijnS95: It seems that'd improve a lot with `quote!` and the way it can expand arrays or iterators, and in general colorize target code in the IDE. But it appears the need for `//` comments with nested `/**/` completely breaks this unfortunately :(.

GuillaumeGomez: No, `quote!` didn't improve the readability at all.

sdroege: Yeah I tried to use quote here but it only made things worse in practice. Maybe I just used it wrongly though...

sdroege (approved):
The PR looks good for me now, let's worry about those string formatting details some other time.

@GuillaumeGomez ?

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

src/codegen/enums.rs
MarijnS95: I'd argue:

```rust
        let type_name = format_ident!("{}", enum_.name);

        let members = members.iter().map(|m| {
            let name = &m.name;
            let ident = format_ident!("{}", name);
            // TODO: version_condition_tokens(env, m.version, false, 3) -> Option<TokenStream>
            quote!(Self::#ident => #name)
        });

        quote! {
            impl fmt::Display for #type_name {
                fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
                    match *self {
                        #( #members, )*
                        _ => "Unknown",
                    }
                }
            }
        }
```
Is way easier to read than:
```rust
        writeln!(
            w,
            "impl fmt::Display for {0} {{\n\
             \tfn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {{\n\
             \t\twrite!(f, \"{0}::{{}}\", match *self {{",
            enum_.name
        )?;
        for member in &members {
            version_condition(w, env, member.version, false, 3)?;
            writeln!(w, "\t\t\t{0}::{1} => \"{1}\",", enum_.name, member.name)?;
        }
        writeln!(
            w,
            "\t\t\t_ => \"Unknown\",\n\
             \t\t}})\n\
             \t}}\n\
             }}\n"
        )?;
```
But then there are quite a few things that are not ergonomic:

- Not able to expand members in-line (ie. `#(m.name)`);
- Not able to convert strings to identifiers within the macro.

These in turn make it impossible to have something like (pseudo-syntax) `#( Self::#(/*turn into identifier*/ #(members).name) => #(#(members).name), )*` directly in place of the `map`.

Other than that the ability to expand iterators and deal with `Option<>` in expansion is pretty neat.

PR #994, opened 2020-11-18
codegen/enum: Do not generate doc(cfg) on match arms

MarijnS95:
Simplify some duplicate code around cfg and version handling, and fix the following warnings when compiling with the `dox` feature:

    warning: unused doc comment
    --> gstreamer/src/auto/enums.rs:761:41
        |
    761 |             #[cfg_attr(feature = "dox", doc(cfg(feature = "v1_18")))]
        |                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
    762 |             66817 => EventType::InstantRateSyncTime,
        |             --------------------------------------- rustdoc does not generate documentation for match arms

Fixes: c63d1fa ("Generate doc cfg attributes for more beautiful docs")

### DIFF
--- a/src/codegen/enums.rs
+++ b/src/codegen/enums.rs
@@ -1,7 +1,8 @@
 use crate::{
     analysis::{imports::Imports, namespaces},
     codegen::general::{
-        self, cfg_deprecated, derives, version_condition, version_condition_string,
+        self, cfg_deprecated, derives, version_condition, version_condition_no_doc,
+        version_condition_string,
     },
     config::gobjects::GObject,
     env::Env,
@@ -174,7 +175,7 @@ fn generate_enum(
             enum_.name
         )?;
         for member in &members {
-            version_condition(w, env, member.version, false, 3)?;
+            version_condition_no_doc(w, env, member.version, false, 3)?;
             writeln!(w, "\t\t\t{0}::{1} => \"{1}\",", enum_.name, member.name)?;
         }
         writeln!(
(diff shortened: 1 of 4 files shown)


MarijnS95:
Should this be `dox` to match `not_version_condition_no_dox`?

MarijnS95:
Switching attributes to work directly on ifs was worth a shot, but that apparently only landed on 1.43 whereas gtk-rs's MSRV is https://github.com/gtk-rs/gtk-rs/runs/1420232906. If we can't bump this on the short term I'll push the long way around, which is still in local `reflog` :wink:

sdroege (approved):

sdroege:
> Switching attributes to work directly on ifs was worth a shot, but that apparently only landed on 1.43 whereas gtk-rs's MSRV is https://github.com/gtk-rs/gtk-rs/runs/1420232906. If we can't bump this on the short term I'll push the long way around, which is still in local `reflog` wink

We're updating to 1.48 today or tomorrow, so feel free to add back that change if it makes anything nicer :)

MarijnS95:
> We're updating to 1.48 today or tomorrow, so feel free to add back that change if it makes anything nicer :)

Awesome! The change is still here, I held off removing it until receiving this reply :grin:. Wasn't expecting an immediate jump from 1.40 all the way to 1.48 though, but that definitely makes everything a lot nicer.

sdroege:
Then let's get this in here now, and regenerate everything later/tomorrow as part of moving to 1.48.

Issue #746, opened 2019-04-08
Rust 2018 support


a1ien:
Can we add rust 2018 edition support? For now gir generate only 2015 version code.

sdroege:
That would be great indeed. I believe @heftig was looking into this at some point?

heftig:
I got stuck trying to disentangle the import (`use`) handling so it would create 2018-style statements.

Boiethios:
Is there anything new about this? I'd like to use `async`/`.await` in the crate, but I cannot without the crate being in 2018 edition.

About the imports, a solution would be to use systematically the full path: either `crate::full::absolute::path` or `self::full::relative::path`, so it shouldn't be a problem.

sdroege:
No news here but you can use the gtk-rs crates in 2018 crates, there's no problem with that. I have some async/await code using gtk-rs (and there's also https://github.com/gtk-rs/examples/blob/pending/src/bin/gio_futures_await.rs )

Boiethios:
> No news here but you can use the gtk-rs crates in 2018 crates, there's no problem with that. I have some async/await code using gtk-rs (and there's also https://github.com/gtk-rs/examples/blob/pending/src/bin/gio_futures_await.rs )

I'm aware of that: that's the point of the editions. My purpose is to add some asynchronous functions beside the `auto` module containing the auto-generated code, though. Your solution would require to create another crate.

sdroege:
> I'm aware of that: that's the point of the editions.

Well until the very latest glib release that was not possible for the macros :)



> My purpose is to add some asynchronous functions beside the `auto` module containing the auto-generated code, though. Your solution would require to create another crate.

Yes in that case the code generator needs to be updated to output 2018 code. Do you want to work on that?

heftig:
> Yes in that case the code generator needs to be updated to output 2018 code. Do you want to work on that?

I've tried upgrading the code generator at the last Rust+GNOME hackfest, and the biggest stumbling block was the [`use` path change](https://doc.rust-lang.org/edition-guide/rust-2018/module-system/path-clarity.html#use-paths).

Boiethios:
> Yes in that case the code generator needs to be updated to output 2018 code. Do you want to work on that?

Sure, I can give it a try!

> I've tried upgrading the code generator at the last Rust+GNOME hackfest, and the biggest stumbling block was the [`use` path change](https://doc.rust-lang.org/edition-guide/rust-2018/module-system/path-clarity.html#use-paths).

[...]

PR #996, opened 2020-11-20
Switch 2018

GuillaumeGomez:
Fixes https://github.com/gtk-rs/gir/issues/746

### DIFF
--- a/src/analysis/ffi_type.rs
+++ b/src/analysis/ffi_type.rs
@@ -6,6 +6,7 @@ use crate::{
     },
     env::Env,
     library::*,
+    nameutil::{use_glib_if_needed, use_glib_type},
     traits::*,
 };
 use log::{info, trace};
@@ -73,16 +74,7 @@ fn ffi_inner(env: &Env, tid: TypeId, inner: &str) -> Result {
             use crate::library::Fundamental::*;
             let inner = match fund {
                 None => "libc::c_void",
-                Boolean => {
-                    return Ok(format!(
-                        "{}ffi::gboolean",
-                        if env.library.is_glib_crate() {
-                            ""
-                        } else {
-                            "glib::"
-                        }
-                    ))
-                }
+                Boolean => return Ok(use_glib_if_needed(env, "ffi::gboolean")),
                 Int8 => "i8",
                 UInt8 => "u8",
                 Int16 => "i16",
@@ -106,16 +98,7 @@ fn ffi_inner(env: &Env, tid: TypeId, inner: &str) -> Result {
                 UniChar => "u32",
                 Utf8 => "libc::c_char",
                 Filename => "libc::c_char",
-                Type => {
-                    return Ok(format!(
-                        "{}ffi::GType",
-                        if env.library.is_glib_crate() {
-                            ""
-                        } else {
-                            "glib::"
-                        }
-                    ))
-                }
+                Type => return Ok(use_glib_if_needed(env, "ffi::GType")),
                 IntPtr => "libc::intptr_t",
                 UIntPtr => "libc::uintptr_t",
                 _ => return Err(TypeError::Unimplemented(inner.into())),
@@ -192,28 +175,13 @@ fn fix_name(env: &Env, type_id: TypeId, name: &str) -> Result {
             | Type::PtrArray(..)
             | Type::List(..)
             | Type::SList(..)
-            | Type::HashTable(..) => Ok(format!(
-                "{}ffi::{}",
-                if env.library.is_glib_crate() {
-                    ""
-                } else {
-                    "glib::"
-                },
-                name
-            )),
+            | Type::HashTable(..) => Ok(use_glib_if_needed(env, &format!("ffi::{}", name))),
             _ => Ok(name.into()),
         }
     } else {
         let sys_crate_name = &env.namespaces[type_id.ns_id].sys_crate_name;
         let sys_crate_name = if sys_crate_name == "gobject_ffi" {
-            format!(
-                "{}::gobject_ffi",
-                if env.library.is_glib_crate() {
-                    "crate"
-                } else {
-                    "glib"
-                }
-            )
+            use_glib_type(env, "gobject_ffi")
         } else if type_id.ns_id == MAIN_NAMESPACE {
             sys_crate_name.to_owned()
         } else {
--- a/src/analysis/ffi_type.rs
+++ b/src/analysis/ffi_type.rs
@@ -73,7 +73,16 @@ fn ffi_inner(env: &Env, tid: TypeId, inner: &str) -> Result {
             use crate::library::Fundamental::*;
             let inner = match fund {
                 None => "libc::c_void",
-                Boolean => return Ok("glib_sys::gboolean".into()),
+                Boolean => {
+                    return Ok(format!(
+                        "{}ffi::gboolean",
+                        if env.library.is_glib_crate() {
+                            ""
+                        } else {
+                            "glib::"
+                        }
+                    ))
+                }
                 Int8 => "i8",
                 UInt8 => "u8",
                 Int16 => "i16",
@@ -97,7 +106,16 @@ fn ffi_inner(env: &Env, tid: TypeId, inner: &str) -> Result {
                 UniChar => "u32",
                 Utf8 => "libc::c_char",
                 Filename => "libc::c_char",
-                Type => "glib_sys::GType",
+                Type => {
+                    return Ok(format!(
+                        "{}ffi::GType",
+                        if env.library.is_glib_crate() {
+                            ""
+                        } else {
+                            "glib::"
+                        }
+                    ))
+                }
                 IntPtr => "libc::intptr_t",
                 UIntPtr => "libc::uintptr_t",
                 _ => return Err(TypeError::Unimplemented(inner.into())),
@@ -174,14 +192,37 @@ fn fix_name(env: &Env, type_id: TypeId, name: &str) -> Result {
             | Type::PtrArray(..)
             | Type::List(..)
             | Type::SList(..)
-            | Type::HashTable(..) => Ok(format!("glib_sys::{}", name)),
+            | Type::HashTable(..) => Ok(format!(
+                "{}ffi::{}",
+                if env.library.is_glib_crate() {
+                    ""
+                } else {
+                    "glib::"
+                },
+                name
+            )),
             _ => Ok(name.into()),
         }
     } else {
-        let name_with_prefix = format!(
-            "{}::{}",
-            &env.namespaces[type_id.ns_id].sys_crate_name, name
-        );
+        let sys_crate_name = &env.namespaces[type_id.ns_id].sys_crate_name;
+        let sys_crate_name = if sys_crate_name == "gobject_ffi" {
+            format!(
+                "{}::gobject_ffi",
+                if env.library.is_glib_crate() {
+                    "crate"
+                } else {
+                    "glib"
+                }
+            )
+        } else if type_id.ns_id == MAIN_NAMESPACE {
+            sys_crate_name.to_owned()
+        } else {
+            format!(
+                "{}::{}",
+                env.namespaces[type_id.ns_id].crate_name, sys_crate_name
+            )
+        };
+        let name_with_prefix = format!("{}::{}", sys_crate_name, name);
         if env
             .type_status_sys(&type_id.full_name(&env.library))
             .ignored()
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -731,9 +731,6 @@ fn analyze_function(
     }
 
     if r#async && status.need_generate() && !commented {
-        if env.config.library_name != "Gio" {
-            imports.add_with_constraint("gio", version, None);
-        }
         imports.add("std::ptr");
         imports.add("std::boxed::Box as Box_");
         imports.add("std::pin::Pin");
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -200,7 +200,14 @@ fn fixup_gpointer_parameter(
             ref_mode: RefMode::ByRef,
             to_glib_extra: String::new(),
             explicit_target_type: format!("{} {}", pointer_type, ffi_name),
-            pointer_cast: " as glib_sys::gconstpointer".into(),
+            pointer_cast: format!(
+                " as {}ffi::gconstpointer",
+                if env.library.is_glib_crate() {
+                    ""
+                } else {
+                    "glib::"
+                }
+            ),
             in_trait: false,
             nullable: false,
         },
@@ -209,7 +216,6 @@ fn fixup_gpointer_parameter(
 
 fn fixup_special_functions(
     env: &Env,
-    imports: &mut Imports,
     name: &str,
     type_tid: library::TypeId,
     is_boxed: bool,
@@ -222,7 +228,6 @@ fn fixup_special_functions(
         && parameters.c_parameters[0].c_type == "gconstpointer"
     {
         fixup_gpointer_parameter(env, type_tid, is_boxed, parameters, 0);
-        imports.add("glib_sys");
     }
 
     if (name == "compare" || name == "equal" || name == "is_equal")
@@ -232,7 +237,6 @@ fn fixup_special_functions(
     {
         fixup_gpointer_parameter(env, type_tid, is_boxed, parameters, 0);
         fixup_gpointer_parameter(env, type_tid, is_boxed, parameters, 1);
-        imports.add("glib_sys");
     }
 }
 
@@ -592,14 +596,7 @@ fn analyze_function(
         }
     }
 
-    fixup_special_functions(
-        env,
-        imports,
-        name.as_str(),
-        type_tid,
-        is_boxed,
-        &mut parameters,
-    );
+    fixup_special_functions(env, name.as_str(), type_tid, is_boxed, &mut parameters);
 
     // Key: destroy callback index
     // Value: associated user data index
@@ -739,11 +736,8 @@ fn analyze_function(
 
     if r#async && status.need_generate() && !commented {
         if env.config.library_name != "Gio" {
-            imports.add("gio_sys");
             imports.add_with_constraint("gio", version, None);
         }
-        imports.add("glib_sys");
-        imports.add("gobject_sys");
         imports.add("std::ptr");
         imports.add("std::boxed::Box as Box_");
         imports.add("std::pin::Pin");
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -201,12 +201,8 @@ fn fixup_gpointer_parameter(
             to_glib_extra: String::new(),
             explicit_target_type: format!("{} {}", pointer_type, ffi_name),
             pointer_cast: format!(
-                " as {}ffi::gconstpointer",
-                if env.library.is_glib_crate() {
-                    ""
-                } else {
-                    "glib::"
-                }
+                " as {}",
+                nameutil::use_glib_if_needed(env, "ffi::gconstpointer")
             ),
             in_trait: false,
             nullable: false,
--- a/src/analysis/imports.rs
+++ b/src/analysis/imports.rs
@@ -1,5 +1,6 @@
 use super::namespaces;
 use crate::{library::Library, nameutil::crate_name, version::Version};
+use std::borrow::Cow;
 use std::cmp::Ordering;
 use std::collections::btree_map::BTreeMap;
 use std::ops::{Deref, DerefMut};
@@ -119,21 +120,41 @@ impl Imports {
         self.defaults.clear();
     }
 
+    /// The goals of this function is to discard unwanted imports like "ffi" and "crate". It
+    /// also extends the checks in case you are implementing "X". For example, you don't want to
+    /// import "X" or "crate::X" in this case.
+    fn common_checks(&self, name: &str) -> bool {
+        if name == "crate" || name == "ffi" {
+            false
+        } else if let Some(ref defined) = self.defined {
+            !((name.starts_with("crate::") && &name[7..] == defined) || name == defined)
+        } else {
+            true
+        }
+    }
+
     /// Declares that name should be available through its last path component.
     ///
     /// For example, if name is `X::Y::Z` then it will be available as `Z`.
     /// Uses defaults.
     pub fn add(&mut self, name: &str) {
-        if let Some(ref defined) = self.defined {
-            if name == defined {
-                return;
-            }
+        if !self.common_checks(name) {
+            return;
         }
-        if let Some(name) = self.strip_crate_name(name) {
+        if let Some(mut name) = self.strip_crate_name(name) {
+            if name == "xlib" {
+                name = if self.crate_name == "gdk_x11" {
+                    // Dirty little hack to allow to have correct import for GDKX11.
+                    Cow::Borrowed("x11::xlib")
+                } else {
+                    // gtk has a module named "xlib" which is why this hack is needed too.
+                    Cow::Borrowed("crate::xlib")
+                };
+            }
             let defaults = &self.defaults;
             let entry = self
                 .map
-                .entry(name.to_owned())
+                .entry(name.into_owned())
                 .or_insert_with(|| defaults.clone());
             entry.update_version(self.defaults.version);
             entry.update_constraints(&self.defaults.constraints);
@@ -144,16 +165,17 @@ impl Imports {
     ///
     /// For example, if name is `X::Y::Z` then it will be available as `Z`.
     pub fn add_with_version(&mut self, name: &str, version: Option<Version>) {
-        if let Some(ref defined) = self.defined {
-            if name == defined {
-                return;
-            }
+        if !self.common_checks(name) {
+            return;
         }
         if let Some(name) = self.strip_crate_name(name) {
-            let entry = self.map.entry(name.to_owned()).or_insert(ImportConditions {
-                version,
-                constraints: Vec::new(),
-            });
+            let entry = self
+                .map
+                .entry(name.into_owned())
+                .or_insert(ImportConditions {
+                    version,
+                    constraints: Vec::new(),
+                });
             entry.update_version(version);
             // Since there is no constraint on this import, if any constraint
             // is present, we can just remove it.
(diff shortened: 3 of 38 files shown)


src/analysis/ffi_type.rs
```diff
@@ -73,7 +73,16 @@ fn ffi_inner(env: &Env, tid: TypeId, inner: &str) -> Result {
             use crate::library::Fundamental::*;
             let inner = match fund {
                 None => "libc::c_void",
-                Boolean => return Ok("glib_sys::gboolean".into()),
+                Boolean => {
+                    return Ok(format!(
+                        "{}ffi::gboolean",
+                        if env.library.is_glib_crate() {
+                            ""
```
sdroege, line 80: `crate::` if you want to make it more explicit, and we could have a helper function somewhere that returns `glib::` or `crate::` depending on `is_glib_crate()`. That should simplify this and a few other places.

src/analysis/functions.rs
```diff
@@ -209,7 +216,7 @@ fn fixup_gpointer_parameter(
 
 fn fixup_special_functions(
     env: &Env,
-    imports: &mut Imports,
+    _imports: &mut Imports,
```
sdroege, line 219: Maybe remove the parameter completely then :)

src/analysis/imports.rs
```diff
@@ -119,21 +120,37 @@ impl Imports {
         self.defaults.clear();
     }
 
+    fn common_checks(&self, name: &str) -> bool {
```
sdroege, line 123: Document what this function actually does :) When does it return `true` / `false`?

src/analysis/imports.rs
```diff
@@ -119,21 +120,37 @@ impl Imports {
...
+            return;
         }
-        if let Some(name) = self.strip_crate_name(name) {
+        if let Some(mut name) = self.strip_crate_name(name) {
+            // Dirty little hack to allow to have correct import for GDKX11.
+            if name == "xlib" {
+                name = if self.crate_name == "gdk_x11" {
+                    Cow::Borrowed("x11::xlib")
+                } else {
+                    Cow::Borrowed("crate::xlib")
```
sdroege, line 147: `x11::xlib` will also work from inside `gdkx11`, no need to complicate it (and then you can also remove the `pub(crate) use x11::xlib as xlib` from there)

src/analysis/imports.rs
```diff
@@ -231,27 +253,27 @@ impl Imports {
...
         if !name.starts_with(prefix) {
-            return Some(name);
+            return Some(Cow::Borrowed(name));
         }
         let rest = &name[prefix.len()..];
         if rest.is_empty() {
             None
-        } else if let Some(prefix) = rest.strip_prefix("::") {
-            Some(prefix)
+        } else if rest.starts_with("::") {
```
sdroege, line 272: Where would `::Foo` come from? Let's maybe just not generate that anymore and directly have it create `crate::Foo` instead?

src/analysis/object.rs
```diff
@@ -61,8 +61,11 @@ pub fn class(env: &Env, obj: &GObject, deps: &[library::TypeId]) -> Option<Info>
     let klass: &library::Class = type_.maybe_ref()?;
 
     let mut imports = Imports::with_defined(&env.library, &name);
+    if !env.library.is_glib_crate() {
+        imports.add("glib");
```
sdroege, line 65: This shouldn't be needed any more. If `glib` is in the dependencies you can directly use `glib::blabla` everywhere

src/analysis/record.rs
```diff
@@ -80,10 +80,13 @@ pub fn new(env: &Env, obj: &GObject) -> Option<Info> {
...
-    imports.add(env.main_sys_crate_name());
-    if record.glib_get_type.is_some() && is_boxed {
-        imports.add("gobject_sys");
+    imports.add(&format!("crate::{}", env.main_sys_crate_name()));
+    if !env.library.is_glib_crate() {
+        imports.add("glib");
     }
+    // if record.glib_get_type.is_some() && is_boxed {
+    //     imports.add("gobject_sys");
+    // }
```
sdroege, line 89: Why commented out?

src/analysis/rust_type.rs
```diff
@@ -285,7 +286,7 @@ pub fn rust_type_full(
                             } else {
                                 "&"
                             },
-                            if y != "GString" {
+                            if !y.ends_with("GString") {
```
sdroege, line 289: This could give false positives

src/analysis/rust_type.rs
```diff
@@ -160,11 +161,11 @@ pub fn rust_type_full(
                         ok("std::ffi::OsString")
                     }
                 }
-                Type if env.namespaces.glib_ns_id == library::MAIN_NAMESPACE => ok("types::Type"),
+                Type if env.library.is_glib_crate() => ok("crate::types::Type"),
```
sdroege, line 164: Nicer if we could go with `format!("{}::types::Type", some_helper_function(&env.library))` here and elsewhere

src/analysis/signals.rs
```diff
@@ -89,7 +89,6 @@ fn analyze_signal(
 
     let action_emit_name = if signal.is_action {
         imports.add("glib");
```
sdroege, line 91: Not needed any longer (maybe grep for that generally here). Also `use gio` would not be needed if we have that somewhere

src/codegen/enums.rs
```diff
@@ -273,11 +273,12 @@ impl FromGlib<{sys_crate_name}::{ffi_name}> for {name} {{
...
                 writeln!(
                     w,
-                    "        static QUARK: once_cell::sync::Lazy<glib_sys::GQuark> = once_cell::sync::Lazy::new(|| unsafe {{
-            glib_sys::g_quark_from_static_string(b\"{}\\0\".as_ptr() as *const _)
+                    "        static QUARK: once_cell::sync::Lazy<{0}ffi::GQuark> = once_cell::sync::Lazy::new(|| unsafe {{
+            {0}ffi::g_quark_from_static_string(b\"{1}\\0\".as_ptr() as *const _)
         }});
         from_glib(*QUARK)",
-                    quark = quark
+                    if env.library.is_glib_crate() { "" } else { "glib::" },
```
sdroege, line 280: `crate::` instead of empty string also here (and basically everywhere), and also remember my comment about the helper function :)

src/codegen/flags.rs
```diff
@@ -35,8 +35,9 @@ pub fn generate(env: &Env, root_path: &Path, mod_rs: &mut Vec<String>) {
     let path = root_path.join("flags.rs");
     file_saver::save_to_file(path, env.config.make_backup, |w| {
         let mut imports = Imports::new(&env.library);
-        imports.add(env.main_sys_crate_name());
+        imports.add(&format!("crate::{}", env.main_sys_crate_name()));
         imports.add("glib::translate::*");
+        imports.add("bitflags::bitflags");
```
sdroege, line 40: We re-export `bitflags` from `glib` btw, so maybe use that version directly. Then all the crates don't have to depend on it separately anymore.

src/codegen/function_body_chunk.rs
```diff
@@ -409,7 +409,7 @@ impl Builder {
             let nullable = trampoline.parameters.rust_parameters[par.ind_rust].nullable;
             let is_fundamental =
                 add_chunk_for_type(env, par.typ, par, &mut body, &ty_name, nullable);
-            if ty_name == "GString" {
+            if ty_name.ends_with("GString") {
```
sdroege, line 412: Also potential false positive, check if you have more such checks

src/codegen/function_body_chunk.rs
```diff
@@ -1338,15 +1366,15 @@ fn add_chunk_for_type(
...
             let type_name;
-            if ty_name == "GString" {
+            if ty_name.ends_with("GString") {
                 if *nullable {
                     if par.conversion_type == ConversionType::Borrow {
-                        type_name = String::from(": Borrowed<Option<GString>>");
+                        type_name = String::from(": Borrowed<Option<glib::GString>>");
                     } else {
-                        type_name = String::from(": Option<GString>");
+                        type_name = String::from(": Option<glib::GString>");
```
sdroege, line 1374: Always `glib::`? Might also have to be `crate::` if this ever gets called from inside glib :)

src/codegen/general.rs
```diff
@@ -322,7 +338,15 @@ pub fn define_auto_boxed_type(
 ) -> Result<()> {
     let sys_crate_name = env.main_sys_crate_name();
     writeln!(w)?;
-    writeln!(w, "glib_wrapper! {{")?;
+    writeln!(
+        w,
+        "{}::glib_wrapper! {{",
```
sdroege, line 343: As you call all macros explicitly now, make sure to also remove all `#[macro_use]` from the crates after regenerating

src/codegen/property_body.rs
```diff
@@ -73,9 +78,12 @@ impl Builder {
         let mut params = Vec::new();
 
         let cast_target = if self.is_child_property {
-            "gtk_sys::GtkContainer"
+            "ffi::GtkContainer".to_owned()
```
sdroege, line 81: If there are ever child properties in other crates (libhandy and libdazzle seem good candidates), this would have to become `gtk::ffi` (vs. now `crate::ffi`). So maybe add a case for this already

sdroege: Also elsewhere below

src/codegen/property_body.rs
```diff
@@ -131,7 +139,10 @@ impl Builder {
         let unwrap = if self.is_nullable {
             // This one is strictly speaking nullable, but
             // we represent that with an empty Vec instead
-            if self.type_ == "Vec<GString>" {
+            if ["Vec<GString>", "Vec<crate::GString>", "Vec<glib::GString>"]
```
sdroege, line 142: Better check than the `ends_with()` you have elsewhere. Sounds like this could use a helper function `is_gstring()` :)

src/codegen/sys/lib_.rs
```diff
@@ -97,7 +97,7 @@ fn generate_extern_crates(w: &mut dyn Write, env: &Env) -> Result<()> {
 
 fn get_extern_crate_string(library: &ExternalLibrary) -> String {
     format!(
-        "extern crate {}_sys as {};\n",
+        "use {}_sys as {};\n",
```
sdroege, line 100: You could directly change that name inside the `Cargo.toml` with 2018 edition:

```toml
ffi = { package = "glib-sys", version = "..." }
```

for example. That seems like a nicer solution. Same for these imports in the non-sys crates.

sdroege (commented):
Looks mostly ok, just various smaller fixes needed. I'll review the generated code etc once you fixed that and updated it accordingly.

src/analysis/imports.rs
GuillaumeGomez: Well, from the current crate obviously. :smile:

src/analysis/object.rs
GuillaumeGomez: :o

src/analysis/imports.rs
sdroege: Why would someone pass `::Foo` in here then instead of directly going with `crate::Foo`?

src/analysis/record.rs
GuillaumeGomez: Forgot to remove it.

src/analysis/rust_type.rs
GuillaumeGomez: Good idea!

src/analysis/rust_type.rs
GuillaumeGomez: Do you one in mind? I could rewrite it like this:

```rust
if y != "GString" && !y.ends_with("::GString") {
```

src/analysis/signals.rs
GuillaumeGomez: It's actually the last one remaining. :)

src/analysis/imports.rs
GuillaumeGomez: No one. The point of this function is to strip the crate name. So if you receive "glib::Foo", `rest` will be "::Foo".

src/codegen/flags.rs
GuillaumeGomez: I suggest to do it in another PR (maybe the one that will remove all unneeded sys dependencies?)

src/codegen/function_body_chunk.rs
GuillaumeGomez: It seems not because I never encountered the case but it'd be better to not do it by default.

GuillaumeGomez: But just a few lines above we have this check:

```rust
library::Type::Alias(_) if ty_name == "glib::Pid" => {
```

So I wonder if it's worth it to change it?

src/codegen/property_body.rs
GuillaumeGomez: Ok!

src/codegen/sys/lib_.rs
GuillaumeGomez: I approve, however I won't do it here. ;)

sdroege: Create an issue then :)

src/codegen/function_body_chunk.rs
sdroege: How do you mean?

src/codegen/flags.rs
sdroege: Create an issue then, but that should be exactly two lines of changes in gir or not? :)

src/analysis/rust_type.rs
sdroege: Works for me, but do it consistently everywhere please

src/analysis/imports.rs
sdroege: Ok then

src/analysis/rust_type.rs
GuillaumeGomez: Ok, new helper function incoming!

sdroege: I like helper functions

src/codegen/sys/lib_.rs
GuillaumeGomez: https://github.com/gtk-rs/gir/issues/998

src/codegen/flags.rs
GuillaumeGomez: https://github.com/gtk-rs/gir/issues/997

src/analysis/functions.rs
```diff
@@ -739,11 +732,8 @@ fn analyze_function(
 
     if r#async && status.need_generate() && !commented {
         if env.config.library_name != "Gio" {
-            imports.add("gio_sys");
             imports.add_with_constraint("gio", version, None);
```
sdroege, line 735: Can be removed, can't it? Check if there are more `gio` imports being added elsewhere :)

GuillaumeGomez: Oh good point! More imports removed! \o/

src/codegen/function_body_chunk.rs
```diff
@@ -861,15 +862,22 @@ impl Builder {
...
             },
             Param {
                 name: "res".to_string(),
-                typ: "*mut gio_sys::GAsyncResult".to_string(),
+                typ: format!(
+                    "*mut {}ffi::GAsyncResult",
+                    if env.library.is_crate("Gio") {
+                        ""
+                    } else {
+                        "gio::"
```
sdroege, line 874: Maybe this also wants a helper function like for glib :)

src/codegen/function_body_chunk.rs
```diff
@@ -861,15 +862,22 @@ impl Builder {
...
-                typ: "*mut gobject_sys::GObject".to_string(),
+                typ: format!("*mut {}", use_glib_if_needed(env, "gobject_ffi::GObject")),
             },
             Param {
                 name: "res".to_string(),
-                typ: "*mut gio_sys::GAsyncResult".to_string(),
+                typ: format!(
+                    "*mut {}ffi::GAsyncResult",
+                    if env.library.is_crate("Gio") {
+                        ""
```
sdroege, line 872: Why not `crate::`?

src/codegen/property_body.rs
```diff
@@ -73,9 +90,23 @@ impl Builder {
...
+                    "crate"
+                } else {
+                    "gtk"
+                }
+            )
         } else {
-            "gobject_sys::GObject"
+            format!(
+                "{}::gobject_ffi::GObject",
+                if self.env.library.is_glib_crate() {
```
sdroege, line 104: You added a helper function for this :)

src/codegen/property_body.rs
```diff
@@ -73,9 +90,23 @@ impl Builder {
         let mut params = Vec::new();
 
         let cast_target = if self.is_child_property {
-            "gtk_sys::GtkContainer"
+            format!(
+                "{}::ffi::GtkContainer",
+                if self.env.library.is_crate("Gtk") {
```
sdroege, line 95: And maybe this can get another helper function

src/codegen/property_body.rs
```diff
@@ -156,9 +190,23 @@ impl Builder {
...
+                    "crate"
+                } else {
+                    "gtk"
+                }
+            )
         } else {
-            "gobject_sys::GObject"
+            format!(
+                "{}::gobject_ffi::GObject",
+                if self.env.library.is_glib_crate() {
```
sdroege, line 204: Also here

src/codegen/property_body.rs
```diff
@@ -156,9 +190,23 @@ impl Builder {
         let mut params = Vec::new();
 
         let cast_target = if self.is_child_property {
-            "gtk_sys::GtkContainer"
+            format!(
+                "{}::ffi::GtkContainer",
+                if self.env.library.is_crate("Gtk") {
```
sdroege, line 195: And here

src/codegen/property_body.rs
```diff
@@ -209,17 +257,45 @@ impl Builder {
 
     fn get_ffi_func(&self) -> String {
         if self.is_child_property {
-            "gtk_sys::gtk_container_child_get_property".to_owned()
+            format!(
+                "{}::ffi::gtk_container_child_get_property",
+                if self.env.library.is_crate("Gtk") {
```
sdroege, line 262: And here

src/codegen/property_body.rs
```diff
@@ -209,17 +257,45 @@ impl Builder {
...
+                    "crate"
+                } else {
+                    "gtk"
+                }
+            )
         } else {
-            "gobject_sys::g_object_get_property".to_owned()
+            format!(
+                "{}::gobject_ffi::g_object_get_property",
+                if self.env.library.is_glib_crate() {
```
sdroege, line 271: And here

src/codegen/property_body.rs
```diff
@@ -209,17 +257,45 @@ impl Builder {
...
+            )
         }
     }
 
     fn set_ffi_func(&self) -> String {
         if self.is_child_property {
-            "gtk_sys::gtk_container_child_set_property".to_owned()
+            format!(
+                "{}::ffi::gtk_container_child_set_property",
+                if self.env.library.is_crate("Gtk") {
```
sdroege, line 284: and here

src/codegen/property_body.rs
```diff
@@ -209,17 +257,45 @@ impl Builder {
...
+                    "crate"
+                } else {
+                    "gtk"
+                }
+            )
         } else {
-            "gobject_sys::g_object_set_property".to_owned()
+            format!(
+                "{}::gobject_ffi::g_object_set_property",
```
sdroege, line 292: And here

src/codegen/function_body_chunk.rs
GuillaumeGomez: No reason, I'll add it for consistency.

src/nameutil.rs
```diff
@@ -149,6 +150,34 @@ pub fn shared_libs_to_links(shared_libs: &[String]) -> String {
...
+        },
+        import
+    )
+}
+
+pub fn use_glib_if_needed(env: &crate::env::Env, import: &str) -> String {
+    format!(
+        "{}{}",
+        if env.library.is_glib_crate() {
+            ""
```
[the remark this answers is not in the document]
GuillaumeGomez: Because in some cases (a lot of them actually), it's already imported in the current scope. Dark gir magic... Once we remove all imports, this function will very likely be useless.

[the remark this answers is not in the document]
sdroege: Why is that a problem to use `crate::` then?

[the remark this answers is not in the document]
GuillaumeGomez: Honestly, no idea. I tried to keep the code as close to the original as possible. I think this is too much for just one PR. I can open an issue to investigate and remove it if needed (I think it's too much now but who knows?).

src/codegen/return_value.rs
```diff
@@ -97,7 +97,7 @@ fn out_parameter_as_return_parts(
             if num_outs > 1 {
                 if analysis.ret.nullable_return_is_error.is_some() {
                     if is_glib_crate {
-                        ("Result<(", "), BoolError>")
+                        ("Result<(", "), crate::BoolError>")
                     } else {
                         ("Result<(", "), glib::BoolError>")
```
sdroege, line 102: You have a function for this now :)

src/codegen/return_value.rs
```diff
@@ -106,7 +106,7 @@ fn out_parameter_as_return_parts(
                 }
             } else if analysis.ret.nullable_return_is_error.is_some() {
                 if is_glib_crate {
-                    ("Result<", ", BoolError>")
+                    ("Result<", ", crate::BoolError>")
                 } else {
                     ("Result<", ", glib::BoolError>")
```
sdroege, line 111: Also here

src/codegen/return_value.rs
```diff
@@ -120,7 +120,7 @@ fn out_parameter_as_return_parts(
                 (
                     "Result<",
                     if is_glib_crate {
-                        ", Error>"
+                        ", crate::Error>"
                     } else {
                         ", glib::Error>"
```
sdroege, line 125: And here

src/codegen/return_value.rs
```diff
@@ -129,7 +129,7 @@ fn out_parameter_as_return_parts(
                 (
                     "Result<(",
                     if is_glib_crate {
-                        "), Error>"
+                        "), crate::Error>"
```
sdroege, line 132: And here

src/nameutil.rs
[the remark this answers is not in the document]
sdroege: Create an issue then, yes. That seems strange :)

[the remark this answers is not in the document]
GuillaumeGomez: https://github.com/gtk-rs/gir/issues/999

sdroege (approved):
Good to go once the gtk-rs changes are reviewed too

sdroege:
This creates unused imports in -sys crates: https://gitlab.freedesktop.org/slomo/gstreamer-rs/-/jobs/5735521

```
error: unused import: `gstreamer_base_sys as gst_base`
  --> gstreamer-controller/sys/src/lib.rs:15:5
   |
15 | use gstreamer_base_sys as gst_base;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `-D unused-imports` implied by `-D warnings`
```

See https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/638 for the code

sdroege:
```
src/codegen/translate_from_glib.rs
84:                                "{}.{}.ok_or_else(|| glib_bool_error!(\"{}\"))",
102:                            format!("{}.ok_or_else(|| glib_bool_error!(\"{}\"))", res.1, msg),
```

These should use `glib::glib_bool_error!`

sdroege:
Also
```
src/codegen/translate_from_glib.rs
93:                    use_glib_type(env, "glib_result_from_gboolean!("),
```

GuillaumeGomez:
> Also
> 
> ```
> src/codegen/translate_from_glib.rs
> 93:                    use_glib_type(env, "glib_result_from_gboolean!("),
> ```

What's the problem here?

sdroege:
> What's the problem here?

Needs to be `glib::glib_result_from_gboolean!`, `glib::` prefix

sdroege:
> Needs to be `glib::glib_result_from_gboolean!`, `glib::` prefix

Oh it does, nevermind.

sdroege:
Another one in gstreamer-rs, this time not my fault :)

```
warning: unused import: `glib::gobject_ffi`
 --> gstreamer/src/auto/enums.rs:7:5
  |
7 | use glib::gobject_ffi;
  |     ^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default
```

bilelmoussaoui:
I have noticed this new addition by re-generating with the latest commit 
```diff
@ -845,7 +845,7 @@ impl<O: IsA<CellArea>> CellAreaExt for O {
             P: IsA<CellArea>,
         {
             let f: &F = &*(f as *const F);
-            let path = from_glib_full(ffi::gtk_tree_path_new_from_string(path));
+            let path = from_glib_full(gtk_sys::gtk_tree_path_new_from_string(path));
             f(
                 &CellArea::from_glib_borrow(this).unsafe_cast_ref(),
                 &from_glib_borrow(renderer),
```
Not sure what caused it?

sdroege:
```
warning: unused import: `glib::gobject_ffi`
 --> gstreamer/src/auto/enums.rs:7:5
  |
7 | use glib::gobject_ffi;
  |     ^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

```

Still happens. You changed it for flags but not for enums @GuillaumeGomez :)

bilelmoussaoui:
This one https://github.com/gtk-rs/gir/blob/master/src/codegen/trampoline.rs#L219 should become a `ffi::` instead of using `gtk_sys::`

sdroege (approved):

bilelmoussaoui:
The changes from the latest commits looks all good to me except an unnecessary import 
```
error: unused import: `gdk4_sys as gdk`
  --> gdk4-wayland/sys/src/lib.rs:13:5
   |
13 | use gdk4_sys as gdk;
   |     ^^^^^^^^^^^^^^^
   |
   = note: `-D unused-imports` implied by `-D warnings`
```

PR #1003, opened 2020-11-23
config/members: add manual to wanted check

fengalin:
Missed this in https://github.com/gtk-rs/gir/pull/977

Not a big deal, but that will remove some error logs.


sdroege:
@GuillaumeGomez The CI fails here because (AFAIU) you didn't update the tests to Rust 2018

PR #1006, opened 2020-11-25
codegen: Explicitly use glib::Value instead of importing

BrainBlasted:
Fixes an issue where glib::Value could conflict with other
things named "Value"

### DIFF
--- a/src/analysis/properties.rs
+++ b/src/analysis/properties.rs
@@ -181,7 +181,6 @@ fn analyze_property(
             imports.add_used_type(s);
         }
         if type_string.is_ok() {
-            imports.add("glib::Value");
             imports.add("glib::StaticType");
         }
 
@@ -208,17 +207,14 @@ fn analyze_property(
             imports.add_used_type(s);
         }
         let set_bound = PropertyBound::get(env, prop.typ);
-        if type_string.is_ok() {
-            imports.add("glib::Value");
-            if set_bound.is_some() {
-                imports.add("glib::object::IsA");
-                if !*nullable {
-                    //TODO: support nonnulable setter if found any
-                    warn!(
-                        "Non nulable setter for property generated as nullable \"{}.{}\"",
-                        type_name, name
-                    );
-                }
+        if type_string.is_ok() && set_bound.is_some() {
+            imports.add("glib::object::IsA");
+            if !*nullable {
+                //TODO: support non-nullable setter if found any
+                warn!(
+                    "Non nullable setter for property generated as nullable \"{}.{}\"",
+                    type_name, name
+                );
             }
         }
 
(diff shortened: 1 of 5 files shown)


src/codegen/enums.rs
```diff
@@ -354,7 +353,7 @@ impl FromGlib<{sys_crate_name}::{ffi_name}> for {name} {{
         writeln!(
             w,
             "impl<'a> FromValueOptional<'a> for {name} {{
-    unsafe fn from_value_optional(value: &Value) -> Option<Self> {{
+    unsafe fn from_value_optional(value: &glib::Value) -> Option<Self> {{
```
sdroege, line 356: This needs to use `use_glib_type("Value")` and that would also fix the CI (partially at least).

Same in the other places below

src/analysis/properties.rs
```diff
@@ -209,7 +208,6 @@ fn analyze_property(
         }
         let set_bound = PropertyBound::get(env, prop.typ);
         if type_string.is_ok() {
-            imports.add("glib::Value");
             if set_bound.is_some() {
```
sdroege, line 211: clippy wants these two `if`s merged now

sdroege:

src/analysis/properties.rs
```diff
@@ -208,17 +207,14 @@ fn analyze_property(
...
-                        "Non nulable setter for property generated as nullable \"{}.{}\"",
-                        type_name, name
-                    );
-                }
+        if type_string.is_ok() && set_bound.is_some() {
+            imports.add("glib::object::IsA");
+            if !*nullable {
+                //TODO: support nonnulable setter if found any
+                warn!(
+                    "Non nulable setter for property generated as nullable \"{}.{}\"",
```
sdroege, line 215:
```suggestion
                    "Non nullable setter for property generated as nullable \"{}.{}\"",
```

src/analysis/properties.rs
```diff
@@ -208,17 +207,14 @@ fn analyze_property(
...
-                    //TODO: support nonnulable setter if found any
-                    warn!(
-                        "Non nulable setter for property generated as nullable \"{}.{}\"",
-                        type_name, name
-                    );
-                }
+        if type_string.is_ok() && set_bound.is_some() {
+            imports.add("glib::object::IsA");
+            if !*nullable {
+                //TODO: support nonnulable setter if found any
```
sdroege, line 213:
```suggestion
                //TODO: support non-nullable setter if found any
```

sdroege:
Not your typos, but as you've moved that code around let's just fix them now :) Otherwise good to me. @GuillaumeGomez?

GuillaumeGomez (approved):

sdroege:
Thanks!

Issue #1009, opened 2020-11-27
Switch to doc cfg instead of feature dox


GuillaumeGomez:
It was first implemented in https://github.com/gtk-rs/gir/pull/1005 but proved impossible to work when the [PR on gtk-rs](https://github.com/gtk-rs/gtk-rs/pull/177)  was opened.

This is currently not possible because of two things (but only one of them fixed would allow us to make it work):

* https://github.com/rust-lang/cargo/issues/8902
* https://github.com/rust-lang/cargo/issues/8905

MarijnS95:
@GuillaumeGomez Is it time to flip this back on? I read somewhere that `#[cfg]` is now picked up without requiring `#[doc(cfg())]` too? Is the second issue still a problem in practice?

GuillaumeGomez:
It's nightly only for now and requires to enable the feature.

MarijnS95:
@GuillaumeGomez Sure, the "advantage" of the current solution is that you can still build (incomplete) docs on `stable` as long as `dox` is not set. docs.rs already builds with nightly though.

bilelmoussaoui:
Note that our current docs require nightly as well

GuillaumeGomez:
gtk-rs documentation generated on docs.rs is mostly useless unfortunately. Also, we'd need to use `cfg_hide` to not display too many `cfg`s (because we might have some only needed for docs in any case). So not a straightforward conversion.

> Note that our current docs require nightly as well

True, but relying on a very recent nightly feature which is not complete yet (I need to send some follow-ups) is not the best idea here I think.

MarijnS95:
In GStreamer we already have a (by accident, in the GL windowing modules I added probably right during this back-and-forth transition) `#![cfg_attr(all(not(doctest), doc), feature(doc_cfg))]` that doesn't compile on stable rustdoc anyway.

@GuillaumeGomez Yes, we'd have to do the transition atomically. If we just start relying on not needing `doc(cfg())` anymore the `feature="dox"` will show up until we can replace that with `doc` as cfg option.

GuillaumeGomez:
There is `cfg_hide` for that. :3

MarijnS95:
@GuillaumeGomez Sure, but I don't like that :|

GuillaumeGomez:
[...]

PR #1005, opened 2020-11-23
Switch to doc cfg instead of feature dox

GuillaumeGomez:
When rustdoc is running, it provides alongside `cfg(doc)`, which is all we need, making the `dox` feature unnecessary.

I need to check on gtk-rs first if I didn't forget it there so please wait before merging!

PR #1010, opened 2020-11-27
Revert "Switch to doc cfg instead of feature dox"

GuillaumeGomez:
Reverts gtk-rs/gir#1005 as explained in https://github.com/gtk-rs/gir/issues/1009


sdroege:
Sadness

sdroege (approved):

PR #1011, opened 2020-11-28
cargo_toml: Do not overwrite library name for unversioned system-deps

MarijnS95:
See: https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/565#note_709394

In the case of GL EGL/Wayland/X11 we use gstreamer-gl-1.0 for every version except from 1.18 onwards, where these are renamed to `gstreamer-gl-{egl,wayland,x11}-1.0`.
This requires us to manually change the name of the "unversioned" system dep in Cargo.toml, without the generator overwriting it afterwards.

---

Let me know if this should be a separate helper function next to `set_string` instead.


sdroege:
CC @gdesmott

sdroege (approved):
Makes sense to me

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

Issue #963, opened 2020-08-11
Rename `get_` methods


hyarsan:
The `get_` methods should be renamed to fit the amazing [Rust API Guidelines].

This would apply to the other crates as well.
Yes, there are a lot of methods, but it would lead to nicer, more rusty code.

## Examples
`ButtonExt.get_text()` -> `ButtonExt.text()`
`LabelExt.get_angle()` -> `LabelExt.angle()`

Perhaps methods like the following could look different?
`ButtonExt.get_always_show_image()` -> `ButtonExt.always_shows_image()`
`ButtonExt.get_use_underline()` ->`ButtonExt.uses_underline()`
Ehh...probably not.... it's just something to consider.

[Rust API Guidelines]: https://rust-lang.github.io/api-guidelines/naming.html#c-getter

sdroege:
So @fengalin created https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/639 and I think this improves the API a lot, making it more Rust-y.

I think we should do the same here for gtk-rs: *most* getters should have their `get_` prefix removed, very few should keep it and some need special handling: `get_is_foo` -> `is_foo`, and others where the second word is a verb. We can implement heuristics for that and the 5 cases not covered by one of the heuristics can be manually fixed via Gir.toml.

@bilelmoussaoui @GuillaumeGomez What do you think?

sdroege:
Also this should be done together with https://github.com/gtk-rs/gir/issues/962

bilelmoussaoui:
I totally agree on this on renaming both cases. Other language bindings do make the properties usage seamless, for example Python can lets you modify the properties using `object.props.some_prop = value`, the same +/- for GJS & Vala. It would be nice if we can do all the renaming changes once though to avoid having to re-update everything multiple times though.

See #936 #974 #905

fengalin:
Based on my experience of doing that manually in `gstreamer-rs`, bedsides implementing this in `gir`, we also need an automatic process for manual code conversion. I'm not skilled at `batch` text processing at all, but I think that goes beyond its capabilities. If we all agree on this, I'll take a look at implementing a parser / filter for this where we could setup the heuristics. Maybe we could also fix the call places in the process. Hopefully, this parser / filter could be useful for other future changes.

sdroege:
> we also need an automatic process for manual code conversion

[...]

Issue #909, opened 2020-04-28
Generate doc aliases to C equivalents


GuillaumeGomez:
`rustdoc` has the `doc_alias` feature that you can see in action [here](https://doc.rust-lang.org/nightly/src/core/ops/bit.rs.html#113) (the two first items are aliases).

Considering that most examples for GNOME libraries are in C, I assumed that users could be looking for the C function in our doc. By adding aliases on our items, we would also return the rust one.

fengalin:
Just to let you know, I've already done (at least part of) this for https://github.com/gtk-rs/gir/issues/963. Will post a PR this we hopefully.

GuillaumeGomez:
I think it can be closed now.

PR #1014, opened 2020-12-05
Doc aliases for C & renamed functions

fengalin:
(At least part of) https://github.com/gtk-rs/gir/issues/909.

Was handled while preparing https://github.com/gtk-rs/gir/issues/963.

### DIFF
--- a/src/codegen/function.rs
+++ b/src/codegen/function.rs
@@ -1,7 +1,8 @@
 use super::{
     function_body_chunk,
     general::{
-        cfg_condition, cfg_deprecated, doc_hidden, not_version_condition, version_condition,
+        cfg_condition, cfg_deprecated, doc_alias, doc_hidden, not_version_condition,
+        version_condition,
     },
     parameter::ToParameter,
     return_value::{out_parameters_as_return, ToReturnValue},
@@ -82,6 +83,9 @@ pub fn generate(
     version_condition(w, env, analysis.version, commented, indent)?;
     not_version_condition(w, analysis.not_version, commented, indent)?;
     doc_hidden(w, analysis.doc_hidden, comment_prefix, indent)?;
+    if !in_trait || only_declaration {
+        doc_alias(w, &analysis.glib_name, comment_prefix, indent)?;
+    }
     writeln!(
         w,
         "{}{}{}{}{}",
@@ -167,7 +171,10 @@ pub fn declaration(env: &Env, analysis: &analysis::functions::Info) -> String {
 
     format!(
         "fn {}{}({}){}",
-        analysis.name, bounds, param_str, return_str,
+        analysis.codegen_name(),
+        bounds,
+        param_str,
+        return_str,
     )
 }
 
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -69,6 +69,7 @@ pub struct AsyncFuture {
 #[derive(Debug)]
 pub struct Info {
     pub name: String,
+    pub new_name: Option<String>,
     pub glib_name: String,
     pub status: GStatus,
     pub kind: library::FunctionKind,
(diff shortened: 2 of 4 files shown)


src/codegen/function.rs
```diff
@@ -74,6 +75,12 @@ pub fn generate(
     version_condition(w, env, analysis.version, commented, indent)?;
     not_version_condition(w, analysis.not_version, commented, indent)?;
     doc_hidden(w, analysis.doc_hidden, comment_prefix, indent)?;
+    if !in_trait || only_declaration {
+        doc_alias(w, &analysis.glib_name, comment_prefix, indent)?;
+        if analysis.new_name.is_some() {
+            doc_alias(w, &analysis.name, comment_prefix, indent)?;
```
GuillaumeGomez, line 81: An example for this second doc_alias:

```rust
#[doc(alias = "gdk_pixbuf_new_from_bytes")]
#[doc(alias = "new_from_bytes")]
pub fn from_bytes() {}
```

I think there will be a lot of of equivalents for `new_from_bytes` and I'm sure there is much point into having it...

fengalin: Yes indeed. See new version.

GuillaumeGomez:
Looks much better, thanks!

src/codegen/function.rs
GuillaumeGomez: Looks much better, thanks!

sdroege:
Let's get this in after https://github.com/gtk-rs/gir/pull/991 :) One step at a time

src/codegen/function.rs
```diff
@@ -82,6 +83,12 @@ pub fn generate(
...
```
GuillaumeGomez, line 89: Actually: isn't it still generated doc alias we don't want? Do you have an example now please?

fengalin: https://github.com/fengalin/gtk-rs/blob/309d221a728c5d706cbd553ebe96874a8ec2cba9/glib/src/auto/date_time.rs#L47

GuillaumeGomez: I only see one `doc(alias = "...")` in this link. Did I miss something?

fengalin: Isn't that what you wanted: https://github.com/gtk-rs/gir/pull/1014#discussion_r536735301?

fengalin: For explicit renames in Gir.toml, you get this: https://github.com/fengalin/gtk-rs/blob/309d221a728c5d706cbd553ebe96874a8ec2cba9/gtk/src/auto/widget.rs#L340

GuillaumeGomez: At yes, I meant the second, sorry, should have been more explicit. And once again, we fall into the same issue, it generates very common names such `get_name`. :-/

fengalin: Yes, that might be the case with many names. I can remove this, but I thought that would help users figure out changes in names. This is particularly helpful for getters renaming (upcoming).

GuillaumeGomez: I think it's not a good idea (even less if it's meant to be temporary). What do you think @sdroege ?

fengalin: It's nice that we can serch by the C name, but too bad we don't get results by the actual function name, nor by searching something like "Widget name".

fengalin: Removed

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me! Just waiting for @sdroege and we're good to go. :)

fengalin:
Last minute inspiration... sorry.
Added a function so that we don't duplicate the `new_name` / `name` selection code in codegen.

sdroege:
Please send the regenerate PRs for gtk-rs and gstreamer :)

fengalin:
I'll update the PR I already pushed now, but please don't merge to master in those repositories while I update.

fengalin:
Well I guess gtk-rs is already done by 0ef36307474678fdd11e09d995b28e5ae32e35d9.
So skipping this one.

Issue #1013, opened 2020-12-04
Extend the "final type" checks


bilelmoussaoui:
See https://github.com/gtk-rs/gtk-rs/pull/191#discussion_r536066829 & https://gitlab.gnome.org/GNOME/glib/-/issues/2261 for details

PR #1015, opened 2020-12-05
Improve final type heuristic to also check for unknown instance structs

sdroege:
These always show up in the .gir file but for final types there wouldn't
be any fields inside it. A proper instance struct needs to contain at
least the field for the parent type's instance struct.

Fixes https://github.com/gtk-rs/gir/issues/1013


sdroege:
Doesn't seem to work correctly yet.

sdroege:
Ah no, actually works correctly. Please review :)

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #1017, opened 2020-12-07
Use `cargo:warning` instead of `eprintln!` for printing warnings

sdroege:
The latter just disappears instead of being printed.

Fixes https://github.com/gtk-rs/gir/issues/1016


sdroege:
Also need to update `cairo/sys/build.rs` as that's manually managed

GuillaumeGomez:
Thanks!

PR #1021, opened 2020-12-09
Remove get  for getters

fengalin:
This is an attempt at fixing https://github.com/gtk-rs/gir/issues/963 & https://github.com/gtk-rs/gir/issues/962.

The changes are fairly minimal because most of the "hard work" is done in [`fix-getters-rules`](https://github.com/fengalin/fix-getters/tree/main/rules). I decided to implement those rules in a separate crate because they are not `gir` specific:

* They can be used for non-`gir` based crates.
* They are required for manual code update.

However, I took good care at keeping this dependency minimal. The package used in `gir` only pulls in `once_cell` which is already pulled in by `gir`. I also introduced a feature to remove a module which makes sense in this package, but which is of no use for `gir`.

Please take a look at the `README`s in [`fix-getters`](https://github.com/fengalin/fix-getters). They explain the assumptions and decisions I had to make to achieve this. Of course, these assumption are all subject to debate.

Changes to `gtk-rs` can be found in [this PR](https://github.com/gtk-rs/gtk-rs/pull/211).

Please see the request in [this PR](https://github.com/gtk-rs/gtk-rs/pull/211).

Fixes https://github.com/gtk-rs/gir/issues/963


fengalin:
Of course, CI failure is expected.

sdroege:
@GuillaumeGomez What do you think? :)

README.md
```diff
@@ -466,6 +466,19 @@ detected by GIR is different than what was implemented and conversion to the C
...
+bypass the automatic renaming mechanism, use `bypass_auto_rename = true`:
+
+```toml
+[[object]]
+name = "Gtk.TextBuffer"
+[...]
+    [[object.function]]
+    name = "get_insert"
+    # Avoid clash with the `insert` operation.
+    bypass_auto_rename = true
```
GuillaumeGomez, line 480: That's exactly what I'm afraid of in the end... Two questions about this:

1. How much name conflicts would be there?
2. Couldn't we automatically detect those conflicts?

fengalin: > How much name conflicts would be there?

See https://github.com/gtk-rs/gtk-rs/pull/211/commits/4fb1e7b4b61f3ada35cee9988a58b9a3b48639aa & https://github.com/gtk-rs/gtk-rs/pull/211/commits/021ee0930b9cd7dd4448fa3647a095c0334e70e2. Please take a look at https://github.com/gtk-rs/gtk-rs/pull/211.

> Couldn't we automatically detect those conflicts?

That's already been done for `get_property_...`, others need some kind of human cognition IMHO.

fengalin:
This is ready. See https://github.com/gtk-rs/gtk-rs/pull/211 for an application on 2 hours ago's `gtk-rs` master.

**Important: when this is merge, other projects will not be able to regen before the `gtk-rs` PR is merged too, otherwise it will break everything up.**

GuillaumeGomez (approved):

GuillaumeGomez:
So just to sum things up for others: it could create conflicts between a trait method and a type method (even if unlikely). But that's pretty much it. For such cases, a nice thing could be to add a lint in clippy but that is not a blocker.

Waiting for @slomo or @bilelmoussaoui approval and then let's merge!

Cargo.toml
```diff
@@ -13,6 +13,7 @@ edition = "2018"
 [dependencies]
 bitflags = "1.0"
 getopts = "0.2.21"
+getter_rules = { default-features = false, package = "fix-getters-rules", git = "https://github.com/fengalin/fix-getters", tag = "0.1.0" }
```
sdroege, line 16: This should probably get a better name so it can go on crates.io without looking strange :) Any suggestions?

sdroege: Not a blocker for merging this now though, just before the release.

src/analysis/functions.rs
```diff
@@ -532,24 +550,32 @@ fn analyze_function(
...
-                name = name[4..].to_string();
+
+    let bypass_auto_rename = configured_functions.iter().any(|f| f.bypass_auto_rename);
+    if !bypass_auto_rename && new_name.is_none() {
+        match func.kind {
+            library::FunctionKind::Method => {
+                new_name = try_getter_rename(func, &name);
+            }
+            library::FunctionKind::Constructor => {
+                if name.starts_with("new_from") || name.starts_with("new_with") {
```
sdroege, line 566: `new_for` would be another pattern btw

Cargo.toml
fengalin: The actual crate name is `fix-getters-rules`, is that better?

sdroege (commented):
Apart from these comments here and the ones on the gtk-rs PR, this looks all good to me. Good work! :)

Cargo.toml
sdroege: Yeah

fengalin:
This is ready for merge now I think. See gtk-rs/gtk-rs#211 for an application on lastest gtk-rs's master.

**Important: when this is merged, please wait until I update gtk-rs/gtk-rs#211 with the merged Gir before merging anything else that could perform a regen on `gtk-rs/gtk-rs`.**

fengalin:
@sdroege @bilelmoussaoui this PR needs your approvals. Then we can merge!

bilelmoussaoui (approved):

PR #1022, opened 2020-12-11
Fixed issue generating the -sys/Cargo.toml

deanleggo:
#1020 
Changed the way `package.metadata.system-deps` `name` is checked and inserted.

Issue #1020, opened 2020-12-09
Puts invalid pkg-config package names into Cargo.toml (dashes replaced with underscores)


deanleggo:
Following the [Gir Tutorial](https://gtk-rs.org/docs-src/tutorial/gir_tutorial) and continuing from the discussion on [GNOME's Discourse](https://discourse.gnome.org/t/rust-bindings-for-tracker-3-from-gir/4968). I am unable to build tracker-sys. On discourse I referenced the test/abi.rs file, but this file is (of course) not ran during the build step.

This time I built Tracker from source and placed the generated gir file in tracker-sys/gir-files along with the required librarys from /usr/share/gir-1.0/ on Fedora 33. The code I have is on [my Gitlab - tracker-rs](https://gitlab.com/dleggo/tracker-rs/-/tree/master/tracker-sys).

During the build process, it is trying to find the package `tracker_sparql_3_0` but the name should be `tracker-sparql-3.0`. For some reason the builder is modifying the package name found inside the tracker gir file when searching pkg-config for the package.

```
error: failed to run custom build command for `tracker-sys v0.0.1 (/home/dean/Projects/tracker-rs/tracker-sys)`

Caused by:
  process didn't exit successfully: `/home/dean/Projects/tracker-rs/tracker-sys/target/debug/build/tracker-sys-52e25f5b766c5977/build-script-build` (exit code: 1)
  --- stdout
  cargo:rerun-if-env-changed=TRACKER_SPARQL_3_0_NO_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG
  cargo:rerun-if-env-changed=TRACKER_SPARQL_3_0_STATIC
  cargo:rerun-if-env-changed=TRACKER_SPARQL_3_0_DYNAMIC
  cargo:rerun-if-env-changed=PKG_CONFIG_ALL_STATIC
  cargo:rerun-if-env-changed=PKG_CONFIG_ALL_DYNAMIC
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64-unknown-linux-gnu
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH_x86_64_unknown_linux_gnu
  cargo:rerun-if-env-changed=HOST_PKG_CONFIG_PATH
  cargo:rerun-if-env-changed=PKG_CONFIG_PATH
  cargo:rerun-if-env-changed=PKG_CONFIG_LIBDIR_x86_64-unknown-linux-gnu
[...]

PR #1023, opened 2020-12-12
cargo_toml: Do not overwrite system-deps version if already set

MarijnS95:
For [1] we need to override the target library version because 1.20 does not exist yet. Gir keeps overwriting this value making it impossible to check such changes in.

As cleanup remove the unnecessary get_mut in set_string: insert already replaces an existing value if the given key exists.

[1]: https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/649#note_723814

### DIFF
--- a/src/codegen/sys/cargo_toml.rs
+++ b/src/codegen/sys/cargo_toml.rs
@@ -117,10 +117,11 @@ fn fill_in(root: &mut Table, env: &Env) {
         let lib_name = ns.package_name.as_ref().unwrap();
 
         let meta = upsert_table(meta, nameutil::lib_name_to_toml(lib_name));
-        if !meta.contains_key("name") {
-            set_string(meta, "name", lib_name);
-        }
-        set_string(meta, "version", env.config.min_cfg_version.to_string());
+        // Allow both the name and version of a system dep to be overridden by hand
+        meta.entry("name")
+            .or_insert_with(|| Value::String(lib_name.to_owned()));
+        meta.entry("version")
+            .or_insert_with(|| Value::String(env.config.min_cfg_version.to_string()));
 
         // Old version API
         unset(meta, "feature-versions");
@@ -130,7 +131,10 @@ fn fill_in(root: &mut Table, env: &Env) {
             .filter(|(&v, _)| v > env.config.min_cfg_version)
             .for_each(|(v, lib_version)| {
                 let version_section = upsert_table(meta, &v.to_feature());
-                set_string(version_section, "version", lib_version.to_string());
+                // Allow system-deps version for this feature level to be overridden by hand
+                version_section
+                    .entry("version")
+                    .or_insert_with(|| Value::String(lib_version.to_string()));
             });
     }
 
@@ -186,10 +190,6 @@ fn get_crate_name(config: &Config, root: &Table) -> String {
 }
 
 fn set_string<S: Into<String>>(table: &mut Table, name: &str, new_value: S) {
-    if let Some(v) = table.get_mut(name) {
-        *v = Value::String(new_value.into());
-        return;
-    }
     table.insert(name.into(), Value::String(new_value.into()));
 }
 


src/codegen/sys/cargo_toml.rs
```diff
@@ -117,10 +117,10 @@ fn fill_in(root: &mut Table, env: &Env) {
         let lib_name = ns.package_name.as_ref().unwrap();
 
         let meta = upsert_table(meta, nameutil::lib_name_to_toml(lib_name));
-        if !meta.contains_key("name") {
-            set_string(meta, "name", lib_name);
-        }
-        set_string(meta, "version", env.config.min_cfg_version.to_string());
+        meta.entry("name")
+            .or_insert_with(|| Value::String(lib_name.to_owned()));
```
sdroege, line 121: Doesn't this basically revert https://github.com/gtk-rs/gir/pull/1022 ?

MarijnS95: This time I fixed my mistake from #1011. Since that PR didn't bother to state an actual bug has been fixed _nor how_, I couldn't be bothered to explain either.

Hint: the difference is already in the signature for `fn entry` vs `fn get_mut`.

GuillaumeGomez: Please add a code comment explaining all this so that we know exactly what is going on in case we need to update it once again in the future.

MarijnS95: @GuillaumeGomez Added an extra comment explaining this allows overriding `name` and `version` by hand. More explanation and links are in the commit, unless you want those inline as well.

I doubt you intend to explain the difference between `.get_mut` and `.entry` in an inline comment though, that is better suited for the commit addressing the bug itself.

sdroege: > Since that PR didn't bother to state an actual bug has been fixed nor how, I couldn't be bothered to explain either.

You mean #1022? The full explanation is in the linked issue #1020 :)

MarijnS95: Not that I can see, and even if it did, such an explanation does not show up in `git blame` nor `git show`.

sdroege: As #1020 explains, autogenerating the -sys crate put a dependency with a wrong pkg-config name into `Cargo.toml`: `tracker-sparql-3.0` it should've been, `tracker_sparql_3.0` it was. That's what #1022 fixed.

MarijnS95: @sdroege Thanks for providing a perfect example to the problem (lacking commit titles/messages) I'm highlighting here.

My mistake in #1011 causes no `name = "name-with-hyphens"` override to show up _at all_, because `get_mut` does not allow modifying the underlying map with insertions (it returns `Option<&mut ItemType>`, so only a mutable reference to the item _if it exists_). That's what `fn entry` is for.
Note that `set_string` - reintroduced by #1022 - does not perform this conversion - `lib_name` is already "canonicalized" with hyphens.

I think the author is seeing underscores because it falls back to `[lib] name = "name_with_underscores"`, again because the unversioned `[...system-deps...]` table is generated without `name` entry.

GuillaumeGomez (approved):

deanleggo:
Hey @MarijnS95, I'm new to this and I didn't know my #1022 completely overrides your #1011 (from a week or two before). Sorry, next time I'll check the history and make sure the PR and commit message states why the change was made (not relying on a linked issue/discussion) and discuss with the author of the code about the change.

MarijnS95:
@deanleggo No worries, you didn't override #1022 but rather fix a serious regression introduced by it. However, according to the  discussion in #1020 and subsequent changes made in #1022 it seems that the underlying issue was not understood. This is why I advocate for well thought out commit titles (and descriptions where necessary).

Furthermore using git blame to find an eventual isolated change to those lines might help understand the problem and find the previous author, though in this case the maintainers likely remembered who (what PR) changed it just under 2 weeks ago. CC-ing that author is never a bad idea so that they are aware of the change, are able to explain their situation/use-case and most importantly: learn from their mistakes.

In this PR I have applied an alternate solution, the solution I intended to write in #1011 and would have suggested if I were aware of #1022 before it got merged. No harm done, just that these small and nasty regressions are a great learning source for everyone.

PR #1025, opened 2020-12-17
Use `wrapper!` instead of `glib_wrapper!`

ids1024:
Goes along with https://github.com/gtk-rs/gtk-rs/pull/220.

### DIFF
--- a/src/codegen/general.rs
+++ b/src/codegen/general.rs
@@ -106,7 +106,7 @@ pub fn define_object_type(
         .collect();
 
     writeln!(w)?;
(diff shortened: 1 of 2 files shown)


sdroege (approved):

sdroege:
@GuillaumeGomez ?

bilelmoussaoui (approved):

GuillaumeGomez:
Oh right, we talked about it a while ago when switching to 2018 edition. Thanks for doing it!

PR #1027, opened 2020-12-23
Add an error in case a name field has content of the pattern field

GuillaumeGomez:
It's a common issue when edition the gir files so instead of silently fail, it now displays an error.

cc @sdroege

### DIFF
--- a/src/config/functions.rs
+++ b/src/config/functions.rs
@@ -565,12 +565,13 @@ pattern = 'func\d+'
 "#,
         );
         let fns = Functions::parse(Some(&toml), "a");
-        assert_eq!(fns.len(), 4);
+        assert_eq!(fns.len(), 3);
 
         assert_eq!(fns.matched("func1").len(), 2);
         assert_eq!(fns.matched("func2").len(), 2);
         assert_eq!(fns.matched("func3").len(), 1);
-        assert_eq!(fns.matched("f1.5").len(), 1);
+        // "f1.5" is not a valid name
+        assert_eq!(fns.matched("f1.5").len(), 0);
         assert_eq!(fns.matched("none").len(), 0);
     }
 
--- a/src/config/ident.rs
+++ b/src/config/ident.rs
@@ -46,10 +46,20 @@ impl Ident {
                     e
                 })
                 .ok(),
-            None => toml
-                .lookup("name")
-                .and_then(Value::as_str)
-                .map(|s| Ident::Name(s.into())),
+            None => match toml.lookup("name").and_then(Value::as_str) {
+                Some(name) => {
+                    if name.contains(['.', '+', '*'].as_ref()) {
+                        error!(
+                            "Should be `pattern` instead of `name` in {} for `{}`",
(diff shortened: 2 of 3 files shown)


src/config/ident.rs
```diff
@@ -46,10 +46,20 @@ impl Ident {
                     e
                 })
                 .ok(),
-            None => toml
-                .lookup("name")
-                .and_then(Value::as_str)
-                .map(|s| Ident::Name(s.into())),
+            None => match toml.lookup("name").and_then(Value::as_str) {
+                Some(name) => {
+                    if name.chars().any(|c| c == '.' || c == '+' || c == '*') {
```
sdroege, line 51:
```suggestion
                    if name.contains(&['.', '+', '*']) {
```

sdroege (commented):
Looks good to me otherwise, but CI doesn't like it for some reason

src/config/functions.rs
```diff
@@ -565,12 +565,13 @@ pattern = 'func\d+'
 "#,
         );
         let fns = Functions::parse(Some(&toml), "a");
-        assert_eq!(fns.len(), 4);
```
sdroege, line 568: Why did this change?

src/config/properties.rs
```diff
@@ -184,12 +184,13 @@ pattern = 'prop\d+'
 "#,
         );
         let props = Properties::parse(Some(&toml), "a");
-        assert_eq!(props.len(), 4);
```
sdroege, line 187: Why did this change?

src/config/ident.rs
```diff
@@ -46,10 +46,23 @@ impl Ident {
...
                 })
                 .ok(),
-            None => toml
-                .lookup("name")
-                .and_then(Value::as_str)
-                .map(|s| Ident::Name(s.into())),
+            None => match toml.lookup("name").and_then(Value::as_str) {
+                Some(name) => {
+                    // we're forced to do it in two steps otherwise we get a nice compiler error:
+                    // expected a `std::ops::Fn<(char,)>` closure, found `[char; 3]`
```
sdroege, line 52: Not really forced, that's your decision :) The other options are all shorter!

src/config/functions.rs
GuillaumeGomez: The explanation is in the comment just below. ;)

src/config/properties.rs
GuillaumeGomez: The explanation is in the comment just below. ;)

sdroege: Which comment?

src/config/functions.rs
GuillaumeGomez: This one: https://github.com/gtk-rs/gir/pull/1027/files#diff-ee42c4f4c96eb712946d909affc0319ee2aa8c12db41446d441a5144191269e4R573 :)

sdroege: Oh the test was broken before...

sdroege (approved):

PR #1031, opened 2021-01-07
Do not generate functions ending with _finish_utf8

mbiggio:
Such a function is present in
https://lazka.github.io/pgi-docs/Gio-2.0/classes/DataInputStream.html.
Another option would be to explicitly add an exception for this function
in the Gir.toml of gtk-rs/gio.

### DIFF
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -104,7 +104,7 @@ impl Info {
             .rust_parameters
             .iter()
             .any(|param| param.typ.full_name(&env.library) == "Gio.AsyncResult");
-        self.name.ends_with("_finish") && has_async_result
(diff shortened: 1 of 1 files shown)


mbiggio:
This change is required for https://github.com/gtk-rs/gtk-rs/pull/245

sdroege:
Hmm as these functions are manually bound I would say they should be ignored manually in `Gir.toml` too. Sorry for missing that before, I thought all the `_finish()` functions caused this problems now.

mbiggio:
> Hmm as these functions are manually bound I would say they should be ignored manually in `Gir.toml` too. Sorry for missing that before, I thought all the `_finish()` functions caused this problems now.

No problem, I will drop this PR and put back the `Gir.toml` rule :)

PR #1032, opened 2021-01-07
Remove use_boxed_functions override

MarijnS95:
This override only works when `get_type` is available and is intended when copy and free are not available, to call the generic `g_boxed_copy/free` functions (which under the hood resolve the right `copy/free` function from its `GType`, as retrieved with get_type).

However, this behaviour is already implied if `copy/free` nor `(un)ref` are found. Hence it is only useful to forcibly use generic `g_boxed_copy/free` even if type-specific functions are directly available in G-IR and Rust. That can almost never be the right thing.

The functionality isn't used for nearly 3 years, and wasn't even used for all that long in gstreamer-rs; it was introduced [here][1] and removed a mere [15 days after][2].

[1]: https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/commit/3a0c9723044fb3c9063fcef60e0270171576ba92
[2]: https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/commit/92016945fc7882979664bbea6e36a5585a2fea1d

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -49,7 +49,9 @@ jobs:
           path: target
           key: checks-${{ runner.os }}-cargo-target-dir-${{ steps.toolchain.outputs.rustc_hash }}
       - name: Install packages from apt
-        run: sudo apt install libgtk-3-dev libssh2-1-dev libglib2.0-dev libgraphene-1.0-dev libcairo-gobject2 libcairo2-dev
+        run: |
+          sudo apt update
+          sudo apt install libgtk-3-dev libssh2-1-dev libglib2.0-dev libgraphene-1.0-dev libcairo-gobject2 libcairo2-dev
         if: matrix.os == 'ubuntu-20.04'
       - name: "Run clippy"
         uses: actions-rs/cargo@v1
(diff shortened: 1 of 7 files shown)


sdroege (approved):

sdroege:
The CI failure looks like it needs an `apt update` before trying to installed the packages... @bilelmoussaoui ?

bilelmoussaoui:
Looks like a random network failure to me? There's a "re-run all checks" button on the right at https://github.com/gtk-rs/gir/pull/1032/checks, do you mind re-triggering it?

sdroege:
Tried that a few times :)

GuillaumeGomez (approved):

GuillaumeGomez:
I really wonder what's going on with the CI...

MarijnS95:
Ouch, `cargo update` automatically pushing `regex` from 1.4.2 to 1.4.3 breaks as well:

```
error: large size difference between variants
Error:   --> src/config/ident.rs:10:5
   |
10 |     Pattern(Regex),
   |     ^^^^^^^^^^^^^^ this variant is 552 bytes
   |
   = note: `-D clippy::large-enum-variant` implied by `-D warnings`
note: and the second-largest variant is 24 bytes:
  --> src/config/ident.rs:9:5
   |
9  |     Name(String),
   |     ^^^^^^^^^^^^
```

Guess we should put this in a `Box` after all.

PR #1038, opened 2021-01-14
Add extra_dox_features config setting

GuillaumeGomez:
In `gdkx11`, the `cairo` dependency is optional but is still being used with `dox`. We need to be able to generate extra `dox` features and this feature does it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,10 @@ version = "3.16"
 dependencies = [
   "glib-sys/v3_16"
 ]
+# Add features to the "dox" feature declaration in `Cargo.toml`. So with the following
+# config, it'll generate:
+# dox = ["whatever"]
+dox_feature_dependencies = ["whatever"]
 ```
 
 You can mark some functions that has suffix `_utf8` on Windows:
--- a/src/config/config.rs
+++ b/src/config/config.rs
@@ -45,6 +45,7 @@ pub struct Config {
     pub extra_versions: Vec<Version>,
     pub lib_version_overrides: HashMap<Version, Version>,
     pub feature_dependencies: HashMap<Version, Vec<String>>,
+    pub dox_feature_dependencies: Vec<String>,
 }
 
 impl Config {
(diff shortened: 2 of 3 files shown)


src/analysis/function_parameters.rs
```diff
@@ -78,14 +78,16 @@ pub enum TransformationType {
...
-            | ToGlibUnknown { .. }
-            | ToSome(_)
-            | IntoRaw(_))
+                | ToGlibScalar { .. }
+                | ToGlibPointer { .. }
+                | ToGlibBorrow
+                | ToGlibUnknown { .. }
+                | ToSome(_)
+                | IntoRaw(_)
+        )
```
sdroege, line 90: What's this change?

README.md
```diff
@@ -78,6 +78,8 @@ version = "3.16"
 dependencies = [
   "glib-sys/v3_16"
 ]
+# Add features to the "dox" one:
+extra_dox_features = ["whatever"]
```
sdroege, line 82: Needs some explanation what this actually does. That it adds them to the `Cargo.toml` and not to the `#[cfg]`

src/analysis/function_parameters.rs
GuillaumeGomez: Arf, damn fmt... T_T

README.md
GuillaumeGomez: Good idea, I'll extend the explanations a bit.

src/config/config.rs
```diff
@@ -362,6 +365,23 @@ fn read_extra_versions(toml: &toml::Value) -> Result<Vec<Version>, String> {
     }
 }
 
+fn read_extra_dox_features(toml: &toml::Value) -> Result<Vec<String>, String> {
+    match toml.lookup("options.extra_dox_features") {
```
sdroege, line 369: `toml.lookup(...).and_then(|a| your code).unwrap_or_else(Vec::new)` is probably nicer and shorter

README.md
sdroege: Maybe should also contain something about `Cargo.toml` in the name of the configuration

GuillaumeGomez: I'm open to suggestions. :)

sdroege: `dox_feature_dependencies`? It's dependencies after all

src/config/config.rs
GuillaumeGomez: Hum, no. The code is quite horrible. ^^'

GuillaumeGomez:
Updated!

sdroege (approved):

PR #1041, opened 2021-01-21
Extend --gir-directory (into --gir-directories) to be able to pass multiple gir directories

### LINKED ISSUE #1040: Allow to provide more than one folder for gir files
As discussed in https://github.com/gtk-rs/gir-files/pull/75

GuillaumeGomez:
Fixes #1040.

cc @hfiguiere @bilelmoussaoui


sdroege (approved):

bilelmoussaoui:
Looks good to me, the readme could use an update to mention that multiple directories are supported as well

GuillaumeGomez:
@bilelmoussaoui: Actually, I just realized that `girs_dir` should be taking an array too.

GuillaumeGomez:
So I also renamed the config field (from `girs_dir` into `girs_dirs`), which is why the CI is breaking.

GuillaumeGomez:
I actually renamed it to `girs_directories` to make it easier to read.

src/config/config.rs
```diff
@@ -109,14 +110,25 @@ impl Config {
...
+            .filter(|x| !x.is_empty())
+            .map(|x| PathBuf::from(&x))
+            .collect();
+        if girs_dirs.is_empty() {
+            let dirs = toml.lookup_vec("options.girs_directories", "No options.girs_directories")?;
+            for dir in dirs {
+                let dir = dir.as_str().ok_or_else(|| {
+                    "options.girs_dirs expected to be array of string".to_string()
+                })?;
+                girs_dirs.push(config_dir.join(dir));
```
sdroege, line 124:
```
 /usr/share/rust/.cargo/bin/cargo fmt -- --check
Diff in /home/runner/work/gir/gir/src/config/config.rs at line 116:
             .map(|x| PathBuf::from(&x))
             .collect();
         if girs_dirs.is_empty() {
-            let dirs = toml.lookup_vec("options.girs_directories", "No options.girs_directories")?;
+            let dirs =
+                toml.lookup_vec("options.girs_directories", "No options.girs_directories")?;
             for dir in dirs {
                 let dir = dir.as_str().ok_or_else(|| {
                     "options.girs_dirs expected to be array of string".to_string()
Error: The process '/usr/share/rust/.cargo/bin/cargo' failed with exit code 1
```

README.md
```diff
@@ -125,7 +125,7 @@ This mode requires you to write another TOML file. [gtk/Gir.toml](https://github
 
 ```toml
 [options]
-girs_dir = "gir-files"
+girs_directories = ["gir-files"]
```
sdroege, line 128: Can we make it so that existing `Gir.toml` that just have a string continue to work? I.e. you'd check if there's a string or an array and use them accordingly? Seems like useless churn to have to update this everywhere.

src/main.rs
```diff
@@ -44,7 +44,12 @@ fn build_config() -> Result<RunKind, String> {
         "CONFIG",
     );
     options.optflag("h", "help", "Show this message");
-    options.optopt("d", "gir-directory", "Directory for girs", "GIRSPATH");
+    options.optmulti(
+        "d",
+        "gir-directories",
```
sdroege, line 49: Elsewhere you call this girs-dirs (note the double plural). Let's make this consistent in one way or another

README.md
GuillaumeGomez: Well, since we're updating the name of the option in any case...

src/main.rs
GuillaumeGomez: Good catch!

README.md
sdroege: Can we maybe continue allowing the old one then? Doesn't seem like a very useful change otherwise

GuillaumeGomez: I'd rather not so that this parameter remains the reflection of the command-line argument. If you're just bored to make the change in gstreamer-rs, I'll send you a PR. But if it's more than that, we can see what @bilelmoussaoui thinks about it too (and maybe other gir users too).

sdroege: I can change it just with sed, but it seems like potentially confusing for all the outside users of gir. What would the behaviour be when running it with the old name?

GuillaumeGomez: I can add an error saying that the parameter was renamed?

bilelmoussaoui: Why not just call it `sources` instead of girs_directories ?

GuillaumeGomez: `sources` could be about almost anything.

bilelmoussaoui: Sure, but it's in a Gir.toml file and it's literally the only input you give to gir (as in files) ? but I agree that sources isn't very clear. But I find girs_directories to be extra-verbose.

src/config/config.rs
```diff
@@ -109,14 +110,26 @@ impl Config {
             }
         };
 
-        let girs_dir: PathBuf = match girs_dir.into() {
-            Some("") | None => {
-                let path = toml.lookup_str("options.girs_dir", "No options.girs_dir")?;
-                config_dir.join(path)
+        let mut girs_dirs: Vec<PathBuf> = girs_dirs
```
EPashkin, line 113: Why not leave `git_dir` as shortcut for single item list?

sdroege:
What's the status here?

GuillaumeGomez:
Agreement about the option naming, if it should still support the old name, etc...

MarijnS95:
@sdroege is this going to enable the use of `gtk-rs/gir-files` in `gstreamer-rs` for files like `GLib`/`GObject` in addition to GStreamer-specific files that remain in checked in to the main repository?

sdroege:
Maybe, that doesn't really seem like an urgent issue to me but that's why @bilelmoussaoui wanted this implemented for gtk4-rs.

MarijnS95:
@sdroege The most annoying issue there is the mis-use of git hashes of the `gir-files` directory (that's not a submodule). Looks like this PR cheats its way out by taking the hash of the first directory.

sdroege:
Yes that part is annoying :) I'd prefer to not have that hash in there if the gir-files directory is not a git submodule.

MarijnS95:
I bet that can be improved as part of or on top of this change:

- Generate a (potentially empty) list of hashes per submodule passed to `--gir-directories`;
- Omit hash if `git rev-parse --show-toplevel` is identical to the root directory, not identical to the directory passed to `--gir-directories` or something like that.

bilelmoussaoui:
> Maybe, that doesn't really seem like an urgent issue to me but that's why @bilelmoussaoui wanted this implemented for gtk4-rs.

Yes, that's exactly what I had in mind. The same thing should be done for webkit2gtk-rs imho as it doesn't make sense to have a ton of gir-files in the gir-files repository if we don't need them/use them.

GuillaumeGomez:
So is there anything else I need to do here?

MarijnS95:
@GuillaumeGomez Perhaps see if the git hash header/version can be improved as outlined in https://github.com/gtk-rs/gir/pull/1041#issuecomment-781220647, part of this PR or a new one on top? Would be nice to emit a hash for every unique directory that is a git repo, excluding the root the generated result is put into.

GuillaumeGomez:
Hum... Do you have suggestion for the formatting? Because we need to include both the folder name and the hash.

GuillaumeGomez:
Sounds good to me!

src/config/config.rs
GuillaumeGomez: I think we shouldn't have both options because it tends to make think that the other doesn't exist.

GuillaumeGomez:
I won't be able to provide such a nice URL so I'll instead use:

```
// This file was generated by gir (https://github.com/gtk-rs/gir @ 5726e16)
// from gir-files (https://github.com/gtk-rs/gir-files @ f059d75d)
// from gst-gir-files (@ <hash>)
```

MarijnS95:
> I won't be able to provide such a nice URL

If the hash can be grabbed it's easy to retrieve the URL of a single pull remote? Granted, using `origin` or `upstream` or whatever default when multiple exist is debatable.

GuillaumeGomez:
Look at my remotes then:

```bash
$ git remote -v
origin	git@github.com:GuillaumeGomez/gtk-rs.git (fetch)
origin	git@github.com:GuillaumeGomez/gtk-rs.git (push)
upstream	git@github.com:gtk-rs/gtk-rs (fetch)
upstream	git@github.com:gtk-rs/gtk-rs (push)
```

That's definitely not great and we can't be sure it'll be better for others either.

MarijnS95:
Same here. I guess `upstream` is the best bet if it exists next to `origin`, but getting into the semantics of that is probably not worth it for a simple hash annotation - let's put that effort towards not spewing these disruptive changes in `gstreamer-rs` where it inadvertently takes the hash of the root repo that is obviously constantly changing.

GuillaumeGomez:
@MarijnS95 Updated. Wanna give it a try? :)

src/codegen/general.rs
```diff
@@ -16,32 +16,71 @@ use std::{
...
+        conf.girs_version
+            .iter()
+            .map(|info| {
+                format!(
+                    "// from {} ({}{})\n",
+                    info.gir_dir.display(),
+                    match info.get_repository_url() {
+                        Some(url) => format!("{} @ ", url),
+                        None => String::new(),
+                    },
```
MarijnS95, line 35: If I'm being `option_if_let_else`:

```suggestion
                    info.get_repository_url().map_or_else(String::new, |url| format!("{} @ ", url))
```

`gir`/`gstreamer-rs` really need more strict lints turned on - if it wasn't so dreadful without a crate-shareable config file.

src/codegen/general.rs
```diff
@@ -16,32 +16,71 @@ use std::{
 
 pub fn start_comments(w: &mut dyn Write, conf: &Config) -> Result<()> {
     if conf.single_version_file.is_some() {
-        return start_comments_no_version(w);
+        return start_comments_no_version(w, conf);
     }
-    writeln!(
+    write!(
```
MarijnS95, line 21: You could possibly just call `single_version_file` and save some duplication. (Perhaps rename the function as well)

src/config/config.rs
```diff
@@ -125,11 +144,14 @@ impl Config {
...
         }
-        let girs_version = girs_dirs
+        let mut girs_version = girs_dirs
             .iter()
-            .filter_map(repo_hash)
-            .next()
-            .unwrap_or_else(|| "???".into());
+            .map(|d| GirVersion {
+                gir_dir: d.clone(),
+                hash: repo_hash(d),
```
MarijnS95, line 151: Probably for a separate PR: can we make this return `None` if `d` is not a git repository (or simply a directory within the root repository)?

MarijnS95:
Had to give it a try, and:

```diff
-// This file was generated by gir (https://github.com/gtk-rs/gir @ 5726e16)
-// from gir-files (https://github.com/gtk-rs/gir-files @ f059d75d)
+// This file was generated by gir (https://github.com/gtk-rs/gir @ 8fe37c0+)
+// from gstreamer/sys/../../gir-files (https://github.com/gtk-rs/gir-files @ 21f7670)
+// from gstreamer/sys/../../gst-gir-files (???)
 // DO NOT EDIT
-
 #include "manual.h"
```
I guess that's pretty okay, maybe see if we can get a canonical path relative to the root repo.

Wasn't aware that `-d` on the command line overrides `girs_dir`... @sdroege we should clean up `gstreamer-rs` to either use `girs_directories` which is tedious given that we have crates with different (1-3 folders deep: `gl/egl/sys`) nesting, or passing both root directories straight from `generator.py`. That also cleans up the comments again:

```diff
-from gstreamer-audio/../gir-files (https://github.com/gtk-rs/gir-files @ @ 21f7670)
-from gstreamer-audio/../gst-gir-files (@ ???)
+from gir-files (https://github.com/gtk-rs/gir-files @ @ 21f7670)
+from gst-gir-files (@ ???)
```
This is from a `version.txt` file by the way... Hence why I ask in the review to not duplicate code like that as an `@` bug slipped into one of the implementations!

src/codegen/general.rs
```diff
@@ -16,32 +16,71 @@ use std::{
...
+{}",
+        VERSION,
+        conf.girs_version
+            .iter()
+            .map(|info| {
+                format!(
+                    "from {} ({}@ {})\n",
+                    info.gir_dir.display(),
+                    match info.get_repository_url() {
+                        Some(url) => format!("{} @ ", url),
```
MarijnS95, line 77: Here you're emitting double `@`'s. Just move the code from `start_comments` into `single_version_file` and call that instead.

src/codegen/general.rs
GuillaumeGomez: Nice!

src/config/config.rs
GuillaumeGomez: For now I'll apply the current rule: "???". But please open an issue if you want it to changed.

src/codegen/general.rs
MarijnS95: @GuillaumeGomez Whoops, forgot to say that `single_version_file` does not emit `//` before every line, breaking all rust files :hand_over_mouth:

src/codegen/general.rs
```diff
@@ -16,32 +16,48 @@ use std::{
...
-        VERSION, conf.girs_version
+{}",
+        VERSION,
+        conf.girs_version
+            .iter()
+            .map(|info| {
+                format!(
+                    "from {} ({}@ {})\n",
+                    info.gir_dir.display(),
+                    info.get_repository_url().unwrap_or(""),
```
MarijnS95, line 56: If we're going all-out, `.unwrap_or_default()` :grimacing:

GuillaumeGomez: :laughing:

src/codegen/general.rs
```diff
@@ -16,32 +16,48 @@ use std::{
...
         "Generated by gir (https://github.com/gtk-rs/gir @ {})
-from gir-files (https://github.com/gtk-rs/gir-files @ {})",
-        VERSION, conf.girs_version
+{}",
+        VERSION,
+        conf.girs_version
+            .iter()
+            .map(|info| {
+                format!(
+                    "from {} ({}@ {})\n",
```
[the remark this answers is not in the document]
MarijnS95: Or otherwise add it in the format string below. Looks like this now:

```
from /home/marijn/Code/aanmelder/gstreamer-rs/gir-files (https://github.com/gtk-rs/gir-files@ 21f7670)
from /home/marijn/Code/aanmelder/gstreamer-rs/gst-gir-files (@ ???)
```

MarijnS95:
> In the last update, I also canonicalized the gir dir path so it makes more beautiful paths. :)

Canonicalization always gives an absolute path, that'll change depending on where the user has the repo checked out. Seeing `/home/marijn/.../gstreamer-rs/gir-files` might not be beautiful for everyone :grimacing:

GuillaumeGomez:
Anything else to be done here?

bilelmoussaoui:
Never mind, I forgot to update the generator script

MarijnS95:
@GuillaumeGomez You might want to fix this first: https://github.com/gtk-rs/gir/pull/1041#discussion_r590543201 :grimacing:

src/codegen/general.rs
MarijnS95: The easy fix is of course:

```diff
diff --git a/src/codegen/general.rs b/src/codegen/general.rs
index e956c4f..de75ddd 100644
--- a/src/codegen/general.rs
+++ b/src/codegen/general.rs
@@ -18,7 +18,7 @@ pub fn start_comments(w: &mut dyn Write, conf: &Config) -> Result<()> {
     if conf.single_version_file.is_some() {
         start_comments_no_version(w, conf)
     } else {
-        single_version_file(w, conf)
+        single_version_file(w, conf, "// ")
     }
 }

@@ -41,17 +41,19 @@ pub fn start_comments_no_version(w: &mut dyn Write, conf: &Config) -> Result<()>
     )
 }

-pub fn single_version_file(w: &mut dyn Write, conf: &Config) -> Result<()> {
+pub fn single_version_file(w: &mut dyn Write, conf: &Config, prefix: &str) -> Result<()> {
     write!(
         w,
-        "Generated by gir (https://github.com/gtk-rs/gir @ {})
+        "{}Generated by gir (https://github.com/gtk-rs/gir @ {})
 {}",
+        prefix,
         VERSION,
         conf.girs_version
             .iter()
             .map(|info| {
                 format!(
-                    "from {} ({}@ {})\n",
+                    "{}from {} ({}@ {})\n",
+                    prefix,
                     info.gir_dir.display(),
                     info.get_repository_url()
                         .map_or_else(String::new, |u| format!("{} ", u)),
diff --git a/src/codegen/mod.rs b/src/codegen/mod.rs
index 6303210..177e3be 100644
--- a/src/codegen/mod.rs
+++ b/src/codegen/mod.rs
@@ -72,7 +72,7 @@ pub fn generate_mod_rs(env: &Env, root_path: &Path, mod_rs: &[String], traits: &
 pub fn generate_single_version_file(env: &Env) {
     if let Some(ref path) = env.config.single_version_file {
         save_to_file(path, env.config.make_backup, |w| {
-            general::single_version_file(w, &env.config)
+            general::single_version_file(w, &env.config, "")
         });
     }
 }
```

MarijnS95: We lost `// DO NOT EDIT` as well, btw.

src/codegen/general.rs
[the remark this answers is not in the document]
GuillaumeGomez: Weird, it shouldn't be the case:

```rust
info.get_repository_url()
    .map_or_else(String::new, |u| format!("{} ", u)),
```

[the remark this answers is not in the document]
MarijnS95: I think you pushed a fix for this right as I was posting this review. Let me re-check, should be good now.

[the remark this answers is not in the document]
MarijnS95: Yeah, looks good now:

```
+// Generated by gir (https://github.com/gtk-rs/gir @ 2a4d7f6+)
+// from gir-files (https://github.com/gtk-rs/gir-files @ 21f7670)
+// from gst-gir-files (@ ???)
```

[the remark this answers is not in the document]
GuillaumeGomez: Yeay! \o/

GuillaumeGomez:
@MarijnS95 I also applied your patch to add the missing "//".

MarijnS95:
@GuillaumeGomez Neat! I'll re-check and set up a prospective MR over at GStreamer-rs.

You might want to fix up remaining `girs_dir` mentions in `tests/sys` in this repo.

GuillaumeGomez:
I ocmpletely forgot about those tests, shame on me! Thanks for the reminder!

README.md
GuillaumeGomez: You only need to write it once, you'll survive. :p And better extra-verbose than super confused.

MarijnS95: I've just removed this entirely because it makes little sense:

- Hard to do proper path "canonicalization" (without getting a full root path);
- No need to find-replace the `../gir-files`, `../../gir-files` and `../../../gir-files` variants;
- Already passing these paths with `-d` in `generator.py`, those are relative to the root directory already.
  Besides, these _override_ the value set by `girs_dirs`!!(!!)

GuillaumeGomez:
So, what do we do in here?

MarijnS95 (approved):
@GuillaumeGomez I guess this is all okay now and we can continue with #1068. The projects using `gir` will have to see whether they switch to `gir_directories = ["../../../somewhere"]` or (better, IMO) directly use the `-d` flag from `generator.py`, like https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/719/diffs?commit_id=670976e745d074ef5b70c40c1843371e4f033ac8.

GuillaumeGomez:
Ok, then merging and sending PRs to gtk-rs and gtk4-rs. Thanks for your reviews!

MarijnS95:
@GuillaumeGomez PR's are already including review, at least gtk4-rs: https://github.com/gtk-rs/gtk4-rs/pull/212

PR #1043, opened 2021-01-29
analysis: Move global imports to object-specific analysis and under versioning constraints

MarijnS95:
While adding some new classes to gstreamer-rs gir keeps emitting `use glib::translate::*;` for classes that do not have any members that could possibly use translation functionality, resulting in obvious "unused import" linter warnings. ~Simply adding this import only if there are members (functions, signals or properties) solves that problem for now.~ All these cases have been addressed by adding `with_defaults` import wrappers, and importing "global" imports under these so that they have the appropriate version constraints. In cases where no members exist (or they are commented out) this code does not even run at all, and the import is completely omitted (the original issue this PR was about).

### DIFF
--- a/src/analysis/enums.rs
+++ b/src/analysis/enums.rs
@@ -46,6 +46,7 @@ pub fn new(env: &Env, obj: &GObject, imports: &mut Imports) -> Option<Info> {
     imports.add_defined(&format!("crate::{}", name));
 
     let imports = &mut imports.with_defaults(enumeration.version, &None);
+    imports.add("glib::translate::*");
 
     let has_get_quark = enumeration.error_domain.is_some();
     if has_get_quark {
--- a/src/analysis/enums.rs
+++ b/src/analysis/enums.rs
@@ -45,6 +45,8 @@ pub fn new(env: &Env, obj: &GObject, imports: &mut Imports) -> Option<Info> {
     // Mark the type as available within the enum namespace:
     imports.add_defined(&format!("crate::{}", name));
 
+    let imports = &mut imports.with_defaults(enumeration.version, &None);
+
     let has_get_quark = enumeration.error_domain.is_some();
     if has_get_quark {
         imports.add("glib::Quark");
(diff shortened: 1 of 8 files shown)


MarijnS95:
This isn't solving all of it though. In a different bringup there are classes (3 of them are shared classes, one is an object) with parameters that cannot be generated (`/*Ignored*/` and `/*Unimplemented*/`). Probably need to find a solution to those before merging as well..

sdroege:
> Probably need to find a solution to those before merging as well.

Should be enough to have a separate bool for those that also checks in addition if the item should be commented out, or not?

MarijnS95:
> Should be enough to have a separate bool for those that also checks in addition if the item should be commented out, or not?

@sdroege The total endeavour was more fun than that, but all my `"unused import"` errors in `gstreamer-rs` have been resolved now and there are quite a few nice fixes in it for `gtk-rs` as well! Not that those aren't hidden by `#[allow(unused_imports)]` though...

sdroege:
>  [@MarijnS95](/MarijnS95) [MarijnS95](/MarijnS95) marked this pull request as draft [16 minutes ago](#event-4269040387)

What's left to be done to make it complete? Can you put a small todo list into the issue description?

MarijnS95:
@sdroege Noticed that after moving the `glib::translate` import to `signals`/`properties`/`special_functions` it can and should simply be removed from `analysis/object`. Wanted to test and implement that theory before someone got to the merge button, hence marked it as draft :)

MarijnS95:
Cool, the tests for gtk-rs and all clippy tests for gstreamer-rs (where `#[allow(unused_imports)]` has been removed) succeed: https://gitlab.freedesktop.org/MarijnS95/gstreamer-rs/-/pipelines/263964.

Note that somehow `gstreamer-base` was forgotten about, which still had a bunch of unconstrained imports in `enums.rs`. Should all be fixed now!

sdroege (approved):
Looks good to me, thanks!

@GuillaumeGomez ?

GuillaumeGomez (approved):
That's super cool! Thanks!

PR #1044, opened 2021-01-29
codegen: Omit mut_override from boxed copy if parameter is *const

MarijnS95:
While not harmful at all (clippy doesn't even flinch when writing `as *mut _` where a `*const` is required; not to mention opaque `mut_override` function calls) it is misleading to see `mut_override` in calls like `copy` where this is not required at all.

---

This is a rather quick and gross bit of duct-tape for something that's not really harming anything but readability and manual analysis of autogenerated files; feedback and suggestions welcome!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -23,9 +23,9 @@ checksum = "cf1de2fe8c75bc145a2f577add951f8134889b4795d47466a54a5c846d691693"
 
 [[package]]
 name = "cfg-if"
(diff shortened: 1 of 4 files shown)


sdroege (approved):

sdroege:
@GuillaumeGomez What do you think? Seems like a small improvement at least :)

GuillaumeGomez (approved):

Issue #1047, opened 2021-02-02
Rework the ABI tests for sys crates


pbor:
When adding the Windows CI (with VisualStudio, non mingw), I noticed that the ABI tests fail. At a first glance this is because the do not find a C compiler.

I spent a little bit of time looking into it, but it seems it is a pretty deep rabbit hole, so I am putting some notes on a possible plan here so that it can be discussed/reviewed before starting on it.

Observations:

1. The code compiles C programs when running `cargo test`, so it has a `Compiler` struct and it invokes `pkg-config` manually. That abstraction does not have all the complexity required to handle different OSes, cross-compilation etc. Ideally we should use the `cc` and the `pkg-config` crates which already have all of that
2. The `cc` crate however only builds C libraries, not C executables
3. The `cc` and `pkg-config` crates are designed to be run from `build.rs` and rely on the env set by `cargo`. I managed to make a small experiment forwarding env vars from `build.rs` to the test code and then run `cc` there, but this seems ugly.
4. There is no way to have code in `build.rs` which is conditional to tests. This is by design since the build should be the same for tests and normal builds. It is however possible to check env vars to take conditional paths based on cargo features.

Possible plan:
1. Refactor the C tests in a single C library instead of an executable per test
2. Add a `abi_tests` feature to the generated `Cargo.toml`. This will be off for the library users, while can be turned on by CI
3. Use `cc` crate from `build.rs`, conditionally to the feature to build the C libary
4. Use ffi in the abi tests to call the C library

pbor:
Sebastian suggested it might be worth looking at bindgen (not cbindgen): it can generate various ABI tests for the stuff it binds, we can probably learn from that (and it tests more than our minimal tests currently)

PR #1050, opened 2021-02-06
codegen: rework ABI tests to use a single C program for each test

pbor:
This is a first step in the plan outlined in https://github.com/gtk-rs/gir/issues/1047

Instead of compiling a C program for each layout and one for each
constant to check, generate a single layout.c and a single
constant.c which output all the values, one per line.
On the rust side, it is easier to compile and run the C program
just once and then parse its output one line at a time to compare
with the corresponding rust record.

Note also that C programs are now taken by the system's header
files (eg atk.h) rather than from `-DATK_FOO=1` passed by the
test itself. I think this is actually the right thing to check
since that is the real value we want to compare with.
However this also means that the test will fail when run on
a system older than the one used to generate the gir files, since
some constants may not present in the headers.

### DIFF
--- a/src/codegen/sys/tests.rs
+++ b/src/codegen/sys/tests.rs
@@ -317,6 +317,8 @@ impl Compiler {
     pub fn new() -> Result<Compiler, Box<dyn Error>> {
         let mut args = get_var("CC", "cc")?;
         args.push("-Wno-deprecated-declarations".to_owned());
+        // For _Generic
+        args.push("-std=c11".to_owned());
         // For %z support in printf when using MinGW.
         args.push("-D__USE_MINGW_ANSI_STDIO".to_owned());
         args.extend(get_var("CFLAGS", "")?);
--- a/src/codegen/sys/tests.rs
+++ b/src/codegen/sys/tests.rs
@@ -46,12 +46,12 @@ pub fn generate(env: &Env, crate_name: &str) {
 
     let layout_c = tests.join("layout.c");
     save_to_file(&layout_c, env.config.make_backup, |w| {
-        generate_layout_c(env, &layout_c, w)
+        generate_layout_c(env, &layout_c, w, &ctypes)
     });
 
     let constant_c = tests.join("constant.c");
     save_to_file(&constant_c, env.config.make_backup, |w| {
-        generate_constant_c(env, &constant_c, w)
+        generate_constant_c(env, &constant_c, w, &cconsts)
     });
 
     let abi_rs = tests.join("abi.rs");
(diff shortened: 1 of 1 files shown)


pbor:
A successful run after a regen can be seen here: https://github.com/pbor/gtk-rs/runs/1845572278?check_suite_focus=true

sdroege:
Can you put a WIP PR of gtk-rs up with the changes this would do?

sdroege:
> Note also that C programs are now taken by the system's header
> files (eg atk.h) rather than from `-DATK_FOO=1` passed by the
> test itself. I think this is actually the right thing to check
> since that is the real value we want to compare with.
> However this also means that the test will fail when run on
> a system older than the one used to generate the gir files, since
> some constants may not present in the headers.

This was also the case before as mentioned on IRC, but AFAIU you unconfused yourself since then? :)

src/codegen/sys/tests.rs
```diff
@@ -233,27 +246,37 @@ fn generate_constant_c(env: &Env, path: &Path, w: &mut dyn Write) -> io::Result<
...
-                    unsigned long long: "###gir test###%llu###gir test###\n",
-                    double: "###gir test###%f###gir test###\n",
-                    long double: "###gir test###%ld###gir test###\n"),
-           ABI_CONSTANT_NAME);
-    return 0;
-}"####
-    )
+#define PRINT_CONSTANT(CONSTANT_NAME) \
+    printf("%s;", #CONSTANT_NAME); \
+    printf(_Generic((CONSTANT_NAME), \
```
sdroege, line 251: OOC, does MSVC support that? It's "new" in C11.

Also we should probably make sure to pass `--std=c11` or so to the compiler.

src/codegen/sys/tests.rs
```diff
@@ -403,111 +407,102 @@ impl Results {
...
 #[test]
 fn cross_validate_constants_with_c() {
-    let tmpdir = Builder::new().prefix("abi").tempdir().expect("temporary directory");
-    let cc = Compiler::new().expect("configured compiler");
+    let mut c_constants: Vec<(String, String)> = Vec::new();
+
+    for l in get_c_output("constant").unwrap().lines() {
+        let mut words = l.trim().split(";");
+        let name = words.next().unwrap().to_owned();
+        let value = words.next().unwrap().parse().unwrap();
```
sdroege, line 415: Too bad we don't have https://github.com/rust-lang/rust/issues/74674 yet :)

sdroege: You probably want to replace the `unwrap`s with `expect`s

src/codegen/sys/tests.rs
```diff
@@ -403,111 +407,102 @@ impl Results {
...
+        c_constants.push((name, value));
+    }
+
+    let mut results: Results = Default::default();
+
+    for (i, (&(rust_name, rust_value), &(ref c_name, ref c_value))) in
```
sdroege, line 421: The pattern here looks a bit strange. Why all the `&` and `ref`? :)

src/codegen/sys/tests.rs
```diff
@@ -403,111 +407,102 @@ impl Results {
...
+    let mut c_constants: Vec<(String, String)> = Vec::new();
+
+    for l in get_c_output("constant").unwrap().lines() {
+        let mut words = l.trim().split(";");
+        let name = words.next().unwrap().to_owned();
+        let value = words.next().unwrap().parse().unwrap();
+        c_constants.push((name, value));
+    }
+
+    let mut results: Results = Default::default();
```
sdroege, line 419:
```suggestion
    let mut results = Results::default();
```

Maybe?

sdroege:
Looks OK otherwise, even simpler and cleaner than before in `gir` and the generated code will be so much faster.

src/codegen/sys/tests.rs
pbor: Do you mean `_Generic`? note that it was there before as well. I have not started to worry about MSVC yet :)

sdroege: Yes

pbor:
> Can you put a WIP PR of gtk-rs up with the changes this would do?

https://github.com/gtk-rs/gtk-rs/pull/293

sdroege (approved):
Looks good to me apart from https://github.com/gtk-rs/gtk-rs/pull/293#discussion_r571588226

@GuillaumeGomez ?

src/codegen/sys/tests.rs
```diff
@@ -202,58 +202,80 @@ fn generate_manual_h(env: &Env, path: &Path, w: &mut dyn Write) -> io::Result<()
...
+                    long double: "%ld"), \
+           CONSTANT_NAME); \
+    printf("\n");
+"####
+    )?;
+
+    writeln!(w, "{}", r"int main() {")?;
+
+    for cconst in cconsts {
+        writeln!(w, "\tPRINT_CONSTANT({name});", name = cconst.name,)?;
```
GuillaumeGomez, line 274: You used whitespaces before, please don't use tabs here.

PR #1051, opened 2021-02-06
Add documentation in case this is an abstract class

### LINKED ISSUE #966: Add note when a type is `abstract = 1` in documentation
Something like:

```
"Is a Abstract Base Class, you can't instantiate it" ?
and if you try, you'll get a critical warning and NULL
```

GuillaumeGomez:
Fixes #966.


sdroege:
Thanks, that reminded me of creating an issue for https://gitlab.gnome.org/GNOME/glib/-/issues/2321

sdroege (approved):

PR #1052, opened 2021-02-07
WIP: move abi checks to a feature and build the C code at build time

pbor:
The abi tests are now conditional to the abi-tests feature,
when that is enabled, the C code for the abi tests is now built
at compile time, using the cc crate rather than having our own
Compiler abstraction and having to compile stuff at runtime and
deal with temp directories etc.


src/codegen/sys/build.rs
```diff
@@ -55,10 +60,43 @@ fn main() {} // prevent linking libraries to avoid documentation failure
...
-    if let Err(s) = system_deps::Config::new().probe() {
+    let libs = system_deps::Config::new().probe();
+    if let Err(s) = libs {
         println!("cargo:warning={}", s);
         process::exit(1);
     }
+
+    let libs = libs.unwrap();
+
+    if env::var("CARGO_FEATURE_ABI_TESTS").is_ok() {
```
sdroege, line 71: Why not `#[cfg("abi-tests")]`?

src/codegen/sys/tests.rs
```diff
@@ -213,20 +213,45 @@ fn generate_layout_c(
...
 
-    writeln!(w, "    return 0;")?;
-    writeln!(w, "{}", r"}")
+    writeln!(
+        w,
+        "{}",
+        r####"};
+
+const Layout *c_layouts(size_t *n) {
+    *n = sizeof(LAYOUTS) / sizeof(Layout);
```
sdroege, line 251:
```suggestion
    *n = G_N_ELEMENTS(LAYOUTS);
```

src/codegen/sys/tests.rs
```diff
@@ -287,98 +342,86 @@ fn generate_abi_rs(
...
-        .unwrap_or_else(|| OsString::from("pkg-config"));
-    let mut cmd = Command::new(pkg_config);
-    cmd.arg("--cflags");
-    cmd.args(packages);
-    let out = cmd.output()?;
-    if !out.status.success() {
-        return Err(format!("command {:?} returned {}",
-                           &cmd, out.status).into());
+    unsafe {
+        let mut n = 0;
```
sdroege, line 383: Or `mem::MaybeUninit` if you feel like it

sdroege:
Seems like a good idea

src/codegen/sys/tests.rs
pbor: dunno... just init to 0 seem easier/cleaner, but no strong opinion

src/codegen/sys/tests.rs
pbor: This file does not include glib... I would keep it minimal C, but no strong opinion

sdroege: The other C code uses `g_strdup()` instead of `strdup()`, etc., that's why I was suggesting to use glib API here. Your call :)

pbor: yeah... in the other test I had to pull in glib, because C without `g_strdup_printf` is a nightmare :)
While it is true that we are pulling in glib, I would still keep this file as minimal as possible.

pbor:
This currently fails for gdk-pixbuf due to https://github.com/gdesmott/system-deps/issues/28

With regard to Windows CI, the good news is that this approach works and the C programs are compiled and the layout test work, the bad news is the MSVC does not support `_Generic` so the current code in `constant.c` does not work... I am not sure if we have a better idea about that (use C++?) or simply skip that test on windows

sdroege:
It's available since Visual Studio 2019 version 16.7.0 from what I can see, and you can enable it via `/std:c11`. Should maybe just require that version for running the tests? What version is available with GitHub actions?

pbor:
> It's available since Visual Studio 2019 version 16.7.0 from what I can see, and you can enable it via `/std:c11`. Should maybe just require that version for running the tests? What version is available with GitHub actions?

Actually it seems that the Visual Studio version is new enough and does have `_Generic`, but however it does not like how we use it:

```
type 'char' in _Generic association compatible with previous association type 'char'
```

sdroege:
That means we have `char` in there twice, which we probably shouldn't and it's gcc's fault to not complain about that

pbor:
the problem is that we have:

```
        char: "%c", \
        signed char: "%hhd", \
        unsigned char: "%hhu", \
```

so we are printing them differently

sdroege:
Ah that's annoying and otherwise needs conditional compilation for platforms where signed or unsigned chars are the default...

sdroege:
OTOH, only the two cases (signed / unsigned) should be there. `char` alone is not really needed

pbor:
That has helped, but now it complains about not finding `<stdalign.h>`, which as far as I understand should be there in MSVC when using c11. Meh.

PR #1054, opened 2021-02-10
sys: use system-deps 3.0

gdesmott:
The return type of the probe() method changed, so this will affect user
code doing manual work in their build.rs, the one we generate does not.

I tried regenerating `gtk-rs` and it build fine.

### DIFF
--- a/src/codegen/sys/cargo_toml.rs
+++ b/src/codegen/sys/cargo_toml.rs
@@ -83,7 +83,7 @@ fn fill_in(root: &mut Table, env: &Env) {
 
     {
         let build_deps = upsert_table(root, "build-dependencies");
-        set_string(build_deps, "system-deps", "2.0");
+        set_string(build_deps, "system-deps", "3");
     }
 
     {


src/codegen/sys/cargo_toml.rs
```diff
@@ -83,7 +83,7 @@ fn fill_in(root: &mut Table, env: &Env) {
 
     {
         let build_deps = upsert_table(root, "build-dependencies");
-        set_string(build_deps, "system-deps", "2.0");
+        set_string(build_deps, "system-deps", "3.0");
```
gdesmott, line 86: Maybe we could just use `3` here? `3.x` will stay API compatible.

sdroege: I don't mind either way, but IIRC @GuillaumeGomez has opinions

sdroege: Also `3.0` and `3` are absolutely equivalent in how they're handled by cargo

sdroege (approved):

src/codegen/sys/cargo_toml.rs
GuillaumeGomez: Then let's just use `3`. :)

sdroege: @gdesmott You update it?

GuillaumeGomez (approved):

PR #1055, opened 2021-02-14
doc_alias: don't generate if the enum c_name = rust_name

bilelmoussaoui:
Avoids `#[doc(alias = "...")]` is the same as the item's name errors


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #1056, opened 2021-02-15
Update fmt

GuillaumeGomez:


sdroege:
What happened here?

GuillaumeGomez:
The stable version changed (I realized that when looking at #1055).

sdroege:
> The stable version changed (I realized that when looking at #1055).

Indeed, that somehow changed between 1.49 and 1.50. Interesting

PR #1057, opened 2021-02-17
Consider nullability of out parameters in normal and async functions

sdroege:
Both commits needed for https://github.com/gtk-rs/gtk-rs/pull/307

### DIFF
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -97,7 +97,10 @@ impl Bounds {
                     let func_name = func.c_identifier.as_ref().unwrap();
                     let finish_func_name = finish_function_name(func_name);
                     if let Some(function) = find_function(env, &finish_func_name) {
-                        let mut out_parameters = find_out_parameters(env, function);
+                        // FIXME: This should work completely based on the analysis of the finish() function
+                        // but that a) happens afterwards and b) is not accessable from here either.
+                        let mut out_parameters =
+                            find_out_parameters(env, function, configured_functions);
                         if use_function_return_for_result(
                             env,
                             function.ret.typ,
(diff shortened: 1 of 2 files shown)


sdroege:
Generated code is also in that PR.

src/analysis/bounds.rs
```diff
@@ -97,7 +97,10 @@ impl Bounds {
                     let func_name = func.c_identifier.as_ref().unwrap();
                     let finish_func_name = finish_function_name(func_name);
                     if let Some(function) = find_function(env, &finish_func_name) {
-                        let mut out_parameters = find_out_parameters(env, function);
+                        // FIXME: This should work completely based on the analysis of the finish() function
+                        // but that a) happens afterwards and b) is not accessable from here either.
```
GuillaumeGomez, line 101:
```suggestion
                        // but that a) happens afterwards and b) is not accessible from here either.
```

src/analysis/bounds.rs
```diff
@@ -312,7 +319,16 @@ fn find_out_parameters(env: &Env, function: &Function) -> Vec<String> {
                 && param.direction == ParameterDirection::Out
                 && param.name != "error"
         })
-        .map(|(_, param)| rust_type_nullable(env, param.typ, param.nullable).into_string())
+        .map(|(_, param)| {
+            // FIXME: This should work completely based on the analysis of the finish() function
+            // but that a) happens afterwards and b) is not accessable from here either.
```
GuillaumeGomez, line 324:
```suggestion
            // but that a) happens afterwards and b) is not accessible from here either.
```

GuillaumeGomez (approved):

PR #1060, opened 2021-02-20
signals: rename emit to emit_by_name

bilelmoussaoui:
ObjectExt::emit should take a SignalId instead of the signal name
see https://developer.gnome.org/gobject/stable/gobject-Signals.html#g-signal-emit-by-name

Needed for https://github.com/gtk-rs/gtk-rs/pull/302


sdroege (approved):
Good to go once the other PR is ready

PR #1061, opened 2021-02-24
codegen: Avoid a dead code warning in ABI tests

heftig:
When there are no C types, we don't need the crate imports.


heftig:
Hmm, actually the entire test should be skipped when there aren't any types to test.

PR #1062, opened 2021-02-26
Add support for connecting to signals with details

zec:
Some GObject signals use a non-default [_detail_ argument](https://developer.gnome.org/gobject/stable/signal.html#signal-detail). This allows one to specify that a closure should only be invoked on a subset of that signal's emissions.

There is an attribute in the GIR annotations that specifies when a signal uses non-default values of _detail_. As of time of writing, the code generated by the `gir` tool does not allow one to specify the value of _detail_ in the relevant connect methods&mdash;the closure passed to `connect_<signal name>` will get invoked on all emissions of that signal, regardless of the value of _detail_. This PR contains an attempt at adding such support in. For signals that use _detail_, the relevant method's signature is changed from, e.g.,

```rust
fn connect_state_change<F: Fn(&Self, &str, bool) + 'static>(
    &self,
    f: F
) -> SignalHandlerId
```
to

```rust
fn connect_state_change<F: Fn(&Self, &str, bool) + 'static>(
    &self,
    detail: Option<&str>,
    f: F
) -> SignalHandlerId
```

`Option<&str>` is used because not specifying a _detail_ value in `g_signal_connect*` is valid in these cases (it means the closure will be invoked regardless of the value of _detail_).

This has been tested out on code generation for gtk-rs/gtk-rs and gtk-rs/webkit2gtk-rs, and it produces the expected changes. I have even tested out the use of `Some(_)` and `None` for such a modified connection method (`webkit2gtk::UserContentManagerExt::connect_script_message_received`) and it appears to work as desired.

Some possible issues with my PR as it stands:

- This makes backward-incompatible changes to the API for these signals. Is that OK? If not, what approach(es) would be preferable?
- Is `Option<&str>` the best type to specify the _detail_ value with, or would a bespoke enum serve better?
- This is my first time working in this codebase. I wouldn't mind a sanity check of my work.

Constructive critique of this PR is appreciated!

### DIFF
--- a/src/codegen/signal_body.rs
+++ b/src/codegen/signal_body.rs
@@ -5,6 +5,7 @@ pub struct Builder {
     signal_name: String,
     trampoline_name: String,
     in_trait: bool,
+    is_detailed: bool,
 }
 
 impl Builder {
@@ -24,6 +25,11 @@ impl Builder {
 
     pub fn in_trait(&mut self, value: bool) -> &mut Builder {
         self.in_trait = value;
+        self
+    }
+
+    pub fn is_detailed(&mut self, value: bool) -> &mut Builder {
+        self.is_detailed = value;
         self
     }
 
@@ -55,6 +61,7 @@ impl Builder {
             signal: self.signal_name.clone(),
             trampoline: self.trampoline_name.clone(),
             in_trait: self.in_trait,
+            is_detailed: self.is_detailed,
         }
     }
 }
--- a/src/writer/to_code.rs
+++ b/src/writer/to_code.rs
@@ -128,17 +128,34 @@ impl ToCode for Chunk {
                 ref signal,
                 ref trampoline,
                 in_trait,
+                is_detailed,
             } => {
-                let s1 = format!(
-                    "connect_raw(self.as_ptr() as *mut _, b\"{}\\0\".as_ptr() as *const _,",
-                    signal
-                );
+                let mut v: Vec<String> = Vec::with_capacity(6);
+                if is_detailed {
+                    v.push(format!(
+                        r#"let detailed_signal_name = detail.map(|name| {{ format!("{0}::{{}}\0", name) }});"#,
+                        signal
+                    ));
+                    v.push(format!(
+                        r#"let signal_name: &[u8] = detailed_signal_name.as_ref().map_or(&b"{0}\0"[..], |n| n.as_bytes());"#,
+                        signal
+                    ));
+                    v.push(
+                        "connect_raw(self.as_ptr() as *mut _, signal_name.as_ptr() as *const _,"
+                            .to_string(),
+                    );
+                } else {
+                    v.push(format!(
+                        "connect_raw(self.as_ptr() as *mut _, b\"{}\\0\".as_ptr() as *const _,",
+                        signal
+                    ));
+                }
                 let self_str = if in_trait { "Self, " } else { "" };
-                let s2 = format!(
+                v.push(format!(
                     "\tSome(transmute::<_, unsafe extern \"C\" fn()>({}::<{}F> as *const ())), Box_::into_raw(f))",
                     trampoline, self_str
-                );
-                vec![s1, s2]
+                ));
+                v
             }
             Name(ref name) => vec![name.clone()],
             ExternCFunc {
--- a/src/analysis/properties.rs
+++ b/src/analysis/properties.rs
@@ -270,6 +270,7 @@ fn analyze_property(
                     destroy: None,
                 },
                 is_action: false,
+                is_detailed: false, // well, technically this *is* an instance of a detailed signal, but we "pre-detailed" it
                 version: prop_version,
                 deprecated_version: prop.deprecated_version,
                 doc: None,
@@ -301,6 +302,7 @@ fn analyze_property(
                 version: prop_version,
                 deprecated_version: prop.deprecated_version,
                 doc_hidden: false,
+                is_detailed: false, // see above comment
             })
         } else {
             None
--- a/src/analysis/signals.rs
+++ b/src/analysis/signals.rs
@@ -17,6 +17,7 @@ pub struct Info {
     pub version: Option<Version>,
     pub deprecated_version: Option<Version>,
     pub doc_hidden: bool,
+    pub is_detailed: bool,
 }
 
 pub fn analyze(
@@ -115,6 +116,7 @@ fn analyze_signal(
         version,
         deprecated_version,
         doc_hidden,
+        is_detailed: signal.is_detailed,
     };
 
     info
--- a/src/chunk/chunk.rs
+++ b/src/chunk/chunk.rs
@@ -58,6 +58,7 @@ pub enum Chunk {
         signal: String,
         trampoline: String,
         in_trait: bool,
+        is_detailed: bool,
     },
     Name(String),
     ExternCFunc {
--- a/src/codegen/signal.rs
+++ b/src/codegen/signal.rs
@@ -191,7 +191,11 @@ fn function_type_string(
 
 fn declaration(analysis: &analysis::signals::Info, function_type: &Option<String>) -> String {
     let bounds = bounds(function_type);
-    let param_str = "&self, f: F";
+    let param_str = if !analysis.is_detailed {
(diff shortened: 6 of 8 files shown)


sdroege:
Thanks for the change, I think that's a useful addition.

> This makes backward-incompatible changes to the API for these signals. Is that OK? If not, what approach(es) would be preferable?

That's generally ok, yes.

> Is `Option<&str>` the best type to specify the _detail_ value with, or would a bespoke enum serve better?

What kind of enum are you thinking of?

----

To make reviewing this and understanding the impact a bit easier, can you also submit a PR to `gtk-rs/gtk-rs` that regenerates everything with this version of gir? Thanks :)

src/writer/to_code.rs
```diff
@@ -128,17 +128,33 @@ impl ToCode for Chunk {
...
-                    signal
-                );
+                let mut v: Vec<String> = Vec::with_capacity(4);
+                if is_detailed {
+                    v.push(format!(
+                        r#"let detailed_signal_name = detail.map(|name| {{ format!("{0}::{{}}\0", name) }});"#,
+                        signal
+                    ));
+                    v.push(format!(
+                        r#"let signal_name_ptr: *const u8 = (if let Some(ref name) = detailed_signal_name {{ name.as_bytes() }} else {{ &b"{0}\0"[..] }}).as_ptr();"#,
```
MarijnS95, line 140:
```suggestion
                        r#"let signal_name_ptr = detailed_signal_name.as_ref().map_or(&b"{0}\0"[..], |name| name.as_bytes());"#,
```

Something like that should be a bit more idiomatic in Rust. Then move the `as_ptr()` into the function call to loose the lifetime as late as possible.

src/writer/to_code.rs
```diff
@@ -128,17 +128,33 @@ impl ToCode for Chunk {
...
+                    ));
+                }
+                v.push(if !is_detailed {
+                    format!(
+                        "connect_raw(self.as_ptr() as *mut _, b\"{}\\0\".as_ptr() as *const _,",
+                        signal
+                    )
+                } else {
+                    "connect_raw(self.as_ptr() as *mut _, signal_name_ptr as *const _,".to_string()
+                });
```
MarijnS95, line 151: Perhaps you could just merge this with the `if` block above, and add an `else` for it?

```suggestion
                    v.push("connect_raw(self.as_ptr() as *mut _, signal_name_ptr.as_ptr() as *const _,".to_string());
                } else {
                    v.push(format!(
                        "connect_raw(self.as_ptr() as *mut _, b\"{}\\0\".as_ptr() as *const _,",
                        signal
                    ));
                }
```

zec:
> > Is `Option<&str>` the best type to specify the _detail_ value with, or would a bespoke enum serve better?
> 
> What kind of enum are you thinking of?

Well, what I'm wondering there is that the variant name `None` might be a little misleading in this context. When you pass in `None` as the `detail` argument, the semantics are actually "get invoked for _any_ value of _detail_", not "get invoked for _no_ value of _detail_". As an alternative, I was thinking something along the lines of:

```rust
pub enum DetailSpecification_ExactNamesTBD {
    /// Get invoked when the detail argument matches the `str`
    Specific(&str),
    /// Get invoked for any value of the detail argument
    Any
}
```

The advantage of a new enum is that the variant names match the semantics of this use-case. The advantage of using `Option<&str>` is that it's familiar to everyone who's written non-trivial Rust.

> To make reviewing this and understanding the impact a bit easier, can you also submit a PR to `gtk-rs/gtk-rs` that regenerates everything with this version of gir? Thanks :)

I'll try to make that today (UTC-5).

src/writer/to_code.rs
zec: That makes sense. I'll unify those two conditionals.

src/writer/to_code.rs
zec: Heh, at one point in my development I had something similar at that point, then wondered if it was too convoluted. I'll make something like that change, and clarify the lifetime of `detailed_signal_name` as well.

src/writer/to_code.rs
MarijnS95: :+1: Sorry for CI noise, this diff already included the move of `.as_ptr()` to the call site, which requires the other diff to be applied as well.

zec: No prob, will make sure all's good in the next commit.

src/writer/to_code.rs
MarijnS95: That's fair. I always prefer `map_or` over `if let` as it has a nice default case, but always get bitten by the missing `.as_ref()` to not move out of the `Option`.

I don't think it's necessary to "clarify" the lifetime, just as long as `signal_name_ptr` is _not_ a raw pointer. That way it carries the lifetime of `detailed_signal_name` during the remainder of the function scope, disallowing that to be moved or mutated.

zec: See what you think with commit 65cb60e50efa.

MarijnS95: You don't need the explicit drop at the end, `detailed_signal_name` will go out of scope anyway. Rather, my suggestion to solve it this way (by making `signal_name` a `&[u8]` borrow) is to disallow `detailed_signal_name` from being mutably borrowed or moved at the same time. By the way, you don't need to explicitly specify the `&[u8]` type, as shown in the suggestion above.

zec: My thought with the drop is that it makes it explicit that `detailed_signal_name` outlasts the call to `connect_raw()`, but if you don't think it needs to be explicit, I can change it back.

MarijnS95: It doesn't need to be explicit, and by placing `.as_ptr()` inside the call expression (rather than before it) that is already guaranteed.

zec: OK, changed in 3d8dc87.

MarijnS95: Cool, thanks! I'm not a maintainer (can't approve this PR in a sensible way), but they'll likely ask you to squash the patches.

sdroege: Yes, please clean up the commit history a bit :)

zec: OK, I'll do that once everyone's done reviewing changes.

sdroege:
@bilelmoussaoui Does this do the right thing on `gtk4`?

sdroege (approved):

sdroege:
@MarijnS95 Thanks for the review, there was nothing left for me to complain about :)

bilelmoussaoui (approved):

bilelmoussaoui:
> @bilelmoussaoui Does this do the right thing on `gtk4`?

Seems to work fine

bilelmoussaoui:
I wonder why the `ObjectExt:: emit_with_details` takes the details as `glib::Quark` while we use a `str` here, does it make sense?

sdroege:
> I wonder why the `ObjectExt:: emit_with_details` takes the details as `glib::Quark` while we use a `str` here, does it make sense?

`emit_by_name()` (the string version) can also take the detail as part of the string (`"signal_name::detail"`). `emit_with_details()` works with the signal id and quark. Usually if you care about using the signal id for performance reasons, you would also care about caching the quark for the detail string for multiple calls.

Using a quark here in this PR would make it harder for the majority of callers.

sdroege:
Waiting for @GuillaumeGomez and then let's get this in :)

MarijnS95 (approved):

src/codegen/signal_body.rs
```diff
@@ -27,6 +28,11 @@ impl Builder {
         self
     }
 
+    pub fn is_detailed(&mut self, value: bool) -> &mut Builder {
```
GuillaumeGomez, line 31: That's a weird name for a setter... What about `detailed` or something along the line?

GuillaumeGomez (approved):
Just a nit for me but otherwise looks good!

zec:
OK, since everyone's approved the PR as it currently stands (unless I'm misinterpreting @GuillaumeGomez's comments), I've squashed the history.

GuillaumeGomez:
It's fine if merged as is, which is why I approved it. I'd just prefer a better naming for the method though but you can ignore me. :p

PR #1068, opened 2021-03-13
git: Read remote url from upstream or origin

MarijnS95:
Depends on #1041

With the new separate gir directories it is useful to know what repo (url) a git hash came from, without hardcoding it here. Consensus shows that these repos are either directly cloned through submodules with a default remote named `origin` or a manual rename to `upstream` with the personal remote (from which PRs are commonly created) named `origin`. In that case the url of `upstream` is preferred.

CC @GuillaumeGomez @sdroege

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -36,15 +36,34 @@ fn dirty(path: impl AsRef<Path>) -> bool {
     }
 }
 
+fn gitmodules_config(subcommand: &[&str]) -> Option<String> {
+    let mut args = vec!["config", "-f", ".gitmodules", "-z"];
+    args.extend(subcommand);
+    let output = git_command(Path::new("."), &args).ok()?;
+
+    if !output.status.success() {
+        return None;
+    }
+
+    let mut result = String::from_utf8(output.stdout).ok()?;
+    assert_eq!(result.pop(), Some('\0'));
+    Some(result)
+}
+
 // This file is also compiled from build.rs where this function is unused
 #[allow(dead_code)]
 pub(crate) fn repo_remote_url(path: impl AsRef<Path>) -> Option<String> {
-    let output = ["upstream", "origin"].iter().find_map(|remote| {
-        let output = git_command(path.as_ref(), &["remote", "get-url", remote]).ok()?;
-        // XXX: Use `.then_some` when it is stabilized
-        output.status.success().then(|| output)
-    })?;
-    String::from_utf8(output.stdout)
-        .ok()
-        .map(|s| s.trim_end_matches('\n').into())
+    // Find the subsection that defines the module for the given path:
+    let key_for_path = gitmodules_config(&[
+        "--name-only",
+        "--get-regexp",
+        r"submodule\..+\.path",
+        &format!("^{}$", path.as_ref().display()),
+    ])?;
+
+    let subsection = key_for_path
+        .strip_suffix(".path")
+        .expect("submodule.<subsection>.path should end with '.path'");
+
+    gitmodules_config(&["--get", &format!("{}.url", subsection)])
 }
--- a/src/git.rs
+++ b/src/git.rs
@@ -2,12 +2,14 @@ use std::path::Path;
 use std::process::Command;
 
 pub fn repo_hash<P: AsRef<Path>>(path: P) -> Option<String> {
-    let git_path = path.as_ref().to_str();
-    let mut args = match git_path {
-        Some(path) => vec!["-C", path],
-        None => vec![],
-    };
+    let git_path = path
+        .as_ref()
+        .to_str()
+        .expect("Repository path must be a valid UTF-8 string");
+
+    let mut args = vec!["-C", git_path];
     args.extend(&["rev-parse", "--short", "HEAD"]);
+
     let output = Command::new("git").args(&args).output().ok()?;
     if !output.status.success() {
         return None;
--- a/src/git.rs
+++ b/src/git.rs
@@ -35,3 +35,16 @@ fn dirty(path: impl AsRef<Path>) -> bool {
         Err(_) => false,
     }
 }
+
+// This file is also compiled from build.rs where this function is unused
+#[allow(dead_code)]
+pub(crate) fn repo_remote_url(path: impl AsRef<Path>) -> Option<String> {
+    let output = ["upstream", "origin"].iter().find_map(|remote| {
+        let output = git_command(path.as_ref(), &["remote", "get-url", remote]).ok()?;
+        // XXX: Use `.then_some` when it is stabilized
+        output.status.success().then(|| output)
+    })?;
+    String::from_utf8(output.stdout)
+        .ok()
+        .map(|s| s.trim_end_matches('\n').into())
+}
--- a/src/git.rs
+++ b/src/git.rs
@@ -1,16 +1,21 @@
+use std::io::Result;
 use std::path::Path;
-use std::process::Command;
+use std::process::{Command, Output};
 
-pub fn repo_hash<P: AsRef<Path>>(path: P) -> Option<String> {
+fn git_command(path: impl AsRef<Path>, subcommand: &[&str]) -> Result<Output> {
     let git_path = path
         .as_ref()
         .to_str()
         .expect("Repository path must be a valid UTF-8 string");
 
     let mut args = vec!["-C", git_path];
-    args.extend(&["rev-parse", "--short", "HEAD"]);
+    args.extend(subcommand);
 
-    let output = Command::new("git").args(&args).output().ok()?;
+    Command::new("git").args(&args).output()
+}
+
+pub fn repo_hash(path: impl AsRef<Path>) -> Option<String> {
+    let output = git_command(path.as_ref(), &["rev-parse", "--short", "HEAD"]).ok()?;
     if !output.status.success() {
         return None;
     }
@@ -24,14 +29,8 @@ pub fn repo_hash<P: AsRef<Path>>(path: P) -> Option<String> {
     }
 }
 
-fn dirty<P: AsRef<Path>>(path: P) -> bool {
-    let path = path.as_ref().to_str();
-    let mut args = match path {
-        Some(path) => vec!["-C", path],
-        None => vec![],
-    };
-    args.extend(&["ls-files", "-m"]);
-    match Command::new("git").args(&args).output() {
+fn dirty(path: impl AsRef<Path>) -> bool {
+    match git_command(path.as_ref(), &["ls-files", "-m"]) {
         Ok(modified_files) => !modified_files.stdout.is_empty(),
         Err(_) => false,
     }
--- a/src/config/config.rs
+++ b/src/config/config.rs
@@ -4,7 +4,7 @@ use super::{
 };
 use crate::{
     config::error::TomlHelper,
-    git::repo_hash,
+    git::{repo_hash, repo_remote_url},
     library::{self, Library},
     nameutil::set_crate_name_overrides,
     version::Version,
@@ -63,16 +63,18 @@ pub struct GirVersion {
     pub gir_dir: PathBuf,
     file_name: Option<String>,
     hash: Option<String>,
+    url: Option<String>,
 }
 
 impl GirVersion {
-    fn new(gir_dir: &Path, hash: Option<String>) -> Self {
+    fn new(gir_dir: &Path) -> Self {
         Self {
             gir_dir: normalize_path(gir_dir),
             file_name: gir_dir
                 .file_name()
                 .map(|s| s.to_str().expect("OsStr::to_str failed").to_owned()),
-            hash,
+            hash: repo_hash(gir_dir),
+            url: repo_remote_url(gir_dir),
         }
     }
 
@@ -81,10 +83,7 @@ impl GirVersion {
     }
 
     pub fn get_repository_url(&self) -> Option<&str> {
-        match self.file_name {
-            Some(ref r) if r == "gir-files" => Some("https://github.com/gtk-rs/gir-files"),
-            _ => None,
-        }
+        self.url.as_deref()
     }
 }
 
@@ -198,7 +197,7 @@ impl Config {
         }
         let mut girs_version = girs_dirs
             .iter()
-            .map(|d| GirVersion::new(d, repo_hash(d)))
+            .map(|d| GirVersion::new(d))
             .collect::<Vec<_>>();
         girs_version.sort_by(|a, b| a.gir_dir.partial_cmp(&b.gir_dir).unwrap());
 


MarijnS95:
I'm probably going blind, even the CI here throws this:

```rust
error: function is never used: `repo_remote_url`
Error:   --> src/git.rs:30:15
   |
30 | pub(crate) fn repo_remote_url(path: impl AsRef<Path>) -> Option<String> {
   |               ^^^^^^^^^^^^^^^
   |
   = note: `-D dead-code` implied by `-D warnings`
```

But the function is definitely there and used. Comment it out and this doesn't compile anymore with the function missing in `config.rs`.

sdroege:
Yeah that seems to make no sense at all :D Maybe just add a `#[allow(unused)]` and we can worry about that later...

sdroege (commented):
Looks good to me otherwise

GuillaumeGomez:
Can you provide an example of the output please?

MarijnS95:
@GuillaumeGomez Sorry, forgot to paste the link yesterday:

https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/719

src/git.rs
```diff
@@ -27,6 +27,18 @@ pub fn repo_hash(path: impl AsRef<Path>) -> Option<String> {
     }
 }
 
+pub(crate) fn repo_remote_url(path: impl AsRef<Path>) -> Option<String> {
+    let get_url = |remote: &str| {
+        let output = git_command(path.as_ref(), &["remote", "get-url", remote]).ok()?;
+        // XXX: Use `.then_some` when it is stabilized
+        output.status.success().then(|| output)
+    };
+    let output = get_url("upstream").or_else(|| get_url("origin"))?;
```
GuillaumeGomez, line 36: That really doesn't sound like a good idea... The best way would be to check for the default remote and use this one instead. I don't like much harcoded things that depends on users' config.

MarijnS95: Is there such a thing as a "default remote"? I know every branch can have its own "tracking remote" but that's different. In any case it is more likely to have ones fork as default remote, as that's what you'll end up pushing to. Maybe looking at the pull-remote for `master` is a good idea?

Anyway, let me quote a piece from the PR description again:

> Consensus shows that these repos are either directly cloned through submodules with a default remote named `origin` or a manual rename to `upstream` with the personal remote (from which PRs are commonly created) named `origin`.

It seems like that's the best we can do, bar hardcoding the gstreamer URL (and future ones) in `gir` or not providing a URL at all.

GuillaumeGomez: Why is `upstream` the default then? For git submodules, it seems like you instead want to use `origin`. And in case none of the two remotes are there, we are back to the original situation. To prevent that, we could use `git remote -v` (which could actually prevent to run multiple `git` commands since all remotes are listed at once).

And finally: if this is a `git@` URL, it's definitely not very useful, which is why I didn't add it in the first place. :-/

sdroege: You can get something like that via `git branch -vv`. Not sure how to properly read what `git branch --set-upstream-to` points at.

However I think this all does not work with detached branches, like what `git submodule` would create. I don't think git actually knows what the "canonical" origin is where the data came from.

I guess a more correct way for the hacks you and @MarijnS95 are doing would be to set a different push and fetch URL for the `origin`, and leave the `fetch` one intact, and use `origin` the way how it's supposed to be used ;)

See e.g.

```
$ git remote -v
origin	https://gitlab.freedesktop.org/gstreamer/gstreamer.git (fetch)
origin	git@gitlab.freedesktop.org:slomo/gstreamer.git (push)
```

MarijnS95: > Why is `upstream` the default then?

Since both [you](https://github.com/gtk-rs/gir/pull/1041#issuecomment-793849521) and me seem to have renamed the "upstream" repo to that, and it seems to be a common pattern when working with a fork to create PRs from (or `origin`+`fork`, which this also works for). Indeed, ssh links make little sense and I have it cloned as `https` because it is a public repo; I guess you are pushing (new branches) directly to it every once in a while? A https-fetch and ssh-push URL as Sebastian make most sense here, but both to your own fork.

> (which could actually prevent to run multiple `git` commands since all remotes are listed at once).

I prefer running multiple commands providing just the information we need rather than attempting to parse human-readable output that is not deemed stable to parse in the first place.

> You can get something like that via `git branch -vv`. Not sure how to properly read what `git branch --set-upstream-to` points at.

There are multiple tricks: https://stackoverflow.com/a/9753364/2844473

> However I think this all does not work with detached branches, like what `git submodule` would create. I don't think git actually knows what the "canonical" origin is where the data came from.

We'd have to find a `master` branch; does it create one and track the initial `origin` upon clone/checkout?

> I guess a more correct way for the hacks you and @MarijnS95 are doing would be to set a different push and fetch URL for the `origin`, and leave the `fetch` one intact, and use `origin` the way how it's supposed to be used ;)

That afaik has complications when pushing to and pulling from my own remote :grimacing: 

Fwiw `get-url` returns the fetch-url by default, unless `--push` is passed.

sdroege: For submodules, just let git do its thing (i.e. it will become `origin`). Who knows what other side effects changing the `origin` remote has with the other `git submodule` commands. It's fragile enough even without such things :)

Personally I have a separate clone of the submodules in a different place, which I use for changes to them. For other repos I have a setup like above, and my fork is write-only.

GuillaumeGomez: I really think that overall we're trying to do clever things that are way too focus on the existing code. And even that isn't working out so well. Not sure this is worth it...

sdroege: I would just use `origin` here and if someone manually broke the submodule configuration then that's their problem ;)

MarijnS95: > For submodules, just let git do its thing (i.e. it will become `origin`). Who knows what other side effects changing the `origin` remote has with the other `git submodule` commands. It's fragile enough even without such things :)

No need to mess with submodules from `gir` itself (unless you want to query whether the specified directory is a submodule and extract a url from that), and anyway preferring `upstream` over `origin` still works just as well on the `origin`-only checkout doesn't it?

> Personally I have a separate clone of the submodules in a different place, which I use for changes to them. For other repos I have a setup like above, and my fork is write-only.

I find that massively inconvenient if for example working on the combination of gir, gstreamer and gir-files in one spot - all changes have to be in the same place in order to even work.

> I would just use `origin` here and if someone manually broke the submodule configuration then that's their problem ;)

Again, `origin` is still supported just as well; does it really hurt to prefer `upstream`'s url over that?

MarijnS95:
> Yeah that seems to make no sense at all :D Maybe just add a `#[allow(unused)]` and we can worry about that later...

Do you happen to know if there's an existing Rust issue open about this. It happens up till nightly, with different names and reordering functions in this file...

src/git.rs
sdroege: > does it really hurt to prefer upstream's url over that?

I'm fine with your change

sdroege:
@GuillaumeGomez Can we move this forward? This is blocked on your currently and I'd like to move forward in gstreamer-rs with all this so we can finally close this topic :)

src/git.rs
sdroege: My compromise proposal here would be to always get the submodule URL via `.gitmodules`. That should always be correct

sdroege: Or maybe if @GuillaumeGomez insists on having the URL configurable in `Gir.toml`, we shouldn't use any git submodules at all and have `gir` clone the repo / commit that is configured in `Gir.toml` for each of the directories?

sdroege:
> Yeah that seems to make no sense at all :D Maybe just add a `#[allow(unused)]` and we can worry about that later...

This happens because the function is only used in `main.rs` but not from anything that comes from `lib.rs`. So it complains for the latter.

src/git.rs
sdroege: As discussed with @GuillaumeGomez yesterday, we would go with looking at the git configuration directly for the submodule. So basically for the submodule at path `gir-files` we would get the submodule name

```sh
# note the escaping for the backslashes so they end up in the command and the dot is not considered as an any-character
git config --file .gitmodules --get-regexp 'submodule\\..*\\.path' '^gir-files$'
```

This prints `submodule.gir-files.path gir-files`, the first `gir-files` is the submodule name.

Then

```sh
git config --file .gitmodules --get-regexp 'submodule\\.gir-files\\.url'
```

This prints `submodule.gir-files.url https://github.com/gtk-rs/gir-files` and there's the URL.

For this URL to be wrong for the submodule, you'd have to actually change the submodule configuration and then it would also show up as a change in the main git repository.

It would be resilient to whatever hacks you're doing inside the submodule repository itself.

MarijnS95:
> This happens because the function is only used in `main.rs` but not from anything that comes from `lib.rs`. So it complains for the latter.

`src/config/config.rs` uses `repo_remote_url` and "comes from" `lib.rs`, not `main.rs`?

Note that `repo_hash` is used from `build.rs` which is the only reason it is not showing up as well.

sdroege:
Hm unclear then. But I think the problem is related to this crate being a library, an executable, and also being used from build.rs.

MarijnS95:
@sdroege Bingo! The problem is of course:

https://github.com/gtk-rs/gir/blob/2ad58e9e20e5adbdb51f56d9159bc6ee4c0da02e/build.rs#L3-L4

The function is unused no matter `pub(crate)` or `pub`, since an executable doesn't export anything.

src/git.rs
sdroege: @MarijnS95 What do you think?

MarijnS95: @sdroege I think that I may not have enough time to fully dig into the suggested change, but will try...

I guess that's a smart way, yes. First map the path to a submodule, then back to a URL. Can we do it in a more rusty way though, ie. https://docs.rs/git2/0.13.17/git2/struct.Config.html? Perhaps we can convert the whole `git` command mangling over to a proper git wrapper while at it.

sdroege: That's where we were at some point, see https://github.com/gtk-rs/gir/pull/855 . I don't mind but some people are allergic to dependencies :)

For this case it seems the parsing of the output is easy enough with the `&str` iterators and functions like `strip_prefix()`, `split_once()`, etc.

MarijnS95: Yikes, maybe that is resolved by now by trimming down the `-sys` crate in some way? Surprisingly `gstreamer-rs` with all its autogenerated code compiles even faster than that :joy:

sdroege: It seems easy enough for now to paste the git output in this case. I'd worry about that in the future if something complicated happens :)

MarijnS95: @sdroege It is finally weekend and I finally have some time :partying_face:! `split_once` would have been awesome but that's still unstable :rage:, but there's an even better solution: `--name-only`! (For some reason there is no `--value-only` equivalent when using `--get-regexp`, fortunately we don't need it in the end).

The end result is pretty elegant, if I may say so myself (bar some string manipulation to trim `\0` etc, improvement suggestions are welcome). I'll leave some self-review comments.

src/git.rs
```diff
@@ -22,15 +27,41 @@ pub fn repo_hash<P: AsRef<Path>>(path: P) -> Option<String> {
...
+    match git_command(path.as_ref(), &["ls-files", "-m"]) {
         Ok(modified_files) => !modified_files.stdout.is_empty(),
         Err(_) => false,
     }
 }
+
+fn gitmodules_config(subcommand: &[&str]) -> Option<String> {
+    let mut args = vec!["config", "-f", ".gitmodules", "-z"];
+    args.extend(subcommand);
+    let output = git_command(Path::new("."), &args).ok()?;
```
MarijnS95, line 40: Perhaps should just use `Command::new` here directly. Saves an extra `vec![]` heap allocation + copying over values from the slice, and unnecessary `["-C", "."]` argument.

src/git.rs
```diff
@@ -22,15 +27,41 @@ pub fn repo_hash<P: AsRef<Path>>(path: P) -> Option<String> {
...
+
+    if !output.status.success() {
+        return None;
+    }
+
+    let mut result = String::from_utf8(output.stdout).ok()?;
+    assert_eq!(result.pop(), Some('\0'));
```
MarijnS95, line 47: I swear there should be something better for this _while retaining an **owned** string_ (`strip_suffix` and friends don't work).

src/git.rs
```diff
@@ -22,15 +27,41 @@ pub fn repo_hash<P: AsRef<Path>>(path: P) -> Option<String> {
...
+}
+
+// This file is also compiled from build.rs where this function is unused
+#[allow(dead_code)]
+pub(crate) fn repo_remote_url(path: impl AsRef<Path>) -> Option<String> {
+    // Find the subsection that defines the module for the given path:
+    let key_for_path = gitmodules_config(&[
+        "--name-only",
+        "--get-regexp",
+        r"submodule\..+\.path",
```
MarijnS95, line 58: Considering to replace `.` with `[\w-]` since we probably don't use any weird characters... On the other hand anything besides newlines and null char is allowed (assuming it is escaped properly):

https://git-scm.com/docs/git-config#:~:text=Subsection%20names%20are%20case,you%20don%E2%80%99t%20need%20to.

src/git.rs
sdroege: Where does the `\0` come from and why didn't that cause problems for any of the other git commands before?

src/git.rs
sdroege: Seems fine to me as is. We'll notice when it fails :)

src/git.rs
sdroege: But probably is more code? I don't think performance concerns are valid here for this code. Write it simple and understandable, not fast :)

sdroege: Maybe it should also take an `Option` for the path?

sdroege (commented):
Seems fine to me

src/git.rs
sdroege: Oh I thought `split_once` was finally stable in 1.51. It's 1.52 apparently, oh well!

src/git.rs
MarijnS95: `-z` makes it so that the string is terminated by `\0`. Now that no string splitting is needed anymore it's probably fine to leave it terminated by `\n`. The existing `repo_hash` uses `.trim_end_matches('\n')` which returns a slice reference and has to be cloned again, when we really want to resize the underlying `Vec<u8>` of a string (size -= 1, capacity stays the same).

Oh well, performance. Might as well use that and `.into()`/`.to_string()`/`.to_owned()`.

src/git.rs
MarijnS95: > But probably is more code?

Fortunately not really, this is all we need from `git_command`:

```suggestion
    let output = Command::new("git").args(&args).output().ok()?;
```

Not so much about performance, just seems cleaner that way. `git_command` is strangely using an `Option` already to hide when the path is not a valid UTF-8 string, rather than panicking :raised_eyebrow:

sdroege: That seems broken (the `Option` thing). It should rather panic in that case...

MarijnS95: Indeed. Fixed it as part of this PR, just in case.

src/git.rs
sdroege: Makes sense

MarijnS95:
@GuillaumeGomez Any remaining comments on this PR? This is our only blocker to migrate `gstreamer-rs` cleanly over to finally use `gir-files` as submodule (and have another submodule for gst-specific files).

sdroege:
Yeah it would be good to get this in so we can get unblocked :)

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

src/git.rs
MarijnS95: Now that `split_once` lands today, anywhere we can use this? :P

PR #1072, opened 2021-03-18
add an option to always generate builder patterns

bilelmoussaoui:
In some crates like gtk3/gtk4, creating a builder pattern is often set to true. It makes the configuration file super huge. This is an effort to try to reduce that by adding an option to always enable the builder pattern generation. 
Note that you can still disable builder pattern generation individually using `ignore_builder = true`

See the following commit for a real use case: https://github.com/gtk-rs/gtk4-rs/pull/219/commits/e3eac0f3adc5852b68cffc814445fd7627b31614

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,10 @@ trust_return_value_nullability = false
 # Disable running `cargo fmt` on generated files
 # (defaults to false)
 disable_format = true
+# Always generate a Builder if possible. This is mostly a convenient setter as most of the 
+# time you might want the Builder to be generated. Ignoring none-desired ones can still be done with per object `generate_builder` configuration.
+# (defaults to false)
+generate_builder = true
 ```
 
 This mode generates only the specified objects. You can either add the object's fullname to the `generate` array or add it to the `manual` array (but in this case, it won't be generated, just used in other functions/methods instead of generating an "ignored" argument). Example:
--- a/README.md
+++ b/README.md
@@ -177,13 +177,13 @@ status = "generate"
 generate_builder = true
 ```
 
-If you want to remove warning messages about the not bound `Builders` during the generation you don't want to be generated, you can ignore them with the `ignore_builder` flag in object configuration:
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -153,6 +153,10 @@ trust_return_value_nullability = false
 # Disable running `cargo fmt` on generated files
 # (defaults to false)
 disable_format = true
+# Always generate a Builder if possible. This is mostly a convenient setter as most of the 
+# time you might want the Builder to be generated. Ignoring none-desired one can still be done with `ignore_builder`
+# (defaults to false)
+always_generate_builder = true
```
sdroege, line 159: Shouldn't this be just `generate_builder`? It's the same configuration as per-object, but in the global scope.

Also it's not "always", it would still be overridden by per-object configurations setting it to `false` or not?

bilelmoussaoui: > Shouldn't this be just `generate_builder`? It's the same configuration as per-object, but in the global scope.

Probably

> Also it's not "always", it would still be overridden by per-object configurations setting it to `false`?

Yes, you can still use `ignore_builder` per object.

sdroege: >  Yes, you can still use `ignore_builder` per object.

Not `generate_builder = false`?

bilelmoussaoui: I can refactor it to use that instead of having a different option for exactly the same thing?

sdroege: Better to use the same name in global and per-object scope, otherwise it's just confusing :)

bilelmoussaoui: I refactored the code to remove the use of `ignore_builder` and renamed the global option to `generate_builder` as well.

sdroege (approved):
Much easier to understand now :)

bilelmoussaoui:
cc @GuillaumeGomez, what do you think about this change?

GuillaumeGomez (approved):

GuillaumeGomez:
Looks great, thanks!

PR #1074, opened 2021-03-19
"canonicalize" paths

GuillaumeGomez:
So for example, if we set "a/../b", it'll write up "b" (because "a" is "eaten" since there is a ".."). If there is no parent folder before a "..", it does nothing.

### DIFF
--- a/src/config/config.rs
+++ b/src/config/config.rs
@@ -13,24 +13,55 @@ use log::warn;
 use std::{
     collections::HashMap,
     fs,
-    path::{Path, PathBuf},
+    path::{Component, Path, PathBuf},
     str::FromStr,
 };
 
+fn canonicalize(path: &Path) -> String {
+    let mut parts: Vec<String> = Vec::new();
+
+    for component in path.components() {
+        match component {
+            Component::RootDir | Component::CurDir => {}
+            Component::ParentDir => {
+                parts.pop();
+            }
+            c => parts.push(
+                c.as_os_str()
+                    .to_str()
+                    .expect("OsStr::to_str failed")
+                    .to_owned(),
+            ),
+        }
+    }
+    parts.join("/")
+}
+
 #[derive(Debug)]
 pub struct GirVersion {
-    pub gir_dir: PathBuf,
+    pub gir_dir: String,
+    file_name: Option<String>,
     hash: Option<String>,
 }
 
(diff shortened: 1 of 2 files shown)


src/config/config.rs
```diff
@@ -13,24 +13,55 @@ use log::warn;
...
+
     pub fn get_hash(&self) -> &str {
         self.hash.as_deref().unwrap_or("???")
     }
 
     pub fn get_repository_url(&self) -> Option<&str> {
-        match self.gir_dir.file_name() {
-            Some(r) if r == "gir-files" => Some("https://github.com/gtk-rs/gir-files"),
+        match self.file_name {
+            Some(ref r) if r == "gir-files" => Some("https://github.com/gtk-rs/gir-files"),
```
sdroege, line 64: Why hardcode the gtk-rs gir-files URL here?

src/config/config.rs
```diff
@@ -13,24 +13,55 @@ use log::warn;
 use std::{
     collections::HashMap,
     fs,
-    path::{Path, PathBuf},
+    path::{Component, Path, PathBuf},
     str::FromStr,
 };
 
+fn canonicalize(path: &Path) -> String {
```
sdroege, line 20: Strange that this doesn't exist in `std` in one way or another. But looks like this should work at least for a while until we find a case where it breaks...

src/config/config.rs
GuillaumeGomez: It was done like this, keeping the same logic. Can be removed in the other PR.

src/config/config.rs
GuillaumeGomez: I love your positive thinking. :rofl:

MarijnS95: There exists a crate that takes care of this, which might be a bit more complete: https://docs.rs/pathdiff/0.2.0/src/pathdiff/lib.rs.html#34-77

Anyway IMO this function does too many copies that could have been `&Path` slices coalesced into a single owned `PathBuf` in the end, instead of returning a `String` (it's still a path, no?).

GuillaumeGomez: Performance-wise, it's not the best, however it's done once for all paths at the beginning and that's it. However, we don't use it as a `Path` anyway so there's not much point into storing it as one.

PR #1075, opened 2021-03-19
config: "Canonicalize" ../ away from relative git paths

MarijnS95:
@GuillaumeGomez This is the implementation I intended to make, back when originally noticing and reporting this issue. Unfortunately you merged the PR just before I was able to send it as a review suggestion, so here it is as a PR instead.

- Accepts an `impl AsRef<Path>`;
- Returns a `PathBuf` since it's still a path, so no changes are needed in the other files. Those have been reverted;
- No unnecessary clones, `as_os_str` / `to_str` / `to_owned` confusing code: just playing around with path components and copying them to a new buffer (owned `PathBuf`) at once in the end;
- Fails on paths that go out of their "root/current" directory, ie. `../`;
- Tests, but we'll have to see what to do with silently removing `CurrentDir` and `RootDir`!

### DIFF
--- a/src/config/config.rs
+++ b/src/config/config.rs
@@ -13,24 +13,76 @@ use log::warn;
 use std::{
     collections::HashMap,
     fs,
-    path::{Path, PathBuf},
+    path::{Component, Path, PathBuf},
     str::FromStr,
 };
+
+/// Performs canonicalization by removing `foo/../` and `./` components
+/// from `path`, without hitting the file system. It does not turn relative
+/// paths into absolute paths.
+fn normalize_path(path: impl AsRef<Path>) -> PathBuf {
+    let mut parts: Vec<Component<'_>> = vec![];
+
+    for component in path.as_ref().components() {
+        match (component, parts.last()) {
+            (Component::CurDir, _) | (Component::ParentDir, Some(Component::RootDir)) => {}
+            (Component::ParentDir, None) | (Component::ParentDir, Some(Component::ParentDir)) => {
+                parts.push(Component::ParentDir)
+            }
+            (Component::ParentDir, Some(_)) => {
+                parts
+                    .pop()
+                    .expect("Cannot navigate outside of base directory!");
+            }
+            (c, _) => parts.push(c),
+        }
+    }
+    parts.iter().collect()
+}
+
+#[test]
+fn test_normalize_path() {
+    assert_eq!(normalize_path("foo/../bar").as_os_str(), "bar");
+    assert_eq!(normalize_path("foo/./bar").as_os_str(), "foo/bar");
+    assert_eq!(normalize_path("./foo").as_os_str(), "foo");
+    assert_eq!(
+        normalize_path("foo/../bar/baz/../qux").as_os_str(),
+        "bar/qux"
+    );
+    assert_eq!(
+        normalize_path("foo/bar/baz/../../qux").as_os_str(),
+        "foo/qux"
+    );
+    assert_eq!(normalize_path("/foo/../bar").as_os_str(), "/bar");
+    assert_eq!(normalize_path("/../bar").as_os_str(), "/bar");
+    assert_eq!(normalize_path("foo/../../bar").as_os_str(), "../bar");
+}
 
 #[derive(Debug)]
 pub struct GirVersion {
     pub gir_dir: PathBuf,
+    file_name: Option<String>,
     hash: Option<String>,
 }
 
 impl GirVersion {
+    fn new(gir_dir: &Path, hash: Option<String>) -> Self {
+        Self {
+            gir_dir: normalize_path(gir_dir),
+            file_name: gir_dir
+                .file_name()
+                .map(|s| s.to_str().expect("OsStr::to_str failed").to_owned()),
+            hash,
+        }
+    }
+
     pub fn get_hash(&self) -> &str {
         self.hash.as_deref().unwrap_or("???")
     }
 
     pub fn get_repository_url(&self) -> Option<&str> {
-        match self.gir_dir.file_name() {
-            Some(r) if r == "gir-files" => Some("https://github.com/gtk-rs/gir-files"),
+        match self.file_name {
+            Some(ref r) if r == "gir-files" => Some("https://github.com/gtk-rs/gir-files"),
             _ => None,
         }
     }
@@ -146,10 +198,7 @@ impl Config {
         }
         let mut girs_version = girs_dirs
             .iter()
-            .map(|d| GirVersion {
-                gir_dir: d.clone(),
-                hash: repo_hash(d),
-            })
+            .map(|d| GirVersion::new(d, repo_hash(d)))
             .collect::<Vec<_>>();
         girs_version.sort_by(|a, b| a.gir_dir.partial_cmp(&b.gir_dir).unwrap());
 
--- a/src/config/config.rs
+++ b/src/config/config.rs
@@ -13,55 +13,24 @@ use log::warn;
 use std::{
     collections::HashMap,
     fs,
-    path::{Component, Path, PathBuf},
+    path::{Path, PathBuf},
     str::FromStr,
 };
 
-fn canonicalize(path: &Path) -> String {
-    let mut parts: Vec<String> = Vec::new();
-
-    for component in path.components() {
-        match component {
-            Component::RootDir | Component::CurDir => {}
-            Component::ParentDir => {
-                parts.pop();
-            }
-            c => parts.push(
-                c.as_os_str()
-                    .to_str()
-                    .expect("OsStr::to_str failed")
-                    .to_owned(),
-            ),
-        }
-    }
-    parts.join("/")
-}
-
 #[derive(Debug)]
 pub struct GirVersion {
-    pub gir_dir: String,
-    file_name: Option<String>,
+    pub gir_dir: PathBuf,
     hash: Option<String>,
 }
 
 impl GirVersion {
-    fn new(gir_dir: &Path, hash: Option<String>) -> Self {
-        Self {
-            gir_dir: canonicalize(gir_dir),
-            file_name: gir_dir
-                .file_name()
-                .map(|s| s.to_str().expect("OsStr::to_str failed").to_owned()),
-            hash,
-        }
-    }
-
     pub fn get_hash(&self) -> &str {
         self.hash.as_deref().unwrap_or("???")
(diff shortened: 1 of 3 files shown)


GuillaumeGomez:
The renaming and the tests: great idea!

However, the rest is just over-engineering in my opinion.

MarijnS95:
Less code, less diff, less clones, less conversions. How's that over-engineering?

src/config/config.rs
```diff
@@ -17,37 +17,52 @@ use std::{
...
-    let mut parts: Vec<String> = Vec::new();
+fn relative_canonicalize(path: impl AsRef<Path>) -> PathBuf {
+    let mut parts: Vec<Component<'_>> = vec![];
 
-    for component in path.components() {
+    for component in path.as_ref().components() {
         match component {
-            Component::RootDir | Component::CurDir => {}
+            // :shrug:
+            // Component::RootDir | Component::CurDir => {}
```
GuillaumeGomez, line 26: `CurDir` is basically ".", there is no point into keeping it.

src/config/config.rs
```diff
@@ -17,37 +17,52 @@ use std::{
...
+    assert_eq!(
+        relative_canonicalize("foo/../bar/baz/../bav").as_os_str(),
+        "bar/bav"
+    );
+    assert_eq!(
+        relative_canonicalize("foo/bar/baz/../../bav").as_os_str(),
+        "foo/bav"
+    );
+    // Losing root...
+    assert_eq!(relative_canonicalize("/foo").as_os_str(), "foo");
```
GuillaumeGomez, line 50: How come we lose the root with this implementation?

GuillaumeGomez:
@MarijnS95 No, you're right, I judged too quickly.

src/config/config.rs
MarijnS95: Fair, `foo/./bar` should just become `foo/bar`. I would like to preserve `RootDir` though, for correctness (even if it is unlikely to be used).

src/config/config.rs
MarijnS95: Not anymore, because I commented it out.

MarijnS95:
> The renaming

@GuillaumeGomez I am still not sure if it is fully correct though. Normalization, canonicalization, the terms seem to be used interchangeably.

src/config/config.rs
```diff
@@ -17,37 +17,53 @@ use std::{
...
-    for component in path.components() {
+    for component in path.as_ref().components() {
         match component {
-            Component::RootDir | Component::CurDir => {}
+            Component::CurDir => {}
             Component::ParentDir => {
-                parts.pop();
+                parts
+                    .pop()
+                    .expect("Cannot navigate outside of base directory!");
```
MarijnS95, line 32: Instead of panicking we can also `parts.push(c)`. It should not be forbidden to have `../` at the beginning of the string, but then we should make sure that the thing we're `pop`'ing in subsequent iterations is _not_ a `ParentDir` :)

MarijnS95:
```console
foo:bar/ $ tests/sys/test.sh
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
    Finished release [optimized] target(s) in 0.02s
    Finished release [optimized] target(s) in 0.02s
     Running `/home/runner/work/gir/gir/target/release/gir -c gir-atk.toml`
thread 'main' panicked at 'Cannot navigate outside of base directory!', src/config/config.rs:31:22
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Quite glad I added the assert!

src/config/config.rs
GuillaumeGomez: Please remove the `expect` part. This is mostly about display after all. One thing I just thought: `/../whatever` should be transformed into `/whatever`, no `whatever`.

MarijnS95: > Please remove the `expect` part. This is mostly about display after all.

I strongly dislike such inconsistencies, and making such assumptions by imposing arbitrary undocumented limitations on the future. There's nothing preventing this from being used in more serious usecases later on.

> One thing I just thought: `/../whatever` should be transformed into `/whatever`, no `whatever`.

Correct, apparently `/..` is valid... Though for our intents a relative path outside of the path we know about should not! Otherwise we might as well canonicalize the root path and gir directory, and remove the former as prefix from the latter, that solves the problem in the same way.
Anyway, this is now fixed.

GuillaumeGomez: > I strongly dislike such inconsistencies, and making such assumptions by imposing arbitrary undocumented limitations on the future. There's nothing preventing this from being used in more serious usecases later on.

I don't see any case where it would extended. The current case can be seen in https://github.com/gtk-rs/gtk-rs/pull/366. If we keep this expect, all our `Gir.toml` files become invalid because they all use a path starting with "..". Again: this is about display **only**.

MarijnS95: > If we keep this expect, all our `Gir.toml` files become invalid because they all use a path starting with "..".

Negative. The path entering this function is the subdirectory concatenated with the gir directory:

```console
foo:gtk-rs/ $ ./generator.py 2>&1 | rg as_ref
[src/config/config.rs:26] path.as_ref() = "gdk/../gir-files"
[src/config/config.rs:26] path.as_ref() = "gdk/sys/../../gir-files"
[src/config/config.rs:26] path.as_ref() = "pango/../gir-files"
[src/config/config.rs:26] path.as_ref() = "pango/sys/../../gir-files"
[src/config/config.rs:26] path.as_ref() = "gdkx11/../gir-files"
[src/config/config.rs:26] path.as_ref() = "gdkx11/sys/../../gir-files"
[src/config/config.rs:26] path.as_ref() = "graphene/../gir-files"
[src/config/config.rs:26] path.as_ref() = "graphene/sys/../../gir-files"
[src/config/config.rs:26] path.as_ref() = "gtk/../gir-files"
[src/config/config.rs:26] path.as_ref() = "gtk/sys/../../gir-files"
[src/config/config.rs:26] path.as_ref() = "gio/../gir-files"
[src/config/config.rs:26] path.as_ref() = "gio/sys/../../gir-files"
[src/config/config.rs:26] path.as_ref() = "gdk-pixbuf/../gir-files"
[src/config/config.rs:26] path.as_ref() = "gdk-pixbuf/sys/../../gir-files"
[src/config/config.rs:26] path.as_ref() = "glib/../gir-files"
[src/config/config.rs:26] path.as_ref() = "glib/../gir-files"
[src/config/config.rs:26] path.as_ref() = "glib/sys/../../gir-files"
[src/config/config.rs:26] path.as_ref() = "glib/gobject-sys/../../gir-files"
[src/config/config.rs:26] path.as_ref() = "pangocairo/../gir-files"
[src/config/config.rs:26] path.as_ref() = "pangocairo/sys/../../gir-files"
[src/config/config.rs:26] path.as_ref() = "atk/../gir-files"
[src/config/config.rs:26] path.as_ref() = "atk/sys/../../gir-files"
```

The only breakage is observed in the CI, where `gir-*.toml` files are specified directly in the `tests/sys` (so no subdirectory prepended), and the gir files are cloned at `../gir-files`.

> Again: this is about display **only**.

For now.

GuillaumeGomez: So if you run your change on the current gtk-rs, is it working?

MarijnS95: > So if you run your change on the current gtk-rs, is it working?

Correct, it was also working with the `expect` (because none of the paths above try to walk outside of their directory).

GuillaumeGomez: Well then, no problem! We can keep the `expect`. :)

GuillaumeGomez:
> Quite glad I added the assert!

It would have failed in any case because it wouldn't have found the folder.

src/config/config.rs
MarijnS95: Well, it's already fixed to not need the expect...

And it breaks the tests in `tests/src/test.sh`, as mentioned below.

MarijnS95:
> It would have failed in any case because it wouldn't have found the folder.

Why not? Anyway, as the documentation indicates, this function shouldn't even touch/read the filesystem at all.

sdroege:
How should we proceed here @GuillaumeGomez @MarijnS95 ?

MarijnS95:
@sdroege I don't know. It seems @GuillaumeGomez prefers a simplified function that gracefully succeeds on unsupported cases (returning the wrong result) as it is "for display only" while I prefer a more fully featured function that either does the right thing or panics. In this case that panic would break the CI tests which is why the issue has been resolved **_and_ covered by the tests**. That way there are no surprises in the incredibly rare case that a random contributor needs a function like this and starts using it outside of pretty-printing purposes.

GuillaumeGomez:
I stand on my position. :)

sdroege:
I agree that this should panic if it doesn't work instead of silently doing the wrong thing. Silently doing the wrong thing is never the right behaviour, we're not writing JavaScript here.

MarijnS95:
@GuillaumeGomez I guess this PR will not move any further then since I won't submit the variant that returns the wrong path under some occasions "because it doesn't matter, it's only for display". IMO the path displayed in these files should still be a valid path relative the root. You get either the `.expect` (that breaks the tests) or stacking of `../` from me (current, working _and tested_ implementation).

GuillaumeGomez:
Damn, you're hard on business! Let's go with this version then, still better than the correct version in any case. I restart the CI.

MarijnS95:
> Damn, you're hard on business!

I guess favouring correctness has really hit me with Rust :grimacing: 

> still better than the correct version in any case

Not sure what you mean with "_than_ the correct version" - this is supposed to be the correct (rather than silently incorrect) version. If there's any case that is not handled correctly, now is the time to say so that they can be added to the tests and addressed!

GuillaumeGomez:
Sorry, it's a typo. I meant current, not "correct" (which makes no sense in here).

MarijnS95:
@GuillaumeGomez Thanks! I'm fixing the Python issue as we speak, backporting some clean stuff made for `gstreamer-rs`. Stay tuned!

MarijnS95:
@GuillaumeGomez Took about 10 tries because `github-cli` is utter garbage (manual push + PR next time again), https://github.com/gtk-rs/gtk-rs/pull/385 should fix the CI issue.

GuillaumeGomez (approved):

GuillaumeGomez:
Please just rebase on top of master to make the CI happy. :)

Issue #1079, opened 2021-03-23
`doc_cfg` hints not shown on reexports from private modules


MarijnS95:
Hi @GuillaumeGomez! I am currently revisiting [the](https://github.com/gtk-rs/gir/pull/991#issuecomment-738978667) [idea](https://github.com/gtk-rs/gir/pull/991#issuecomment-738988457) to omit `cfg`/`doc_cfg` attributes when the parent (`impl` block or `mod` in its entirety) already has a version constraint that is at least as strict. This was sparked after going through https://doc.rust-lang.org/std/os/index.html, where modules and types do **not** have `#[doc(cfg(unix))]` but seem to inherit the constraint from their parent module, solving exactly the issue you were describing.
In other words, we have version information on the module, any modules within it, and any `struct`/`trait` that has its own page (right at the top, and on the module overview page) - without recursively duplicating `#[doc(cfg())]` etc :partying_face:

I have hence [rebased](https://github.com/MarijnS95/gir/compare/deduplicate-cfg-guards) those commits back and regenerated, and, to no surprise, the features are indeed propagated into the files/modules... but not in the way I expected.

"Fortunately" the same issue seems to be affecting `master` already. `doc(cfg)` is currently only used on `mod`s and `fn`s, not on `struct`s in ie. `glib::wrapper!`. As it seems `rustdoc` is not able to put these feature requirement labels on items in a private module (`mod auto;`) that are then re-exported (`pub use auto::*;`) - even if the same version constraint is on both. Let's take for example `gdk::DeviceTool`. On `master` it has a feature constraint on the `mod` and `fn`s, and looks like this (with the following `rustc` nightly):

```console
me:gtk-rs/ $ rustc +nightly -Vv
rustc 1.53.0-nightly (5d04957a4 2021-03-22)
binary: rustc
commit-hash: 5d04957a4b4714f71d38326fc96a0b0ef6dc5800
commit-date: 2021-03-22
host: x86_64-unknown-linux-gnu
release: 1.53.0-nightly
LLVM version: 12.0.0
```

There's no version constraint on the module overview page:

![image](https://user-images.githubusercontent.com/2325264/112217812-3ff99f80-8c23-11eb-9af0-b0d5e42e97f0.png)

On the page for that type, there is no feature requirement on the `struct`, but there is on the `fn`s:

![image](https://user-images.githubusercontent.com/2325264/112217852-4a1b9e00-8c23-11eb-8d97-9298e219764c.png)

[...]

PR #1080, opened 2021-03-23
Omit version constraints if the surrounding scope already guards these

MarijnS95:
Originally committed and discussed in https://github.com/gtk-rs/gir/pull/991#issuecomment-738978667 / https://github.com/gtk-rs/gir/pull/991#issuecomment-738988457, these patches can finally be submitted as `rustdoc` automatically propagates feature requirements on modules to the items within it, showing it conveniently on the top of the page without a single `doc(cfg())` residing inside this file anymore:

image

Note that rustdoc automatically coalesces to this already! This PR just cleans up the code so that it's easier to read through: a loose `#[cfg]` statement makes it seem like something special is going on, even though it is simply repeating what is already declared on the `mod` as a whole. See: https://github.com/MarijnS95/gtk-rs/compare/deduplicate-cfg-guards

Unfortunately a bug with publicly reexporting private modules (ie. our `mod auto; pub use crate::auto::*;`) prevents `rustdoc` from figuring out this type constraint on `struct`s, and instead opts to propagate the requirement to _each and every member_ :( (#1079):

image

That is already the case on `master` (see it for yourself!) and this PR does nothing to alleviate the situation.


GuillaumeGomez (approved):
Nice!

MarijnS95:
@GuillaumeGomez Thanks! Upon re-reading the PR description, do note that those screenshots _do not_ show the current situation! This PR _only cleans up the code_ but the main issue is either solved by applying the temporary workaround mentioned in https://github.com/gtk-rs/gir/issues/1079 or automagically when it is addressed in rustdoc: https://github.com/rust-lang/rust/issues/83428. Apologies for selling "misleading" pretty pictures :sweat_smile:

PR #1084, opened 2021-03-28
doc: Use intra-doc-links instead of relative paths to html pages

MarijnS95:
Using intra-doc-links saves quite a bit of code in gir, looks cleaner in the markdown files and resulting docs (`[src]` button mostly), and has the advantage of being checked by `rustdoc` for validity.

That last point is not totally fair as most of these trait "implements" / "implementors" segments are generated inside macros, for which`rustdoc` does not seem to emit any warnings or errors currently.

---

This has only been tested on GStreamer-rs, and only sampled at random because of aforementioned "no-warnings-in-macros" issue :disappointed:

### DIFF
--- a/src/codegen/doc/mod.rs
+++ b/src/codegen/doc/mod.rs
@@ -256,8 +256,8 @@ fn create_object_doc(w: &mut dyn Write, env: &Env, info: &analysis::object::Info
                 .filter(|&tid| !env.type_status(&tid.full_name(&env.library)).ignored())
                 .map(|tid| {
                     format!(
-                        "[`{name}`](struct.{name}.html)",
-                        name = env.library.type_(tid).get_name()
+                        "[`{n}`](struct@crate::{n})",
+                        n = env.library.type_(tid).get_name()
                     )
                 })
                 .collect::<Vec<_>>();
@@ -635,23 +635,12 @@ fn get_type_trait_for_implements(env: &Env, tid: TypeId) -> String {
         format!("{}Ext", env.library.type_(tid).get_name())
     };
     if tid.ns_id == MAIN_NAMESPACE {
-        implements_link(&trait_name)
+        format!("[`{n}`](trait@crate::{n})", n = &trait_name)
     } else if let Some(symbol) = env.symbols.borrow().by_tid(tid) {
-        let mut full_trait_name = symbol.full_rust_name();
-        let crate_path = if let Some(crate_name) = symbol.crate_name() {
-            if crate_name == "gobject" {
-                "../glib/object".to_owned()
-            } else {
-                format!("../{}", crate_name)
-            }
-        } else {
-            error!("Type {} don't have crate", tid.full_name(&env.library));
-            "unknown".to_owned()
-        };
-        full_trait_name.push_str("Ext");
-        implements_ext_link(&full_trait_name, &trait_name, &crate_path)
+        let full_trait_name = symbol.full_rust_name();
+        format!("[`trait@{}Ext`]", &full_trait_name)
     } else {
-        error!("Type {} don't have crate", tid.full_name(&env.library));
+        error!("Type {} doesn't have crate", tid.full_name(&env.library));
         format!("`{}`", trait_name)
     }
 }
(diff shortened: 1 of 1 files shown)


MarijnS95:
@sophie-h Rebased and conflict automatically resolved (first commit was already applied), please review on gtk-rs :)

sophie-h:
I try. Great chance to understand the separated lgpl-docs.

src/codegen/doc/mod.rs
```diff
@@ -254,12 +254,7 @@ fn create_object_doc(w: &mut dyn Write, env: &Env, info: &analysis::object::Info
                 .into_iter()
                 .chain(env.class_hierarchy.subtypes(info.type_id))
                 .filter(|&tid| !env.type_status(&tid.full_name(&env.library)).ignored())
-                .map(|tid| {
-                    format!(
-                        "[`{name}`](struct.{name}.html)",
-                        name = env.library.type_(tid).get_name()
-                    )
-                })
+                .map(|tid| format!("[`{n}`](crate::{n})", n = env.library.type_(tid).get_name()))
```
sophie-h, line 257: I would suggest to use `"[`{n}`](struct@crate::{n})"` everywhere.

If `crate::x` does not exist, rustdoc dumps `href="crate::x"` into the html code which is in turn motivates browser to use xdg-open on this. With `struct@` it just links to an empty page.

sophie-h (changes requested):
Everything else looks great. (Not an expert in this at all though.)

src/codegen/doc/mod.rs
MarijnS95: @sophie-h that's interesting, I don't get any links in GStreamer-rs for missing references at all: do you have an example? Unfortunately all our uses of this code sits within `glib::wrapper` and does not get warnings for broken references, making it hard to find one here.

That said, besides improved linking it seems like a good idea to explicitly specify the type with `struct@`, thanks for the suggestion! Done the same for `trait@` and updated the PR :grin:

sophie-h: I only managed to get broken links by not having the newest gtk version enabled for `cargo doc` and then having links to objects from more recent versions from the LGPL docs I guess?

The online published version of gtk-rs shouldn't have any gir generated broken links because I wasn't getting any warning with all features enabled. So this is really just a precaution.

sophie-h (approved):

src/codegen/doc/mod.rs
MarijnS95: > I wasn't getting any warning with all features enabled

It doesn't seem like there are much if any mistakes in GStreamer (so far only `gst::ObjectExt` should have been `gst::GstObjectExt`, that's not really the pattern we should follow). I bet `gtk` also only has these links generated inside the `glib::wrapper!` macro hence no warnings show up.

sophie-h: The LGPL docs are copied from another repo into the actual source code in gtk-rs before generating the docs.

MarijnS95: Similar in GStreamer, the docs are committed in the repo and built into a separate crate. Either way we'll only get warnings once the documentation is actually embedded in the code and built by `rustdoc`, and we won't see them when they're inside macro expansions :/

sophie-h: Just to clarify: I regenerated the LGPL docs with your branch and then generated the gtk-rs docs based on that to test this PR.

MarijnS95: Great, good to know - thanks!

MarijnS95:
This supposedly also fixes broken links in `gtk4` where no documentation for external crates is generated and uploaded, like a reference to `glib::object::ObjectExt`. Most of the crate relies on `--extern-html-root-url=glib=https://gtk-rs.org/gtk-rs/` to point all `glib` references to a different link, but the explicit relative paths emitted here are (obviously) not caught by that.

CC @bilelmoussaoui :)

sdroege (approved):

bilelmoussaoui (approved):

GuillaumeGomez (approved):

GuillaumeGomez:
Nice, thanks!

PR #1085, opened 2021-03-28
[3/3] Emit intra-doc-links for symbol references

MarijnS95:
The docs look like a wasteland with all these symbols (containing `::`) in monospace without being clickable. Thanks to intra-doc-links we can simply wrap them in `[]` and color up the whole page!

Things that are still broken:
- Links to manually overridden types: `GstBinExt::iterate_elements()` should use `GstBinExtManual`;
- Signals. `` [`Bin::element-added`](crate::Bin::element-added) `` doesn't exist. Messing with the regex and `` [`Bin`](crate::Bin)::element-added `` might be better for the time being (if monospace is added back);
- The C docs use a format like `#GstElementClass.set_context()` which are not yet parsed correctly;
- (For GStreamer) various documentation fixes that are merged in `master` only and will probably only come around with the v1.20 release.
- `@` references other "members": this could be struct fields or function parameters to which we cannot and do not want to emit docs. However, they are also used to reference other functions _on the same type_ and enum variants on the same type. For clarity, correctness and brevity we should omit `crate::<type>::<member>` here and substitute that with `Self::<member>` - assuming we could look up and substitute the C name in the first place!

---

Not sure how we should land this. In GStreamer (just the GStreamer base crate, nothing outside that) there are already ~100 errors for broken links, currently showing up as `` [`foo`] `` in the code. Some require fixes upstream, others require more iteration here. We could drag this PR on for very long, not use intra-doc-links at all, or accept the warnings and stray `[]` (makes the docs more ugly) and fix remaining issues over time.

### DIFF
--- a/src/codegen/doc/format.rs
+++ b/src/codegen/doc/format.rs
@@ -134,44 +134,42 @@ fn replace_c_types(entry: &str, symbols: &symbols::Info, in_type: &str) -> Strin
     let out = SYMBOL.replace_all(&out, |caps: &Captures<'_>| {
         let member = caps.get(4).map(|m| m.as_str()).unwrap_or("");
         let sym = symbols.by_c_name(&caps[3]);
-        match &caps[2] {
-            /* References to members, enum variants or methods within the same type. */
-            "@" => {
-                if let Some(sym) = sym {
+
+        if let Some(sym) = sym {
+            if sym.owner_name() == Some(in_type) {
+                // `#` or `%` symbols should probably have been `@` to denote
+                // that it is a reference within the current type.
+                format!(
+                    "{}[`{n}{m}`](Self::{n}{m})",
+                    &caps[1],
+                    n = sym.name(),
+                    m = member
+                )
+            } else {
+                match &caps[2] {
                     // Catch invalid @ references that have a C symbol available but do not belong
                     // to the current type (and can hence not use `Self::`). For now generate XXX
                     // but with a valid global link so that the can be easily spotted in the code.
                     // assert_eq!(sym.owner_name(), Some(in_type));
-                    if sym.owner_name() != Some(in_type) {
-                        format!(
-                            "{}[`crate::{}{}`] (XXX: @-reference does not belong to {}!)",
-                            &caps[1],
-                            sym.full_rust_name(),
-                            member,
-                            in_type,
-                        )
-                    } else {
-                        format!(
-                            "{}[`{n}{m}`](Self::{n}{m})",
-                            &caps[1],
-                            n = sym.name(),
-                            m = member
-                        )
-                    }
-                } else {
-                    format!("{}`{}{}`", &caps[1], &caps[3], member)
+                    "@" => format!(
+                        "{}[`crate::{}{}`] (XXX: @-reference does not belong to {}!)",
+                        &caps[1],
+                        sym.full_rust_name(),
+                        member,
+                        in_type,
+                    ),
+                    "#" => format!(
+                        "{}[`{n}{m}`](crate::{n}{m})",
+                        &caps[1],
+                        n = sym.full_rust_name(),
+                        m = member,
+                    ),
+                    "%" => format!("{}[`{}{}`]", &caps[1], sym.full_rust_name(), member,),
+                    c => panic!("Unknown symbol reference {}", c),
                 }
             }
-            "#" | "%" => {
-                format!(
-                    "{}`{}{}`",
-                    &caps[1],
-                    sym.map(symbols::Symbol::full_rust_name)
-                        .unwrap_or_else(|| caps[3].into()),
-                    member
-                )
-            }
-            u => panic!("Unknown reference type {}", u),
+        } else {
+            format!("{}`{}{}`", &caps[1], &caps[3], member)
         }
     });
     let out = GDK_GTK.replace_all(&out, |caps: &Captures<'_>| {
--- a/src/codegen/doc/format.rs
+++ b/src/codegen/doc/format.rs
@@ -1,6 +1,6 @@
 use crate::analysis::symbols;
 use once_cell::sync::Lazy;
-use regex::{Captures, Regex};
+use regex::{Captures, Match, Regex};
 
 const LANGUAGE_SEP_BEGIN: &str = "<!-- language=\"";
 const LANGUAGE_SEP_END: &str = "\" -->";
@@ -78,7 +78,8 @@ fn format(mut input: &str, symbols: &symbols::Info, in_type: &str) -> String {
 
 static SYMBOL: Lazy<Regex> =
     Lazy::new(|| Regex::new(r"(^|[^\\])([@#%])(\w+\b)([:.]+[\w-]+\b)?").unwrap());
-static FUNCTION: Lazy<Regex> = Lazy::new(|| Regex::new(r"(\b[a-z0-9_]+)\(\)").unwrap());
+static FUNCTION: Lazy<Regex> =
+    Lazy::new(|| Regex::new(r"([@#%])?(\w+\b[:.]+)?(\b[a-z0-9_]+)\(\)").unwrap());
 static GDK_GTK: Lazy<Regex> = Lazy::new(|| Regex::new(r"G[dt]k[A-Z]\w+\b").unwrap());
 static TAGS: Lazy<Regex> = Lazy::new(|| Regex::new(r"<[\w/-]+>").unwrap());
 static SPACES: Lazy<Regex> = Lazy::new(|| Regex::new(r"[ ]{2,}").unwrap());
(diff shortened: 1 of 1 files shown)


src/codegen/doc/format.rs
```diff
@@ -77,32 +77,51 @@ fn format(mut input: &str, symbols: &symbols::Info) -> String {
...
+                if let Some(sym) = lookup_opt(&caps[3]) {
+                    format!("{}[`{n}{t}`](crate::{n}{t})", &caps[1], n = sym, t = trail)
+                } else {
+                    format!("{}`{}{}`", &caps[1], &caps[3], trail)
+                }
+            }
+            // Global constants
+            "%" => {
+                if let Some(sym) = lookup_opt(&caps[3]) {
+                    format!("{}[`{}{}`]", &caps[1], sym, trail)
```
MarijnS95, line 108: We could use `crate::` here, but there are quite some constants like `` [`None`] `` and `` [`true`] `` that link to Rust primitives automatically, which all need to be filtered out.

src/codegen/doc/format.rs
```diff
@@ -77,32 +77,51 @@ fn format(mut input: &str, symbols: &symbols::Info) -> String {
...
     let out = SYMBOL.replace_all(entry, |caps: &Captures<'_>| {
-        format!(
-            "{}`{}{}`",
-            &caps[1],
-            lookup(&caps[2]),
-            caps.get(3).map(|m| m.as_str()).unwrap_or("")
-        )
+        let trail = caps.get(4).map(|m| m.as_str()).unwrap_or("");
+        match &caps[2] {
+            // Parameter (and member?) names
```
MarijnS95, line 94: I don't really know what `@/#/%` mean _exactly_ just yet. Is it a doxygen thing?

src/codegen/doc/format.rs
```diff
@@ -77,32 +77,51 @@ fn format(mut input: &str, symbols: &symbols::Info) -> String {
...
+            // TODO: If member names, these can probably be emitted for simple structs where the fields are pub?
+            "@" => format!("{}`{}{}`", &caps[1], &caps[3], trail),
+            // Global symbols
+            "#" => {
+                if let Some(sym) = lookup_opt(&caps[3]) {
+                    format!("{}[`{n}{t}`](crate::{n}{t})", &caps[1], n = sym, t = trail)
```
MarijnS95, line 100: This is pretty ugly (but it's not in the source, probably not worth worrying about). Types that are in scope don't need `crate::` and `Self::` is even prettier where possible :grimacing:

src/codegen/doc/format.rs
```diff
@@ -77,32 +77,51 @@ fn format(mut input: &str, symbols: &symbols::Info) -> String {
...
     });
     let out = GDK_GTK.replace_all(&out, |caps: &Captures<'_>| {
         format!("`{}`", lookup(&caps[0]))
     });
     let out = FUNCTION.replace_all(&out, |caps: &Captures<'_>| {
-        format!("`{}`", lookup(&caps[1]))
+        if let Some(func) = lookup_opt(&caps[1]) {
+            format!("[`{f}()`](crate::{f}())", f = &func)
+        } else {
+            format!("`{}()`", &caps[1])
```
MarijnS95, line 123: Added `()` here:

- It is already in the docs;
- More clear when reading the docs, knowing that this references a function and not some other symbol;
- Clearer error messages by `rustdoc`;
- Rustdoc probably also enforces the link to point to a function, and fails otherwise (not tested).

src/codegen/doc/format.rs
sdroege: `@` is struct member/parameter/enum value in struct/function/enum documentation. `#` is for referencing types, `%` for referencing constants and enum values. It's a gtk-doc thing.

sdroege:
@GuillaumeGomez This needs some input from you for how to proceed here with the open issues :)

src/codegen/doc/format.rs
MarijnS95: Right. `@` is a tricky one because we _can_ reference enum variants, but not parameters nor struct members. Perhaps the same `if let Some(name) = lookup_opt()` can be used here to tell whether it is referencing a valid C symbol or not.

MarijnS95: By the way it appears C code mixes and matches `@#%` for "global names" like (enum) constants, functions, and types. I guess `@` simply means "associated to" because it is also used to reference functions on the same type?

If we have a valid C symbol for it, perhaps this should emit Self::<member>` instead of `crate::<type>::<member>`.

sdroege: > but not parameters nor struct members.

Not reference, but usually these are used to document those directly and then you would put the doc comment directly on that parameter or struct member or not?

sdroege: And for mixed usage in C code, gtk-doc is very forgiving with making mistakes and I think nobody (including me) knows how these things would be done correctly :)

MarijnS95: > Not reference, but usually these are used to document those directly and then you would put the doc comment directly on that parameter or struct member or not?

I've previously seen a list of ` * @GST_ENUM_VARIANT: the docs` as a list on top of the type, rather than on the variant directly. Not sure how that works for struct members. Anyway, this issue is about _linking to_ members and fields from the documentation itself, say if you have a large guiding paragraph on some type explaining how to use and what members to look at. You don't want sentences cut out and moved to the member directly :wink: 

Besides, structs are less relevant for GStreamer-rs since FFI fields are generally hidden from the user.

> And for mixed usage in C code, gtk-doc is very forgiving with making mistakes and I think nobody (including me) knows how these things would be done correctly :)

:/ that makes it all the tougher to do these links correctly, and leads us back to the question in the last paragraph of the PR description: how would we land this? Accept that there are many broken links? Try to filter them out or tell rustdoc to omit `[]` - if at all possible?

MarijnS95: Okay, gave `Self::<memb>` a try since we have neat access to 

- `@GstGLContext`/`@GstRTSPSessionMedia`: We have references directly to `Self`, is that correct or should these use `#`? There are not too many...;
- `@gst_video_decoder_set_output_state()` on `VideoCodecState` (needs upstream fixing, `@` should be dropped);

Other than that I might almost submit a minor PR that comes before this taking care of just `Self` links, way less issues :tada:

GuillaumeGomez:
From the code alone, it's hard to judge. @MarijnS95 Did you generate the code somewhere? If so, please open a PR to gtk-rs so I can take a look on the output. :)

MarijnS95:
@GuillaumeGomez Only locally now, maybe there is more to share in the evening. I _can_ however share some diffs for GStreamer showcasing `@` links mentioned in https://github.com/gtk-rs/gir/pull/1085#discussion_r603880127. These are much more forgiving (no errors besides mistakes in the C docs!) and I'll submit a separate PR to add just those as an initial implementation - let's call it intra-doc-links lite!

https://gitlab.freedesktop.org/MarijnS95/gstreamer-rs/-/compare/master...generator-docs

GuillaumeGomez:
An important thing to note is that now rustdoc intra-doc links are stable, meaning that all URLs generated like this: `[some type](relative url)` can be re-written like this: `[some type]`.

MarijnS95:
> An important thing to note is that now rustdoc intra-doc links are stable, meaning that all URLs generated like this: `[some type](relative url)` can be re-written like this: `[some type]`.

@GuillaumeGomez Yes, I'm aware, see also https://github.com/gtk-rs/gir/pull/1084/files.

However, if we don't want to riddle the documentation with `Self::` and `crate::` those links still have to be used: `` [`something`](Self::something) `` and `` [`something`](crate::something) ``. I'm all-ears to a more concise solution!

GuillaumeGomez:
No, `Self::` and `crate::` are fine by me. :)

MarijnS95:
Let's see what the rest thinks, I find it ugly while reading the docs.

The most horrible thing is `crate::<some type>::` where `Self::` could have been used, but fortunately for another change I'm already threading through the "owning type" into this bit of code and can now decide when to emit `Self::` over `crate::<some type>::` alleviating that a bit.

GuillaumeGomez:
I don't understand the issue here. It's in the markdown, not in the generated HTML documentation, so what's the issue?

MarijnS95:
> I don't understand the issue here. It's in the markdown, not in the generated HTML documentation, so what's the issue?

Except that it's not culled from HTML? On `rustc 1.53.0-nightly (07e0e2ec2 2021-03-24)`:

```markdown
Allocators are identified by name and can be retrieved with
[`Self::find()`]. [`crate::AllocatorExt::set_default()`] can be used to change the
default allocator.
```

This is rendered as:

image

And that is why I have opted to generate this instead:

```markdown
Allocators are identified by name and can be retrieved with
[`find()`](Self::find()). [`AllocatorExt::set_default()`](crate::AllocatorExt::set_default()) can be used to change the
default allocator.
```

GuillaumeGomez:
Yes, I'm perfectly fine with `crate::` and `Self::`. :)

sdroege:
What's the status here?

GuillaumeGomez (approved):

PR #1086, opened 2021-03-28
codegen/doc: Omit "Feature: `vXXX`" text in favour of doc_cfg

MarijnS95:
Nowadays we have the `doc_cfg` feature rendering clean, consistent and obvious markers for feature requirements through `doc(cfg())` attributes in the code (perhaps automatically derived from `cfg()` attributes in the not so distant future). With this it is not necessary to repeat the requirement in the documentation text once again, which wasn't really obvious nor good-looking to begin with.

Besides, this "Feature:" text is currently erroneously generated on the last member of an enum, instead of on the enum itself (parts omitted for brevity):

```rust
#[cfg(any(feature = "v1_14_1", feature = "dox"))]
#[cfg_attr(feature = "dox", doc(cfg(feature = "v1_14_1")))]
pub enum WebRTCFECType {
    /// none
    None,
    /// ulpfec + red
    ///
    /// Feature: `v1_14_1`
    ///
    UlpRed,
}
```
Note that other documentation comments on the enum _are_ correctly generated on the enum itself.

### DIFF
--- a/src/codegen/doc/mod.rs
+++ b/src/codegen/doc/mod.rs
@@ -217,9 +217,6 @@ fn create_object_doc(w: &mut dyn Write, env: &Env, info: &analysis::object::Info
                 "\nThis is an Abstract Base Class, you cannot instantiate it."
             )?;
         }
-        if let Some(version) = info.version {
-            writeln!(w, "\nFeature: `{}`", version.to_feature())?;
-        }
 
         let impl_self = if has_trait { Some(info.type_id) } else { None };
         let mut implements = impl_self
@@ -245,10 +242,6 @@ fn create_object_doc(w: &mut dyn Write, env: &Env, info: &analysis::object::Info
                 write!(w, "`[Deprecated]` ")?;
             }
             writeln!(w, "Trait containing all `{}` methods.", ty.name)?;
-
-            if let Some(version) = info.version {
-                writeln!(w, "\nFeature: `{}`", version.to_feature())?;
-            }
 
             let mut implementors = Some(info.type_id)
                 .into_iter()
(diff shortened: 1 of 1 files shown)


sdroege (approved):

GuillaumeGomez:
The problem is that `doc_cfg` is unstable and that it'll be removed in a not so far future. I really don't think this is a good idea here...

MarijnS95:
@GuillaumeGomez You mean this was just some nightly experiment that will disappear entirely? Will it be superseded by automatically detecting the feature from `#[cfg()]` directly or have no direct replacement at all until much later?

GuillaumeGomez:
The goal is to make rustdoc aware of the `cfg` attributes directly so there is no need for the `doc(cfg)`. A PR is already in progress. The end goal being to completely remove it.

MarijnS95:
> The goal is to make rustdoc aware of the `cfg` attributes directly so there is no need for the `doc(cfg)`. A PR is already in progress. The end goal being to completely remove it.

@GuillaumeGomez That's fine, but that still means the `vx_xx` features are _rendered in the HTML_ right? That is still a valid reason to remove this now-superfluous comment from the documentation, now?

GuillaumeGomez:
I'm not sure of what you mean. My point is that I'm worried that using an unstable feature might prevent users to generate doc locally or even break our crates' code. The update on rustdoc side will add it automatically depending on the `cfg`.

sdroege:
> My point is that I'm worried that using an unstable feature

That's already the case with the current docs. They only build with nightly.

MarijnS95:
> I'm not sure of what you mean. My point is that I'm worried that using an unstable feature might prevent users to generate doc locally or even break our crates' code. The update on rustdoc side will add it automatically depending on the `cfg`.

The unstable feature is already fully in use and docs (if `dox` is enabled) already build with that. This PR does _nothing_ to change or prolong that requirement. It only removes some redundant (again, if building with `doc_cfg`) text from the documentation body.

In other words: if you want the _textual_ `Feature: vXXX` message to stay until `doc_cfg` is stabilized and everyone can build (implicitly) with the `dox` feature without nightly, this PR should be stashed for a bit.

bilelmoussaoui (approved):

PR #1088, opened 2021-03-30
[1/3] Emit intra-doc links for self (`@`) references

MarijnS95:
Or self-dubbed intra-docs lite (wrt gir). This PR is the safest of the bunch: `@` is only used in limited quantities and pretty much every C symbol (in GStreamer) links to a valid item. There's even extra validation to ensure that the type/parent name of the symbol matches that of the type the docs are generated for!

We still have the debate about seeing or hiding `crate::` and `Self::` in the generated HTML docs going on: https://github.com/gtk-rs/gir/pull/1085#issuecomment-810171414. @GuillaumeGomez doesn't mind and prefers shorter gir and markdown code, what do the other maintainers think? See the screenshots how having `crate::` and `Self::` in the HTML looks.

### DIFF
--- a/src/analysis/symbols.rs
+++ b/src/analysis/symbols.rs
@@ -38,6 +38,10 @@ impl Symbol {
     pub fn crate_name(&self) -> Option<&String> {
         self.crate_name.as_ref()
     }
+
+    pub fn name(&self) -> &str {
+        &self.name
+    }
 }
 
 #[derive(Debug)]
--- a/src/analysis/symbols.rs
+++ b/src/analysis/symbols.rs
@@ -16,7 +16,11 @@ impl Symbol {
     pub fn full_rust_name(&self) -> String {
         let mut ret = String::new();
         if let Some(ref s) = self.crate_name {
-            ret.push_str(s);
+            if s == "gobject" {
+                ret.push_str("glib::object");
+            } else {
+                ret.push_str(s);
+            }
             ret.push_str("::");
         }
         if let Some(ref s) = self.owner_name {
--- a/src/analysis/symbols.rs
+++ b/src/analysis/symbols.rs
@@ -39,6 +39,10 @@ impl Symbol {
         self.crate_name.as_ref()
     }
 
+    pub fn owner_name(&self) -> Option<&str> {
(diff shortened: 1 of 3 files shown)


GuillaumeGomez:
In the HTML, they should not be present, only in the markdown link for intra-doc linking.

MarijnS95:
> In the HTML, they should not be present, only in the markdown link for intra-doc linking.

The screenshot in https://github.com/gtk-rs/gir/pull/1085#issuecomment-810457153 proves otherwise, no? Or has that been added recently (more recent than `rustc 1.53.0-nightly (07e0e2ec2 2021-03-24)`)?

GuillaumeGomez:
No I mean that **we** should not make it appear. ;)

GuillaumeGomez:
Well, unless for `Self::`. I think it might be confusing to just add a link to a method if we don't know if it's the current object.

MarijnS95:
> No I mean that **we** should not make it appear. ;)

Right! You meant to say that you were fine with having `crate::` (in a ``[`{type}`](crate::{type})`` contraption) in the _markdown_, not necessarily (confirmed here: not at all) in the HTML!

> Well, unless for `Self::`. I think it might be confusing to just add a link to a method if we don't know if it's the current object.

That might be obvious enough if done consistently. No explicit type -> must be the current type. But that's up for discussion.

There are some fixes to be pushed later to reintroduce the members (the 4th capturing group) back into the links, and unfortunately `|` match syntax within tuples is only experimental on nightly :angry:

MarijnS95:
@GuillaumeGomez Allright, reworked a bit to get rid of `|`! It's  bit ugly/messy, the final result in #1085 where `[#%]` are also taken into account is cleaner, but we really need this PR as a step-up because all `@` references (at least in GStreamer) are valid, contrary to `#` and `%`.

GuillaumeGomez (approved):

GuillaumeGomez (approved):

GuillaumeGomez:
Funny bug, I approved it twice. Looks good to me! @sdroege anything else for you?

sdroege:
I usually don't have opinions on docs things ;)

PR #1089, opened 2021-03-30
[2/3] Emit intra-doc links for function references

MarijnS95:
The same as #1088 but for functions. This one is more tricky since not all functions (ie. memory management functions like `free` and `unref`) are generated, hence result in invalid links.  Certain functions are implemented manually and do not exist under `<type>Ext` but `<type>ExtManual` - this might be a shortcoming of the existing C symbol mapping that should simply not map those functions to an `Ext` type. Other than that the gains in GStreamer are massive over the few tens of broken links.


sdroege:
What's the status here? @GuillaumeGomez

PR #1090, opened 2021-04-02
config: Normalize submodule path before looking it up in .gitmodules

MarijnS95:
`normalize_path` was introduced in [1] for pretty-printing relative paths with `../` ParentDir components.  It was not intended for use outside of displaying scenarios but thanks to tests we can safely use it to solve a mismatch [2] when searching for submodule paths in `.gitmodules`.

This functionality introduced with [3] finds the submodule associated with a path based on string matching in order to pull out the URL, but was only tested with GStreamer-rs where "clean" paths are passed in through gir's `-d` option.  Gtk-rs and gtk4-rs use `girs_directories =` in `Gir.toml` with relative paths like `gtk4/../gir-files`: these cannot be found in `.gitmodules` with simple string matching and need the path simplified first.

[1]: https://github.com/gtk-rs/gir/pull/1075
[2]: https://github.com/gtk-rs/gtk4-rs/pull/251#discussion_r606147518
[3]: https://github.com/gtk-rs/gir/pull/1068

### DIFF
--- a/src/config/config.rs
+++ b/src/config/config.rs
@@ -67,14 +67,15 @@ pub struct GirVersion {
 }
 
 impl GirVersion {
-    fn new(gir_dir: &Path) -> Self {
+    fn new(gir_dir: impl AsRef<Path>) -> Self {
+        let gir_dir = normalize_path(gir_dir);
         Self {
-            gir_dir: normalize_path(gir_dir),
             file_name: gir_dir
                 .file_name()
                 .map(|s| s.to_str().expect("OsStr::to_str failed").to_owned()),
-            hash: repo_hash(gir_dir),
-            url: repo_remote_url(gir_dir),
+            hash: repo_hash(&gir_dir),
+            url: repo_remote_url(&gir_dir),
+            gir_dir,
         }
     }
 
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
As much as I like your contributions, maybe a bit less salty/mocker would be nicer.

GuillaumeGomez:
I'm not sure to understand how this is an improvement though: you're adding an indirection on internal code. Unless I missed something, you just updated existing code, didn't add anything new, right?

MarijnS95:
> As much as I like your contributions, maybe a bit less salty/mocker would be nicer.

Was intended to be anecdotal but I agree it is also offensive. Removed it, apologies!

> I'm not sure to understand how this is an improvement though: you're adding an indirection on internal code. Unless I missed something, you just updated existing code, didn't add anything new, right?

The detail is in this `let` statement shadowing `gir_dir`:

```rust
let gir_dir = normalize_path(gir_dir);
```
Now the normalized `gir_dir` is passed into `repo_remote_url`, which would receive `gtk4/../gir-files` prior (for which no entry in `.gitmodules` exists).

GuillaumeGomez:
Oh I see, which explains the modifications. Then all good for me and thanks for the extra explanation!

GuillaumeGomez:
This is a draft currently, do you want to update it or can I merge it?

MarijnS95:
@GuillaumeGomez Should be good. Was mainly a draft since we originally settled on using `normalize_path` for display purposes only, but then:

- I don't think we have any better alternative;
- The rest of these - the URL and git hash - end up in the same comment (display) anyway.

GuillaumeGomez:
No I'm fine with it. I didn't expect these changes so I'm happy we went with your solution in the end. Thanks!

PR #1092, opened 2021-04-05
TryFromGlib support

fengalin:
This is the continuation of https://github.com/gtk-rs/gtk-rs/pull/11. See also details in https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/issues/234.

This PR adds new variants to `conversion_type` so that `gir` can generate the types implementing `TryFromGlib` properly.

Note: since there is no identified implementer for `TryFromGlib<Error=GlibNoneOrInvalidError>` and to avoid introducing untested code, the `ResultOption` variant is not handled in this PR.

## conversion_type "Option"

The `Option` variant is available for types `T` implementing `TryFromGlib<Error=GlibNoneError>`. As a reminder, this allows implementing `FromGlib` for `Option<T>` and `ToGlib` for both `T` and `Option<T>`. In this case, `Option<T>` will be used for return values (including ffi output arguments). For in-arguments, except if the parameter is declared `mandatory`, `impl Into<Option<T>>` so that either an `Option<T>` or `T` can be used.

Ex. from `gstreamer-rs`:

``` rust
[[object]]
name = "Gst.ClockTime"
status = "manual"
conversion_type = "Option"
```

The type `ClockTime` implements `TryFromGlib<Error=GlibNoneError>` (and `OptionToGlib`), which means that its Rust representation can take advantage of `Option<ClockTime>`.

Additionally, the user can instruct `gir` to `expect` `Some` or `Ok` results for specific arguments or return values. E.g.:

``` rust
[[object]]
name = "Gst.Clock"
status = "generate"
manual_traits = ["ClockExtManual"]
    [[object.function]]
    name = "get_calibration"
        [[object.function.parameter]]
        name = "internal"
        mandatory = true
```

In the above example, the user instructs gir to consider the `internal` argument (which also happens to be an out argument) with type gir `Gst.ClockTime` can be represented as a `ClockTime` without the `Option`. This argument is actually part of a set of output arguments. With the above gir declaration, the generated signature is the following (the implementation takes care of `expect`ing the value to be defined):

``` rust
    fn get_calibration(
        &self,
    ) -> (
        ClockTime,
        Option<ClockTime>,
        Option<ClockTime>,
        Option<ClockTime>,
    );
```

For a return value, the mandatory declaration reads:

``` rust
    [[object.function]]
    name = "util_get_timestamp"
    /.../
        [object.function.return]
        # always returns a value
        mandatory = true
```

Note: these examples are presented for illustration purposes only and are intended at testing the mechanism for now, the identification of the actual candidates for mandatory arguments and return values in `gstreamer-rs` is a work in progress.

See resulting code generations for [the out argument](https://gitlab.freedesktop.org/fengalin/gstreamer-rs/-/blob/opt-int-types-rework/gstreamer/src/auto/clock.rs#L273) (see also [`this example`](https://gitlab.freedesktop.org/fengalin/gstreamer-rs/-/blob/opt-int-types-rework/gstreamer/src/auto/clock.rs#L264) for an in argument) and for the [return value](https://gitlab.freedesktop.org/fengalin/gstreamer-rs/-/blob/opt-int-types-rework/gstreamer/src/auto/functions.rs#L269). This is part of a [`gstreamer-rs` branch](https://gitlab.freedesktop.org/fengalin/gstreamer-rs/-/tree/opt-int-types-rework) aiming at implementing https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/issues/234.

## conversion_type "Result"

The `Result` variant is available for types `T` implementing `TryFromGlib<Error=Err>` where `Err` is neither `GlibNoneError` nor `GlibNoneOrInvalidError`. In this case, `Result<T, ErrorType>` will be used for return values (including ffi output arguments) and the type itself in argument position.

In `gstreamer-rs`, the C type `GstStateChangeReturn` can represent both a successful or an error return value. In Rust, the `Result` `enum` is the idiomatic way of returning an error. In `gstreamer-rs`, bindings to functions returning `GstStateChangeReturn` had to be manually implemented so as to return `Result<StateChangeSuccess, StateChangeError>`. Note that in this case, the type implementing `TryFromGlib` is `StateChangeSuccess` and not `GstStateChangeReturn`. Thanks to this PR, these functions can now be auto-generated.

``` rust
[[object]]
name = "Gst.StateChangeReturn"
status = "generate"
must_use = true
    [object.conversion_type]
    variant = "Result"
    ok_type = "StateChangeSuccess"
    err_type = "StateChangeError"
```

See [`gst::Element::change_state`]()https://gitlab.freedesktop.org/fengalin/gstreamer-rs/-/blob/opt-int-types-rework/gstreamer/src/auto/element.rs#L77) auto-generation as an example.

For in-arguments user can either use `StateChangeReturn` directly, the `Result` variants or even `StateChangeSuccess` / `StateChangeError` directly. See the [generated code for `gst::Element::continue_state`](https://gitlab.freedesktop.org/fengalin/gstreamer-rs/-/blob/d80d282895fcfb40b8a0b5770105bf516416672b/gstreamer/src/auto/element.rs#L81).

## Non-regression

- [`gtk-rs` regen](https://github.com/fengalin/gtk-rs/tree/try_from_glib-improvement).
- [`gtk4-rs` regen](https://github.com/fengalin/gtk4-rs/tree/try_from_glib-improvement).
- [`gstreamer-rs` regen](https://gitlab.freedesktop.org/fengalin/gstreamer-rs/-/tree/test-regen-try_from_glib).

I didn't write documentation for this yet. It was already more involving than I thought it would be. I'll had an entry in the README once the solution gets accepted.

### DIFF
--- a/src/analysis/conversion_type.rs
+++ b/src/analysis/conversion_type.rs
@@ -1,11 +1,26 @@
-use crate::{config::gobjects::GObject, env, library::*};
+use crate::{env, library::*};
 
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+use std::sync::Arc;
+
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub enum ConversionType {
-    Direct,  //coded without conversion
-    Scalar,  //coded with from_glib
-    Pointer, //coded with from_glib_xxx
-    Borrow,  //same as Pointer, except that use from_glib_borrow instead from_glib_none
+    /// Coded without conversion.
+    Direct,
+    /// Coded with from_glib.
+    Scalar,
+    /// Type implementing TryFromGlib<Error=GlibNoneError>.
+    Option,
+    /// Type implementing TryFromGlib<Err> where Err is neither GlibNoneError
+    /// nor GlibNoneOrInvalidError. Embeds the Error type name.
+    /// Defaults to the object's type for the `Ok` variant if `ok_type` is `None`.
+    Result {
+        ok_type: Arc<str>,
+        err_type: Arc<str>,
+    },
+    /// Coded with from_glib_xxx.
+    Pointer,
+    // Same as Pointer, except that use from_glib_borrow instead from_glib_none.
+    Borrow,
     Unknown,
 }
 
@@ -19,16 +34,17 @@ impl ConversionType {
     pub fn of(env: &env::Env, type_id: TypeId) -> ConversionType {
         let library = &env.library;
 
-        if let Some(&GObject {
-            conversion_type: Some(conversion_type),
-            ..
-        }) = env.config.objects.get(&type_id.full_name(library))
+        if let Some(conversion_type) = env
+            .config
+            .objects
+            .get(&type_id.full_name(library))
+            .and_then(|gobject| gobject.conversion_type.clone())
         {
             return conversion_type;
         }
 
         use crate::library::{Fundamental::*, Type::*};
-        match *library.type_(type_id) {
+        match library.type_(type_id) {
             Fundamental(fund) => match fund {
                 Boolean => ConversionType::Scalar,
                 Int8 => ConversionType::Direct,
@@ -82,8 +98,12 @@ impl ConversionType {
             Function(_) => ConversionType::Direct,
             Custom(super::library::Custom {
                 conversion_type, ..
-            }) => conversion_type,
+            }) => conversion_type.clone(),
             _ => ConversionType::Unknown,
         }
+    }
+
+    pub fn can_use_to_generate(&self) -> bool {
+        matches!(self, ConversionType::Option | ConversionType::Result { .. })
     }
 }
--- a/src/analysis/try_from_glib.rs
+++ b/src/analysis/try_from_glib.rs
@@ -1,0 +1,101 @@
+use std::{borrow::Cow, sync::Arc};
+
+use crate::{
+    analysis::conversion_type::ConversionType,
+    config,
+    library::{self, Infallible, Mandatory},
+    Env,
+};
+
+#[derive(Clone, Debug)]
+pub enum TryFromGlib {
+    Default,
+    NotImplemented,
+    Option,
+    OptionMandatory,
+    Result {
+        ok_type: Arc<str>,
+        err_type: Arc<str>,
+    },
+    ResultInfallible {
+        ok_type: Arc<str>,
+    },
+}
+
+impl Default for TryFromGlib {
+    fn default() -> Self {
+        TryFromGlib::Default
+    }
+}
+
+impl TryFromGlib {
+    fn _new(
+        env: &Env,
+        type_id: library::TypeId,
+        mut config_mandatory: impl Iterator<Item = Mandatory>,
+        mut config_infallible: impl Iterator<Item = Infallible>,
+    ) -> Self {
+        let conversion_type = ConversionType::of(env, type_id);
+        match conversion_type {
+            ConversionType::Option => {
+                if *config_mandatory.next().unwrap_or(Mandatory(false)) {
+                    TryFromGlib::OptionMandatory
+                } else {
+                    TryFromGlib::Option
+                }
+            }
+            ConversionType::Result { ok_type, err_type } => {
+                if *config_infallible.next().unwrap_or(Infallible(false)) {
+                    TryFromGlib::ResultInfallible {
+                        ok_type: Arc::clone(&ok_type),
+                    }
+                } else {
+                    TryFromGlib::Result {
+                        ok_type: Arc::clone(&ok_type),
+                        err_type: Arc::clone(&err_type),
+                    }
+                }
+            }
+            _ => TryFromGlib::NotImplemented,
+        }
+    }
+
+    pub fn from_type_defaults(env: &Env, type_id: library::TypeId) -> Self {
+        Self::_new(env, type_id, None.into_iter(), None.into_iter())
+    }
+
+    pub fn or_type_defaults(&self, env: &Env, type_id: library::TypeId) -> Cow<'_, Self> {
+        match self {
+            TryFromGlib::Default => Cow::Owned(Self::from_type_defaults(env, type_id)),
+            other => Cow::Borrowed(other),
+        }
+    }
+
+    pub fn from_parameter(
+        env: &Env,
+        type_id: library::TypeId,
+        configured_parameters: &[&config::functions::Parameter],
+    ) -> Self {
+        Self::_new(
+            env,
+            type_id,
+            configured_parameters.iter().filter_map(|par| par.mandatory),
+            configured_parameters
+                .iter()
+                .filter_map(|par| par.infallible),
+        )
+    }
+
+    pub fn from_return_value(
+        env: &Env,
+        type_id: library::TypeId,
+        configured_functions: &[&config::functions::Function],
+    ) -> Self {
+        Self::_new(
+            env,
+            type_id,
+            configured_functions.iter().filter_map(|f| f.ret.mandatory),
+            configured_functions.iter().filter_map(|f| f.ret.infallible),
+        )
+    }
+}
--- a/README.md
+++ b/README.md
@@ -483,6 +483,101 @@ name = "Gtk.TextBuffer"
     # Avoid clash with the `insert` operation.
     bypass_auto_rename = true
 ```
+
+#### conversion_type "Option"
+
+The `conversion_type` variant `Option` is available for types `T` implementing
+`glib::TryFromGlib<Error=GlibNoneError>`. As a reminder, this allows
+implementing `FromGlib` for `Option<T>` and usually goes alongside with `ToGlib`
+for both `T` and `Option<T>`. In this case, `Option<T>` will be used for return
+values (including ffi output arguments). For in-arguments, except if the
+parameter is declared `mandatory`, `impl Into<Option<T>>` so that either an
+`Option<T>` or `T` can be used.
+
+Ex. from `gstreamer-rs`:
+
+``` rust
+[[object]]
+name = "Gst.ClockTime"
+status = "manual"
+conversion_type = "Option"
+```
+
+The type `ClockTime` implements `glib::TryFromGlib<Error=GlibNoneError>` (and
+`OptionToGlib`), which means that its Rust representation can take advantage of
+`Option<ClockTime>`.
+
+Additionally, the user can instruct `gir` to `expect` `Some` or `Ok` results for
+specific arguments or return values. E.g.:
+
+``` rust
+[[object]]
+name = "Gst.Clock"
+status = "generate"
+manual_traits = ["ClockExtManual"]
+    [[object.function]]
+    name = "get_calibration"
+        [[object.function.parameter]]
+        name = "internal"
+        mandatory = true
+```
+
+In the above example, the user instructs gir to consider the `internal` argument
+(which also happens to be an out argument) with type gir `Gst.ClockTime` can be
+represented as a `ClockTime` without the `Option`. This argument is actually
+part of a set of output arguments. With the above gir declaration, the generated
+signature is the following (the implementation takes care of `expect`ing the
+value to be defined):
+
+``` rust
+    fn get_calibration(
+        &self,
+    ) -> (
+        ClockTime,
+        Option<ClockTime>,
+        Option<ClockTime>,
+        Option<ClockTime>,
+    );
+```
+
+For a return value, the mandatory declaration reads:
+
+``` rust
+    [[object.function]]
+    name = "util_get_timestamp"
+    /.../
+        [object.function.return]
+        # always returns a value
+        mandatory = true
+```
+
+#### conversion_type "Result"
+
+The `conversion_type` variant `Result` is available for types `T` implementing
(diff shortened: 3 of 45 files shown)


src/analysis/conversion_type.rs
```diff
@@ -1,11 +1,18 @@
...
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+use std::sync::Arc;
+
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub enum ConversionType {
-    Direct,  //coded without conversion
-    Scalar,  //coded with from_glib
-    Pointer, //coded with from_glib_xxx
-    Borrow,  //same as Pointer, except that use from_glib_borrow instead from_glib_none
+    Direct,           // coded without conversion
```
GuillaumeGomez, line 7: At this point, we could turn those comments into doc comments I think. That'd also make the formatting easier.

GuillaumeGomez:
I think this is a good idea. So I guess you now need to add the documentation in the readme and maybe make a PR which uses it so we can double-check that it's doing what it's supposed to. :)

src/analysis/conversion_type.rs
fengalin: Sure

fengalin:
> maybe make a PR which uses it so we can double-check that it's doing what it's supposed to. :)

You mean like the 3 uses cases for which I gave links to the `gstreamer-rs` branch? ;) For `Result` and `ResultOption`, I don't know of any actual current uses cases yet, so they could fail. At least, I added test cases in gir to make sure parsing is done properly and I tested the generation by changing the `conversion_type` for `ClockTime` and applying the variants to the `mandatory` / `infallible` attributes. But this is not something that I can keep in a repo since it doesn't match the actual use of this type.

GuillaumeGomez:
> For `Result` and `ResultOption`, I don't know of any actual current uses cases yet, so they could fail.

If they're not used yet, better to not add them for the moment then.

fengalin:
> If they're not used yet, better to not add them for the moment then.

The traits are available in `glib` because they appear to be a logical generalization of the `Option` variant. So I thought this was a good idea to implement their handling in gir too. Should I have focused on the `Option` variant only, I would probably have come by a solution which would not have been flexible enough to handle all these uses cases. Now that the code is there, I think that makes sense to keep it and save the cognitive burden if someone needs this. I really think these are valid representations in Rust which have been overlooked so far because of how C handles options and errors.

GuillaumeGomez:
I'm not too keen keeping unused code around, even if I understand your logic. Unused code is future broken code. So I'm not really sure what to do here... Is there really no case where we could use these two variants so that they are still "checked"?

fengalin:
> Is there really no case where we could use these two variants so that they are still "checked"?

TBH I don't know gtk-rs (and gstreamer-rs) enough to answer. I already handled a variation on this for the [`UniChar` / `u32` to `char` and `Option<char>` conversions](https://github.com/gtk-rs/gir/pull/977) but I think it's different since it uses fundamental types and the `std` lib conversion, so it's hard coded in gir like it was before the change. The cases would be:

- `Result`: a C representation with an invalid range.
- `ResultOption`: a C representation with an invalid range AND a sentinel value for Undefined.

Does it ring a bell? @sdroege @bilelmoussaoui, somebody else?

sdroege:
I currently can only think of things that are special cases or are handled via manual bindings anyway, but there are lots of such situations really.

> `Result`: a C representation with an invalid range.

Signal ids, signal handler ids, `GSource` ids, etc. that have a single value as "invalid".

> `ResultOption`: a C representation with an invalid range AND a sentinel value for Undefined.

Maybe strings in general? `NULL` is `None`, but they can also be invalid UTF-8 and that would be invalid. Doesn't really happen with GLib-style APIs though as strings must be UTF-8 on the C side too there so it's valid to simply panic if this is not the case.

fengalin:
> Signal ids, signal handler ids, GSource ids, etc. that have a single value as "invalid".

Can you elaborate plz? Is this a value that would motivate a `Result<_, _>`? Can they really be returned by the C API as invalid?

> Maybe strings in general? NULL is None, but they can also be invalid UTF-8 and that would be invalid. Doesn't really happen with GLib-style APIs though as strings must be UTF-8 on the C side too there so it's valid to simply panic if this is not the case.

That would probably be a good candidate though. Maybe something reversed like: default to `infallible` (so `Option<_>`) and opt-in for `Result<Option<_>, _>` for specific APIs that might return an invalid UTF-8 String. However, this seems like a huge change.

sdroege:
> Can you elaborate plz? Is this a value that would motivate a `Result<_, _>`? Can they really be returned by the C API as invalid?

Connecting to an invalid signal id would return 0 (invalid) for the signal handler id. That would be an `Err`-case. Registering a signal with an invalid name would return 0 (invalid). Things like that, but we need manual bindings in those places anyway and need to check even more things in addition to what the C code would check so it doesn't seem very useful there.

>  However, this seems like a huge chang

Also for these specific APIs you will likely have to implement manual bindings anyway because the API pattern is "uncommon" and might be different for every case. It's unlikely that it would be covered in a generic way by the code generator.

fengalin:
Ok, so they are probably not candidate for an illustration with gir.

I'll check the `Result` cases in `gstreamer-rs` (since that was the starting point for all this) and see if I come up with something.

fengalin:
`enum`s are probably the obvious cases for `Result`. Currently, the idea is to use an `__Unknown(i32)` variant, which is an indication to the application that the Rust binding is unaware of the variant. On the other hand, it allows an application to pass this variant even if the binding is not up to date... So I'm not 100% sure about this too.

A simple test for this could use the [`ClockEntryType`'s getter](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/blob/master/gstreamer/src/clock.rs#L80). Why is this manual code currently? What's the point in the `as *mut _` cast?

sdroege:
> A simple test for this could use the [`ClockEntryType`'s getter](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/blob/master/gstreamer/src/clock.rs#L80). Why is this manual code currently? What's the point in the `as *mut _` cast?

`ClockId` is an opaque type. That's a `ffi::GstClockId as *mut ffi::GstClockEntry` cast there.

fengalin:
This is ready for another review. Last changes since 2 weeks:

* Found candidates in `gstreamer-rs` for the `Result` variant. C functions returning `*Return` `enum`s can now be auto-generated while returning `Result<*Success, *Error>`. This previously required manual implementation.
* Removed the `ResultOption` variant because I couldn't find type implementers so far.
* Reworked the `rust_type` module so it uses a `RustTypeBuilder` instead of defining functions for each combination of arguments required to build the Rust type.

Please see the commits message for details and the PR message for examples and links to implementations & non-regression verifications.

fengalin:
I forgot to mention that I have a bigger plan for a slight rework that should allow more code factorization. The idea is to apply the principles I used for `TryFromGlib` to `ConversionType` as a whole:

- Retrieve the modifiers from the configuration during analysis in constructors. For instance, that would allow getting `nullable` in one place so this is not scattered everywhere we deal with parameters or return values. As an extension, introducing a general purpose modifier, like `mandatory` which I consider more appropriate than `nullable` in many cases, and performing potential consistency checks could be factorized as well.
- Merge the `TryFromGlib` machinery with `ConversionType` and pass it when building `RustType`s. This makes sense when looking at the `RustTypeBuilder::try_build` implementation and that would avoid passing `with_nullable` or any additional future options.

However, I think this PR already contains a lot of changes, so I propose to try this in a separate PR and focus on [the `gstreamer-rs` PR which motivated that in the first place](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/issues/234) first.

Also, IMHO, there is a need for a bigger rework where `codegen` "sees" the result of the analysis by default. Current approach consists in relying on the definitions from `library` and retrieving `config` on a case by case basis. The `analysis` phase tends to apply computations that go beyond applying the mere `config`. The strategy from renaming getters comes to mind: not only does it apply heuristics for renames, it also performs conflicts resolution with other functions with the target name as well taking care of deprecation & availability of said functions. See also [this PR](https://github.com/gtk-rs/gir/pull/1098) which fixes renames functions not being available in `doc` mode.

sdroege:
The results in gstreamer look very good.

sdroege:
@GuillaumeGomez Does it seem fine to you implementation-wise?

src/analysis/try_from_glib.rs
```diff
@@ -0,0 +1,101 @@
...
+};
+
+#[derive(Clone, Debug)]
+pub enum TryFromGlib {
+    Default,
+    NotImplemented,
+    Option,
+    OptionMandatory,
+    Result {
+        ok_type: Arc<str>,
```
GuillaumeGomez, line 17: Why `Arc` and not `Rc`?

src/analysis/try_from_glib.rs
```diff
@@ -0,0 +1,101 @@
...
+        err_type: Arc<str>,
```
GuillaumeGomez, line 18: Same.

src/analysis/try_from_glib.rs
```diff
@@ -0,0 +1,101 @@
...
+    },
+    ResultInfallible {
+        ok_type: Arc<str>,
```
GuillaumeGomez, line 21: Same.

src/analysis/try_from_glib.rs
fengalin: I guess `gir` must be using some non-`Sync` mechanisms somewhere.

`rustc` forced me to use `Arc` instead of `Rc`. Didn't want to investigate why that was needed as it was not the point of this PR and I had already been through a lot to get there.

src/analysis/try_from_glib.rs
fengalin: See https://github.com/gtk-rs/gir/pull/1092#pullrequestreview-648012795

src/analysis/try_from_glib.rs
fengalin: See https://github.com/gtk-rs/gir/pull/1092#pullrequestreview-648012795

fengalin:
Make sure to read https://github.com/gtk-rs/gir/pull/1092#issuecomment-824933920 & https://github.com/gtk-rs/gir/pull/1092#issuecomment-825520946 for an overview of newest features and a vision for future improvements.

src/analysis/try_from_glib.rs
sdroege: The reason is simply because we store values of these things in `static`s:

```
error[E0277]: `Rc<str>` cannot be sent between threads safely
  --> src/codegen/sys/functions.rs:15:1
   |
15 | static DEFAULT_OBJ: Lazy<GObject> = Lazy::new(Default::default);
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `Rc<str>` cannot be sent between threads safely
   |
   = help: within `GObject`, the trait `std::marker::Send` is not implemented for `Rc<str>`
   = note: required because it appears within the type `ConversionType`
   = note: required because it appears within the type `std::option::Option<ConversionType>`
   = note: required because it appears within the type `GObject`
   = note: required because of the requirements on the impl of `Sync` for `once_cell::imp::OnceCell<GObject>`
   = note: required because it appears within the type `once_cell::sync::OnceCell<GObject>`
   = note: required because of the requirements on the impl of `Sync` for `once_cell::sync::Lazy<GObject>`
   = note: shared static variables must have a type that implements `Sync`
```

If we don't want `Arc` then we'd have to replace this on all usage sites with `GObject::default()`. If that's better or not, I don't know, but it's certainly not the point of this PR.

sdroege:
Are there any other concerns here?

src/analysis/try_from_glib.rs
fengalin: Yes, that's the one. Sorry for not trying that again to illustrate.
It's probably just a matter of using [`unsync::Lazy`](https://docs.rs/once_cell/1.7.2/once_cell/unsync/struct.Lazy.html).

I can try this and update if the PR doesn't stall too long after that :)

sdroege:
@GuillaumeGomez How should we proceed here?

GuillaumeGomez:
Is there anything else to be done here? Otherwise I think it's good to go.

fengalin:
Thanks, I'll perform a last check tomorrow and tell you when I think it's Ok to merge.

GuillaumeGomez (approved):

fengalin:
Last updates:

* rebased on master.
* checked regen non-regression for `gtk-rs`, `gtk4-rs` & `gstreamer-rs` `master`s. See links to branches in the PR's comment.
* rebased and CI checked [`gstreamer-rs`'s "Use TryFromGlib for *Return types"](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/759).
* rebased and CI checked [`gstreamer-rs`'s "WIP: Rework ClockTime & optional formatted values"](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/760).
* Added README entries. Maybe a little verbose, but I don't really see how to be more concise.

I think it should be ready to be merged now!

PR #1097, opened 2021-04-12
Import generator.py from gtk-rs

MarijnS95:
This is a direct import of all the commits from `gtk-rs` (the original source, and it's the most complete) with https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/739 for `docs/` generation and https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/740 to run all `gir` processes in parallel since they have no inter-dependencies and give a **massive** speed boost the more cores one has.

We'd like to place this script with `gir` so that it can be shared between `gtk-rs`, `gtk4-rs` and `gstreamer-rs` (and perhaps other projects that wish to invoke `gir` for every `Gir.toml` found in their workspace) without having to copypaste the file/changes (yuck) or transplant commits with `git format-patch`/`git am` (less disgusting, but still annoying) whenever we want to interchange improvements.

Things to consider before we merge this:
- gtk-rs/gtk4-rs currently do not check their `docs/` into the repo. This should perhaps not be the default?
- We previously had an extra `-d gst-gir-files` in GStreamer-rs, but that's not possible anymore in a shared `generator.py` script. We should either:
  1. `ln -s ./gir/generator.py` and use `girs_directories = ["../gir-files", "../gst-gir-files"]` _and_ its `../../` _and_ its `../../../` variant in `Gir.toml`;
  2. Use a simple script like `./gir/generator.py --gir-files-directories ./gir-files --gir-files-directories ./gst-gir-files "$@"`.

---

Unfortunately I'm not allowed to make a PR for unrelated histories so there's this ugly `Merge branch 'extract-generator' into HEAD` commit, perhaps a maintainer can push a more pretty merge commit straight to master so that we don't get two, once this is accepted.

### DIFF
--- a/generator.py
+++ b/generator.py
@@ -1,0 +1,237 @@
+#!/usr/bin/env python3
+
+from pathlib import Path
+import argparse
+import subprocess
+import sys
+import asyncio
+
+DEFAULT_GIR_FILES_DIRECTORY = Path("./gir-files")
+DEFAULT_GIR_DIRECTORY = Path("./gir/")
+DEFAULT_GIR_PATH = DEFAULT_GIR_DIRECTORY / "target/release/gir"
+
+
+def run_command(command, folder=None):
+    return subprocess.run(command, cwd=folder, check=True)
+
+
+async def spawn_process(exe, args):
+    p = await asyncio.create_subprocess_exec(
+        str(exe),
+        *(str(arg) for arg in args),
+        stdout=asyncio.subprocess.PIPE,
+        stderr=asyncio.subprocess.PIPE,
+    )
+
+    stdout, stderr = await p.communicate()
+    stdout = stdout.decode("utf-8")
+    stderr = stderr.decode("utf-8")
+    assert p.returncode == 0, stderr.strip()
+    return stdout, stderr
+
+
+async def spawn_gir(gir_exe, args):
+    stdout, stderr = await spawn_process(gir_exe, args)
+    # Gir doesn't print anything to stdout. If it does, this is likely out of
+    # order with stderr, unless the printer/logging flushes in between.
+    assert not stdout, "`gir` printed unexpected stdout: {}".format(stdout)
+    if stderr:
+        return "===> stderr:\n\n" + stderr + "\n"
+    return ""
+
+
+def update_workspace():
+    return run_command(["cargo", "build", "--release"], "gir")
+
+
+def ask_yes_no_question(question, conf):
+    question = "{} [y/N] ".format(question)
+    if conf.yes:
+        print(question + "y")
+        return True
+    line = input(question)
+    return line.strip().lower() == "y"
+
+
+def update_submodule(submodule_path, conf):
+    if any(submodule_path.iterdir()):
+        return False
+    print("=> Initializing {} submodule...".format(submodule_path))
+    run_command(["git", "submodule", "update", "--init", submodule_path])
+    print("<= Done!")
+
+    if ask_yes_no_question(
+        "Do you want to update {} submodule?".format(submodule_path), conf
+    ):
+        print("=> Updating submodule...")
+        run_command(["git", "reset", "--hard", "HEAD"], submodule_path)
+        run_command(["git", "pull", "-f", "origin", "master"], submodule_path)
+        print("<= Done!")
+        return True
+    return False
+
+
+def build_gir():
+    print("=> Building gir...")
+    update_workspace()
+    print("<= Done!")
+
+
+async def regenerate_crate_docs(gir_exe, crate_dir, base_gir_args):
+    doc_path = "docs.md"
+    # Generate into docs.md instead of the default vendor.md
+    doc_args = base_gir_args + ["-m", "doc", "--doc-target-path", doc_path]
+
+    # The above `gir -m doc` generates docs.md relative to the directory containing Gir.toml
+    doc_path = crate_dir / doc_path
+    embed_args = ["-g", "-m", "-o", doc_path, "-d", crate_dir / "src"]
+
+    logs = "==> Regenerating documentation for `{}` into `{}`...\n".format(
+        crate_dir, doc_path
+    )
+    logs += await spawn_gir(gir_exe, doc_args)
+
+    logs += "==> Embedding documentation from `{}` into `{}`...\n".format(
+        doc_path, crate_dir
+    )
+    stdout, stderr = await spawn_process("rustdoc-stripper", embed_args)
+    if stdout:
+        logs += "===> stdout:\n\n" + stdout + "\n"
+    if stderr:
+        logs += "===> stderr:\n\n" + stderr + "\n"
+
+    return logs
+
+
+def regen_crates(path, conf):
+    processes = []
+    if path.is_dir():
+        for entry in path.rglob("Gir*.toml"):
+            processes += regen_crates(entry, conf)
+    elif path.match("Gir*.toml"):
+        args = ["-c", path, "-o", path.parent] + [
+            d for path in conf.gir_files_paths for d in ("-d", path)
+        ]
+
+        is_sys_crate = path.parent.name.endswith("sys")
+
+        if conf.embed_docs:
+            # Embedding documentation only applies to non-sys crates
+            if is_sys_crate:
+                return processes
+
+            processes.append(regenerate_crate_docs(conf.gir_path, path.parent, args))
+        else:
+            if is_sys_crate:
+                args.extend(["-m", "sys"])
+
+            async def regenerate_crate(path, args):
+                return "==> Regenerating `{}`...\n".format(path) + await spawn_gir(
+                    conf.gir_path, args
+                )
+
+            processes.append(regenerate_crate(path, args))
+
+    else:
+        raise Exception("`{}` is not a valid Gir*.toml file".format(path))
+
+    return processes
+
+
+def valid_path(path):
+    path = Path(path)
+    if not path.exists():
+        raise argparse.ArgumentTypeError("`{}` no such file or directory".format(path))
+    return path
+
+
+def directory_path(path):
+    path = Path(path)
+    if not path.is_dir():
+        raise argparse.ArgumentTypeError("`{}` directory not found".format(path))
+    return path
+
+
+def file_path(path):
+    path = Path(path)
+    if not path.is_file():
+        raise argparse.ArgumentTypeError("`{}` file not found".format(path))
+    return path
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description="Helper to regenerate gtk-rs crates using gir.",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+    )
+
+    parser.add_argument(
+        "path",
+        nargs="*",
+        default=[Path(".")],
+        type=valid_path,
+        help="Paths in which to look for Gir.toml files",
+    )
+    parser.add_argument(
+        "--gir-files-directories",
+        nargs="+",  # If the option is used, we expect at least one folder!
+        dest="gir_files_paths",
+        default=[],
+        type=directory_path,
+        help="Path of the gir-files folder",
+    )
+    parser.add_argument(
+        "--gir-path",
+        default=DEFAULT_GIR_PATH,
+        type=file_path,
+        help="Path of the gir executable to run",
+    )
+    parser.add_argument(
+        "--yes",
+        action="store_true",
+        help=" Always answer `yes` to any question asked by the script",
+    )
+    parser.add_argument(
+        "--no-fmt",
+        action="store_true",
+        help="If set, this script will not run `cargo fmt`",
+    )
+    parser.add_argument(
+        "--embed-docs",
+        action="store_true",
+        help="Build documentation with `gir -m doc`, and embed it with `rustdoc-stripper -g`",
+    )
+
+    return parser.parse_args()
+
+
+async def main():
+    conf = parse_args()
+
+    if not conf.gir_files_paths:
+        update_submodule(DEFAULT_GIR_FILES_DIRECTORY, conf)
+
+    if conf.gir_path == DEFAULT_GIR_PATH:
+        update_submodule(DEFAULT_GIR_DIRECTORY, conf)
+        build_gir()
+
+    print("=> Regenerating crates...")
+    for path in conf.path:
+        print("=> Looking in path `{}`".format(path))
+        # Collect and print the results as soon as they trickle in, one process at a time:
+        for coro in asyncio.as_completed(regen_crates(path, conf)):
+            print(await coro, end="")
+
+    if not conf.no_fmt and not run_command(["cargo", "fmt"]):
+        return 1
+    print("<= Done!")
+    print("Don't forget to check if everything has been correctly generated!")
+    return 0
+
+
+if __name__ == "__main__":
+    try:
+        asyncio.run(main())
+    except Exception as e:
+        print("Error: {}".format(e), file=sys.stderr)
+        sys.exit(1)
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ authors = [
 	"Guillaume Gomez <guillaume1.gomez@gmail.com>",
 ]
 build = "build.rs"
-exclude = ["Gir*.toml", "tests/**/*", "*.md"]
+exclude = ["Gir*.toml", "tests/**/*", "*.md", "generator.py"]
 edition = "2018"
 
 [dependencies]
--- a/README.md
+++ b/README.md
@@ -6,6 +6,8 @@ The `GIR` is used to generate both the sys level crate and a safe API crate to u
 
 This README files is more about the options and little overview. If you want a tutorial on how to generate a crate using `gir`, we recommend you to read this [tutorial](https://gtk-rs.org/docs-src/tutorial/gir_tutorial) instead.
 
+`gir` includes a wrapper script `./generator.py` that detects `Gir.toml` configurations in the current directory (or the path(s) passed on the command-line) and generates "normal" or "sys" crates for it. Alternatively `--embed-docs` can be passed to prepare source-code for a documentation build by moving all documentation into it. For a complete overview of available options, pass `--help`.
+
 ## Introduction to `gir` generation
 
 Using `gir` requires both a `*.toml` and a `*.gir` for generation of the bindings.
@@ -153,7 +155,7 @@ trust_return_value_nullability = false
 # Disable running `cargo fmt` on generated files
 # (defaults to false)
 disable_format = true
-# Always generate a Builder if possible. This is mostly a convenient setter as most of the 
+# Always generate a Builder if possible. This is mostly a convenient setter as most of the
 # time you might want the Builder to be generated. Ignoring none-desired ones can still be done with per object `generate_builder` configuration.
 # (defaults to false)
 generate_builder = true


GuillaumeGomez:
> gtk-rs/gtk4-rs currently do not check their `docs/` into the repo. This should perhaps not be the default?

This cannot be done in gtk-rs/gtk4-rs because of the license of the docs.

> - We previously had an extra `-d gst-gir-files` in GStreamer-rs, but that's not possible anymore in a shared `generator.py` script. We should either:
>    1. `ln -s ./gir/generator.py` and use `girs_directories = ["../gir-files", "../gst-gir-files"]` _and_ its `../../` _and_ its `../../../` variant in `Gir.toml`;
>    2. Use a simple script like `./gir/generator.py --gir-files-directories ./gir-files --gir-files-directories ./gst-gir-files "$@"`.

We added this information into the `Gir.toml` files directly. Please do not override by default and use it instead.

MarijnS95:
> This cannot be done in gtk-rs/gtk4-rs because of the license of the docs.

The same applies to GStreamer-rs, the docs are not embedded in the crates by default. Gtk-rs can't have it in the repo _at all_ for other reasons, perhaps the repository-global `LICENSE` file?

So the answer is: yes please, ad a default-false `--generate-docs` option to `generator.py` before this is mergeable?

> We added this information into the `Gir.toml` files directly. Please do not override by default and use it instead.

It was convenient for GStreamer-rs given that we have quite significant nesting and these directories are the defaults for _all our crates_, but I can see how these flags are supposed to override and it's better to configure it in `Gir.toml` for every crate. Fortunately I already have a commit adding back `girs_directories` to GStreamer-rs, @sdroege I think that's our way forward.

GuillaumeGomez:
> So the answer is: yes please, ad a default-false `--generate-docs` option to `generator.py` before this is mergeable?

The answer is more "it wouldn't be used so don't add dead code"? :)

> It was convenient for GStreamer-rs given that we have quite significant nesting and these directories are the defaults for _all our crates_, but I can see how these flags are supposed to override and it's better to configure it in `Gir.toml` for every crate. Fortunately I already have a commit adding back `girs_directories` to GStreamer-rs, @sdroege I think that's our way forward.

:+1:

MarijnS95:
> The answer is more "it wouldn't be used so don't add dead code"? :)

It would be used, daily, on GStreamer-rs ;)

sdroege:
> The same applies to GStreamer-rs, the docs are not embedded in the crates by default. Gtk-rs can't have it in the repo _at all_ for other reasons, perhaps the repository-global `LICENSE` file?

I think this is just gtk-rs being overly cautious here. What matters is what is distributed with the application, not what is inside the gtk-rs repository. But whatever makes people happy :)

> So the answer is: yes please, ad a default-false `--generate-docs` option to `generator.py` before this is mergeable?

It would be used by gstreamer-rs so it's not dead code.

bilelmoussaoui:
IIRC, I'm fine with moving the gtk4 specific docs to a subdir rather than to a completely separate repository.

sdroege:
> EDIT: But then I also asked @sdroege why the docs themselves aren't just built at compile-time in the CI like gtk-rs.

I'm OK with changing it to do that at compile time, it just makes it more annoying to prepare the docs crate for uploading to crates.io. That would be mostly automated here now.

GuillaumeGomez:
> I think this is just gtk-rs being overly cautious here. What matters is what is distributed with the application, not what is inside the gtk-rs repository. But whatever makes people happy :)

No, the documentation is under LGPL, which isn't compatible. We simply can't put it in the same repository.

Also, small detail but quite important: the docs are **HUGE**. So for my disk and git history sake, please let's keep things separate.

> It would be used by gstreamer-rs so it's not dead code.

I'm starting to worry about having one common generator file for all use cases...

MarijnS95:
> I'm OK with changing it to do that at compile time, it just makes it more annoying to prepare the docs crate for uploading to crates.io. That would be mostly automated here now.

It would be nicer, and if automated by the CI (the deploy already is?) should be no extra hassle for you or anyone else. I agree with @GuillaumeGomez, saves on diskspace, commit noise, _and_ outdated docs!

> I'm starting to worry about having one common generator file for all use cases...

Well 99% of the file is still identical and we keep copy-pasting improvements back and forth which is why I'm advocating for a generic, shared file. Your call in the end though.

GuillaumeGomez:
> Well 99% of the file is still identical and we keep copy-pasting improvements back and forth which is why I'm advocating for a generic, shared file. Your call in the end though.

For now it's not a blocker, just a worry. If it goes too far (too specific to a given project), then we'll have to put limits. For now it remains acceptable.

sdroege:
> No, the documentation is under LGPL, which isn't compatible. We simply can't put it in the same repository.

The (L)GPL is a license that applies to distribution. As long as you don't distribute the docs together with your actual proprietary application there is no problem at all here. That's what I meant with overly cautious. But that also doesn't matter, I don't care if gtk-rs changes that or not :)

> It would be nicer, and if automated by the CI (the deploy already is?) should be no extra hassle for you or anyone else. I agree with @GuillaumeGomez, saves on diskspace, commit noise, _and_ outdated docs!

Yes if you could make the whole release process automatic that would be best. I'm OK with removing the docs if that makes anything easier. Doing a release of the docs crate happens not often enough to waste a lot of time on this.

MarijnS95:
> Yes if you could make the whole release process automatic that would be best. I'm OK with removing the docs if that makes anything easier. Doing a release of the docs crate happens not often enough to waste a lot of time on this.

I guess that's the right way forward, then we'll drop 6b6429c8b01cd3358c5bcfa230f7b9077f09ccaf and move all that into the CI, in  the usual `for` loop. However:

- There's no way to test this locally. Perhaps we need to have a shell script outside the CI (same for gtk-rs) to run `gir -m doc && cargo doc --features=dox,embed-lgpl-docs` without having to extract the commands from `CI.yml` and running them locally.

Thoughts?

bilelmoussaoui:
> > Yes if you could make the whole release process automatic that would be best. I'm OK with removing the docs if that makes anything easier. Doing a release of the docs crate happens not often enough to waste a lot of time on this.
> 
> I guess that's the right way forward, then we'll drop [6b6429c](https://github.com/gtk-rs/gir/commit/6b6429c8b01cd3358c5bcfa230f7b9077f09ccaf) and move all that into the CI, in the usual `for` loop. However:
> 
>     * There's no way to test this locally. Perhaps we need to have a shell script outside the CI (same for gtk-rs) to run `gir -m doc && cargo doc --features=dox,embed-lgpl-docs` without having to extract the commands from `CI.yml` and running them locally. The CI can simply invoke this script.
> 
> 
> Thoughts?

Where you would need such script? to test that gir works properly for the `doc` mode or to generate the docs locally as an end-user. If it's the later, i don't think it's worth it, at least for gtk-rs/gtk4-rs once we have proper stable/git docs.

bilelmoussaoui:
> > I think this is just gtk-rs being overly cautious here. What matters is what is distributed with the application, not what is inside the gtk-rs repository. But whatever makes people happy :)
> 
> No, the documentation is under LGPL, which isn't compatible. We simply can't put it in the same repository.
> 
> Also, small detail but quite important: the docs are **HUGE**. So for my disk and git history sake, please let's keep things separate.

The docs won't be moved anywhere as they can be generated on the CI from the gir-files. It's what I meant with having them in gtk4-rs. Of course the few manual ones if there are any in the future would have to be in a separate subdirectory.

MarijnS95:
> Where you would need such script? to test that gir works properly for the `doc` mode or to generate the docs locally as an end-user. If it's the later, i don't think it's worth it, at least for gtk-rs/gtk4-rs once we have proper stable/git docs.

To easily see the results when working on docs generation in `gir`, for example while working on #1089 and #1085. Or to build crates locally to view the docs. I could create/maintain such a script locally (or invoke the generator by hand for a single file) but it'd benefit everyone (and stay updated) if used in the respective repositories, right from the CI. We don't need to change this for gtk-rs/gtk4-rs though, but we can start doing it in GStreamer-rs.

MarijnS95:
Updated. Docs are now not generated anymore unless `--with-docs` is specified. This still regenerates normal files which might not be optimal for the CI documentation step; perhaps we should rename it to `--docs` and do one _or_ the other?

Also added `--docs-out` for GStreamer-rs so that we can have our `docs/` folder. I wanted to make `--with-docs` a tri-state but couldn't figure out how to:

- Not specified: don't generate docs at all;
- `--with-docs`: generate docs to the default path;
- `--with-docs docs/`: generate docs to `docs/`.

Suggestions welcome.

MarijnS95:
Figured it out: `nargs="?"`, with a value (`default=False`) when the flag is not passed, `const=True` when the flag is passed, and the regular `type=directory_path` will be invoked when a string argument is passed.

sdroege:
How should we proceed here, what are the open questions?

MarijnS95:
@sdroege The only remaining point is making docs generation mutually exclusive. Currently for GStreamer Rust files are always generated, and `--with-docs` _also_ generates docs.

That is not ideal when we want to use this script to regenerate just the documentation in the CI. Our current plan is to put this script to use in gtk-rs and gtk4-rs which currently both publish documentation based on the "outdated" lgpl-docs crate (The CI script generates `docs.md` but never substitute the sources for that crate, leading to missing or outdated documentation).
That is however for another PR, and will be discussed with @GuillaumeGomez how best to set this up.

MarijnS95:
@sdroege Turned this into `--docs` and fixed the remaining issues. Now we generate `*.rs` _or_ the docs when `--docs` is specified, not both at the same time. And as bonus we create the output directory as well instead of complaining it doesn't exist.

MarijnS95:
This PR now also calls `rustdoc-stripper` after `gir -m doc`, effectively doing the entire documentation process in one go. We have opted for this because it makes little sense to check in intermediate `docs.md` files, and the generator is already in charge of finding `Gir.toml` and running commands on it.

However, that cannot easily be parallelized with `subprocess` since we really want to:

1. Print a helpful line;
2. Invoke `gir` and print its `stderr`;
3. Print another helpful line;
4. Invoke `rustdoc-stripper` and print its `stdout`/`stderr`.

That was much easier when we had just one process per crate, we could just spawn it (`subprocess.Popen`) then later wait for all processes to complete (`.communicate()`) and print their path and `stderr` one by one, in order.

Now I have instead opted to move to `asyncio` that nicely wraps it all. The resulting code isn't too bad (IMO), but the remaining problem is commit history. I've pretty much dumped `rustdoc-stripper` + conversion to `asyncio` into into one commit, which is bad for readability. How should we tackle this?

1. - Add a `rustdoc-stripper` invocation to the `--with-docs` commit (5547e5a), effectively turning that into a simplified, synchronous `--embed-docs` variant;
    - Squash the `asyncio` change into the `subprocess` change (1d7e839) effectively hiding that interim state;
    - And have error handling simplifications "somewhere inbetween" :nauseated_face: .
2. - Leave the original commits from GStreamer, followed by;
    - Converting `subprocess` to `asyncio`;
    - Adding `rustdoc-stripper` on top.
3. The same as 2., except adding `rustdoc-stripper` first in a hacky way, then turn it all into `asyncio` (this is the current commit history).

Following 2. is easiest as I won't have to squash into older commits followed by _massive_ conflict resolution. Having that history that briefly resided in GStreamer shouldn't hurt either.

sophie-h:
Really looking forward to the rustdoc-stripper support :)

One quick question: rustdoc-stripper would usually be installed via cargo because it rarely changes in contrast to gir?

And one pedantic question: If generator.py changes the checkout in the gir git submodule, shouldn't it warn about it, because generator.py itself could have changed? (Comparing generator.py hashes would probably the most effective and simplest solution, if this is even a thing.)

MarijnS95:
Went for option 2. and already wrote some commentary in the commits.

> One quick question: rustdoc-stripper would usually be installed via cargo because it rarely changes in contrast to gir?

Correct, that matches what the CI does. We can always change it though, but `gir` seems a bit closer tied to the generation process and target crates than `rustdoc-stripper`, which is only used to embed documentation after the fact (and only for deployed code) and not really core to the development/code itself.

> And one pedantic question: If generator.py changes the checkout in the gir git submodule, shouldn't it warn about it, because generator.py itself could have changed?

Well, yes, that's a more tricky problem. I don't really like automagic submodule updating (it doesn't even work unless the folder is empty!) and it (should anyway) always follow the hash as committed in the surrounding git repository. That's better controlled by the user IMO, but let's see what the rest thinks.

sophie-h:
> That's better controlled by the user IMO, but let's see what the rest thinks.

Also, it's totally not needed for merging this. Maybe let's not do this here :laughing:

sdroege:
Yes, please create a separate PR for that or we'll never get this one here in. (FWIW: I agree that this behaviour should be removed from the script)

sdroege:
> Went for option 2.

That what I would've suggested. Easiest to read.

MarijnS95:
> > That's better controlled by the user IMO, but let's see what the rest thinks.
> 
> Also, it's totally not needed for merging this. Maybe let's not do this here

It _is_ worth looking into, because this is working as expected right now: pull/update the parent repo and the submodule hash + `generator.py` updates, before the user runs it. After this change the script won't be updated and restarted if the submodule changes.

However again, that has never worked because the generator only updates the submodule when the folder is empty, so on initial sync when `generator.py` doesn't exist yet.

sdroege:
Well, then remove it and discuss that with @GuillaumeGomez :) I think everybody else agrees that this would be the correct behaviour but he added that so he probably has a reason for that.

MarijnS95:
> Well, then remove it and discuss that with @GuillaumeGomez :) I think everybody else agrees that this would be the correct behaviour but he added that so he probably has a reason for that.

I will wait for his reply/justification. Perhaps it is better to reinstantiate this functionality, but that will suffer from the update/restart problem.

Anyway, the idea is to either symlink `./gir/generator.py` after this PR, or create a thin wrapper script that could take care of it, since updating repositories and discovering+regenerating crates are quite distinct tasks.

generator.py
```diff
@@ -0,0 +1,233 @@
...
+        ]
+
+        is_sys_crate = path.parent.name.endswith("sys")
+
+        if conf.embed_docs:
+            # Embedding documentation only applies to non-sys crates
+            if is_sys_crate:
+                return processes
+
+            processes.append(regenerate_crate_docs(conf.gir_path, path.parent, args))
```
MarijnS95, line 119:
```suggestion
            if not is_sys_crate:
                processes.append(regenerate_crate_docs(conf.gir_path, path.parent, args))
```

generator.py
```diff
@@ -0,0 +1,233 @@
...
+    logs += await spawn_gir(gir_exe, doc_args)
+
+    logs += "Embedding documentation from `{}` into `{}`...\n".format(
+        doc_path, crate_dir
+    )
+    stdout, stderr = await spawn_process("rustdoc-stripper", embed_args)
+    logs += stdout
+    logs += stderr
+
+    return logs
```
MarijnS95, line 99: Perhaps we should delete `docs.md` again after embedding instead of leaving unused cruft?

Otherwise:
- Generate docs to `/tmp`?
- Allow `gir` to print to `stdout`, and `rustdoc-stripper` to read from `stdin`?

The problem here, again, is that `--doc-target-path` is relative to the directory containing `Gir.toml`.

generator.py
```diff
@@ -0,0 +1,233 @@
...
+        run_command(["git", "pull", "-f", "origin", "master"], submodule_path)
+        print("<= Done!")
+        return True
+    return False
+
+
+def build_gir():
+    print("=> Building gir...")
+    update_workspace()
+    print("<= Done!")
```
MarijnS95, line 75: Could be merged into a single function.

MarijnS95 (commented):
Left a few self-comments, let me know if you like any of these suggestions.

generator.py
```diff
@@ -0,0 +1,233 @@
...
+async def spawn_process(exe, args):
+    p = await asyncio.create_subprocess_exec(
+        str(exe),
+        *args,
+        stdout=asyncio.subprocess.PIPE,
+        stderr=asyncio.subprocess.PIPE,
+    )
+
+    stdout, stderr = await p.communicate()
+    assert p.returncode == 0, stderr.strip()
```
GuillaumeGomez, line 27: What happens in case it panics? Also: `stderr.decode("utf-8").strip()`.

generator.py
```diff
@@ -0,0 +1,233 @@
...
+    return False
+
+
+def build_gir():
+    print("=> Building gir...")
+    update_workspace()
+    print("<= Done!")
+
+
+async def regenerate_crate_docs(gir_exe, crate_dir, base_gir_args):
```
GuillaumeGomez, line 78: Since we are now going to generate docs only from our side and not use lgpl-docs anymore, should we make `gir` generates doc comments directly into the code?

cc @sdroege @bilelmoussaoui

generator.py
```diff
@@ -0,0 +1,233 @@
...
+    logs = "Regenerating documentation for `{}` into `{}`...\n".format(
+        crate_dir, doc_path
+    )
+    logs += await spawn_gir(gir_exe, doc_args)
+
+    logs += "Embedding documentation from `{}` into `{}`...\n".format(
+        doc_path, crate_dir
+    )
+    stdout, stderr = await spawn_process("rustdoc-stripper", embed_args)
+    logs += stdout
```
GuillaumeGomez, line 96: Please add something that would allow to differentiate between stdout and stderr (in case it's not empty) like `=== STDOUT ===` or equivalent.

generator.py
sdroege: As we run `gir` anyway that seems like a good idea... but also will require quite some work on gir. Let's do that as a second step after merging to get rid of rustdoc-stripper completely?

generator.py
```diff
@@ -0,0 +1,233 @@
...
+        build_gir()
+
+    print("=> Regenerating crates...")
+    for path in conf.path:
+        print("=> Looking in path `{}`".format(path))
+        # Collect and print the results as soon as they trickle in, one process at a time:
+        for coro in asyncio.as_completed(regen_crates(path, conf)):
+            print(await coro)
+
+    if not conf.no_fmt and not run_command(["cargo", "fmt"]):
```
GuillaumeGomez, line 221:
```suggestion
    if not conf.no_fmt or not run_command(["cargo", "fmt"]):
```

generator.py
GuillaumeGomez: Not sure it would require that much work. But can be done as a second step.

generator.py
MarijnS95: Oh fair, I blindly moved this assert above the `stderr = stderr.decode("utf-8")`, that's why it was below at first...

> What happens in case it panics?

These asserts bubble up through the call stack and re-raise inside `asyncio.run`, when the task is awaited inside:

```python
        for coro in asyncio.as_completed(regen_crates(path, conf)):
            print(await coro)
```

generator.py
MarijnS95: I don't think we want to `return 1` when `no_fmt` is `False`? This chains `run_command` if `no_fmt` is false, should we apply some De Morgan instead?

```suggestion
    if not (conf.no_fmt or run_command(["cargo", "fmt"])):
```

MarijnS95: Perhaps this is clearer to read when separated in two if statements:
```suggestion
    if not conf.no_fmt:
        if not run_command(["cargo", "fmt"]):
```

GuillaumeGomez: Oh indeed!

MarijnS95: We can also perform the inversion for `no_fmt` inside `argparse` if you want, with `default=True` and `store_false`?

generator.py
MarijnS95: Turned all these (including the `stderr` in `spawn_gir` into:

```python
    if stdout:
        logs += "===> stdout:\n" + stdout
    if stderr:
        logs += "===> stderr:\n" + stderr
```

Okay with you? Will publish the changes later.

GuillaumeGomez: Might need a bit more. I'd prefer to have each of them surrounded by an empty line (to make them easier to spot). So you'll need to check that `logs` ends with `\n\n` to add them yourself. And you also need to add another `\n` after the `:`.

bilelmoussaoui:
One thing that would be nice to have, is a small note/description on the README for gir users about the script and what it does/why it's useful.

generator.py
sdroege: Good, I like incremental improvements :)

MarijnS95:
@bilelmoussaoui Despite the `README` stating that it's there to explain the concept of `gir` rather than how to use it, there is now a comment about `generator.py` regardless.

bilelmoussaoui:
> @bilelmoussaoui Despite the `README` stating that it's there to explain the concept of `gir` rather than how to use it, there is now a comment about `generator.py` regardless.

Indeed, that's an issue to solve some other day. Using something like mdbook as a simple README isn't enough for such complex project  :)

sdroege (approved):

GuillaumeGomez:
Sorry but I don't still don't see the benefit of having async here... For gtk-rs, it takes 5 secs to run (if I don't have to compile `gir`). It adds complexity over the stdout/stderr that isn't very welcome... :-/

MarijnS95:
@GuillaumeGomez The only added complexity is returning strings now, and capturing + separately emitting (as requested) stdout and stderr. While it might be nice to keep stdout in stdout, and stderr in stderr there's little chance separate logs are relevant unless we print all logging to stderr by default (relevant when these tools can write to stdout and read from stdin).

gtk-rs is not the benchmark here. See 1d7e839f7cb65d90e8e10eaa87773b682372ad4d, the impact is much larger on `gstreamer-rs` with more and presumably larger crates. Take off a few seconds again because we're not generating docs together with Rust code anymore, it's still way too much to fit an iterative process when quickly cross-checking the impact of `gir` changes across the codebase.

bilelmoussaoui (approved):

sdroege:
31s to 8s seems like a useful improvement to me and the code complexity increase seems manageable to me.

GuillaumeGomez:
Oh and why fight... I guess I'm the only one seeing issues here so whatever...

MarijnS95:
@GuillaumeGomez I don't want to (start a) fight, just want to show the significance this has on ones time. Maybe I'm the only one working on gstreamer-rs+gir and regeneration a lot, and the time it takes has been a significant barrier to my productivity. Thanks for accepting it regardless, we can always discuss ways to improve this (maintenance-wise)!

GuillaumeGomez:
Don't worry, I think it's mostly a point of view thing. I'll let others update this script in the future because if I'm the only one with this position, it might simply mean I'm in the wrong and don't have a clear sight of what's expected.

Issue #936, opened 2020-06-21
new_from_XXX() / new_with_XXX() functions are not renamed in the docs


sdroege:
Since https://github.com/gtk-rs/gir/pull/929 they are renamed to not have the `new` prefix anymore, but the renaming did not happen for the docs.

Should be fixed before the release :)

See https://gstreamer.pages.freedesktop.org/gstreamer-rs/gstreamer/struct.Pad.html#method.from_template for an example that has no docs anymore.

CC @EPashkin @GuillaumeGomez @gdesmott

gdesmott:
IIRC my patch should use the same code paths as when user manually rename a function using the toml file. Was doc broken for those as well?

sdroege:
I don't know, there are some normal renames in gtk and gio IIRC that you could use for testing

gdesmott:
That's strange, `DBusInterfaceExt.get()` and `WidgetExt.set_widget_name()` both use `rename`. The former is not documented (but neither are the other functions in this trait) but the former is.

sdroege:
Something for @GuillaumeGomez I guess :)

GuillaumeGomez:
To be checked...

sdroege:
@GuillaumeGomez You fixed this or not?

bilelmoussaoui:
cc @fengalin

PR #1098, opened 2021-04-14
Fix renamed functions not being documented

fengalin:
This is a workaround for renamed functions not being matched during doc generation. A better solution would require a rework of `gir` so that codegen processes "analysis augmented" items everywhere.

It's not enough to check the `configured_functions` for a `rename`d value since, among others, final renaming uses heuristics and performs name conflict resolution against existing methods and properties. The renaming logic is implemented in `analysis::functions`. The workaround consists in retrieving the `analysis::functions::Info` to get the `new_name`, if any.

This allows generating new `*.md` files with the expected names. Currently, these files are stored in https://github.com/gtk-rs/lgpl-docs. However, they are only updated when a new version is released. Discussions are in progress so as to generate these files locally on the fly. In the meantime, the result of this PR can only be tested locally by generating the `md` files and applying them using `rustdoc-stripper`, or by relying on a custom branch of https://github.com/gtk-rs/lgpl-docs and patching the target crate's workspace to select this branch.

Fixes https://github.com/gtk-rs/gir/issues/936

### DIFF
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -500,7 +500,7 @@ fn try_getter_rename(func: &library::Function, name: &str) -> Option<String> {
 fn analyze_function(
     env: &Env,
     obj: &config::gobjects::GObject,
-    mut name: String,
+    name: String,
     status: GStatus,
     func: &library::Function,
     type_tid: library::TypeId,
@@ -551,10 +551,7 @@ fn analyze_function(
         commented = true;
     }
 
-    let mut new_name = configured_functions
-        .iter()
-        .filter_map(|f| f.rename.clone())
-        .next();
+    let mut new_name = configured_functions.iter().find_map(|f| f.rename.clone());
 
     let bypass_auto_rename = configured_functions.iter().any(|f| f.bypass_auto_rename);
     if !bypass_auto_rename && new_name.is_none() {
@@ -567,7 +564,7 @@ fn analyze_function(
                     || name.starts_with("new_with")
                     || name.starts_with("new_for")
                 {
-                    name = name[4..].to_string();
+                    new_name = Some(name[4..].to_string());
                 }
             }
             _ => (),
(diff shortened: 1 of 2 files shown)


src/codegen/doc/mod.rs
```diff
@@ -281,8 +281,7 @@ fn create_object_doc(w: &mut dyn Write, env: &Env, info: &analysis::object::Info
             // field could have been renamed.
             if let Some(trait_name) = configured_functions
                 .iter()
-                .filter_map(|f| f.doc_trait_name.as_ref())
-                .next()
+                .find_map(|f| f.doc_trait_name.as_ref())
```
MarijnS95, line 284: We should really start enabling `clippy::pedantic` lints like https://rust-lang.github.io/rust-clippy/master/#filter_map_next, nice catch!

(Or just enable them now and fix the fallout, but there's `62` warnings on stable and `70` warnings on nightly...)

src/codegen/doc/mod.rs
```diff
@@ -291,20 +290,24 @@ fn create_object_doc(w: &mut dyn Write, env: &Env, info: &analysis::object::Info
...
-        create_fn_doc(w, env, function, Some(Box::new(ty)), fn_name)?;
+        if let Some(c_identifier) = &function.c_identifier {
+            // Retrieve the new_name computed during analysis, if any
+            let fn_new_name = info.functions.iter().find_map(|f| {
+                if &f.glib_name == c_identifier {
+                    f.new_name.clone()
+                } else {
+                    None
+                }
+            });
```
MarijnS95, line 301: I don't think you absolutely need to do the cloning inside the `find_map`, would this be more concise:

```suggestion
            let fn_new_name = info
                .functions
                .iter()
                .find(|f| &f.glib_name == c_identifier)
                .and_then(|f| f.new_name.clone());
```

src/analysis/functions.rs
```diff
@@ -567,7 +567,7 @@ fn analyze_function(
                     || name.starts_with("new_with")
```
MarijnS95, line 567: Fyi there's another `filter_map().next()` just 10 lines above this.

MarijnS95 (approved):
Thanks, good catch! See a ton of updated function signatures in the GStreamer-rs docs, looking good :)

src/codegen/doc/mod.rs
fengalin: Much better, thanks! I had a feeling that we could do better (usually the case when returning `None`), but rushed to something else I guess.

src/analysis/functions.rs
fengalin: Couldn't find it.

MarijnS95:
```rust
    let mut new_name = configured_functions
        .iter()
        .filter_map(|f| f.rename.clone())
        .next()
```

At line 554-557.

fengalin: Thanks. Didn't look in the right file...
I think I had more of these in https://github.com/gtk-rs/gir/pull/1092, but that can wait.

MarijnS95: Thanks! I agree, better make a separate PR that enables some more strict clippy lints and fix them there :)

fengalin:
I think this can be merged. @MarijnS95 as cross checked the result.

sdroege (approved):

bilelmoussaoui (approved):

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #1100, opened 2021-04-15
Value trait refactoring

sdroege:
See https://github.com/gtk-rs/gtk-rs/pull/454

### DIFF
--- a/src/analysis/properties.rs
+++ b/src/analysis/properties.rs
@@ -214,6 +214,7 @@ fn analyze_property(
         }
         if type_string.is_ok() {
             imports.add("glib::StaticType");
+            imports.add("glib::ToValue");
         }
 
         Some(Property {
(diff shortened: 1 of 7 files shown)


src/analysis/properties.rs
```diff
@@ -239,6 +240,9 @@ fn analyze_property(
         if let Ok(ref s) = used_rust_type(env, prop.typ, true) {
             imports.add_used_type(s);
         }
+        if type_string.is_ok() {
+            imports.add("glib::ToValue");
```
bilelmoussaoui, line 244: Doesn't this also requires ToValueOptional?

sdroege: No, that's only used for a blanket `ToValue` implementation. As a user you'd always either use the `ToValue` trait or the `Value::get` (etc) API so no other trait needs to be in scope.

bilelmoussaoui: Makes sense, thanks !

PR #1101, opened 2021-04-16
config,git: Omit gir directory hash and URL if not a submodule

MarijnS95:
Intially plaguing GStreamer-rs this code grabs the git hash from the root repository if the specified gir directory is not a submodule.  GStreamer-rs since moved all gir files to subdirectories (because it is possible to have multiple now) but various external projects have a single gir file checked in to the root repo and do not want to see all generated files change the hash after every regeneration while the root repository is checked out on a different commit.

This change uses `git rev-parse --show-toplevel` to resolve where the repository (folder containing `.git`) for a given directory lives and ignores the hash and URL if that is equivalent to the canonical path of the root.

CC @sophie-h for requesting this :)

### DIFF
--- a/src/config/config.rs
+++ b/src/config/config.rs
@@ -4,7 +4,7 @@ use super::{
 };
 use crate::{
     config::error::TomlHelper,
-    git::{repo_hash, repo_remote_url},
+    git::{repo_hash, repo_remote_url, toplevel},
     library::{self, Library},
     nameutil::set_crate_name_overrides,
     version::Version,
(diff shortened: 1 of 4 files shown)


src/config/config.rs
```diff
@@ -69,18 +69,19 @@ pub struct GirVersion {
...
     fn new(gir_dir: impl AsRef<Path>) -> Self {
         let gir_dir = normalize_path(gir_dir);
+        let is_submodule = toplevel(&gir_dir) != Path::new(".").canonicalize().ok();
         Self {
             file_name: gir_dir
                 .file_name()
                 .map(|s| s.to_str().expect("OsStr::to_str failed").to_owned()),
-            hash: repo_hash(&gir_dir),
-            url: repo_remote_url(&gir_dir),
+            hash: is_submodule.then(|| repo_hash(&gir_dir)).flatten(),
```
MarijnS95, line 77: Theoretically we should add back `unwrap_or("???")` here if this is a submodule but we failed to extract the hash.

sdroege (approved):

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #1102, opened 2021-04-16
Get removal round 2

fengalin:
**Important: don't merge as is. There will be a matching release for [`fix-getters`](https://github.com/fengalin/fix-getters) when we are ready to merge this PR.**

This is the continuation of https://github.com/gtk-rs/gir/pull/1021.

While updating `gstreamer-rs`, we [decided](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/639#note_874289) to discuss `get` prefix elision in sublasses in a second round. Initially, I decided to stick by a conservative `get` prefix elision rule inspired by Rust [naming conventions for getter methods](https://doc.rust-lang.org/1.0.0/style/style/naming/README.html#getter/setter-methods-%5Brfc-344%5D):

> A method foo(&self) -> &T for getting the current value of the field.

This meant that a subset of all the `get` functions where not concerned by the elision. The name for functions such as `gdk_display_get_monitor_at_point`, which retrieves the `Monitor` given some coordinates, was kept as `get_monitor_at_point`. My point was that it was not a `getter` in the strict sense of returning the value of a field, like we would be able to do if the `field` was declared `pub`. The operation performs a computation and I thought that removing the `get` prefix for most of these functions would lead to semantic problems. When exchanging with projects maintainers, I realized that the benefits of this approach was not obvious.

While handling get elision in the subclass infrastructure, I reconsidered the API benefits if we didn't stick by the strict rule I enforced before. I must admit that the resulting API is much better and the semantic issues are few. Even though I browsed the changes, I might have missed cases where the `get` prefix should be restored. Should we find such a method in automatic code in the future, it would be just a matter of using `bypass_auto_rename = true` or `rename = "..."`.

I updated the [`fix-getters`](https://github.com/fengalin/fix-getters) migrations tools accordingly. Previous approach is still available as an option using `--conservative` from the command line.

This PR handles the followings:

* Use `type_ =>` in glib macros for consistency with `get_type()` which is renamed as `type_()`.
* Relax the constraints while identifying the `get` function candidates for renaming.
* Added bool getter substitutes for `need` -> `needs` & `shrink` -> `shrinks` in `fix-getters-rules`.
* Fix child-properties not being renamed and allow renaming the getter for 2 levels child-properties so that we can get `child-tab-expand` -> `child_tab_expands` instead of `child_is_tab_expand`.

See also:

* PR for `gtk-rs/gtk-rs`.
* PR for `gstreamer-rs`. I will update `gst-plugins-rs` shortly.
* I will update `gtk4-rs` shortly, but it should be a walk in the park now.


sdroege:
> * [PR](https://github.com/gtk-rs/gtk-rs/pull/455) for `gtk-rs/gtk-rs`.
> 
> * [MR](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/743) for `gstreamer-rs`. I will update `gst-plugins-rs` shortly.

These two MRs look great to me.

sdroege:
@bilelmoussaoui @GuillaumeGomez What do you think?

bilelmoussaoui:
> @bilelmoussaoui @GuillaumeGomez What do you think?

I have left a few comments on the gtk-rs PR, the changes looks fine to me otherwise

GuillaumeGomez:
Looks good to me as well.

Issue #1094, opened 2021-04-11
-m doc does not seem to generate output for manual objects


MarijnS95:
All objects declared manual - either through `status = "manual"` or `manual = ["TheNamespace.TheObject"]` - do not have LGPL documentation generated for them to the markdown files. However, `rustdoc-stripper` still removes all documentation comments from these non-auto files, which is odd.

- Is `gir` supposed to output documentation for these objects?
- Is `embed-lgpl-docs` capable of adding these docs to the right items and functions if they were defined manually?

bilelmoussaoui:
I always wondered if `doc_trait_name` does that. Or at least supposed to? 
```toml

    # write function docs to trait other than default "xxxExt",
    # also works in [object.signal] and [object.property]
    doc_trait_name = "SocketListenerExtManual"
```

MarijnS95:
@bilelmoussaoui These are not manual trait extensions but rather the implementation of the type and functions themselves. Granted, our `impl` block is called for example `GLMemoryRef` for the `GstGL.GLMemory` type, but then the docs should _at least_ show up in markdown, which they don't either.

MarijnS95:
I think I know what is going on here.

The types I'm working with `(GLBuffer`, `GLMemory`) are boxed types without memory management functions bailing in `Missing memory management functions for`, this must also prevent them from generating docs. We recently added `get_type` to the C bindings to fix this and regenerating with this makes the documentation show up!

MarijnS95:
Part one of this issue has been tackled: #1103. Docs for these types now show up in `docs.md`.

However, in GStreamer we define member functions on these special "miniobject" types (`GLBuffer`, `GLMemory`) inside `impl GLMemoryRef` (and associated functions in `impl GLMemory` if available), but `rustdoc-stripper` doesn't know this. We'll need to see how to do that, perhaps this is already possible.

Will close this after having some other place to proceed with `rustdoc-stripper`.

PR #1103, opened 2021-04-17
analysis/record: Do not bail on missing memory functions

MarijnS95:
This constraint is already enforced with a `panic!` during codegen,
effectively disallowing Gir.toml from specifying objects that cannot be
generated.  It is now more obvious instead of being just an error line
in the (already pretty spammy) logs.

Furthermore analysis is also used to generate documentation, including
those for manual types.  Bailing here means no analysis for the type
is available leading to missing documentation.  After this change we get
a bunch of missing text to show up in `docs.md` in GStreamer for the
manual types that lack memory functions.

See also #1094

---

I have been thinking about wrapping `return None;` in `needs_generate()` but that seems such an unnecessary and arbitrary constraint, let's catch it in codegen instead.

### DIFF
--- a/src/analysis/record.rs
+++ b/src/analysis/record.rs
@@ -180,7 +180,6 @@ pub fn new(env: &Env, obj: &GObject) -> Option<Info> {
             }
         } else {
             error!("Missing memory management functions for {}", full_name);
-            return None;
         }
     }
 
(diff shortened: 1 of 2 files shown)


MarijnS95:
This also finds types in `generate = []` that were silently ignored previously: https://github.com/gtk-rs/gtk-rs/pull/459

sdroege (approved):

bilelmoussaoui (approved):

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #1106, opened 2021-04-22
WIP: builder: prefix properties of parent class interfaces with the crate …

### LINKED ISSUE #1099: Potential naming clash of vX_Y features between libraries
While updating libhandy-rs I realized for the first time, that gir generated code contains version numbers that are a mix of all involved dependencies.

Currently, this is not problem because the version numbers of libhandy, glib and gtk are neatly separated. But I was wondering if this is really future proof. Maybe having version names like `gtk_v3_24` could prove helpful at some point. (It will add some extra noise to gtk's Cargo.toml though.)

Example from libhandy-rs

```rust
pub struct ApplicationWindowBuilder {
    show_menubar: Option<bool>,
    #[cfg(any(feature = "v2_4", feature = "dox"))]
    #[cfg_attr(feature = "dox", doc(cfg(feature = "v2_4")))]
    accept_focus: Option<bool>,
    //application: /*Unknown type*/,
    #[cfg(any(feature = "v3_4", feature = "dox"))]
    #[cfg_attr(feature = "dox", doc(cfg(feature = "v3_4")))]
    attached_to: Option<gtk::Widget>,
    #[cfg(any(feature = "v2_4", feature = "dox"))]
    #[cfg_attr(feature = "dox", doc(cfg(feature = "v2_4")))]
    decorated: Option<bool>,
    default_height: Option<i32>,
    default_width: Option<i32>,
```

Activate all gtk3 versions in libhandy-rs

```toml
[features]
default = [ "v1_0" ]
v1_0 = [ "v3_24", "ffi/v1_0" ]
v1_2 = [ "v1_0", "ffi/v1_2" ]
dox = []

# GTK versions
v3_0 = []
v3_2 = [ "v3_0" ]
v3_4 = [ "v3_2" ]
v3_6 = [ "v3_4" ]
v3_8 = [ "v3_6" ]
v3_10 = [ "v3_8" ]
v3_12 = [ "v3_10" ]
v3_14 = [ "v3_12" ]
v3_16 = [ "v3_14" ]
v3_18 = [ "v3_16" ]
v3_20 = [ "v3_18" ]
v3_22 = [ "v3_20" ]
v3_24 = [ "v3_22" ]
```

bilelmoussaoui:
…name

this only should be done if the props are from a different library
fixes #1099

Before: it would generate something like
```rust
    #[cfg(any(feature = "v3_0", feature = "dox"))]
    #[cfg_attr(feature = "dox", doc(cfg(feature = "v3_0")))]
    hexpand_set: Option<bool>,
    #[cfg(any(feature = "v1_0", feature = "dox"))]
    #[cfg_attr(feature = "dox", doc(cfg(feature = "v1_0")))]
```

after:
```rust
    #[cfg(any(feature = "gtk_v3_0", feature = "dox"))]
    #[cfg_attr(feature = "dox", doc(cfg(feature = "gtk_v3_0")))]
    hexpand_set: Option<bool>,
    #[cfg(any(feature = "gtk_v1_0", feature = "dox"))]
    #[cfg_attr(feature = "dox", doc(cfg(feature = "gtk_v1_0")))]
```

### DIFF
--- a/src/analysis/properties.rs
+++ b/src/analysis/properties.rs
@@ -31,6 +31,18 @@ pub struct Property {
     pub set_bound: Option<PropertyBound>,
     pub version: Option<Version>,
     pub deprecated_version: Option<Version>,
+    pub parent_crate: Option<String>,
+}
+
+impl Property {
+    pub fn parent_crate_feature(&self) -> Option<String> {
+        if let Some(parent_crate) = &self.parent_crate {
+            self.version
+                .map(|v| format!("feature = \"{}_{}\"", parent_crate, v.to_feature()))
+        } else {
+            None
+        }
+    }
 }
 
 pub fn analyze(
@@ -106,6 +118,7 @@ fn analyze_property(
         .min()
         .or(prop.version)
         .or(Some(env.config.min_cfg_version));
+
     let generate = configured_properties
         .iter()
         .filter_map(|f| f.generate)
@@ -230,6 +243,7 @@ fn analyze_property(
             bounds: Bounds::default(),
             version: prop_version,
             deprecated_version: prop.deprecated_version,
+            parent_crate: None,
         })
     } else {
         None
@@ -265,6 +279,7 @@ fn analyze_property(
             bounds: Bounds::default(),
             version: prop_version,
             deprecated_version: prop.deprecated_version,
+            parent_crate: None,
         })
     } else {
         None
--- a/src/analysis/class_builder.rs
+++ b/src/analysis/class_builder.rs
@@ -101,8 +101,7 @@ fn analyze_property(
         .iter()
         .filter_map(|f| f.version)
         .min()
-        .or(prop.version)
-        .or(Some(env.config.min_cfg_version));
+        .or(prop.version);
 
     let for_builder = prop.construct_only || prop.construct || prop.writable;
     if !for_builder {
--- a/src/analysis/class_builder.rs
+++ b/src/analysis/class_builder.rs
@@ -24,7 +24,7 @@ pub fn analyze(
     }
 
     let mut names = HashSet::<String>::new();
-    let mut builder_properties = analyze_properties(env, props, obj, imports, &mut names);
+    let mut builder_properties = analyze_properties(env, props, obj, imports, &mut names, None);
 
     for &super_tid in env.class_hierarchy.supertypes(type_tid) {
         let type_ = env.type_(super_tid);
@@ -41,8 +41,14 @@ pub fn analyze(
                 continue;
             };
 
-        let new_builder_properties =
-            analyze_properties(env, super_properties, super_obj, imports, &mut names);
+        let new_builder_properties = analyze_properties(
+            env,
+            super_properties,
+            super_obj,
+            imports,
+            &mut names,
+            Some(super_obj),
+        );
         builder_properties.extend(new_builder_properties);
     }
 
@@ -55,6 +61,7 @@ fn analyze_properties(
     obj: &GObject,
     imports: &mut Imports,
     names: &mut HashSet<String>,
+    parent_object: Option<&GObject>,
 ) -> Vec<Property> {
     let mut builder_properties = Vec::new();
 
@@ -73,7 +80,7 @@ fn analyze_properties(
         if env.is_totally_deprecated(prop.deprecated_version) {
             continue;
         }
-        let builder = analyze_property(env, prop, &configured_properties, imports);
+        let builder = analyze_property(env, prop, &configured_properties, imports, parent_object);
         if let Some(builder) = builder {
             builder_properties.push(builder);
             names.insert(prop.name.clone());
@@ -88,6 +95,7 @@ fn analyze_property(
     prop: &library::Property,
     configured_properties: &[&config::properties::Property],
     imports: &mut Imports,
+    parent_object: Option<&GObject>,
 ) -> Option<Property> {
     let prop_version = configured_properties
         .iter()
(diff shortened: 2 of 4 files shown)


MarijnS95:
Is this prefixing the properties or the features?

bilelmoussaoui:
> Is this prefixing the properties or the features?

the features of course :)

bilelmoussaoui:
To add more details to this PR: 

When generated a Builder for an object that comes from a library based on GTK (or possibly other libraries); the generator assumes that all the properties comes from the same library. Which is wrong as the parent class/interfaces implemented by the said object can come from another library. 

In such case, we want to prefix those features with the library name where they come from so that we can have features to enable/disable them. Although this might add a ton of unwanted features. If you look at the case of libhandy/gtk 3, mentioned on the issue #1099, you will see generated features for properties that were added in gtk 3.0, or gtk 3.2 or other versions that are definitely supported by the gtk 3 crate as the min required version is gtk 3.14. We sadly can't know such information from the libhandy crate because it's defined in gtk's Gir.toml. I don't have any idea how to handle that in a way we avoid adding unnecessary features.

MarijnS95:
I see what this fixes, just that the title makes it sounds as if the property names themselves are prefixed ^^

Hadn't come across builders before, just native ffi types where the surrounding struct at least inherits the feature/version requirement of its members, otherwise things are misaligned and unusable. It is quite interesting to see that this mixes properties from external libraries and even tries to be smart enough to add per-member feature requirements rather than relying on the user to specify (in say, `Gir.toml`) the minimum version requirement for that builder as a whole, expressed as a local feature (and expecting the user to manually set the right feature requirements for its dependencies).

src/analysis/properties.rs
```diff
@@ -106,6 +118,7 @@ fn analyze_property(
         .min()
         .or(prop.version)
         .or(Some(env.config.min_cfg_version));
```
sophie-h, line 120: Not sure if this is the problematic line, but there is an issue that the `min_cfg_version` of the current crate is used for external objects. Therefore I now get

```diff
     can_default: Option<bool>,
+    #[cfg(any(feature = "gtk_v1_0", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "gtk_v1_0")))]
```

in libhandy.

bilelmoussaoui: Indeed. I will drop it

sophie-h:
There are a bunch of things that aren't fixed by this PR

1. The docs are still filled with messages, that things depend on gtk 2.x/gtk 3.x. I would probably just remove all gtk feature lines from the code as part of the doc generation.
2. We still need to put all gtk features into the libhandy cargo manifest including all gtk 2 versions. (Or remove the lines with a script from the source code after generation.)

bilelmoussaoui:
>  1. The docs are still filled with messages, that things depend on gtk 2.x/gtk 3.x. I would probably just remove all gtk feature lines from the code as part of the doc generation.

I have no idea how to handle such thing :) not from code perspective; but whether it makes sense to remove them or not. As the user needs to know which features they have to enable to access such things. The fact that all gtk 3.14 <= features are assumed as enabled is the problematic part but that's gtk3; not gir itself 

> 2. .We still need to put all gtk features into the libhandy cargo manifest including all gtk 2 versions. (Or remove the lines with a script from the source code after generation.)

Not a `gir` issue. But rather the gtk3 gir file should be fixed & drop all the gtk 2 versions. They are meaningless for gtk3 and it doesn't make sense to add working-around here for such specific use case. Regarding the rest of the features, i don't think there's a way around that :)

sophie-h:
libhandy requires something like gtk >= 3.24.1 that's why I think it's useless to keep them for libhandy

bilelmoussaoui:
> libhandy requires something like gtk >= 3.24.1 that's why I think it's useless to keep them for libhandy

Sure, i'm aware of that. But the issue is there's no way for gir to know such information. Unless you add min_cfg_version for the library's dependencies so you can specify that you require at min gtk 3.24 but that's a different issue that will require a ton of more work than this

I think we can patch the gtk3 gir file in libhandy repository for now to drop all those versions in properties but still have this patch merged for generic use cases.

sophie-h:
Yes totally I totally agree that we should merge this first.

This patch adds some missing requirements to gtk3 for example

```
+    #[cfg(any(feature = "gio_v2_58", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "gio_v2_58")))]
     pub fn is_tcrypt_system_volume(mut self, is_tcrypt_system_volume: bool) -> Self {
         self.is_tcrypt_system_volume = Some(is_tcrypt_system_volume);
```

But it also adds some noise to gtkx11 like

```
+    #[cfg(any(feature = "gdk_v3_0", feature = "dox"))]
+    #[cfg_attr(feature = "dox", doc(cfg(feature = "gdk_v3_0")))]
     pub fn display<P: IsA<gdk::Display>>(mut self, display: &P) -> Self {
         self.display = Some(display.clone().upcast());
         self
     }
```

which needs either a code fix after generation or a `gdk_v3_0` feature in gdk.

bilelmoussaoui:
Oh no, that's not great at all :) well i'm kind of lost how to handle this properly without adding a ton of probably unnecessary work. For the gdk-x11 at least, i think we should fix the gir file to drop the version.

any opinions cc @sdroege @fengalin ?

sophie-h:
> For the gdk-x11 at least, i think we should fix the gir file to drop the version.

What do you mean with fixing the gir file? Gdk's Gir.toml is at `min_cfg_version = "3.14"` but gdkx11 knows nothing about that.

bilelmoussaoui:
> For the gdk-x11 at least, i think we should fix the gir file to drop the version.

The real gir file, the one in gir-files repository to drop the version="3.0" from that property.

sophie-h:
Ah, the .gir-file :laughing: got it.

sophie-h:
Reasonable solution imho. Just add this to the fix.sh (for now.)

I think it's good to switch to the new naming convention asap and also get the missing feature guards in that this PR adds.

bilelmoussaoui:
Solved in a newer PR

PR #1108, opened 2021-04-24
doc: Always consider `trait_name` from `Gir.toml` in implementation docs

MarijnS95:
Some types have their trait name overridden for clarity, but this is not always conveyed in the docs.  This bit of code partially solves a case in GStreamer-rs where links show up to `gst::ObjectExt` that should have been `gst::GstObjectExt` instead, as per `trait_name = "GstObjectExt"`.

---

CC @slomo: This doesn't fix our `GstObjectExt` mismatch in GStreamer-rs _yet_! For that we'll have to replicate the override in every `Gir.toml` that uses the type:

```toml
[[object]]
name = "Gst.Object"
status = "manual"
trait_name = "GstObjectExt"
```

Do we have a smarter way of doing so, as well as for the other `trait_name =` overrides?

### DIFF
--- a/src/analysis/symbols.rs
+++ b/src/analysis/symbols.rs
@@ -13,7 +13,7 @@ pub struct Symbol {
 }
 
 impl Symbol {
-    pub fn full_rust_name(&self) -> String {
+    pub fn parent(&self) -> String {
         let mut ret = String::new();
         if let Some(ref s) = self.crate_name {
             if s == "gobject" {
@@ -27,6 +27,11 @@ impl Symbol {
             ret.push_str(s);
             ret.push_str("::");
         }
+        ret
+    }
+
+    pub fn full_rust_name(&self) -> String {
+        let mut ret = self.parent();
         ret.push_str(&self.name);
         ret
     }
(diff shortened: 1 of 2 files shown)


sdroege:
> Do we have a smarter way of doing so, as well as for the other `trait_name =` overrides?

Unfortunately the other crates don't know the configuration of each other. That same problem also exists in other regards already. And for example in gstreamer-rs we don't even have a way to know the `Gir.toml` of gio.

I think until we find a way to solve that, we need to duplicate the configuration and hope it stays a special case.

sdroege (approved):

MarijnS95:
> I think until we find a way to solve that, we need to duplicate the configuration and hope it stays a special case.

Luckily this isn't too bad for gstreamer. `BinExt` isn't referenced from any doc so no need to specify the `GstBinExt` anyhwere just yet. Nothing for `GES*` `ContainerExt`/`PipelineExt`/`TrackExt` renames either. The only remaining type is `ObjectExt` -> `GstObjectExt` in `audio`/`ges`/`rtsp-server`.

MarijnS95:
Fix for `gstreamer-rs` in https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/764. I don't want to add all renames to all `Gir.toml` preliminarily, but broken links (and worse, wrong ones like `glib::ObjectExt` when we want the `Gst` variant instead) easily go unnoticed. At some point we should just start parsing `Gir.toml` of our dependencies and at the very least put all its `status = "generate"` objects in the `manual` list - including said renames. Saves redeclaring every type, too :)
(I have no idea how big a can o' worms that's going to be, though :sweat_smile:).

PR #1111, opened 2021-04-24
docs: manual traits are in prelude only

### LINKED ISSUE #1110: Links to ExtManual traits are broken
See for example [here](https://gtk-rs.org/gtk4-rs/git/docs/gtk4/struct.Button.html#implements) for `WidgetExtManual`

According to @bilelmoussaoui this happens because the ExtManual traits are only in prelude.

bilelmoussaoui:
fixes #1110

### DIFF
--- a/src/codegen/doc/mod.rs
+++ b/src/codegen/doc/mod.rs
@@ -665,6 +665,6 @@ pub fn get_type_manual_traits_for_implements(
     manual_trait_iters
         .into_iter()
         .flatten()
-        .map(|name| format!("[`{n}`](trait@crate::{n})", n = name))
+        .map(|name| format!("[`{n}`](trait@crate::prelude::{n})", n = name))
         .collect()
 }
--- a/src/codegen/doc/mod.rs
+++ b/src/codegen/doc/mod.rs
@@ -635,7 +635,7 @@ fn get_type_trait_for_implements(env: &Env, tid: TypeId) -> String {
         format!("{}Ext", env.library.type_(tid).get_name())
     };
     if tid.ns_id == MAIN_NAMESPACE {
-        format!("[`{n}`](trait@crate::{n})", n = &trait_name)
+        format!("[`{n}`](trait@crate::prelude::{n})", n = &trait_name)
     } else if let Some(symbol) = env.symbols.borrow().by_tid(tid) {
         let full_trait_name = symbol.full_rust_name();
         format!("[`trait@{}Ext`]", &full_trait_name)


MarijnS95:
I don't remember if this holds true for GStreamer-rs too...

bilelmoussaoui:
I did check before submitting the MR; all the manual traits are defined i a prelude module https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/blob/master/gstreamer-video/src/lib.rs#L111, https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/blob/master/gstreamer-gl/src/lib.rs#L47 & https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/blob/master/gstreamer-base/src/lib.rs#L66 for example

MarijnS95:
@bilelmoussaoui Yup, looks good to me.

@sdroege Is it intended that we re-export some (but maybe not all?) traits from the crate root in addition to the prelude?

sdroege:
> @sdroege Is it intended that we re-export some (but maybe not all?) traits from the crate root in addition to the prelude?

The non-manual traits are all exported from the crate root or not? I think we should re-export also the manual ones there...

bilelmoussaoui:
> > @sdroege Is it intended that we re-export some (but maybe not all?) traits from the crate root in addition to the prelude?
> 
> The non-manual traits are all exported from the crate root or not? I think we should re-export also the manual ones there...

`pub use auto::traits::*` is only used inside the prelude module. The same way for all the other crates, isn't that expected? otherwise the list of exports would be huge :)

MarijnS95:
> > @sdroege Is it intended that we re-export some (but maybe not all?) traits from the crate root in addition to the prelude?
> 
> The non-manual traits are all exported from the crate root or not? I think we should re-export also the manual ones there...

The _non-manual_ traits are only exported from the prelude, as @bilelmoussaoui says `pub use crate::auto::traits::*;` is only inside `pub mod prelude`. However, it seems some (perhaps all, haven't checked this) `ExtManual` traits _are_ exposed from the crate root. I'm just wondering if we _should not_ do that and only reexport `ExtManual` from the prelude.

sdroege:
In that case yes, let's re-export all traits only from the prelude.

MarijnS95:
@sdroege That seems like quite a significant and breaking change, but perhaps for the better. More consistency when it comes to traits, I guess it doesn't make much sense to get the manual ones in the root and only get the autogenerated ones from the prelude only.

MarijnS95:
That was easier fixed than writing an actual commit message... https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/751

sdroege:
So that only affects GStreamer?

bilelmoussaoui:
> So that only affects GStreamer?

From a quick look at the various gtk-rs repositories, seems like it

MarijnS95:
In gtk-rs `auto/lib.rs` also exports non-manual `*Ext` traits, which are reexported from the crate root through `pub use crate::auto::*;`. 

Should we adapt `gir` to only leave these auto trait exports in `mod traits {}` so that they can be reexported from `mod prelude` only?

bilelmoussaoui:
> In gtk-rs `auto/lib.rs` also exports non-manual `*Ext` traits, which are reexported from the crate root through `pub use crate::auto::*;`.

The discussion is about the manual traits though, no?

MarijnS95:
Found one, `TlsConnectionManualExt`. That should be `TlsConnectionExtManual` if I'm not mistaken, and moved from `lib.rs` to `mod prelude` in `gio`, right?

> The discussion is about the manual traits though, no?

Correct, but following https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/751#note_893338 we're also exploring what's going on with auto- (`*Ext`) traits, which AFAIK should only be exported from `crate::auto::traits` and `crate::prelude`, not from the crate root? Otherwise it makes little sense to only hide the manual implementations behind the prelude? Seems rather inconsistent to me.

bilelmoussaoui:
> Found one, TlsConnectionManualExt. That should be TlsConnectionExtManual if I'm not mistaken, and moved from lib.rs to mod prelude in gio, right?

Yes, that's correct

> Correct, but following https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/751#note_893338 we're also exploring what's going on with auto- (*Ext) traits, which AFAIK should only be exported from crate::auto::traits and crate::prelude, not from the crate root? Otherwise it makes little sense to only hide the manual implementations behind the prelude? Seems rather inconsistent to me.

Oh, I see what you mean now. We always re-export the non-manual traits indeed. Well, I don't know then, maybe we can just close this PR & we re-export everything.

MarijnS95:
> Oh, I see what you mean now. We always re-export the non-manual traits indeed. Well, I don't know then, maybe we can just close this PR & we re-export everything.

This PR is probably correct, but then we should make sure not to export any traits from the crate root. In my limited understanding `ExtManual` is merely containing functions that could not be represented by the generator (or in some cases extra functions that might be useful?). This split should be oblivious to crate users, hence we should either:

- Remove `*Ext` exports from `auto` and keep everything within `mod traits` (reexported from `mod prelude`) only;
- Reexport all traits, including manual ones, from the crate root to be consistent with reexported `auto` traits.

What do you think @bilelmoussaoui? @sdroege?

sdroege:
Yes it should be one of the two. What would be the advantage of re-exporting the traits from the crate root?

MarijnS95:
> Yes it should be one of the two. What would be the advantage of re-exporting the traits from the crate root?

Less breakage for crates currently using Gtk/GStreamer and relying on those to be available when importing explicitly or `*` from the crate root? Other than that I don't see any. Either way, if we want to reexport from the crate root at some point we could use `pub use crate::auto::traits::*` anyway, if we go through with this cleanup the auto module wouldn't reexport the same traits twice at least.

bilelmoussaoui:
I would go for having everything in prelude. Most of people don't really know the various traits there's and just want to call the method they want; having to look for the right trait & the module to import it from doesn't sound nice in practice. Also not re-exporting every trait in lib.rs would reduce the very large lib.rs docs page. 

The user can still use a specific trait if they want by importing it from the module it comes from.

sdroege:
On Sun, 2021-04-25 at 09:32 -0700, Bilal Elmoussaoui wrote:
> I would go for having everything in prelude. Most of people don't
> really know the various traits there's and just want to call the
> method they want; having to look for the right trait & the module to
> import it from doesn't sound nice in practice. Also not re-exporting
> every trait in lib.rs would reduce the very large lib.rs docs page.
> The user can still use a specific trait if they want by importing it
> from the module it comes from.

I agree. Let's clean that up then

MarijnS95:
> Also not re-exporting every trait in lib.rs would reduce the very large lib.rs docs page.

It would be very nice 

> I agree. Let's clean that up then

We have https://github.com/gtk-rs/gtk-rs/pull/483 for gtk-rs, https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/751 for GStreamer-rs, and let's get these exports removed from `gir` ASAP.

bilelmoussaoui:
Let's merge this one then, I will come with a MR (unless someone wants to take care of that) to remove the exported and only keep them in the `auto::traits` module (which is already used for re-exporting the traits from the prelude).

MarijnS95:
@bilelmoussaoui I think we need more documentation changes because now non-manual traits also come from `prelude::` and not from the crate root anymore.

MarijnS95 (approved):
Works as expected together with #1113, thanks!

This should go in first though, as it is always valid.

sdroege (approved):

PR #1113, opened 2021-04-25
codegen/object: Only re-export traits from `traits`/`prelude` module

MarijnS95:
Following a discussion in [1] we have decided to remove all trait exports - both manual and automatic - from the crate root in favour of having them available to crate users in `prelude` only, as per the intended design.  It was inconsistent to have ExtManual traits available in `prelude` only (and some randomly in the crate root in GStreamer-rs) whereas all automatic traits were re-exported through the crate root as well.

[1]: https://github.com/gtk-rs/gir/pull/1111

CC @bilelmoussaoui

### DIFF
--- a/src/codegen/object.rs
+++ b/src/codegen/object.rs
@@ -402,15 +402,13 @@ pub fn generate_reexports(
         module_name, analysis.name, none_type
     ));
     if need_generate_trait(analysis) {
-        contents.extend_from_slice(&cfgs);
-        contents.push(format!(
-            "pub use self::{}::{};",
-            module_name, analysis.trait_name
-        ));
         for cfg in &cfgs {
             traits.push(format!("\t{}", cfg));
         }
-        traits.push(format!("\tpub use super::{};", analysis.trait_name));
+        traits.push(format!(
+            "\tpub use super::{}::{};",
+            module_name, analysis.trait_name
+        ));
     }
 
     if !analysis.builder_properties.is_empty() {


bilelmoussaoui:
The #1111 is ready as well :) I suppose we need to submit patches against gtk-rs/gtk4-rs with the changes of this PR before we can get it merged. I can take care of gtk4-rs/gtk-rs if you want to handle gstreamer's side of things.

MarijnS95:
@bilelmoussaoui Not sure what order they need to be in, afaik we need to merge this first to have a clean merge-commit hash to refer to?

The build fails because `generator.py` moved into this repo and gtk-rs simply has a symlink to `./gir/generator.py`, but the submodule wasn't checked out. Should be fixed with #1114.

bilelmoussaoui (approved):

bilelmoussaoui:
> @bilelmoussaoui Not sure what order they need to be in, afaik we need to merge this first to have a clean merge-commit hash to refer to?
> 
> The build fails because `generator.py` moved into this repo and gtk-rs simply has a symlink to `./gir/generator.py`, but the submodule wasn't checked out. Should be fixed with #1114.
> 
> GStreamer PR was just updated btw :)

We can merge both at the same time; my PR doesn't affect the generated code so it doesn't matter that much I think.

MarijnS95:
@bilelmoussaoui I meant the order of merging `gir` first, before updating gtk(4)-rs/gstreamer-rs with the fresh hash, regenerating, and merging the result.

bilelmoussaoui:
Normally we send a PR with the regenerated code, gir PR gets merged, we re-update the various PRs and then get them merged. But I'm fine with not doing it here, the changes are pretty contained and my tests locally seems to work fine.

MarijnS95:
Oh that's fine, I already had one open for `ExtManual` trait removal and shoved regeneration with these two PRs into there: https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/751

Let's do the same for gtk(4)-rs then so that we can see the final result (and make sure it compiles, with the few manual changes that need to be made).

bilelmoussaoui:
I have updated gtk4-rs to remove the few exported ExtManual traits at https://github.com/gtk-rs/gtk4-rs/pull/323, I have also regenerated with the changes to gir from this PR and it looks good to me.

sdroege (approved):

Issue #1105, opened 2021-04-21
add doc alias for renamed functions


bilelmoussaoui:
Especially after merging the renamed getters, some of the methods have a completely different name. Using doc_alias for that makes sense and would help users figure out the new method names.

cc @fengalin @GuillaumeGomez

fengalin:
I see that you declare this on `gir`. When I checked, doc alias where available in generated code.
I guess you are talking about manual code where there was no doc alias in the first place.

bilelmoussaoui:
What I meant is functions such as https://github.com/gtk-rs/gtk4-rs/blob/master/gtk4/src/auto/widget.rs#L152 would need another doc alias with `get_can_focus`. So every function that has a `rename = "new_method"` would use a `#[doc(alias = "old_method_name")]`. Same thing for methods that are automatically renamed as part of the getter removal

fengalin:
@GuillaumeGomez argued that it would lead to too many common names returned when searching in doc: https://github.com/gtk-rs/gir/pull/1014#discussion_r537558009

bilelmoussaoui:
what would be the difference between before adding the doc aliases & renaming and now? if you search for `get_X` & you get 10 results before, you would still get 10 results now.

fengalin:
@GuillaumeGomez your call :)

It would also be possible to get the `old_method_name` doc alias in manual code since `fix-getters-def` provides this behaviour as an option (no the default due to https://github.com/gtk-rs/gir/pull/1014#discussion_r537558009). I could apply a commit on current `master` based on the diff of a branch that would replace the `fix-getters-def` passes with one with the doc alias option. If we want this, we should do it before those commits are too deep in the commit stack though.

GuillaumeGomez:
No. :)

bilelmoussaoui:
Do you mind providing the reasoning behind ? Imho some function names are completely different from their original ones and the user will have a hard time finding about them.

sdroege:
I also think this would be useful.

GuillaumeGomez:
There are multiple ones:

 * It forces us to add yet another alias information, making the doc search less efficient (result-wise).
 * You can find an item using its C name.
 * It forces us to support yet another feature for only one release. Lot of efforts for close to no gain.

bilelmoussaoui:
Why for only one release? The renames are here to stay.

GuillaumeGomez:
[...]

PR #1116, opened 2021-04-26
Add rename doc aliases

fengalin:
This is an attempt at fixing gtk-rs/gir#1105. See also [the result for `gtk-rs`](https://github.com/gtk-rs/gtk-rs/pull/490).

It also addresses concerns described by @sdroege regarding some doc aliases which didn't match the name in upstream doc due to name mangling or property naming conventions. It also adds doc aliases for child properties which were not generated before.

Note that aliases from properties are no longer generated when they match the final name, hence the reduction in the "prop name alias" step below.

### DIFF
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -71,6 +71,7 @@ pub struct AsyncFuture {
 #[derive(Debug)]
 pub struct Info {
     pub name: String,
+    pub func_name: String,
     pub new_name: Option<String>,
     pub glib_name: String,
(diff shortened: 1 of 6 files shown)


sdroege:
@GuillaumeGomez Did you have a chance to look at this? Seems all fine to me (the results) and I can't see any downsides at this point.

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #1118, opened 2021-04-27
build: Rerun if git hash changed

MarijnS95:
`gir`'s hash in version files gets out of sync quite frequently with the hash of the submodule.  This happens when generating from a change locally, creating a PR and later regenerating on the merge commit.  When no file changes have been made `gir` doesn't get recompiled with the newer hash, but the submodule hash is still changed resulting in a mismatch.  It is usually caught later on another machine/checkout.

This change ensures `gir`'s hash is always updated when the checkout changes, even if no code changes have been made that would normally cause `cargo` to recompile it.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -6,6 +6,11 @@ use std::{fs::File, io::Write};
 mod git;
 
 fn main() {
+    let repo_path = git::git_dir(".").unwrap();
+    println!(
+        "cargo:rerun-if-changed={}",
+        repo_path.join("HEAD").display()
+    );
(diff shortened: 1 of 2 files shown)


sdroege (approved):

MarijnS95:
This is pretty useless and the title is false when you're on a branch. `HEAD` contains the name of the tracking branch (ie. `ref: refs/heads/dry`) which doesn't change when making or amending commits. Any alternative like `index`?

Issue #1109, opened 2021-04-24
Gir creates broken links to objects


Hofer-Julian:
I you currently look at the gtk4-rs Button docs, you will see quite a few of broken references.
For example:
"Button-like widgets like [classGtk.ToggleButton], [classGtk.MenuButton], [classGtk.VolumeButton], [classGtk.LockButton], [classGtk.ColorButton] or [classGtk.FontButton] use style classes such as .toggle, .popup, .scale, .lock, .color on the button node to differentiate themselves from a plain GtkButton."

https://gtk-rs.org/gtk4-rs/git/docs/gtk4/struct.Button.html

~According to @bilelmoussaoui this happens because the ExtManual traits are only in prelude.~ (we were talking about different things)

MarijnS95:
Gir doesn't  understand this pattern in the [documentation](https://raw.githubusercontent.com/gtk-rs/gir-files/6088bb634a406402f308f01d2f33067951d5095b/Gtk-4.0.gir):

```
Button-like widgets like [class@Gtk.ToggleButton], [class@Gtk.MenuButton],
[class@Gtk.VolumeButton], [class@Gtk.LockButton], [class@Gtk.ColorButton]
or [class@Gtk.FontButton] use style classes such as .toggle, .popup, .scale,
.lock, .color on the button node to differentiate themselves from a plain
`GtkButton`.
```

It triggers on the `@Gtk.ToggleButton` part but doesn't generate a link for it because no symbol with that name is found, afaik, and runs into:

https://github.com/gtk-rs/gir/blob/a972bd6d8d02cd44992811bd0a73b1da01c3a261/src/codegen/doc/format.rs#L121-L123

bilelmoussaoui:
The same way there's also [method@Gtk.Window.set_default_size], [enum@Gtk.SomeEnum], [vfunc@Gtk.Window.some_vfunc] & [signal@Gtk.Widget::some_signal] 

The method/enum/class should be pretty easy to handle I guess, at least in general. The signal/vfunc are not at all :)

bilelmoussaoui:
There seems to be a `property` thing as well

PR #1119, opened 2021-04-27
doc: Link trait functions in `prelude`

MarijnS95:
We recently removed all `Ext` and `ExtManual` traits from crate roots in favour of consistently exporting them from the prelude only, and at about the same time merged intradoc links for function symbols.  Both work as intended in isolation, but result in lots of broken links when put together because the documentation generator doesn't know about the existence of a `prelude` module.  This patch corrects the module path by leveraging existing `make_trait_method` machinery.

---
Old part of the comment, this is all fixed now:

~This breaks function references to "normal" `impl StructName` blocks. Could be a simple check for `Ext`/`ExtManual` suffix, but that's horrible.~

~All in all I'm really lost as to what the best way is to represent where a function lives, this should afaik be available in analysis and propagated into the documentation generator (through `symbol::Info`?). After all the name already contains `*Ext` or `*ExtManual` meaning an earlier step in the process is well aware of the type: this is where `prelude::` should likely be added, or otherwise through an `is_trait: bool` member. Sounds simple, but that has to wait until I have more time available - or someone else comes up with a nice solution :)~

### DIFF
--- a/generator.py
+++ b/generator.py
@@ -77,28 +77,40 @@ def build_gir():
     print("<= Done!")
 
 
-async def regenerate_crate_docs(gir_exe, crate_dir, base_gir_args):
+async def regenerate_crate_docs(conf, crate_dir, base_gir_args):
     doc_path = "docs.md"
     # Generate into docs.md instead of the default vendor.md
     doc_args = base_gir_args + ["-m", "doc", "--doc-target-path", doc_path]
 
     # The above `gir -m doc` generates docs.md relative to the directory containing Gir.toml
     doc_path = crate_dir / doc_path
-    embed_args = ["-g", "-m", "-o", doc_path, "-d", crate_dir / "src"]
+    embed_args = ["-m", "-d", crate_dir / "src"]
 
-    logs = "==> Regenerating documentation for `{}` into `{}`...\n".format(
-        crate_dir, doc_path
-    )
-    logs += await spawn_gir(gir_exe, doc_args)
+    logs = ""
 
-    logs += "==> Embedding documentation from `{}` into `{}`...\n".format(
-        doc_path, crate_dir
-    )
-    stdout, stderr = await spawn_process("rustdoc-stripper", embed_args)
-    if stdout:
-        logs += "===> stdout:\n\n" + stdout + "\n"
-    if stderr:
-        logs += "===> stderr:\n\n" + stderr + "\n"
+    if conf.strip_docs:
+        logs += "==> Stripping documentation from `{}`...\n".format(crate_dir)
+        # -n dumps stripped docs to stdout
+        _, stderr = await spawn_process("rustdoc-stripper", embed_args + ["-s", "-n"])
+        if stderr:
+            logs += "===> stderr:\n\n" + stderr + "\n"
+
+    if conf.embed_docs:
+        logs += "==> Regenerating documentation for `{}` into `{}`...\n".format(
+            crate_dir, doc_path
+        )
+        logs += await spawn_gir(conf.gir_path, doc_args)
+
+        logs += "==> Embedding documentation from `{}` into `{}`...\n".format(
+            doc_path, crate_dir
+        )
+        stdout, stderr = await spawn_process(
+            "rustdoc-stripper", embed_args + ["-g", "-o", doc_path]
+        )
+        if stdout:
+            logs += "===> stdout:\n\n" + stdout + "\n"
+        if stderr:
+            logs += "===> stderr:\n\n" + stderr + "\n"
 
     return logs
 
@@ -115,12 +127,12 @@ def regen_crates(path, conf):
 
         is_sys_crate = path.parent.name.endswith("sys")
 
-        if conf.embed_docs:
+        if conf.embed_docs or conf.strip_docs:
             # Embedding documentation only applies to non-sys crates
             if is_sys_crate:
                 return processes
 
-            processes.append(regenerate_crate_docs(conf.gir_path, path.parent, args))
+            processes.append(regenerate_crate_docs(conf, path.parent, args))
         else:
             if is_sys_crate:
                 args.extend(["-m", "sys"])
@@ -200,6 +212,11 @@ def parse_args():
         "--embed-docs",
         action="store_true",
         help="Build documentation with `gir -m doc`, and embed it with `rustdoc-stripper -g`",
+    )
+    parser.add_argument(
+        "--strip-docs",
+        action="store_true",
+        help="Remove documentation with `rustdoc-stripper -s -n`. Can be used in conjunction with --embed-docs",
     )
 
     return parser.parse_args()
--- a/src/analysis/symbols.rs
+++ b/src/analysis/symbols.rs
@@ -8,6 +8,7 @@ use std::collections::HashMap;
 #[derive(Clone, Debug, Default)]
 pub struct Symbol {
     crate_name: Option<String>,
+    module_name: Option<String>,
     owner_name: Option<String>,
     name: String,
 }
@@ -15,13 +16,17 @@ pub struct Symbol {
 impl Symbol {
     pub fn parent(&self) -> String {
         let mut ret = String::new();
-        if let Some(ref s) = self.crate_name {
-            if s == "gobject" {
-                ret.push_str("glib::object");
-            } else {
+        if Some("gobject") == self.crate_name() {
+            ret.push_str("glib::object::");
+        } else {
+            if let Some(ref s) = self.crate_name {
                 ret.push_str(s);
+                ret.push_str("::");
             }
-            ret.push_str("::");
+            if let Some(ref module) = self.module_name {
+                ret.push_str(module);
+                ret.push_str("::");
+            }
         }
         if let Some(ref s) = self.owner_name {
             ret.push_str(s);
@@ -37,11 +42,15 @@ impl Symbol {
     }
 
     pub fn make_trait_method(&mut self, trait_name: &str) {
+        if self.module_name.replace("prelude".to_string()).is_some() {
+            // .expect_none is not stabilized yet
+            panic!("{:?} already had a module name set!", self)
+        }
         self.owner_name = Some(trait_name.into());
     }
 
-    pub fn crate_name(&self) -> Option<&String> {
-        self.crate_name.as_ref()
+    pub fn crate_name(&self) -> Option<&str> {
+        self.crate_name.as_deref()
     }
 
     pub fn owner_name(&self) -> Option<&str> {
@@ -141,6 +150,7 @@ pub fn run(library: &Library, namespaces: &namespaces::Info) -> Info {
                             crate_name: crate_name.cloned(),
                             owner_name: Some(name.clone()),
                             name: member.name.to_camel(),
+                            ..Default::default()
                         };
                         info.insert(&member.c_identifier, symbol, Some(tid));
                     }
@@ -149,6 +159,7 @@ pub fn run(library: &Library, namespaces: &namespaces::Info) -> Info {
                             crate_name: crate_name.cloned(),
                             owner_name: Some(name.clone()),
                             name: func.name.clone(),
+                            ..Default::default()
                         };
                         info.insert(func.c_identifier.as_ref().unwrap(), symbol, None);
                     }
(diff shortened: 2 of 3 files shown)


bilelmoussaoui:
Figuring out whether a function is in `impl Struct` or in `StructExt`, `StructExtManual` would allow solving the methods part of #1109. But I haven't looked at that one yet.  

I think we can add a new attribute for functions that does the following:
- Check whether the type is final: if it's the case; use `Struct::function_name`
- If not, check whether it has a `doc_trait_name`, if so use that
- Otherwise fallback to `StructExt` trait name 

and then use that attribute on doc generation. Not sure how all of that is very doable :)

MarijnS95:
@bilelmoussaoui `doc_trait_name` is afaik only used for inherits/implements generation, right? This concerns conversion of "girdoc" (or what was it called?) "references" (`foo()`, `#Foo.bar()` etc) to intradoc links for easier navigation.

Afaik the full location/name should be fully defined during analysis: this is where we should feed the `symbols` list? Then there's no need to figure anything out, if it already contains the right names.

bilelmoussaoui:
> @bilelmoussaoui doc_trait_name is afaik only used for inherits/implements generation, right? This concerns conversion of "girdoc" (or what was it called?) "references" (foo(), #Foo.bar() etc) to intradoc links for easier navigation.

Well, I think `doc_trait_name` is what's used to tell to rustdoc-stripper in which trait the new manual implementation of the function is so it can be documented. Not sure about that though. 

I know it sounds not very related, but you need to figure out from a function name, in which trait/struct that function is defined to re-generate an intradoc link for it. Which would require the dance I mentioned above. Unless I misunderstood something?

MarijnS95:
> Well, I think `doc_trait_name` is what's used to tell to rustdoc-stripper in which trait the new manual implementation of the function is so it can be documented. Not sure about that though.

Correct. But we don't need to figure out the trait nor function name, we need to figure out (or rather, correct) the **module** this trait resides in. Only some are implemented directly for a struct (`impl StructName`) and exported from the crate root (these are not actually traits). All the rest - implementations through traits for classes, the `trait *Ext` stuff - resides under the `prelude` module instead.

In that sense I don't think we can (or want, for _every class_) use `doc_trait_name` to add `prelude::`.

> I know it sounds not very related, but you need to figure out from a function name, in which trait/struct that function is defined to re-generate an intradoc link for it. Which would require the dance I mentioned above. Unless I misunderstood something?

We already have that dance, that's the `symbols` list. This maps a raw C (unique) symbol name into a Rust symbol. It simply seems `struct Symbol` is missing a way to store/represent the extra `prelude` module. Should be easy to add, but then I'll need to check how `analysis/symbols.rs` can get hold of that information the same way the actual analysis for classes derives it. Might also be trivial, I just haven't found a single second to read through this file yet (should have better done so _before_ creating this PR).

MarijnS95:
@bilelmoussaoui The answer was right under my nose:

https://github.com/gtk-rs/gir/blob/05168259538c9d78d2cf41016481426a597c0342/src/analysis/symbols.rs#L39-L41

It just took me to figure out where `"Ext"` was appended because it wasn't happening in the analysis when `Type::Class(Class {..})` was encountered.

Now the fix is proper and simple: When that function is called we add `"prelude"` to the modules and emitted string and everything is golden :)

sophie-h:
I have used

```
./generator.py
./generator.py --embed-docs
```

on gtk4 with this branch and the result does not look too good?

```diff
+Generated by gir (https://github.com/gtk-rs/gir @ 0b6d4f5)
...
+    /// [`trait@glib::object::ObjectExt`], [`trait@gio::ListModelExt`]
```

MarijnS95:
Oh yuck, that's a different bit of code (`get_type_trait_for_implements`) that's affected separately :/

Added a hack, this should work (at least it does on GStreamer). We might need to find a more elegant solution like `full_rust_name_with_module("prelude")` or `parent_with_module("prelude")` like the first iteration of this PR, suggestions welcome!

sophie-h:
Now we have

```diff
-    /// [`ShortcutActionExt`](trait@crate::prelude::ShortcutActionExt), [`trait@glib::object::ObjectExt`]
+    /// [`ShortcutActionExt`](trait@crate::prelude::ShortcutActionExt), [`trait@glib::object::prelude::ObjectExt`]
```

which I think is invalid. (I don't know the details of the gobject structure.)

I'm sorry :sob:

bilelmoussaoui:
> which I think is invalid. (I don't know the details of the gobject structure.)

maybe we should just stop using `glib::object` as a prefix for `ObjectExt` here; it's on glib's prelude.

MarijnS95:
Oh yuck, let's try that again.

bilelmoussaoui:
Seems to work fine now :) there are still some broken links but those are unrelated to the prelude change

sdroege:
So this is OK apart from the `HACK` commit message prefix?

MarijnS95:
@sdroege It's not that clean of a change, but works a treat. As mentioned in https://github.com/gtk-rs/gir/pull/1119#issuecomment-830876746 suggestions for improvemenr are welcome! Otherwise, let's remove HACK and add keep it like this.

sdroege:
I don't have any suggestions, but maybe @GuillaumeGomez does.

generator.py
```diff
@@ -84,17 +84,25 @@ async def regenerate_crate_docs(gir_exe, crate_dir, base_gir_args):
...
+    embed_args = ["-m", "-d", crate_dir / "src"]
 
     logs = "==> Regenerating documentation for `{}` into `{}`...\n".format(
         crate_dir, doc_path
     )
     logs += await spawn_gir(gir_exe, doc_args)
 
+    logs += "==> Stripping documentation in `{}`...\n".format(crate_dir)
+    # -n dumps stripped docs to stdout
+    _, stderr = await spawn_process("rustdoc-stripper", embed_args + ["-s", "-n"])
```
GuillaumeGomez, line 96: Why are we running `rustdoc-stripper` twice btw?

MarijnS95: See the commit introducing this:

> generator: Strip docs before embedding new ones
>
> Otherwise documentation is repeatedly added on subsequent runs, and
> rustdoc-stripper-ignore-next won't be needed either like it was before.

We need it to first strip/prune existing docs (usually that from a previous run) before embedding new docs again.

GuillaumeGomez: That's definitely not logical. Currently we only strip to go back to the original state. What you're doing is the opposite: we don't want to have the docs in the code by default. The simplest way would be to run `rustdoc-stripper` to add docs and then `git reset HEAD`.

MarijnS95: Correct, but that's a really annoying workflow. Contributors might also occasionally forget to add `rustdoc-stripper-ignore-next` which is now irrelevant if we never strip docs.

Also, I think you meant `git reset --hard`, `git reset` will only reset the index but not the working tree (where prior modifications from `rustdoc-stripper` reside).

GuillaumeGomez: Indeed. But anyway, maybe let's use two doc commands: one to add the documentation and the other to strip it instead of doing something like this?

MarijnS95: This looses its value for me if I have to run/chain two commands while developing - unless we can run them simultaneously, ie. `--strip-docs --embed-docs` runs both in the right order.

GuillaumeGomez: I don't see in which case you want to run both at the same time...

sdroege: Like explained in the commit message? :)

GuillaumeGomez: I still don't get in which context you want to use this command outside of the doc generation in the CI.

MarijnS95: You actually **don't** want/need to use it during doc generation on the CI: that'll have a clean checkout (but it's a good way to enforce `// rustdoc-stripper-ignore-next`).

It is however useful **locally during development**. You see, I'm doing quite a few docs-related PRs to `gir` lately, and any change I make has to be tested of course (like this PR!). Being able to make a quick change and re-running `./generator.py --embed-docs` is a lot easier than having to `git reset --hard && ./generator.py` and potentially loosing any other changes I might have made to the gtk-rs/gstreamer-rs dir I'm working in.

GuillaumeGomez: I disagree with how you do things here. I don't think it should do both in the same command: what happens if you added documentation on an item and forgot to add `// rustdoc-stripper-ignore-next`? You run the command to **generate** docs and bingo, you got your doc comment removed.

MarijnS95: > I disagree with how you do things here. I don't think it should do both in the same command

Hence the suggestion to put it behind two distinct flags, that can be used together?

> what happens if you added documentation on an item and forgot to add `// rustdoc-stripper-ignore-next`? You run the command to generate docs and bingo, you got your doc comment removed.

That would have happened before when running with `--features=embed-lgpl-docs`, too.

GuillaumeGomez: > Hence the suggestion to put it behind two distinct flags, that can be used together?

Ah my bad, I thought reading you didn't want that. It's an acceptable solution. :+1: 

> That would have happened before when running with `--features=embed-lgpl-docs`, too.

Not with `embed-lgpl-docs`, only with `purge-lpgl-docs`. ;)

MarijnS95: > Not with `embed-lgpl-docs`, only with `purge-lpgl-docs`. ;)



In `gstreamer-rs` we were always calling `lgpl_docs::purge`, also with `embed-lgpl-docs`: https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/commit/3f508963e76942099abb487093a2ae30ac757ac7. It looks like gtk-rs was doing the same: https://github.com/gtk-rs/gtk-rs/pull/465/files

GuillaumeGomez: Ah indeed. Well, then it's good as is. My bad.

MarijnS95: > Ah indeed. Well, then it's good as is. My bad.

Np, it had me confused at first. Would you still like two separate flags or is it fine to leave it as-is?

MarijnS95: If it's fine I'll mark this all as resolved and re-request reviews :)

GuillaumeGomez: I was typing my comment and forgot to finish it haha.

So I was about to add: might be nice to add an option which simply strip the doc comments (then we can call the function from the doc generation too).

MarijnS95: @GuillaumeGomez Done!

It helped find one stray documentation comment in GStreamer without `// rustdoc-stripper-ignore-next`, @sdroege what do you think, shall we do a `./generator.py --strip-docs && git diff --quiet` to combat these in the CI?

MarijnS95: > So I was about to add: might be nice to add an option which simply strip the doc comments (then we can call the function from the doc generation too).

@GuillaumeGomez Not exactly sure what you mean, I think this was the plan outlined above. We have `--embed-docs`, `--strip-docs`, and as obvious convenience, they can be used together like `--embed-docs --strip-docs` to first do a stripping/cleaning pass (for now stuff won't be stored anywhere, so we might stick to `purge-docs` terminology as per the original `build.rs`?) followed by an embedding pass, for rapid development.

bilelmoussaoui: @MarijnS95 we had that in the past with embed & then purge-lgpl docs, it's probably worth re-adding it :-)

MarijnS95: @bilelmoussaoui Yeah it's a good idea, currently we're loosing the docs for `fn deinit()` in `gstreamer/src/lib.rs`. However, we're _not_ loosing the docs for `gstreamer/src/clock_time.rs`. The rest of the manual documentation (only found in `gl_memory.rs` and `gl_filter.rs`, added by yours truly) is annotated `// rustdoc-stripper-ignore-next`. Curious why `clock_time.rs` is not affected...

GuillaumeGomez: I mean that we should have a command to generate docs and another to strip it. And the command which adds docs should call the function which strips it first internally like you currently do.

MarijnS95: @GuillaumeGomez That's what this does currently, except that `--embed-docs` _does not_ strip unless also passing `--strip-docs` which is a nice tradeoff IMO. Are you okay with that?

If so this PR is ready for review, thanks!

bilelmoussaoui: will do once this one gets merged :) thanks!

sdroege (approved):

PR #1120, opened 2021-04-27
out_param: follow aliases for imports

fengalin:
When generating a function with an output parameter which type is an alias, the actual type was not checked for additional imports.

Issue found auto-generating `gst-base::BaseSink::wait`.

### DIFF
--- a/src/analysis/out_parameters.rs
+++ b/src/analysis/out_parameters.rs
@@ -113,22 +113,27 @@ pub fn analyze(
 pub fn analyze_imports(env: &Env, parameters: &[Parameter], imports: &mut Imports) {
     for par in parameters {
         if par.direction == ParameterDirection::Out {
-            match *env.library.type_(par.typ) {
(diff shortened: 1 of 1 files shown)


src/analysis/out_parameters.rs
```diff
@@ -113,22 +113,27 @@ pub fn analyze(
...
-                    _ => imports.add("std::ptr"),
-                },
-                _ => (),
-            }
+            analyze_type_imports(env, par.typ, par.caller_allocates, imports);
+        }
+    }
+}
+
+pub fn analyze_type_imports(env: &Env, typ: TypeId, caller_allocates: bool, imports: &mut Imports) {
```
sdroege, line 121: Why a new `pub` function that is not used anywhere else (yet)? :)

sdroege (approved):

src/analysis/out_parameters.rs
fengalin: You're right, it doesn't need to be `pub`, however it needs to be a new function since it's recursive on a single out param when the type is `Alias`.

PR #1123, opened 2021-05-03
codegen: only use ret if a builder_postprocess is set

bilelmoussaoui:
this allows us to enable clippy::let_and_return lint

PR #1125, opened 2021-05-03
codegen: add missing } for build pattern

bilelmoussaoui:
fix a missing element from #1123

### DIFF
--- a/src/codegen/object.rs
+++ b/src/codegen/object.rs
@@ -287,7 +287,7 @@ impl {}Builder {{
         writeln!(
             w,
             r#"        let ret = {}::Object::new::<{}>(&properties)
-                .expect("Failed to create an instance of {}"");"#,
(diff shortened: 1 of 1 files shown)


src/codegen/object.rs
```diff
@@ -287,7 +287,7 @@ impl {}Builder {{
         writeln!(
             w,
             r#"        let ret = {}::Object::new::<{}>(&properties)
-                .expect("Failed to create an instance of {}"");"#,
+                .expect("Failed to create an instance of {}");"#,
```
sdroege, line 290: The commit message says to add a missing `}`, the commit removes an additional `"` :)

bilelmoussaoui: there are two commits :-)

bilelmoussaoui: sorry, i completely messed with the latest PR

sdroege (approved):

PR #1126, opened 2021-05-04
config/function: allow forcing 'constructor' annotation

fengalin:
Some `constructor`s are not annotated as such in `gir` files. This PR allows configuring a function as `constructor` so that it can be renamed according to the naming conventions for constructors.

Required for: https://github.com/gtk-rs/gtk-rs/pull/505

### DIFF
--- a/README.md
+++ b/README.md
@@ -484,6 +484,17 @@ name = "Gtk.TextBuffer"
     bypass_auto_rename = true
 ```
 
+Some constructors are not annotated as `constructor` in the `gir` files. In
+order for the naming convention to be applied, you can force a function to be
+considered as a constructor:
+
+```toml
+[[object.function]]
+name = "new_for_path"
+# Not annotated as constructor in Gir => force it to apply naming convention
+constructor = true
+```
+
 #### conversion_type "Option"
 
 The `conversion_type` variant `Option` is available for types `T` implementing
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -539,32 +539,37 @@ fn analyze_function(
     }
 
     let mut new_name = configured_functions.iter().find_map(|f| f.rename.clone());
+    let is_constructor = configured_functions.iter().find_map(|f| f.is_constructor);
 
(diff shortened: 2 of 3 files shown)


MarijnS95:
> Some `constructor`s are not annotated as such in `gir` files.

Is this something that could be fixed in the annotations/source-code instead? Or is that too much churn like is usually the case with GStreamer (annotation fixups generally being too noisy for backports to version 18.x where our gir files come from, having to wait for 1.20).

fengalin:
> Is this something that could be fixed in the annotations/source-code instead?

I'm not sure. You're probably right. I can look into this if it is straight-foward, otherwise I don't want to fall in another rabbit hole once again.

bilelmoussaoui:
> > Some `constructor`s are not annotated as such in `gir` files.
> 
> Is this something that could be fixed in the annotations/source-code instead? Or is that too much churn like is usually the case with GStreamer (annotation fixups generally being too noisy for backports to version 18.x where our gir files come from, having to wait for 1.20).

The same thing is needed for gio, a new release will be required before the patches lands on the gir files & we get to update them. Maybe it's worth it to have a configuration here but report the broken annotations upstream so we can drop them someday :)

MarijnS95:
@bilelmoussaoui At some point we need warnings telling that `Gir.toml` overrides are redundant when the source is fixed, so that we can slim down these extra annotations again :)

fengalin:
> @bilelmoussaoui At some point we need warnings telling that `Gir.toml` overrides are redundant when the source is fixed, so that we can slim down these extra annotations again :)

I can already add this to this PR.

sdroege (approved):
Looks good to me but should be reported to gobject-introspection. There is no annotation for this and it should detect this correctly, but does not.

fengalin:
It seems to be a pretty old known issue: https://gitlab.gnome.org/GNOME/gobject-introspection/-/issues/48

bilelmoussaoui (approved):

fengalin:
See also: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2082.

PR #1127, opened 2021-05-04
codegen: Emit `crate::` for all global `%` constants except from prelude

MarijnS95:
Depends on #1119.

Rust prelude symbols - `true`, `false` and `None` - are explicitly added to the list of symbols.  These are the only global constants/symbols that shouldn't receive a `crate::` prefix.  Everything else coming from the current or another crate must be imported from the root as it may reside in another file/module.

### DIFF
--- a/generator.py
+++ b/generator.py
@@ -77,28 +77,40 @@ def build_gir():
     print("<= Done!")
 
 
-async def regenerate_crate_docs(gir_exe, crate_dir, base_gir_args):
+async def regenerate_crate_docs(conf, crate_dir, base_gir_args):
     doc_path = "docs.md"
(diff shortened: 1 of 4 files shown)


sdroege (commented):
This contains the commit from https://github.com/gtk-rs/gir/pull/1119 too, so let's get that one merged first. Otherwise looks good to me.

MarijnS95:
> This contains the commit from #1119 too, so let's get that one merged first. Otherwise looks good to me.

Correct, we need `..Default::default()` for all the `Sybol {}` struct constructors, that was added there too (hence a force-push to include those commits again).

PR #1128, opened 2021-05-04
[RFC] codegen/doc: Simplify links by allowing `Self::` and `crate::` prefix

MarijnS95:
A lot of effort was put in to hide these from the final documentation, by using `[]()` kind of links.  Unfortunately, for symbols that don't exist (broken references in g-ir documentation, types that haven't been implemented yet, functions related to lifetime management and cleanup that are implicit in Rust) these generate clickable links pointing to a Rust item path (`crate::Foo::bar`) which is inconvenient.  On the other hand, without `[]()` indirection but simple `[]` intradoc links rustdoc doesn't emit a link for unknown items at all, leaving a simple monospace Rust path surrounded by brackets.

---

CC @GuillaumeGomez: we discussed this on IRC and it should be a nice simplifications. Seeing `Self::` or `crate::` in docs shouldn't be too bad, on the contrary readers immediately know where to look for a type instead of having to click through to find out where it's at.

Additionally `prelude::` prefixes will also be emitted now (see #1119) which should otherwise be stripped off as well IMO, if we want to go for as concise docs as possible.

### DIFF
--- a/src/codegen/doc/format.rs
+++ b/src/codegen/doc/format.rs
@@ -100,7 +100,7 @@ fn replace_c_types(entry: &str, symbols: &symbols::Info, in_type: &str) -> Strin
             if sym.owner_name() == Some(in_type) {
                 // `#` or `%` symbols should probably have been `@` to denote
                 // that it is a reference within the current type.
-                format!("[`{f}()`](Self::{f}())", f = sym.name())
+                format!("[`Self::{}()`]", sym.name())
             } else {
                 match caps.get(1).as_ref().map(Match::as_str) {
                     // Catch invalid @ references that have a C symbol available but do not belong
@@ -113,7 +113,7 @@ fn replace_c_types(entry: &str, symbols: &symbols::Info, in_type: &str) -> Strin
                         in_type,
                     ),
                     Some("#") | None => {
-                        format!("[`{f}()`](crate::{f}())", f = sym.full_rust_name())
+                        format!("[`crate::{}()`]", sym.full_rust_name())
                     }
                     Some("%") => panic!("% not allowed for {:?}", caps),
                     Some(c) => panic!("Unknown symbol reference {}", c),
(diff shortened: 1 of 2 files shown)


sdroege:
Please rebase on top of master so the already merged commits disappear :)

sdroege (commented):
That's not very beautiful but seems still better than broken links, yes. Maybe rustdoc should be fixed to not generated broken links but behave the same way in this situation?

MarijnS95:
> That's not very beautiful but seems still better than broken links, yes. Maybe rustdoc should be fixed to not generated broken links but behave the same way in this situation?

Indeed. I'm still hesitant about going this route, though `crate::` and `Self::` prefixes do make it clearer where the link ends up. Not sure whether rustdoc cando a whole lot here, the link on the right-hand side might still be valid in some context (hence tries to link the browser to it) despite it failing to parse/match a Rust item path. OTOH ``[`crate::some::Path`]` is unique to intradoc links and not used for anything else.

Perhaps an experimental mode/flag to omit `<a>` links if the link is detected to be a reference to Rust code, but that code doesn't exist?

GuillaumeGomez (approved):
That simplifies things a bit so I think it's a good idea like I already said. :)

MarijnS95:
Simplification _and_ being more explicit. I think we have two points _for_ this change, and _none_ against (ugly/verbose being countered by being more explicit instead).

GuillaumeGomez:
More explicit is debatable because that makes longer links after all. Readability vs explicitness, etc. Just wanted to nitpick. :p

For me there are only positive points in any case.

GuillaumeGomez:
(damn ctrl+enter...)

MarijnS95:
That is fair. `crate::` is obtuse because that is usually implied already (of course it's somewhere within this crate), especially when referencing external crates as `crate::<the crate>::<the item>`. `Self::` in most cases as well, unless the constant/function referred to lives in the same module. Oh well, let's go for simplicity instead of trying to over-engineer this.

PR #1130, opened 2021-05-05
funct async: use codegen name for _future funct

fengalin:
... otherwise we can't take advantage of config and auto-renaming
on non-future function.


sdroege (approved):

fengalin:
@sdroege : wait, there seems to be another pblm...

PR #1131, opened 2021-05-05
Fix some clippy/compiler warnings

sdroege:


src/codegen/property_body.rs
```diff
@@ -12,8 +12,6 @@ pub struct Builder<'a> {
     is_get: bool,
     is_child_property: bool,
     type_: String,
-    is_ref: bool,
```
GuillaumeGomez, line 15: Why is it not needed anymore?

sdroege: Because of the value trait refactoring. No code anywhere is using it and the compiler warns about it.

GuillaumeGomez:
Thanks!

PR #1135, opened 2021-05-13
Use try_from_glib shortcut

fengalin:
See https://github.com/gtk-rs/gtk-rs-core/pull/3.

I also renamed the function specifiers for `RustTypeBuilder` that I introduced in https://github.com/gtk-rs/gir/pull/1092/commits/c43d9a5c8314631ccf7350e5750d7df96e49c625 because I thought they were uselessly verbose.

### DIFF
--- a/src/config/error.rs
+++ b/src/config/error.rs
@@ -21,7 +21,7 @@ impl TomlHelper for toml::Value {
             None => return,
         };
         for (key, _) in table.iter() {
-            if options.iter().find(|&entry| entry == key).is_none() {
+            if !options.contains(&key.as_str()) {
                 ret.push(key.clone());
             }
         }
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -83,7 +83,7 @@ impl Bounds {
         configured_functions: &[&config::functions::Function],
     ) -> (Option<String>, Option<CallbackInfo>) {
         let type_name = RustType::builder(env, par.typ)
-            .with_ref_mode(RefMode::ByRefFake)
+            .ref_mode(RefMode::ByRefFake)
             .try_build();
         if (r#async && async_param_to_remove(&par.name)) || type_name.is_err() {
             return (None, None);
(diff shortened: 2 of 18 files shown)


src/config/error.rs
```diff
@@ -21,7 +21,7 @@ impl TomlHelper for toml::Value {
             None => return,
         };
         for (key, _) in table.iter() {
-            if options.iter().find(|&entry| entry == key).is_none() {
+            if options.iter().any(|&entry| entry == key) {
```
MarijnS95, line 24: Aren't you missing a `!` here?

```suggestion
            if !options.iter().any(|&entry| entry == key) {
```

But slices also have a contains method:
```suggestion
            if !options.contains(&key.as_str()) {
````

MarijnS95 (changes requested):

src/config/error.rs
fengalin: Ah right... I probably rushed the clippy pass and it didn't show up when I regen `gstreamer-rs`.

fengalin:
It seems CI breakage is unrelated, I'll check this later this weekend or next week.

sdroege:
> It seems CI breakage is unrelated, I'll check this later this weekend or next week.

Yes that's leftover work for @bilelmoussaoui and @GuillaumeGomez from splitting the repos. The CI here also needs to be updated.

sdroege (approved):

MarijnS95 (approved):

GuillaumeGomez:
Ah right, just let me fix it quickly and then we can restart the CI.

GuillaumeGomez:
Opened  #1137.

GuillaumeGomez:
Restarted CI.

GuillaumeGomez:
Ah right, you need to rebase to have the "new" CI. ;)

fengalin:
Thanks @GuillaumeGomez

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #1139, opened 2021-05-14
Stop dereferencing followed by reborrowing in `match` and `if let`

MarijnS95:
This is the churn part of #1134, initially intended to remove `derive(Copy)` from `enum Fundamental` to fit a `String` in there. Those were just addressing the compiler errors, but there are a lot - and I mean a lot - more cases where objects are first dereferenced (copied) like `match *self {}` and then later re-(mutably)-borrowed with `ref` and `ref mut` in match arms. This is not necessary anymore (both the `*` and `ref`/`ref mut` are simply removed) and makes the code more pleasant to read.

OTOH doing it this way satisfies https://rust-lang.github.io/rust-clippy/master/#pattern_type_mismatch.... Curious to all your thoughts!

~If this is okay I'll go ahead and finish the remainder of the offenders.~ Finished!

### DIFF
--- a/src/analysis/ffi_type.rs
+++ b/src/analysis/ffi_type.rs
@@ -34,19 +34,17 @@ pub fn ffi_type(env: &Env, tid: TypeId, c_type: &str) -> Result {
                 .maybe_ref_as::<Fundamental>()
                 .is_some()
             {
-                match *env.library.type_(tid) {
+                match env.library.type_(tid) {
                     Type::FixedArray(_, size, _) => {
                         ffi_inner(env, c_tid, c_type).map_any(|rust_type| {
                             rust_type.alter_type(|typ_| format!("[{}; {}]", typ_, size))
                         })
                     }
                     Type::Class(Class {
-                        c_type: ref expected,
-                        ..
+                        c_type: expected, ..
                     })
                     | Type::Interface(Interface {
-                        c_type: ref expected,
-                        ..
+                        c_type: expected, ..
                     }) if is_gpointer(c_type) => {
                         info!("[c:type `gpointer` instead of `*mut {}`, fixing]", expected);
                         ffi_inner(env, tid, expected).map_any(|rust_type| {
@@ -174,7 +172,7 @@ fn ffi_inner(env: &Env, tid: TypeId, inner: &str) -> Result {
 
 fn fix_name(env: &Env, type_id: TypeId, name: &str) -> Result {
     if type_id.ns_id == INTERNAL_NAMESPACE {
-        match *env.library.type_(type_id) {
+        match env.library.type_(type_id) {
             Type::Array(..)
             | Type::PtrArray(..)
             | Type::List(..)
--- a/src/analysis/rust_type.rs
+++ b/src/analysis/rust_type.rs
@@ -337,7 +337,7 @@ impl<'env> RustTypeBuilder<'env> {
                 if ConversionType::of(self.env, inner_tid) == ConversionType::Pointer =>
             {
                 skip_option = true;
-                let inner_ref_mode = match *self.env.library.type_(inner_tid) {
+                let inner_ref_mode = match self.env.library.type_(inner_tid) {
                     Class(..) | Interface(..) => RefMode::None,
                     _ => self.ref_mode,
                 };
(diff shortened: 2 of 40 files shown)


sdroege (commented):
I think this change is good, yes. It improves readability by removing some unnecessary noise.

MarijnS95:
@sdroege Thanks! I think I covered all instances of unnecessarily dereferencing followed by reborrowing, this is now ready for reveiw.

src/analysis/ffi_type.rs
```diff
@@ -127,8 +125,8 @@ fn ffi_inner(env: &Env, tid: TypeId, inner: &str) -> Result {
             }
             fix_name(env, tid, inner)
         }
-        Type::CArray(inner_tid) => ffi_inner(env, inner_tid, inner),
-        Type::FixedArray(inner_tid, size, _) => ffi_inner(env, inner_tid, inner)
+        &Type::CArray(inner_tid) => ffi_inner(env, inner_tid, inner),
+        &Type::FixedArray(inner_tid, size, _) => ffi_inner(env, inner_tid, inner)
```
sdroege, line 129: That's ugly :) Does this maybe want a `Type::FixedArray(ref inner_tid, size, _) => ...` or so?

src/analysis/override_string_type.rs
```diff
@@ -34,11 +34,11 @@ fn apply(env: &Env, type_id: TypeId, string_type: Option<config::StringType>) ->
             OsString => TypeId::tid_os_string(),
         }
     };
-    match *env.library.type_(type_id) {
+    match env.library.type_(type_id) {
         Type::Fundamental(Fundamental::Filename) => replace,
         Type::Fundamental(Fundamental::OsString) => replace,
         Type::Fundamental(Fundamental::Utf8) => replace,
-        Type::CArray(inner_tid) if can_overriden_fundamental(env, inner_tid) => {
+        &Type::CArray(inner_tid) if can_overriden_fundamental(env, inner_tid) => {
```
sdroege, line 41: Also here

src/analysis/rust_type.rs
```diff
@@ -333,11 +333,11 @@ impl<'env> RustTypeBuilder<'env> {
                     }
                 })
             }
-            List(inner_tid) | SList(inner_tid) | CArray(inner_tid) | PtrArray(inner_tid)
+            &List(inner_tid) | &SList(inner_tid) | &CArray(inner_tid) | &PtrArray(inner_tid)
```
sdroege, line 336: And here and in many other places

src/analysis/ffi_type.rs
MarijnS95: That's the other way around :) - realistically we need `Type::FixedArray(&inner_tid, &size, _)` to destructure the borrow on _just_ those two fields.

src/analysis/override_string_type.rs
MarijnS95: Same here, you cannot "destructure" the borrow with `Type::CArray(&inner_tid)` - note that this is _the inverse of_ `ref` :)

sdroege (commented):
Looks good to me, big improvement :)

Only please don't put `&` into patterns but instead use `ref` / `ref mut`.

src/analysis/override_string_type.rs
MarijnS95: Alternative is to write `*inner_tid` and friends everywhere.

MarijnS95:
> Only please don't put `&` into patterns but instead use `ref` / `ref mut`.

`&` is the inverse of `ref`/`ref mut` so that won't work ;)

sdroege:
Yeah so do it the other way around and `match *var` in these cases. Having `&` in patterns is very ugly

MarijnS95:
> Yeah so do it the other way around and `match *var` in these cases. Having `&` in patterns is very ugly

Those are unfortunately exactly the cases I need to make #1134 work. But then we can add `ref` in there too and it should work. I don't find it particularly ugly though, but who am I to say :)

sdroege:
I'm quite sure that even was a clippy lint at some point (or maybe it's disabled by default now). You ideally never want to have `&` in patterns but only use `ref` / `ref mut` and match in the other direction if needed.

MarijnS95:
> I'm quite sure that even was a clippy lint at some point (or maybe it's disabled by default now). You ideally never want to have `&` in patterns but only use `ref` / `ref mut` and match in the other direction if needed.

Not that I recall, I only know about https://rust-lang.github.io/rust-clippy/master/#pattern_type_mismatch. Changes have been reverted now :)

PR #1143, opened 2021-05-19
codgen: generate doc_alias for static_type

bilelmoussaoui:
this allows finding *_type() functions using their c names

See https://github.com/gtk-rs/gtk4-rs/pull/383 for the end result

### DIFF
--- a/src/codegen/enums.rs
+++ b/src/codegen/enums.rs
@@ -421,13 +421,18 @@ impl FromGlib<{sys_crate_name}::{ffi_name}> for {name} {{
         writeln!(
             w,
             "impl StaticType for {name} {{
-    #[inline]
-    fn static_type() -> {glib_type} {{
-        unsafe {{ from_glib({sys_crate_name}::{get_type}()) }}
-    }}
-}}",
-            sys_crate_name = sys_crate_name,
+                #[inline]",
             name = enum_.name,
+        )?;
+        doc_alias(w, get_type, "", 1)?;
+        writeln!(
+            w,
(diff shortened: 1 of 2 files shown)


sdroege (commented):
That of course only helps for enums/flags and not for boxed/shared/object types :) Not sure how useful that is overall.

bilelmoussaoui:
> That of course only helps for enums/flags and not for boxed/shared/object types :) Not sure how useful that is overall.

For boxed/shared/object types, we need a way for glib::wrapper! to add the doc aliases which is not possible as far as I can tell, you can't stringify something and pass it to `#[doc(alias = "x")]` if `type_` is passed 

I don't think it's super useful, but it would reduce the questions of how can I get `get_type()` for this type. Especially if people are copying a C example/looking at the C docs

bilelmoussaoui:
So I suppose this is something we should just close?

sdroege:
Do doc aliases on trait impls actually work? I thought they only work on trait definitions

bilelmoussaoui:
> Do doc aliases on trait impls actually work? I thought they only work on trait definitions

They do, see https://gtk-rs.org/gtk4-rs/stable/latest/docs/?search=gtk_widget_get_css_name and I still think we should land this one btw. The fact we can't get a doc alias for types using `glib::wrapper!` is something that might get fixed in the future but shouldn't block what we can improve already.

sdroege:
Sure, why not. Doesn't seem too useful for only enums/flags but it's something :)

Issue #1140, opened 2021-05-16
docs: Links are shown with crate:: prefix


sophie-h:
While the `crate::` link works fine, it should be hidden in the output. Something like

```
[`Widget`](struct@gio::Widget)
```

That's also important because it would fix confusing output like `crate::glib` in gtk.

sophie-h:
cc @MarijnS95

MarijnS95:
#1128

sophie-h:
I see. That sounds kind of hopeless.

Example for a lot of noise

 https://gtk-rs.org/gtk4-rs/git/docs/gtk4/struct.ApplicationWindow.html#implements

Example for `crate::gio`

https://gtk-rs.org/gtk4-rs/git/docs/gtk4/prelude/trait.GtkWindowExt.html#tymethod.set_application

MarijnS95:
It's a tricky one. I mentioned on IRC I was already coming back from this decision because `crate::` really does look ugly. See for example the GStreamer struct/enum overview page: https://marijns95.pages.freedesktop.org/gstreamer-rs/gstreamer/#structs. OTOH this is the only way (that I know of) to make intradoc links not clickable if they don't point to anything sensible. `[]()` only detects the intradoc pattern, but leaves it to browsers to figure out if it wasn't pointing to a Rust object :/

Maybe `rustdoc` needs to solve that one way or another, or `gir` needs significant effort to check whether a link would be valid given the info it was.

bilelmoussaoui:
one counter argument against using `crate::` everywhere https://gtk-rs.org/gtk4-rs/stable/0.1/docs/gdk4/struct.Rectangle.html#returns

MarijnS95:
@bilelmoussaoui But why is https://gtk-rs.org/gtk4-rs/stable/0.1/docs/gdk4/struct.Rectangle.html#returns-1 not affected?

MarijnS95:
Relevant change for prelude constants: 52a79988204b4a39e10b9ec22729e7bc9904fcef

MarijnS95:
The GTK doc is probably using `#true` here instead of `%true`?

MarijnS95:
@bilelmoussaoui If we were to "fix" this to hide `crate::` again the broken link would still be turned into ``[`true`](crate::true)`` unless the source doc is fixed to use `%`.

bilelmoussaoui:
I wonder whether we should go with Self for a type's methods/constructors instead of the type's name?

[...]

PR #1144, opened 2021-05-19
docs: generate docs for builder properties

### LINKED ISSUE #1095: Add property docs to *Builder methods
The functions of Builders currently don't have any documentation. If that's possible, it would be nice to a the upstream property documentation as documentation for the function. They seemed to be quite suitable for that.

Sometimes the setter method has important additional information. Maybe it would be extra nice to link to the setter docs? (At least the new GTK 4 docs know about the setter of a property.)

/cc @bilelmoussaoui

bilelmoussaoui:
Fixes #1095

example
image

### DIFF
--- a/src/codegen/doc/mod.rs
+++ b/src/codegen/doc/mod.rs
@@ -168,6 +168,7 @@ fn create_object_doc(w: &mut dyn Write, env: &Env, info: &analysis::object::Info
     let signals: &[Signal];
     let properties: &[Property];
     let is_abstract;
+    let has_builder;
 
     let obj = env
         .config
@@ -183,6 +184,7 @@ fn create_object_doc(w: &mut dyn Write, env: &Env, info: &analysis::object::Info
             signals = &cl.signals;
             properties = &cl.properties;
             is_abstract = env.library.type_(info.type_id).is_abstract();
+            has_builder = obj.generate_builder;
         }
         Type::Interface(iface) => {
             doc = iface.doc.as_ref();
@@ -191,6 +193,7 @@ fn create_object_doc(w: &mut dyn Write, env: &Env, info: &analysis::object::Info
             signals = &iface.signals;
             properties = &iface.properties;
             is_abstract = false;
+            has_builder = false;
         }
         _ => unreachable!(),
     }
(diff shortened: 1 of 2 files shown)


src/codegen/doc/mod.rs
```diff
@@ -234,6 +237,59 @@ fn create_object_doc(w: &mut dyn Write, env: &Env, info: &analysis::object::Info
...
+        let builder_ty = TypeStruct::new(SType::Impl, &format!("{}Builder", info.name));
+        let parent_name = builder_ty.name.clone();
+
+        let mut builder_properties: Vec<_> = properties.iter().collect();
+        for parent_info in &info.supertypes {
+            match env.library.type_(parent_info.type_id) {
+                Type::Class(cl) => {
+                    cl.properties
+                        .iter()
+                        .for_each(|p| builder_properties.push(p));
```
MarijnS95, line 250:
```suggestion
                    builder_properties.extend(cl.properties.iter());
```
Would that work?

src/codegen/doc/mod.rs
```diff
@@ -234,6 +237,59 @@ fn create_object_doc(w: &mut dyn Write, env: &Env, info: &analysis::object::Info
...
+                        "{}",
+                        reformat_doc(&fix_param_names(doc, &None), &symbols, &parent_name)
+                    )?;
+                }
+                if let Some(ver) = property.deprecated_version {
+                    writeln!(w, "\n# Deprecated since {}\n", ver)?;
+                } else if property.doc_deprecated.is_some() {
+                    writeln!(w, "\n# Deprecated\n")?;
+                }
+                if let Some(ref doc) = property.doc_deprecated {
```
MarijnS95, line 281: I'm hoping to remove these soon in favour of `#[deprecated = "msg"]` which, by the way, is not used here or is it emitted elsewhere?

bilelmoussaoui: I should try it on gtk3 which has more deprecated items

src/codegen/doc/mod.rs
bilelmoussaoui: That indeed works, thanks!

src/codegen/doc/mod.rs
bilelmoussaoui: The builder pattern doesn't seem to generate a deprecated guard for the generated functions, it doesn't generate anything at all other than the version guard for properties so I guess let's keep this one here until we move to using guard for deprecated elsewhere

MarijnS95: I'm not sure how to remove them anyway. `#[deprecated]` is generated (at least supposed to) during normal generation and not part of the documentation. Since it can't include the reasoning (LGPL, and that information is simply not available there) only a `#[deprecated = "Since XXX"]` attribute is created.

Besides, putting the full deprecation text in there isn't going to fit nor look pretty.

If you have some time feel free to look into the missing deprecation warning for builders and at least get rid of the `Deprecated`/`Deprecated since XXX` message here (leaving `doc_deprecated` in place). Otherwise I'll enqueue it on my list.

bilelmoussaoui:
With the latest changes, all the methods in Builder patterns are properly documented. I wonder if we should add a doc comment to the generated builder as well. Something like `A builder for generating "{name}"`.`

src/codegen/doc/mod.rs
MarijnS95: Found it, its handling is missing from `fn generate_builder` in `codegen/object.rs`. Will PR later.

MarijnS95: We really need to come up with an `fn give_me_all_the_attributes_at_once` :sweat_smile:

bilelmoussaoui: So what I should do here, keep those as is so that you don't forget adding them in your PR? or drop them for now?

bilelmoussaoui: > We really need to come up with an `fn give_me_all_the_attributes_at_once` sweat_smile

What would such fn do? :)

bilelmoussaoui:
Updated version

image

@MarijnS95 can I drop the crate prefix if i'm sure the item is on the same crate?

MarijnS95:
> @MarijnS95 can I drop the crate prefix if i'm sure the item is on the same crate?

Yes please. We still want to fix this across the entire docs, #1140.

bilelmoussaoui:
Thanks, works as expected. Here's a screenshot of the final result:
image

bilelmoussaoui:
> It looks like all the other monospace types aren't properly linked though :/

of course they are not, they are using Gdk/Gtk as a prefix... An issue for another day :-)

MarijnS95:
> of course they are not, they are using Gdk/Gtk as a prefix... An issue for another day :-)

Assuming these are written as `#GdkPaintable` etc in the docs the regex converter should have taken care of that? It put it in backticks already, or is that the C docs using monospace instead of a type reference?

src/codegen/doc/mod.rs
MarijnS95: The removal is taken care of in https://github.com/gtk-rs/gir/pull/1148, and the missing Rust annotations were added in https://github.com/gtk-rs/gir/pull/1147. You can probably remove these now.

bilelmoussaoui: Thanks, done :-)

sdroege (approved):

PR #1145, opened 2021-05-19
docs: ignore not useful function parameters

### LINKED ISSUE #1115: Omit user_data & user_destroy from docs of functions parameters
They are generated for functions that takes a callback, they can be omitted as they are not really useful

image

cc @MarijnS95

bilelmoussaoui:
Fixes #1115 

A one last type of parameters that's not really useful, is the `n_X` followed after parameter `X` that's an array used to specify the length of the array. Not sure how to check for that one properly. I suppose looking for `n_X` param, if X is found and is of type of Array & `n_X` is of type `gsize` we can omit that one as well?

### DIFF
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -89,6 +89,7 @@ pub struct Info {
     pub cfg_condition: Option<String>,
     pub assertion: SafetyAssertionMode,
     pub doc_hidden: bool,
+    pub doc_ignore_parameters: HashSet<String>,
     pub r#async: bool,
     pub unsafe_: bool,
     pub trampoline: Option<AsyncTrampoline>,
@@ -585,6 +586,11 @@ fn analyze_function(
         .iter()
         .find_map(|f| f.cfg_condition.clone());
     let doc_hidden = configured_functions.iter().any(|f| f.doc_hidden);
+    let doc_ignore_parameters = configured_functions
+        .iter()
+        .find(|f| !f.doc_ignore_parameters.is_empty())
+        .map(|f| f.doc_ignore_parameters.clone())
+        .unwrap_or_default();
     let disable_length_detect = configured_functions.iter().any(|f| f.disable_length_detect);
     let no_future = configured_functions.iter().any(|f| f.no_future);
     let unsafe_ = configured_functions.iter().any(|f| f.unsafe_);
@@ -842,6 +848,7 @@ fn analyze_function(
         cfg_condition,
         assertion,
         doc_hidden,
+        doc_ignore_parameters,
         r#async,
         unsafe_,
         trampoline,
--- a/src/codegen/doc/mod.rs
+++ b/src/codegen/doc/mod.rs
@@ -15,7 +15,7 @@ use once_cell::sync::Lazy;
 use regex::{Captures, Regex};
 use std::{
     borrow::Cow,
-    collections::BTreeSet,
+    collections::{BTreeSet, HashSet},
     io::{Result, Write},
 };
 use stripper_lib::{write_file_name, write_item_doc, Type as SType, TypeStruct};
@@ -296,7 +296,20 @@ fn create_object_doc(w: &mut dyn Write, env: &Env, info: &analysis::object::Info
                 .iter()
                 .find(|f| &f.glib_name == c_identifier)
                 .and_then(|analysed_f| analysed_f.new_name.clone());
-            create_fn_doc(w, env, function, Some(Box::new(ty)), fn_new_name)?;
+            let doc_ignored_parameters = info
+                .functions
+                .iter()
+                .find(|f| &f.glib_name == c_identifier)
+                .map(|analyzed_f| analyzed_f.doc_ignore_parameters.clone())
+                .unwrap_or_default();
+            create_fn_doc(
+                w,
+                env,
+                function,
+                Some(Box::new(ty)),
+                fn_new_name,
+                doc_ignored_parameters,
+            )?;
         }
     }
     for signal in signals {
@@ -313,7 +326,7 @@ fn create_object_doc(w: &mut dyn Write, env: &Env, info: &analysis::object::Info
         } else {
             ty.clone()
         };
-        create_fn_doc(w, env, signal, Some(Box::new(ty)), None)?;
+        create_fn_doc(w, env, signal, Some(Box::new(ty)), None, HashSet::new())?;
     }
     for property in properties {
         let ty = if has_trait {
@@ -362,7 +375,14 @@ fn create_record_doc(w: &mut dyn Write, env: &Env, info: &analysis::record::Info
         ..ty
     };
     for function in &record.functions {
-        create_fn_doc(w, env, function, Some(Box::new(ty.clone())), None)?;
+        create_fn_doc(
+            w,
+            env,
+            function,
+            Some(Box::new(ty.clone())),
+            None,
+            HashSet::new(),
+        )?;
     }
     Ok(())
 }
@@ -464,6 +484,7 @@ fn create_fn_doc<T>(
     fn_: &T,
     parent: Option<Box<TypeStruct>>,
     name_override: Option<String>,
+    doc_ignored_parameters: HashSet<String>,
 ) -> Result<()>
 where
     T: FunctionLikeType + ToStripperType,
@@ -537,6 +558,7 @@ where
 
         let in_parameters = no_array_length_params.iter().filter(|param| {
             let ignore = IGNORED_C_FN_PARAMS.contains(&param.name.as_str())
+                || doc_ignored_parameters.contains(&param.name)
                 || param.direction == ParameterDirection::Out
                 // special case error pointer as it's transformed to a Result
                 || (param.name == "error" && param.c_type == "GError**")
@@ -563,6 +585,7 @@ where
             .iter()
             .filter(|param| {
                 param.direction == ParameterDirection::Out
+                    && !doc_ignored_parameters.contains(&param.name)
                     && !(param.name == "error" && param.c_type == "GError**")
             })
             .collect();
--- a/src/codegen/doc/mod.rs
+++ b/src/codegen/doc/mod.rs
@@ -15,11 +15,22 @@ use once_cell::sync::Lazy;
 use regex::{Captures, Regex};
 use std::{
     borrow::Cow,
+    collections::BTreeSet,
     io::{Result, Write},
 };
 use stripper_lib::{write_file_name, write_item_doc, Type as SType, TypeStruct};
 
 mod format;
+
+// A list of C parameters that are not used directly by the Rust bindings
+const IGNORED_C_FN_PARAMS: [&str; 6] = [
+    "user_data",
+    "user_destroy",
+    "destroy_func",
+    "dnotify",
+    "destroy",
+    "user_data_free_func",
+];
 
 trait ToStripperType {
     fn to_stripper_type(&self) -> TypeStruct;
@@ -503,8 +514,39 @@ where
             )?;
         }
 
-        for parameter in fn_.parameters() {
-            if parameter.instance_parameter || parameter.name.is_empty() {
+        // A list of parameter positions to filter out
+        let mut indices_to_ignore: BTreeSet<_> = fn_
+            .parameters()
+            .iter()
+            .filter_map(|param| param.array_length)
+            .collect();
+        if let Some(indice) = fn_.ret().array_length {
+            indices_to_ignore.insert(indice);
+        }
+
+        // The original list of parameters without the ones that specify an array length
+        let no_array_length_params: Vec<_> = fn_
+            .parameters()
+            .iter()
+            .enumerate()
+            .filter_map(|(indice, param)| {
+                (!indices_to_ignore.contains(&(indice as u32))).then(|| param)
+            })
+            .filter(|param| !param.instance_parameter)
+            .collect();
+
+        let in_parameters = no_array_length_params.iter().filter(|param| {
+            let ignore = IGNORED_C_FN_PARAMS.contains(&param.name.as_str())
+                || param.direction == ParameterDirection::Out
+                // special case error pointer as it's transformed to a Result
+                || (param.name == "error" && param.c_type == "GError**")
+                // special case `data` with explicit `gpointer` type as it could be something else (unlike `user_data`)
+                || (param.name == "data" && param.c_type == "gpointer");
+            !ignore
+        });
+
+        for parameter in in_parameters {
+            if parameter.name.is_empty() {
                 continue;
             }
             if let Some(ref doc) = parameter.doc {
@@ -517,13 +559,39 @@ where
             }
         }
 
-        if let Some(ref doc) = fn_.ret().doc {
+        let out_parameters: Vec<_> = no_array_length_params
+            .iter()
+            .filter(|param| {
+                param.direction == ParameterDirection::Out
+                    && !(param.name == "error" && param.c_type == "GError**")
+            })
+            .collect();
+
+        if fn_.ret().doc.is_some() || !out_parameters.is_empty() {
             writeln!(w, "\n# Returns\n")?;
+        }
+        // document function's return
+        if let Some(ref doc) = fn_.ret().doc {
             writeln!(
                 w,
                 "{}",
                 reformat_doc(&fix_param_names(doc, &self_name), &symbols, &parent_name)
             )?;
+        }
+        // document OUT parameters as part of the function's Return
+        for parameter in out_parameters {
+            if let Some(ref doc) = parameter.doc {
+                writeln!(
+                    w,
+                    "\n## `{}`",
+                    nameutil::mangle_keywords(&parameter.name[..])
+                )?;
+                writeln!(
+                    w,
+                    "{}",
+                    reformat_doc(&fix_param_names(doc, &self_name), &symbols, &parent_name),
+                )?;
+            }
         }
         Ok(())
     })
--- a/src/config/functions.rs
+++ b/src/config/functions.rs
@@ -12,7 +12,7 @@ use crate::{
     version::Version,
 };
 use log::error;
-use std::str::FromStr;
+use std::{collections::HashSet, str::FromStr};
 use toml::Value;
 
 #[derive(Clone, Debug)]
@@ -230,6 +230,7 @@ pub struct Function {
     pub parameters: Parameters,
     pub ret: Return,
     pub doc_hidden: bool,
+    pub doc_ignore_parameters: HashSet<String>,
     pub is_windows_utf8: bool,
     pub disable_length_detect: bool,
     pub doc_trait_name: Option<String>,
@@ -263,6 +264,7 @@ impl Parse for Function {
                 "return",
                 "name",
                 "doc_hidden",
+                "doc_ignore_parameters",
                 "is_windows_utf8",
                 "disable_length_detect",
                 "pattern",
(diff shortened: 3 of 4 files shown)


src/codegen/doc/mod.rs
```diff
@@ -503,7 +513,30 @@ where
...
+                    let array_param_name = &param.name[2..];
+                    &param.name[0..2] == "n_"
+                        && fn_
+                            .parameters()
+                            .iter()
+                            .find(|p| p.name == array_param_name)
+                            .is_some()
+                } else {
+                    false
+                };
```
bilelmoussaoui, line 530: this seems to work pretty fine at least on gtk4-rs, there are three types where that convention fails: ListStore, TreeModelFilter & TreeStore. Imho, i would say it's worth it (of course, in such case the unnecessary params are kept)  and the correct solution here would be to rename the params upstream so that we can use such convention without worrying too much.

src/codegen/doc/mod.rs
```diff
@@ -503,7 +513,30 @@ where
...
+
+                !IGNORED_C_FN_PARAMS.contains(&param.name.as_str())
+                    && param.direction != ParameterDirection::Out
+                    && !(param.name == "error" && param.c_type != "GError**")
```
bilelmoussaoui, line 534: This one is something i'm not super sure about :)

bilelmoussaoui:
cc @MarijnS95 , would be nice to check the changes of this on gstreamer & see if nothing useful gets omitted

src/codegen/doc/mod.rs
```diff
@@ -503,7 +513,28 @@ where
...
+            .parameters()
+            .iter()
+            .filter(|param| {
+                let is_array_length_param = if param.name.len() > 3 {
+                    let array_param_name = &param.name[2..];
+                    &param.name[0..2] == "n_"
+                        && fn_.parameters().iter().any(|p| p.name == array_param_name)
+                } else {
+                    false
+                };
```
MarijnS95, line 526:
```suggestion
                let is_array_length_param = if let Some(array_param_name) = param.name.strip_prefix("n_) {                    
                    fn_.parameters().iter().any(|p| p.name == array_param_name)
                } else {
                    false
                };
```

src/codegen/doc/mod.rs
```diff
@@ -503,7 +513,28 @@ where
...
+
+                !IGNORED_C_FN_PARAMS.contains(&param.name.as_str())
```
MarijnS95, line 528: I feel like this is easier to read if you apply De Morgan's law and turn this into a `!(... || ...)`, perhaps even a:


```rust
let ignore = IGNORED_C_FN_PARAMS.contains(&param.name.as_str())
    || param.direction != ParameterDirection::Out
    || ...;
!ignore
```

src/codegen/doc/mod.rs
```diff
@@ -503,7 +513,28 @@ where
...
+                    && param.direction != ParameterDirection::Out
+                    && !(param.name == "error" && param.c_type != "GError**")
+                    && !is_array_length_param
+                    // special case `data` as it could be something else other than `user_data`
+                    && !(param.name == "data" && param.c_type == "gpointer")
+            })
+            .collect();
```
MarijnS95, line 535: No need to collect into a vector if you're iterating these below anyway?

src/codegen/doc/mod.rs
MarijnS95: Shouldn't this be:

```suggestion
                    && !(param.name == "error" && param.c_type == "GError**")
```

MarijnS95 (changes requested):

MarijnS95:
> cc @MarijnS95 , would be nice to check the changes of this on gstreamer & see if nothing useful gets omitted

I think @sdroege could better answer that question, it's not up to me :)

src/codegen/doc/mod.rs
```diff
@@ -503,7 +513,28 @@ where
...
+                let is_array_length_param = if param.name.len() > 3 {
+                    let array_param_name = &param.name[2..];
+                    &param.name[0..2] == "n_"
+                        && fn_.parameters().iter().any(|p| p.name == array_param_name)
+                } else {
+                    false
+                };
+
+                !IGNORED_C_FN_PARAMS.contains(&param.name.as_str())
+                    && param.direction != ParameterDirection::Out
```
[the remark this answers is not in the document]
bilelmoussaoui: That would make sense, I will give it a try and see how that looks like.

src/codegen/doc/mod.rs
bilelmoussaoui: Ah that looks much better. TIL a bout strip_prefix. Thanks!

src/codegen/doc/mod.rs
```diff
@@ -524,6 +551,20 @@ where
                 "{}",
                 reformat_doc(&fix_param_names(doc, &self_name), &symbols, &parent_name)
             )?;
+            // fold out parameters as part of returns
+            let out_parameters = fn_.parameters().iter().filter(|param| {
+                !(param.direction == ParameterDirection::Out
```
MarijnS95, line 556: I don't think this one should be inverted? Now you're getting all but `ParameterDirection::Out` :)

src/codegen/doc/mod.rs
```diff
@@ -524,6 +551,20 @@ where
...
+                    || (param.name == "error" && param.c_type == "GError**"))
```
MarijnS95, line 557:
```suggestion
                param.direction == ParameterDirection::Out
                    && !(param.name == "error" && param.c_type == "GError**")
```

I guess?

src/codegen/doc/mod.rs
bilelmoussaoui: yeah, already fixed. I only pushed to be able to easily test on gtk4-rs

bilelmoussaoui: Oh, it should be properly working now. But one very bad thing is that sometimes a functions returns true or false and we convert that to an option which makes the whole thing not very nice I guess. 

See for example this function that returns ` Option<(i32, i32)>`
```rust
    /// [`true`] if `cell` belongs to `self`.
    /// return location for the horizontal
    ///  position of `cell` within `self`, may be [`None`]
    /// return location for the width of `cell`,
    ///  may be [`None`]
```

bilelmoussaoui: In general out/inout params are just bad for introspection :/

bilelmoussaoui: Note: if we end up keeping this, we should also handle `n_` cases as they are used as out params sometimes as well.

bilelmoussaoui: Okay never mind, i actually made it slightly better 
image

Of course there are some cases that can't be catched and ignored like this one 
image

but in general, it puts the information where it belongs, in the return instead of parameters.

bilelmoussaoui: Oh never mind, I just learned about the `length` annotation... It should be fixable.

bilelmoussaoui: image
I have pushed the fixed version

bilelmoussaoui:
Ok after a bunch of tests, I'm very happy with the results of this one.

src/codegen/doc/mod.rs
```diff
@@ -21,6 +21,16 @@ use stripper_lib::{write_file_name, write_item_doc, Type as SType, TypeStruct};
...
 mod format;
 
+// A list of C parameters that are not used directly by the Rust bindings
+const IGNORED_C_FN_PARAMS: [&str; 6] = [
+    "user_data",
+    "user_destroy",
+    "destroy_func",
+    "dnotify",
+    "destroy",
+    "user_data_free_func",
```
sdroege, line 31: You should probably also check the types or not?

sdroege: And maybe add a configuration (per function etc) to override this?

src/codegen/doc/mod.rs
```diff
@@ -503,8 +513,42 @@ where
...
+        let no_array_length_params: Vec<_> = fn_
+            .parameters()
+            .iter()
+            .enumerate()
+            .filter_map(|(indice, param)| {
+                if indices_to_ignore.contains(&(indice as u32)) {
+                    None
+                } else {
+                    Some(param)
+                }
```
MarijnS95, line 536: I've been shouted at for suggesting `.then` in another repo, but maybe it works here? `(!indices_to_ignore.contains(&(indice as u32))).then(|| param)`.

Alternatively you might as well separate these out over a `.filter(|(index, _)| ...).map(|(_, param)| param)`.

src/codegen/doc/mod.rs
```diff
@@ -517,13 +561,34 @@ where
...
-        if let Some(ref doc) = fn_.ret().doc {
+        let out_parameters: Vec<_> = no_array_length_params
+            .iter()
+            .filter(|param| {
+                param.direction == ParameterDirection::Out
+                    && !(param.name == "error" && param.c_type == "GError**")
+            })
+            .collect();
+
+        if fn_.ret().doc.is_some() || !out_parameters.is_empty() {
```
MarijnS95, line 572: Yeah I was wondering about this one already, now the out parameters neatly show up even if there is no return doc :)

src/codegen/doc/mod.rs
```diff
@@ -517,13 +561,34 @@ where
...
+                writeln!(
+                    w,
+                    "{}",
+                    reformat_doc(&fix_param_names(doc, &self_name), &symbols, &parent_name)
+                )?;
+            }
+            // fold out parameters as part of returns
+            for parameter in out_parameters {
+                if let Some(ref doc) = parameter.doc {
+                    writeln!(w, "## `{}`", nameutil::mangle_keywords(&parameter.name[..]))?;
```
MarijnS95, line 584: Still a bit wary of this name, it doesn't show up in the tuple making it slightly confusing where this is coming from? Adding the index like `` ## {i}: `{name}` `` probably isn't exactly pretty either.

src/codegen/doc/mod.rs
```diff
@@ -517,13 +561,34 @@ where
...
-            )?;
+            if let Some(ref doc) = fn_.ret().doc {
+                writeln!(
+                    w,
+                    "{}",
+                    reformat_doc(&fix_param_names(doc, &self_name), &symbols, &parent_name)
+                )?;
+            }
+            // fold out parameters as part of returns
+            for parameter in out_parameters {
```
MarijnS95, line 582: Should there be a newline between return doc and return parameters just like there are two empty lines around `# Returns`? Doesn't make a difference for the rendered docs, but makes the markdown source easier to read.

src/codegen/doc/mod.rs
```diff
@@ -503,8 +513,42 @@ where
             )?;
         }
 
-        for parameter in fn_.parameters() {
-            if parameter.instance_parameter || parameter.name.is_empty() {
+        // A list of parameters positions to filter out
+        let mut indices_to_ignore: Vec<_> = fn_
```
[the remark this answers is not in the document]
MarijnS95: Whoopsie, I meant `BTreeSet` to contemplate `HashSet` :)

src/codegen/doc/mod.rs
bilelmoussaoui: Not sure what to override exactly ?

src/codegen/doc/mod.rs
bilelmoussaoui: I don't think the indices adds any values nor it makes it clearer. The situation can somehow be imported but would require a bunch of special casing (whether it's an Option, Result.....) which is something not in the scope of this PR.

src/codegen/doc/mod.rs
bilelmoussaoui: Why there would be? they are part of the returns

MarijnS95: Nicer to read. Currently:

```rust
    /// # Returns
    ///
    /// TRUE if the conversion was successful.
    /// ## `dest_val`
    /// pointer to destination value
```
And I'd like:

```rust
    /// # Returns
    ///
    /// TRUE if the conversion was successful.
    ///
    /// ## `dest_val`
    /// pointer to destination value
```
But again, this makes no difference to **rendered markdown output**, which is all we care about. Feel free to ignore this and mark it as resolved.

MarijnS95: Heck, we might even remove the `\n` from `# Returns` to save quite a few characters/lines in this intermediate textual form.

bilelmoussaoui: Ah I see what you mean. Yeah, that would make the docs output readable

src/codegen/doc/mod.rs
MarijnS95: It's not, we should probably leave this and let it to the user to figure out how to interpret the docs. Indeed, not the scope of this PR and probably not something that's really bothersome or fixable.

src/codegen/doc/mod.rs
bilelmoussaoui: In general, this list should only contain parameters that are known per conventions to be not useful at all for the rust bindings (form a user perspective) if something on the list proves to cause issues, it should be special cases like `data` a bit below where it can mean `user_data` or something else.

src/codegen/doc/mod.rs
bilelmoussaoui: I went with then, the filter & then map isn't great imho :)

MarijnS95: Yeah, there's no `.enumerate_filter()` or something like that :/

MarijnS95: OTOH you already have a `filter` over param... :)

bilelmoussaoui: > OTOH you already have a `filter` over param... :)

I want to keep both separate for understandability otherwise the code becomes cryptic

src/codegen/doc/mod.rs
```diff
@@ -503,8 +514,39 @@ where
...
+        let mut indices_to_ignore: BTreeSet<_> = fn_
+            .parameters()
+            .iter()
+            .filter_map(|param| param.array_length)
+            .collect();
+        if let Some(indice) = fn_.ret().array_length {
+            indices_to_ignore.insert(indice);
+        }
+
+        // The original list of parameters without the ones that specifics an array length
```
MarijnS95, line 527:
```suggestion
        // The original list of parameters without the ones that specify an array length
```

src/codegen/doc/mod.rs
```diff
@@ -503,8 +514,39 @@ where
             )?;
         }
 
-        for parameter in fn_.parameters() {
-            if parameter.instance_parameter || parameter.name.is_empty() {
+        // A list of parameters positions to filter out
```
MarijnS95, line 517:
```suggestion
        // A list of parameter positions to filter out
```

src/codegen/doc/mod.rs
```diff
@@ -503,8 +514,39 @@ where
...
+            })
+            .filter(|param| !param.instance_parameter)
+            .collect();
+
+        let in_parameters = no_array_length_params.iter().filter(|param| {
+            let ignore = IGNORED_C_FN_PARAMS.contains(&param.name.as_str())
+                || param.direction == ParameterDirection::Out
+                // special case error pointer as it's transformed to a Result
+                || (param.name == "error" && param.c_type == "GError**")
+                // special case `data` as it could be something else other than `user_data`
```
MarijnS95, line 543:
```suggestion
                // special case `data` with explicit `gpointer` type as it could be something else (unlike `user_data`)
```

I think is more fitting for what you're trying to convey here?

src/codegen/doc/mod.rs
```diff
@@ -517,13 +559,38 @@ where
...
+        let out_parameters: Vec<_> = no_array_length_params
+            .iter()
+            .filter(|param| {
+                param.direction == ParameterDirection::Out
+                    && !(param.name == "error" && param.c_type == "GError**")
+            })
+            .collect();
+
+        if fn_.ret().doc.is_some() || !out_parameters.is_empty() {
             writeln!(w, "\n# Returns\n")?;
```
MarijnS95, line 571: You probably don't need to fold the `if let Some(ref doc) = fn_.ret().doc` and `for parameter in out_parameters` inside this block?

I'm not sure what's easier to read...

src/codegen/doc/mod.rs
bilelmoussaoui: yes, what i wrote was more of frenglish :P

src/codegen/doc/mod.rs
bilelmoussaoui: They are both "return" results... not sure really how to make it any better without over complex-ifying  this PR

src/codegen/doc/mod.rs
sdroege: Hiding other parameters from the documentation, for example length parameters of arrays or destroy notifies with non-default names.

src/codegen/doc/mod.rs
MarijnS95: Taking them outside of the `if fn_.ret().doc.is_some() || !out_parameters.is_empty() {` block as that means the same, and saves a bit of indentation. It doesn't change anything functionally and might, OTOH, make it less readable too...

bilelmoussaoui: Ah I see what you mean

bilelmoussaoui: I somehow thought that wouldn't work but now that I think of :P needs more coffee! Thanks for all the great suggestions :-)

src/codegen/doc/mod.rs
bilelmoussaoui: Ah I see, the array length parameters are already hidden based on the introspection data (the `length` annotation) so if there's some case, it would need an annotation fix instead :)

src/codegen/doc/mod.rs
MarijnS95: If you're doing this, do the same for `if let Some(ref doc) = fn_.ret().doc {`.

MarijnS95:
```rust
        if fn_.ret().doc.is_some() || !out_parameters.is_empty() {
            writeln!(w, "\n# Returns\n")?;
        }
```
Is what you'll have in the end ;)

src/codegen/doc/mod.rs
bilelmoussaoui: I will have a look later today in order to add a config for :)

bilelmoussaoui: Actually, wasn't that hard so I implemented it already :-)

src/analysis/functions.rs
```diff
@@ -585,6 +586,11 @@ fn analyze_function(
         .iter()
         .find_map(|f| f.cfg_condition.clone());
     let doc_hidden = configured_functions.iter().any(|f| f.doc_hidden);
+    let doc_ignore_parameters = configured_functions
+        .iter()
+        .map(|f| f.doc_ignore_parameters.clone())
+        .next()
```
MarijnS95, line 592: Isn't that `.find_map` (inb4 clippy)?

src/config/functions.rs
```diff
@@ -308,6 +310,14 @@ impl Parse for Function {
...
             .and_then(Value::as_bool)
             .unwrap_or(false);
+        let doc_ignore_parameters = toml
+            .lookup_vec("doc_ignore_parameters", "Invalid doc_ignore_parameters")
+            .map(|v| {
+                v.iter()
+                    .filter_map(|v| v.as_str().map(String::from))
+                    .collect()
+            })
+            .unwrap_or_else(|_| Vec::new());
```
MarijnS95, line 320: `unwrap_or_default`?

src/codegen/doc/mod.rs
```diff
@@ -464,6 +484,7 @@ fn create_fn_doc<T>(
     fn_: &T,
     parent: Option<Box<TypeStruct>>,
     name_override: Option<String>,
+    doc_ignored_parameters: Vec<String>,
```
MarijnS95, line 487: Should be a `HashSet` for quicker lookup, if you only ever use `.contains`?

src/codegen/doc/mod.rs
bilelmoussaoui: This was actually a good idea as I found two use cases for it on gtk4-rs where array length annotation isn't properly picked for TreeStore/ListStore

sdroege (approved):

PR #1149, opened 2021-05-22
Use `S: AsRef<str>`, `&[S]` instead of `&[&str]`

rafaelcaricio:
This PR complements the changes in https://github.com/gtk-rs/gtk-rs-core/pull/89 applying the new code pattern in the auto generated code.


MarijnS95:
Would it be possible and shorter to use `&[impl AsRef<str>]` instead?

PR #1150, opened 2021-05-22
Improve display for intra-doc links

### LINKED ISSUE #1140: docs: Links are shown with crate:: prefix

GuillaumeGomez:
Fixes #1140.

### DIFF
--- a/src/codegen/doc/format.rs
+++ b/src/codegen/doc/format.rs
@@ -100,7 +100,7 @@ fn replace_c_types(entry: &str, symbols: &symbols::Info, in_type: &str) -> Strin
             if sym.owner_name() == Some(in_type) {
                 // `#` or `%` symbols should probably have been `@` to denote
                 // that it is a reference within the current type.
-                format!("[`Self::{}()`]", sym.name())
+                format!("[`{0}()`][Self::{0}]", sym.name())
             } else {
                 match caps.get(1).as_ref().map(Match::as_str) {
                     // Catch invalid @ references that have a C symbol available but do not belong
@@ -108,12 +108,12 @@ fn replace_c_types(entry: &str, symbols: &symbols::Info, in_type: &str) -> Strin
                     // but with a valid global link so that the can be easily spotted in the code.
                     // assert_eq!(sym.owner_name(), Some(in_type));
                     Some("@") => format!(
-                        "[`crate::{}()`] (XXX: @-reference does not belong to {}!)",
+                        "[`{0}()`][crate::{0}] (XXX: @-reference does not belong to {1}!)",
                         sym.full_rust_name(),
                         in_type,
                     ),
                     Some("#") | None => {
-                        format!("[`crate::{}()`]", sym.full_rust_name())
+                        format!("[`{0}()`][crate::{0}]", sym.full_rust_name())
                     }
                     Some("%") => panic!("% not allowed for {:?}", caps),
                     Some(c) => panic!("Unknown symbol reference {}", c),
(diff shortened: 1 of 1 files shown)


src/codegen/doc/format.rs
```diff
@@ -100,20 +100,20 @@ fn replace_c_types(entry: &str, symbols: &symbols::Info, in_type: &str) -> Strin
...
+                format!("[`{0}()`][Self::{0}]", sym.name())
             } else {
                 match caps.get(1).as_ref().map(Match::as_str) {
                     // Catch invalid @ references that have a C symbol available but do not belong
                     // to the current type (and can hence not use `Self::`). For now generate XXX
                     // but with a valid global link so that the can be easily spotted in the code.
                     // assert_eq!(sym.owner_name(), Some(in_type));
                     Some("@") => format!(
-                        "[`crate::{}()`] (XXX: @-reference does not belong to {}!)",
+                        "[`{0}()`][crate::{0}] (XXX: @-reference does not belong to {1}!)",
```
MarijnS95, line 111: I'd left these, they're ugly and broken anyway.

MarijnS95:
We were doing the same in parallel but I went for reverting #1128 which is more complete, you missed a bunch in trait/struct extension handling: https://github.com/MarijnS95/compare/new/shorter-links. And as requested, you missed a proper explanation in the commit, which I did provide :wink:

MarijnS95:
The intent of not using `[]()` was that these would otherwise generate dead links. ``[`not::in::scope`]`` would have rendered just the path in monospace, and `[]` surrounding it indicating that something is wrong here. With these changes however, broken paths are fully rendered:

```markdown
/// An element can be retrieved from a bin with [`prelude::GstBinExt::get_by_name()`][crate::prelude::GstBinExt::get_by_name()], using the
/// elements name. [`prelude::GstBinExt::get_by_name_recurse_up()`][crate::prelude::GstBinExt::get_by_name_recurse_up()] is mainly used for internal
/// purposes and will query the parent bins when the element is not found in the
/// current bin.
```

Gets turned into:

image

MarijnS95:
In other words, we're getting rid of `crate::` and `Self::` in working links, but substitute that for massive long duplicates in broken links.

As mentioned in my commit variant of this PR above, it's annoying to curate the C docs properly (we should probably not add symbols to the symbol list if they aren't actually generated in scope?).

GuillaumeGomez:
@MarijnS95 I think it'll be simpler to go with your branch. Closing this PR!

MarijnS95:
@GuillaumeGomez Fine by me :)

Not sure if we want to go ahead given that broken links are now rendered even worse than the usual ``[`crate::prelude::GstBinExt::get_by_name()`]`` we'd get now? Maybe we just have to accept that in favour of losing `crate::`/`Self::` prefixes though - if yes I'll submit that branch :)

GuillaumeGomez:
I don't mind having the prefixes, so maybe let's vote first before changing something again? ;)

MarijnS95:
I have mixed feelings. Broken links (instead of unclickable markdown) are nice, but so is hiding the prefixes because it really looks crowded on the index page:

image

I think most opinions have been captured in https://github.com/gtk-rs/gir/issues/1140 though.

bilelmoussaoui:
I suppose most of the issues are caused by the symbols list containing items that are not generated at all?

MarijnS95:
> I suppose most of the issues are caused by the symbols list containing items that are not generated at all?

Exactly. We don't generate clickable links for items that are not in the `symbols` list, so we should just make sure anything that doesn't have a generate/manual status isn't in there. But there are many more heuristics that determine whether something gets generated or not :/

PR #1152, opened 2021-05-23
codegen: use C value instead of hardcoding it

bilelmoussaoui:
See https://github.com/gtk-rs/gtk4-rs/pull/397 for the resulted changes

cc @GuillaumeGomez, i suppose that's why we had such code in gtk's ResponseType, basically based on what gir generates :)

### DIFF
--- a/src/codegen/enums.rs
+++ b/src/codegen/enums.rs
@@ -58,7 +58,6 @@ fn generate_enum(
     struct Member {
         name: String,
         c_name: String,
-        value: String,
         version: Option<Version>,
         deprecated_version: Option<Version>,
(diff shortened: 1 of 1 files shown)


MarijnS95 (approved):
Nice find! The FFI layer should be in control of the constants, not the safe wrapper layer :grimacing: 

Do change the title of the PR/commit to `codegen/enum: Use FFI constant instead of integer value` though; I had no idea what this PR was about until reading the code ;)

sdroege (approved):

GuillaumeGomez:
MUCH better!

PR #1153, opened 2021-05-23
Use `impl Trait` instead of generic type names with trait bound

MarijnS95:
For https://github.com/gtk-rs/gtk-rs-core/pull/89#issuecomment-846516935

---

`impl` is quite a bit more concise and readable when a type parameter with trait bound is only used once.  Aside that, this simplifies some of the type handling in trampolines by using `P: IsA<>` directly instead of through a `where P: ...` indirection.

---

Related MRs:
https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/778
https://github.com/gtk-rs/gtk-rs-core/pull/95
https://github.com/gtk-rs/gtk3-rs/pull/547
https://github.com/gtk-rs/gtk4-rs/pull/404

### DIFF
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -28,7 +28,15 @@ pub enum BoundType {
 
 impl BoundType {
     pub fn need_isa(&self) -> bool {
-        matches!(*self, BoundType::IsA(_))
+        matches!(*self, Self::IsA(_))
+    }
+
+    // TODO: This is just a heuristic for now, based on what we do in codegen!
+    // Theoretically the surrounding function should determine whether it needs to
+    // reuse an alias (ie. to use in `call_func::<P, Q, R>`) or not.
+    // In the latter case an `impl` is generated instead of a type name/alias.
+    pub fn has_alias(&self) -> bool {
+        matches!(*self, Self::NoWrapper)
     }
 }
 
@@ -36,7 +44,8 @@ impl BoundType {
 pub struct Bound {
     pub bound_type: BoundType,
     pub parameter_name: String,
-    pub alias: char,
+    /// Bound does not have an alias when `param: impl type_str` is used
+    pub alias: Option<char>,
     pub type_str: String,
     pub callback_modified: bool,
 }
@@ -217,7 +226,9 @@ impl Bounds {
         if self.used.iter().any(|n| n.parameter_name == name) {
             return;
         }
-        let alias = self.unused.pop_front().expect("No free type aliases!");
+        let alias = bound_type
+            .has_alias()
+            .then(|| self.unused.pop_front().expect("No free type aliases!"));
         self.used.push(Bound {
             bound_type,
             parameter_name: name.to_owned(),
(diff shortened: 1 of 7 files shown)


MarijnS95:
So this is far less trivial than anticipated, and not how I planned to spend my sunday!

There's still some trouble with trampolines now that I made aliases optional but  the `where TYPE_PARAMETERS_START: IsA` is hardcoded anyway (for traits) so that's what I'll replace the TODO with.

Other than that the current "heuristic" to detect whether to use `impl` is based on `NoWrapper`, which appears to be an explicit type-name/trait-bound. If it's not that we'll use `impl` instead, assuming the type-name isn't used elsewhere in the code. That should probably be enough.

I'll be out for a while, but feel more than welcome to go over the `TODO`s and give suggestions (keeping the above in mind, as I have plans for those `TODO`s already).

src/codegen/bound.rs
```diff
@@ -0,0 +1,56 @@
...
+    /// `nullable` only applies to [`BoundType::IsA`]
+    pub(crate) fn use_as_type(
+        &self,
+        ref_mode: RefMode,
+        nullable: Nullable,
+        r#async: bool,
+    ) -> String {
+        // TODO: Should we assert that Some alias implies NoWrapper, and the other way around?
+        // Alternatively, should the alias be encoded in the bound type, as a properly separated
+        // "mode" entirely where type names are _reused_ in the function to call other functions?
```
MarijnS95, line 28: Same question about the heuristic. I'm thinking to split this up to always return the alias when it exists, otherwise return `impl TheTraitBound`.

src/codegen/function.rs
```diff
@@ -283,47 +248,45 @@ pub fn bounds(
 
     let skip_lifetimes = bounds
         .iter()
-        .filter(|bound| skip.contains(&bound.alias))
+        // TODO: False or true?
+        .filter(|bound| bound.alias.map_or(false, |alias| skip.contains(&alias)))
```
MarijnS95, line 252: Note to self: Try and remember what was up here.

src/codegen/function.rs
```diff
@@ -283,47 +248,45 @@ pub fn bounds(
...
-        .collect();
-
-    if strs.is_empty() {
-        (String::new(), Vec::new())
+        .collect::<Vec<_>>();
+
+    let bounds = bounds.iter().filter(|bound| {
+        bound.alias.map_or(true, |alias| !skip.contains(&alias))
+            && (!filter_callback_modified || !bound.callback_modified)
+    });
```
MarijnS95, line 268: I need to revisit this: can't be a trait bound if it doesn't have an alias, then it's only written in-place with `impl`.

src/codegen/function.rs
```diff
@@ -283,47 +248,45 @@ pub fn bounds(
...
-        (format!("<{}>", strs.join(", ")), bounds)
-    }
+        format!("<{}>", type_names.join(", "))
+    };
+
+    let bounds = lifetimes
+        .into_iter()
+        // TODO: enforce that this is only used on NoWrapper!
+        // TODO: Analyze if alias is used in function, otherwise set to None!
+        .chain(bounds.filter_map(|b| b.alias).map(|a| a.to_string()))
```
MarijnS95, line 286: Ambiguity between `NoWrapper` and `impl` (no alias) again.

src/analysis/bounds.rs
```diff
@@ -28,17 +28,25 @@ pub enum BoundType {
...
 
 #[derive(Clone, Eq, Debug, PartialEq)]
 pub struct Bound {
     pub bound_type: BoundType,
     pub parameter_name: String,
-    pub alias: char,
+    /// Bound does not have an alias when `param: impl type_str` is used
+    pub alias: Option<char>,
     pub type_str: String,
-    pub info_for_next_type: bool,
```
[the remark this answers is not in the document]
bilelmoussaoui: From a quick search, not really

src/analysis/bounds.rs
```diff
@@ -153,14 +160,8 @@ impl Bounds {
...
-                if (!need_is_into_check || !*par.nullable)
-                    && par.c_type != "GDestroyNotify"
-                    && !self.add_parameter(&par.name, &type_string, bound_type, r#async)
-                {
-                    panic!(
-                        "Too many type constraints for {}",
-                        func.c_identifier.as_ref().unwrap()
-                    )
+                if (!need_is_into_check || !*par.nullable) && par.c_type != "GDestroyNotify" {
+                    self.add_parameter(&par.name, &type_string, bound_type, r#async)
```
rafaelcaricio, line 164: Good change, it took me some time to find this call when skimming through the code. Now it's more clear :)

src/analysis/bounds.rs
```diff
@@ -380,34 +341,60 @@ mod tests {
...
         bounds.add_parameter("b", "", typ.clone(), false);
-        assert_eq!(
-            bounds.get_parameter_alias_info("a"),
-            Some(('P', typ.clone()))
-        );
-        assert_eq!(bounds.get_parameter_alias_info("b"), Some(('Q', typ)));
-        assert_eq!(bounds.get_parameter_alias_info("c"), None);
+        let bound = bounds.get_parameter_bound("a").unwrap();
+        assert_eq!(bound.alias, None);
+        assert_eq!(bound.bound_type, typ);
```
rafaelcaricio, line 398: Maybe we could have the same test for `BoundType::AsRef<None>`?

src/analysis/bounds.rs
MarijnS95: Yup, this just panics now when running out of type parameters to use :tada:

src/analysis/bounds.rs
MarijnS95: I'm thinking of extending this with a few comments explaining why/when we do or do not expect a type alias.

In the end it shouldn't be based on `BoundType` as mentioned in a comment above, but I haven't invested the time to figure that one out and refactor it yet. Works well enough as a heuristic for now, but these tests should likely be adjusted in the future.

sdroege:
@GuillaumeGomez Should we get this in now then?

GuillaumeGomez (approved):

GuillaumeGomez:
This is good enough for now. Thanks @MarijnS95 !

sdroege:
@MarijnS95 You update your gtk-rs-core/gtk3-rs/gtk4-rs/gstreamer-rs PRs? :)

MarijnS95:
@sdroege Updated everything except the gtk4-rs PR, that one should be reopened before I can force-push (otherwise GH looses the PR<->branch relationship) :)

PR #1156, opened 2021-05-23
codgen: generate a builder method for objects with a Builder 

### LINKED ISSUE #1070: Add a builder method to gobjects
Adding a builder method to gobjects helps the discoverability of the builder pattern. In my opinion it would also make the usage of the builder pattern a bit more obvious.

For details see https://matklad.github.io/2020/08/12/who-builds-the-builder.html

bilelmoussaoui:
Fixes #1070

### DIFF
--- a/src/codegen/object.rs
+++ b/src/codegen/object.rs
@@ -47,6 +47,24 @@ pub fn generate(
             )?;
         }
 
+        if !analysis.builder_properties.is_empty() {
+            // generate builder method that returns the corresponding builder
+            let builder_name = format!("{}Builder", analysis.name);
+            writeln!(
+                w,
+                "
+            // rustdoc-stripper-ignore-next
+            /// Creates a new builder-style object to construct a [`{name}`]
+            /// This method returns an instance of [`{builder_name}`] which can be used to create a [`{name}`].
+            pub fn builder() -> {builder_name} {{
+                {builder_name}::default()
+            }}
+        ",
+                name = analysis.name,
+                builder_name = builder_name
+            )?;
+        }
+
         if !need_generate_trait(analysis) {
             for func_analysis in &analysis.methods() {
                 function::generate(


src/codegen/object.rs
```diff
@@ -47,6 +47,22 @@ pub fn generate(
...
         }
 
+        if !analysis.builder_properties.is_empty() {
+            // generate builder method that returns the corresponding builder
+            let builder_name = format!("{}Builder", analysis.name);
+            writeln!(
+                w,
+                "
+            // rustdoc-stripper-ignore-next
+            // Create an instance of [`{builder_name}`].
```
bilelmoussaoui, line 57: the comment here could be a bit better, what do you think @Hofer-Julian ?

Hofer-Julian: I like this one
https://docs.rs/http/0.2.1/http/request/struct.Request.html#method.builder

Maybe we can formulate it the same way?

bilelmoussaoui: Much better than what I wrote :-)

src/codegen/object.rs
```diff
@@ -47,6 +47,24 @@ pub fn generate(
...
 
+        if !analysis.builder_properties.is_empty() {
+            // generate builder method that returns the corresponding builder
+            let builder_name = format!("{}Builder", analysis.name);
+            writeln!(
+                w,
+                "
+            // rustdoc-stripper-ignore-next
+            /// Creates a new builder-style object to manufacture a `{name}`
+            /// This method returns an instance of `{builder_name}` which can be used to create a `{name}`.
```
MarijnS95, line 58: These need to be ``[`{foo}`]`` to get nice clickable links, or is that too verbose when the context/scope is already pretty clear?

bilelmoussaoui: I would add a link here just so we are consistent with the rest of the builder docs I added before. It might be a bit verbose though I guess

MarijnS95 (approved):

src/codegen/object.rs
```diff
@@ -47,6 +47,24 @@ pub fn generate(
...
         }
 
+        if !analysis.builder_properties.is_empty() {
+            // generate builder method that returns the corresponding builder
+            let builder_name = format!("{}Builder", analysis.name);
+            writeln!(
+                w,
+                "
+            // rustdoc-stripper-ignore-next
+            /// Creates a new builder-style object to manufacture a [`{name}`]
```
sdroege, line 57:
```suggestion
            /// Creates a new builder-style object to construct a [`{name}`]
```

Maybe? We call `new` functions "constructors", not manufacturers ;)

sdroege (commented):
Looks good to me apart from that

src/codegen/object.rs
bilelmoussaoui: changed :-)

PR #1160, opened 2021-05-24
[refactor] Simplify handling of trait bounds and aliases

MarijnS95:
All the cleanups from #1153 separated in nicely documented, individually reviewable commits.

This includes a simplification of `where T: IsA<Foo>` into specifying the trait bound directly when the type parameter is defined: `func<T: IsA<Foo>>`.

As requested by @GuillaumeGomez the "fix" for `clippy::use_self` is in here too, but in hindsight I'm not too sure how useful it is. This is what it does:

```diff
diff --git a/gdk/src/auto/device.rs b/gdk/src/auto/device.rs
index f53e2a154f..9994f88311 100644
--- a/gdk/src/auto/device.rs
+++ b/gdk/src/auto/device.rs
@@ -450,7 +450,7 @@ impl Device {
     }

     #[doc(alias = "changed")]
-    pub fn connect_changed<F: Fn(&Device) + 'static>(&self, f: F) -> SignalHandlerId {
+    pub fn connect_changed<F: Fn(&Self) + 'static>(&self, f: F) -> SignalHandlerId {
         unsafe extern "C" fn changed_trampoline<F: Fn(&Device) + 'static>(
             this: *mut ffi::GdkDevice,
             f: glib::ffi::gpointer,
```

`F` of the function and `F` of the `unsafe extern "C" fn` still mean the same thing, though they are now textually different. The the nested `unsafe extern "C" fn` cannot reference type parameters (`Self` is a type parameter!) of the surrounding context, which is why it still carries `Device` instead of `Self`.
A bit further don this code-block the pointer of `changed_trampoline::<F>` is taken, to pass type parameter `F` of the surrounding function into the nested one (hence they reference the same type).

### DIFF
--- a/src/codegen/bound.rs
+++ b/src/codegen/bound.rs
@@ -1,0 +1,61 @@
+use crate::{
+    analysis::{
+        bounds::{Bound, BoundType},
+        ref_mode::RefMode,
+    },
+    library::Nullable,
+};
+
+impl Bound {
+    /// Returns the type parameter reference.
+    /// Currently always returns the alias.
+    pub(super) fn type_parameter_reference(&self) -> char {
+        self.alias
+    }
+
+    /// Returns the type parameter reference, with [`BoundType::IsA`] wrapped
+    /// in `ref_mode` and `nullable` as appropriate.
+    pub(super) fn full_type_parameter_reference(
+        &self,
+        ref_mode: RefMode,
+        nullable: Nullable,
+    ) -> String {
+        let t = self.type_parameter_reference();
+        let ref_str = ref_mode.for_rust_type();
+        match self.bound_type {
+            BoundType::IsA(_) if *nullable => {
+                format!("Option<{}{}>", ref_str, t)
+            }
+            BoundType::IsA(_) => format!("{}{}", ref_str, t),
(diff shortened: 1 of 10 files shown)


src/codegen/bound.rs
```diff
@@ -0,0 +1,62 @@
...
+};
+
+impl Bound {
+    /// Returns the type parameter reference.
+    /// Currently always returns the alias.
+    pub(super) fn type_parameter_reference(&self) -> char {
+        self.alias
+    }
+
+    /// TODO: Naming
```
GuillaumeGomez, line 16:
```suggestion
    // TODO: Naming
```

GuillaumeGomez (approved):

GuillaumeGomez:
So I guess if we regen with this, the only changes we should see are the `Self` ones. If so, looks good to me! :)

src/codegen/bound.rs
sdroege: The naming seems fine to me FWIW

sdroege (approved):

src/codegen/bound.rs
MarijnS95: Cool, the `/// TODO` is entirely gone now!

GuillaumeGomez:
Thanks!

PR #1161, opened 2021-05-24
Docs: refactored + GI Docgen support

### LINKED ISSUE #1109: Gir creates broken links to objects

bilelmoussaoui:
What does this PR do:

- Adds support of the news docs linking format used in GTK4 newer Pango/GdkPixbuf and other GTK specific libraries.
See https://gnome.pages.gitlab.gnome.org/gi-docgen/linking.html for the linking spec.

-  Switch to using analysis data instead of the symbols list for figuring out whether a function, object, enum member/bitfield variant should have a doc link which solves the broken doc links

- Handles Markdown codeblocks correctly as we now try to find items in a \`item\` situation and falls back to text language in case no language is defined so that doc tests can still pass.

The advantages of doing so are:
- Support of renamed functions
- Properly ignored enum members/flag variants are handled correctly
- No more #1140 
- Properly working links or nothing.

The disadvantages: 
- Bindings now have to be extra careful with their Gir.toml configurations as that defines what we can link to for the docs or not. Note that this is not super optimal and sometimes you have to duplicate things across multiple Gir.toml configurations. The only way forward to optimize that is by having a way to merge multiple configurations (say use Gio's configuration inside GTK 4 without having to duplicate all the manual/renamed/ and bunch of stuff from there). 
- The approach of abusing manual types by putting everything that comes from other crates on it isn't great, we probably want something like "external" for types that are defined elsewhere, along from where to look at their definitions. Related to the first point.
- Probably a bit slower to run because of all the search required to find the right types/the right methods. Especially when the link doesn't provide us with enough informations. But somehow Rust handles that pretty great and the "slowness" isn't that noticeable at all. It was tested on both GTK4 by me & GStreamer bindings by @MarijnS95 

Still TODO for the future:

- The regex for catching `#Item` can catch a URL such as `https://some-url.com/#info`, the same way the regex for catching `@item` can catch `some@email.com` but those are considered very low priority breakages for now compared to the fixes this PR brings in. Of course, they should be properly handled in the future but the possible solutions we have are pretty limited especially with `rust-regex` not supporting look-ahead/look-behind. Details in https://github.com/gtk-rs/gir/pull/1161#discussion_r640217460
- Functions ignored by gi-docgen because they can't be found are reformatted as `namespace::type::the_method()` but that makes the `FUNCTION` regex catches the items and they get checked again and they end up in a double "`" which slightly breaks the formatting. Again, requires a similar solution to the item in the point 1. 
- Figure out the whole merge different Gir configurations. 

Fixes #1109, #1140

### DIFF
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -104,6 +104,15 @@ impl Info {
         self.new_name.as_ref().unwrap_or(&self.name)
     }
 
+    pub fn is_special(&self) -> bool {
+        self.codegen_name()
+            .trim_end_matches('_')
+            .split('_')
+            .last()
+            .map(|i| i.parse::<special_functions::Type>().is_ok())
+            .unwrap_or(false)
+    }
+
     pub fn is_async_finish(&self, env: &Env) -> bool {
         let has_async_result = self
             .parameters
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -115,10 +115,7 @@ impl Info {
 
     // returns whether the method can be linked in the docs
     pub fn should_be_doc_linked<F: Fn(&Self) -> bool>(&self, env: &Env, search: F) -> bool {
-        self.status != GStatus::Ignore
-            && !self.is_special()
-            && !self.is_async_finish(env)
-            && search(self)
+        !self.status.ignored() && !self.is_special() && !self.is_async_finish(env) && search(self)
     }
 
     pub fn doc_link(&self, parent: Option<&str>, visible_parent: Option<&str>) -> String {
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -124,17 +124,16 @@ impl Info {
         is_self: bool,
     ) -> String {
         if let Some(p) = parent {
-            let visible_parent_name = if is_self {
-                "Self"
+            if is_self {
+                format!("[`{f}()`][Self::{f}()]", f = self.codegen_name())
             } else {
-                visible_parent.unwrap_or(p)
-            };
-            format!(
-                "[`{visible_type_name}::{fn_name}()`][crate::{name}::{fn_name}()]",
-                name = p,
-                visible_type_name = visible_parent_name,
-                fn_name = self.codegen_name(),
-            )
+                format!(
+                    "[`{visible_parent}::{f}()`][crate::{p}::{f}()]",
+                    visible_parent = visible_parent.unwrap_or(p),
+                    p = p,
+                    f = self.codegen_name()
+                )
+            }
         } else {
             format!(
                 "[`{fn_name}()`][crate::{fn_name}()]",
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -107,8 +107,8 @@ impl Info {
     pub fn is_special(&self) -> bool {
         self.codegen_name()
             .trim_end_matches('_')
-            .split('_')
-            .last()
+            .rsplit('_')
+            .next()
             .map(|i| i.parse::<special_functions::Type>().is_ok())
             .unwrap_or(false)
     }
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -118,12 +118,22 @@ impl Info {
         !self.status.ignored() && !self.is_special() && !self.is_async_finish(env) && search(self)
     }
 
-    pub fn doc_link(&self, parent: Option<&str>, visible_parent: Option<&str>) -> String {
+    pub fn doc_link(
+        &self,
+        parent: Option<&str>,
+        visible_parent: Option<&str>,
+        is_self: bool,
+    ) -> String {
         if let Some(p) = parent {
+            let visible_parent_name = if is_self {
+                "Self"
+            } else {
+                visible_parent.unwrap_or(p)
+            };
             format!(
                 "[`{visible_type_name}::{fn_name}()`][crate::{name}::{fn_name}()]",
                 name = p,
-                visible_type_name = visible_parent.unwrap_or(p),
+                visible_type_name = visible_parent_name,
                 fn_name = self.codegen_name(),
             )
         } else {
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -113,6 +113,30 @@ impl Info {
             .unwrap_or(false)
     }
 
+    // returns whether the method can be linked in the docs
+    pub fn should_be_doc_linked<F: Fn(&Self) -> bool>(&self, env: &Env, search: F) -> bool {
+        self.status != GStatus::Ignore
+            && !self.is_special()
+            && !self.is_async_finish(env)
+            && search(self)
+    }
+
+    pub fn doc_link(&self, parent: Option<&str>, visible_parent: Option<&str>) -> String {
+        if let Some(p) = parent {
+            format!(
+                "[`{visible_type_name}::{fn_name}()`][crate::{name}::{fn_name}()]",
+                name = p,
+                visible_type_name = visible_parent.unwrap_or(p),
+                fn_name = self.codegen_name(),
+            )
+        } else {
+            format!(
+                "[`{fn_name}()`][crate::{fn_name}()]",
+                fn_name = self.codegen_name()
+            )
+        }
+    }
+
     pub fn is_async_finish(&self, env: &Env) -> bool {
         let has_async_result = self
             .parameters
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -109,8 +109,7 @@ impl Info {
             .trim_end_matches('_')
             .rsplit('_')
             .next()
-            .map(|i| i.parse::<special_functions::Type>().is_ok())
-            .unwrap_or(false)
+            .map_or(false, |i| i.parse::<special_functions::Type>().is_ok())
     }
 
     // returns whether the method can be linked in the docs
--- a/src/analysis/object.rs
+++ b/src/analysis/object.rs
@@ -10,7 +10,16 @@ use crate::{
     traits::*,
 };
 use log::info;
-use std::ops::Deref;
+use std::{borrow::Cow, ops::Deref};
+
+/// The location of an item within the object
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum LocationInObject {
+    Impl,
+    Ext,
+    ExtManual,
+    Builder,
+}
 
 #[derive(Debug, Default)]
 pub struct Info {
@@ -44,27 +53,42 @@ impl Info {
         self.signals.iter().any(|s| s.action_emit_name.is_some())
     }
 
-    /// Returns [`true`] when the function is implemented directly on the type,
-    /// [`false`] when it is implemented on an extension trait instead.
-    pub fn function_in_impl(&self, fn_info: &functions::Info) -> bool {
-        self.final_type
+    /// Returns the location of the function within this object
+    pub fn function_location(&self, fn_info: &functions::Info) -> LocationInObject {
+        if self.final_type
             || matches!(
                 fn_info.kind,
                 FunctionKind::Constructor | FunctionKind::Function
             )
+        {
+            LocationInObject::Impl
+        } else if fn_info.status == GStatus::Generate {
+            LocationInObject::Ext
+        } else {
+            LocationInObject::ExtManual
+        }
     }
 
-    // Generate a visible doc name based on whether the passed function and whether the type is final
-    pub fn generate_doc_link_info(&self, fn_info: &functions::Info) -> (String, String) {
-        if self.function_in_impl(fn_info) {
-            (self.name.clone(), self.name.clone())
-        } else {
-            let type_name = if fn_info.status == GStatus::Generate {
-                self.trait_name.clone()
-            } else {
-                format!("{}Manual", self.trait_name)
-            };
-            (format!("prelude::{}", type_name), type_name)
+    /// Generate doc name based on function location within this object
+    /// See also [`Self::function_location()`].
+    /// Returns `(item/crate path including type name, just the type name)`
+    pub fn generate_doc_link_info(
+        &self,
+        fn_info: &functions::Info,
+    ) -> (Cow<'_, str>, Cow<'_, str>) {
+        match self.function_location(fn_info) {
+            LocationInObject::Impl => (self.name.as_str().into(), self.name.as_str().into()),
+            LocationInObject::ExtManual => {
+                let trait_name = format!("{}Manual", self.trait_name);
+                (format!("prelude::{}", trait_name).into(), trait_name.into())
+            }
+            LocationInObject::Ext => (
+                format!("prelude::{}", self.trait_name).into(),
+                self.trait_name.as_str().into(),
+            ),
+            LocationInObject::Builder => {
+                panic!("C documentation is not expected to link to builders (a Rust concept)!")
+            }
         }
     }
 }
--- a/src/analysis/object.rs
+++ b/src/analysis/object.rs
@@ -44,14 +44,19 @@ impl Info {
         self.signals.iter().any(|s| s.action_emit_name.is_some())
     }
 
-    // Generate a visible doc name based on whether the passed function and whether the type is final
-    pub fn generate_doc_link_info(&self, fn_info: &functions::Info) -> (String, String) {
-        if self.final_type
+    /// Returns [`true`] when the function is implemented directly on the type,
+    /// [`false`] when it is implemented on an extension trait instead.
+    pub fn function_in_impl(&self, fn_info: &functions::Info) -> bool {
+        self.final_type
             || matches!(
                 fn_info.kind,
                 FunctionKind::Constructor | FunctionKind::Function
             )
-        {
+    }
+
+    // Generate a visible doc name based on whether the passed function and whether the type is final
+    pub fn generate_doc_link_info(&self, fn_info: &functions::Info) -> (String, String) {
+        if self.function_in_impl(fn_info) {
             (self.name.clone(), self.name.clone())
         } else {
             let type_name = if fn_info.status == GStatus::Generate {
--- a/src/analysis/object.rs
+++ b/src/analysis/object.rs
@@ -2,7 +2,13 @@ use super::{
     child_properties::ChildProperties, imports::Imports, info_base::InfoBase,
     signatures::Signatures, *,
 };
-use crate::{config::gobjects::GObject, env::Env, library, nameutil::*, traits::*};
+use crate::{
+    config::gobjects::{GObject, GStatus},
+    env::Env,
+    library::{self, FunctionKind},
+    nameutil::*,
+    traits::*,
+};
 use log::info;
 use std::ops::Deref;
 
@@ -36,6 +42,25 @@ impl Info {
 
     pub fn has_action_signals(&self) -> bool {
         self.signals.iter().any(|s| s.action_emit_name.is_some())
+    }
+
+    // Generate a visible doc name based on whether the passed function and whether the type is final
+    pub fn generate_doc_link_info(&self, fn_info: &functions::Info) -> (String, String) {
+        if self.final_type
+            || matches!(
+                fn_info.kind,
+                FunctionKind::Constructor | FunctionKind::Function
+            )
+        {
+            (self.name.clone(), self.name.clone())
+        } else {
+            let type_name = if fn_info.status == GStatus::Generate {
+                self.trait_name.clone()
+            } else {
+                format!("{}Manual", self.trait_name)
+            };
+            (format!("prelude::{}", type_name), type_name)
+        }
     }
 }
 
--- a/src/analysis/symbols.rs
+++ b/src/analysis/symbols.rs
@@ -169,7 +169,7 @@ pub fn run(library: &Library, namespaces: &namespaces::Info) -> Info {
                             name: member.name.to_camel(),
                             ..Default::default()
                         };
-                        info.insert(&member.c_identifier, symbol, Some(tid));
+                        info.insert(&member.c_identifier, symbol, None);
                     }
                     for func in functions {
                         let symbol = Symbol {
--- a/src/analysis/symbols.rs
+++ b/src/analysis/symbols.rs
@@ -18,7 +18,7 @@ impl Symbol {
     pub fn parent(&self) -> String {
         let mut ret = String::new();
         if Some("gobject") == self.crate_name() {
-            ret.push_str("glib::object::");
+            ret.push_str("glib::");
         } else {
             if let Some(ref s) = self.crate_name {
                 ret.push_str(s);
--- a/src/codegen/doc/format.rs
+++ b/src/codegen/doc/format.rs
@@ -58,8 +58,11 @@ fn get_language<'a>(entry: &'a str, out: &mut String) -> &'a str {
 fn format(input: &str, env: &Env, in_type: &str) -> String {
     let mut ret = String::with_capacity(input.len());
     // We run gi_docgen first because it's super picky about the types it replaces
-    let no_c_types_re = gi_docgen::replace_c_types(input, env, in_type);
-    ret.push_str(&replace_c_types(&no_c_types_re, env, in_type));
+    let out = replace_c_types(input, env, in_type);
+    let out = gi_docgen::replace_c_types(&out, env, in_type);
+    // this has to be done after gi_docgen replaced the various types it knows as it uses `@` in it's linking format
+    let out = PARAM_SYMBOL.replace_all(&out, |caps: &Captures<'_>| format!("`{}`", &caps[2]));
+    ret.push_str(&out);
     ret
 }
 
@@ -99,7 +102,6 @@ fn replace_c_types(entry: &str, env: &Env, _in_type: &str) -> String {
         }
     });
     let out = GDK_GTK.replace_all(&out, |caps: &Captures<'_>| find_struct(&caps[2], env));
-    let out = PARAM_SYMBOL.replace_all(&out, |caps: &Captures<'_>| format!("`{}`", &caps[2]));
     let out = TAGS.replace_all(&out, "`$0`");
     SPACES.replace_all(&out, " ").into_owned()
 }
--- a/src/codegen/doc/format.rs
+++ b/src/codegen/doc/format.rs
@@ -87,7 +87,7 @@ static SPACES: Lazy<Regex> = Lazy::new(|| Regex::new(r"[ ]{2,}").unwrap());
 fn replace_c_types(entry: &str, env: &Env, in_type: Option<&TypeId>) -> String {
     let out = FUNCTION.replace_all(entry, |caps: &Captures<'_>| {
         let name = &caps[3];
-        find_function(name, env, in_type).unwrap_or_else(|| {
+        find_method_or_function_by_ctype(None, name, env, in_type).unwrap_or_else(|| {
             info!("No function found for `{}()`", name);
             format!("`{}()`", name)
         })
@@ -100,8 +100,11 @@ fn replace_c_types(entry: &str, env: &Env, in_type: Option<&TypeId>) -> String {
         symbol_name => match &caps[1] {
             "%" => find_constant_or_variant(symbol_name, env, in_type),
             "#" => {
-                let method_name = caps.get(3).map(|m| m.as_str().trim_start_matches('.'));
-                find_method_or_type(symbol_name, method_name, env, in_type)
+                if let Some(method_name) = caps.get(3).map(|m| m.as_str().trim_start_matches('.')) {
+                    find_method(symbol_name, method_name, env, in_type)
+                } else {
+                    find_type(symbol_name, env, in_type)
+                }
             }
             s => panic!("Unknown symbol prefix `{}`", s),
         }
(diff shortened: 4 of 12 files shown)


src/codegen/doc/format.rs
```diff
@@ -57,48 +55,27 @@ fn get_language<'a>(entry: &'a str, out: &mut String) -> &'a str {
     entry
 }
 
-fn format(mut input: &str, env: &Env, in_type: &str) -> String {
+fn format(input: &str, env: &Env, in_type: &str) -> String {
```
bilelmoussaoui, line 58: This was a bad idea actually i guess, it breaks some stuff but most of those are due to it's abuse a bit. It's used in different use cases like being called on `code_blocks_transformation` which doesn't make much sense tbh. The commit that introduced it's usage just says it's improved the formatting of the codeblocks. But then it's also called with `replace_c_types` which means we replace the C code with Rust equivalent? in codeblocks? that doesn't sound correct. 

We should instead just match single  "```" for codeblocks instead of generalizing the function on all the cases.

src/codegen/doc/mod.rs
```diff
@@ -766,11 +765,11 @@ fn get_type_trait_for_implements(env: &Env, tid: TypeId) -> String {
         format!("{}Ext", env.library.type_(tid).get_name())
     };
     if tid.ns_id == MAIN_NAMESPACE {
-        format!("[`trait@crate::prelude::{}`]", &trait_name)
+        format!("[{name}](crate::prelude::{name})", name = &trait_name)
     } else if let Some(symbol) = env.symbols.borrow().by_tid(tid) {
         let mut symbol = symbol.clone();
         symbol.make_trait(&trait_name);
-        format!("[`trait@{}`]", &symbol.full_rust_name())
+        format!("[{name}](trait@{name})", name = &symbol.full_rust_name())
```
MarijnS95, line 772: `trait@` is already hidden right?

src/codegen/doc/format.rs
```diff
@@ -54,121 +57,258 @@ fn get_language<'a>(entry: &'a str, out: &mut String) -> &'a str {
...
+            .functions
+            .iter()
+            .filter(|fn_info| fn_info.status != GStatus::Ignore)
+            .filter(|fn_info| !fn_info.is_special() && !fn_info.is_async_finish(env))
+            .find(|fn_info| fn_info.glib_name == name)
+            .map(|fn_info| (record_info, fn_info))
     });
-    let out = TAGS.replace_all(&out, "`$0`");
-    SPACES.replace_all(&out, " ").into_owned()
+    let is_globa_func = env.analysis.global_functions.as_ref().and_then(|info| {
```
MarijnS95, line 269:
```suggestion
    let is_global_func = env.analysis.global_functions.as_ref().and_then(|info| {
```
?

src/codegen/doc/format.rs
```diff
@@ -54,121 +57,258 @@ fn get_language<'a>(entry: &'a str, out: &mut String) -> &'a str {
...
+
+fn find_struct(name: &str, env: &Env) -> String {
+    let symbols = env.symbols.borrow();
+
+    let symbol = if let Some(obj) = env
+        .analysis
+        .objects
+        .iter()
+        .find(|(_, o)| o.c_type == name)
+        .map(|(_, o)| o)
```
MarijnS95, line 228: Why not map first? Saves unpacking the tuple in the `find`.

MarijnS95 (changes requested):
From a very quick scroll-through this appears to be really slow code. The `find_` functions all take a single symbol string and then iterate multiple collections to find the item that we want, with extra constraints and whatnot - and that would happen for every symbol occurence. At the very least these collections should be stored in `HashMap`s for `O(1)` lookup, and where possible have predicates applied to it up-front to slim down its size.

src/analysis/functions.rs
```diff
@@ -104,6 +104,15 @@ impl Info {
...
     }
 
+    pub fn is_special(&self) -> bool {
+        self.codegen_name()
+            .trim_end_matches('_')
+            .split('_')
+            .last()
+            .map(|i| i.parse::<special_functions::Type>().is_ok())
+            .unwrap_or(false)
+    }
```
MarijnS95, line 114: This should be `rsplit('_').first() to grab the last item directly. Note that these functions always return the string remainder, so `""` or any other string not containing `_` yields exactly one item (but at least it can easily be parsed as special function).

src/analysis/symbols.rs
```diff
@@ -18,7 +18,7 @@ impl Symbol {
     pub fn parent(&self) -> String {
         let mut ret = String::new();
         if Some("gobject") == self.crate_name() {
-            ret.push_str("glib::object::");
+            ret.push_str("glib::");
```
MarijnS95, line 21: Did it move or do we have a better object path now?

src/codegen/doc/format.rs
```diff
@@ -54,121 +57,258 @@ fn get_language<'a>(entry: &'a str, out: &mut String) -> &'a str {
...
-    Lazy::new(|| Regex::new(r"(^|[^\\])([@#%])(\w+\b)([:.]+[\w-]+\b)?").unwrap());
+static SYMBOL: Lazy<Regex> = Lazy::new(|| Regex::new(r"([#%])(\w+\b)([:.]+[\w-]+\b)?").unwrap());
+static PARAM_SYMBOL: Lazy<Regex> =
+    Lazy::new(|| Regex::new(r"([@])(\w+\b)([:.]+[\w-]+\b)?").unwrap());
 static FUNCTION: Lazy<Regex> =
     Lazy::new(|| Regex::new(r"([@#%])?(\w+\b[:.]+)?(\b[a-z0-9_]+)\(\)").unwrap());
-static GDK_GTK: Lazy<Regex> = Lazy::new(|| Regex::new(r"G[dt]k[A-Z]\w+\b").unwrap());
+// **note**
+// The optional . at the end is to make the regex more relaxed for some weird broken cases on gtk3's docs
+// it doesn't hurt other docs so please don't drop it
```
MarijnS95, line 75: Have an example here? Would be nice to have it in the code at least (or a test) so that future developers know what to keep in mind.

src/analysis/symbols.rs
bilelmoussaoui: it's used for a ton of cases like `glib::object::Value` which is a wrong link because Value is in `glib::value::Value` or `glib::object::ObjectExt` which is already in the prelude. So I removed it :)

bilelmoussaoui:
> From a very quick scroll-through this appears to be really slow code. The `find_` functions all take a single symbol string and then iterate multiple collections to find the item that we want, with extra constraints and whatnot - and that would happen for every symbol occurence. At the very least these collections should be stored in `HashMap`s for `O(1)` lookup, and where possible have predicates applied to it up-front to slim down its size.

It's a very slow code yes, never said it will be super performant. But it gives super great results so far :)

src/codegen/doc/format.rs
bilelmoussaoui: There were some occurrences in gtk with `GtkSomething.` and it wasn't picked, although, that would be a bug to fix in the docs... I don't really mind dropping it :) it was mostly as an explanation for the review phase

src/codegen/doc/mod.rs
```diff
@@ -766,11 +765,11 @@ fn get_type_trait_for_implements(env: &Env, tid: TypeId) -> String {
         format!("{}Ext", env.library.type_(tid).get_name())
     };
     if tid.ns_id == MAIN_NAMESPACE {
-        format!("[`trait@crate::prelude::{}`]", &trait_name)
+        format!("[{name}](crate::prelude::{name})", name = &trait_name)
```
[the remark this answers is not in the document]
bilelmoussaoui: That's true. The change can be reverted although, a huge reason why [][] wasn't a great choice is because of the broken links which we mostly don't have with this approach.

MarijnS95:
> It's a very slow code yes, never said it will be super performant. But it gives super great results so far :)

To put it bluntly those `find_` functions look "fundamentally wrong" but I guess it's good practice to go through a bunch of iterations first to make it work, and refactor for optimization and simplicity again after - at least that's how I generally approach my changesets :grin: 

That said it's not nearly as bad as I thought when running `gir -m doc`. This is with the embed step commented out, only generating `docs.md` on all of GStreamer, before and after:

```sh
./generator.py --embed-docs  9.35s user 0.36s system 441% cpu 2.199 total
./generator.py --embed-docs  9.90s user 0.35s system 457% cpu 2.240 total
```

Well within margin of error. I'm curious how Rust it made that happen :grin:

src/codegen/doc/mod.rs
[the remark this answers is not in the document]
MarijnS95: `[][]` is still probably the better choice, so that a broken link is rendered as the plaintext markdown instead of being clickable and linking to garbage. Let's discuss this elsewhere though :)

bilelmoussaoui:
I just made most of the parsing work and replacing work correctly and I will get back to see how can I optimize the `find` phase. I didn't want to want to try to do something just for the `gi-docgen` and then end up with a completely different requirements for the rest of the docs. 

Now that everything is more or less set on stone,  I will see this weekend how can I simplify things a bit as some of the find functions could be stored as you mentioned in a HashMap to optimize things and simplify the search code for both gi-docgen & the "normal" docs. 

Also note, the PR still says "Very WIP" :P

MarijnS95:
> Also note, the PR still says "Very WIP" :P

I know I know, but I'm never sure when and where to give feedback. Either the feedback is something the author is well aware of and is already planning to tackle, other times the feedback comes after the PR is marked "ready for review" and would incur significant changes that have better been brought up earlier in the progess. It's a fine line :)

MarijnS95:
Noticing a couple more issues:

- None of the links appear to use monospace anymore (``[`this should be monospace`]``), I don't think that removal was intended?
- `()` postfixes are gone from all function links. I like these for two reasons;
  1. Documentation immediately makes it clear this points to a link (just like the g-ir docs);
  2. `rustdoc` ensures the link points to a function (AFAIK also used to distinguish [namespaces](https://doc.rust-lang.org/stable/reference/names/namespaces.html))).
- Local references are not using `Self::` anymore.

Then, some random missing conversions:
- A bunch of `GST_FLOW_*` enums are not turned into `crate::gst::FlowReturn::*` anymore;
- `GstFormat` and `GstClockTime` in `gstreamer-audio/src/audio_info.rs`;
- `Self::` functions in `gstreamer-audio/src/auto/audio_decoder.rs` are not resolved at all, while they are definitely listed as manual functions;
- `Self::` functions in `gstreamer-audio/src/auto/audio_encoder.rs` are now using the verbose `crate::` format.

Could these be missing from the manual list?

---

I guess I could go on but that's just noise at this point. If you can address the points above so that the diff doesn't change nearly as much while dealing with all annotations the right way, we can look at failed resolution later.

We should probably get (a variant of) #1150 in first to massively reduce diff surface, and just `git checkout --theirs` when rebasing this PR.

src/codegen/doc/mod.rs
```diff
@@ -336,7 +335,12 @@ fn create_object_doc(w: &mut dyn Write, env: &Env, info: &analysis::object::Info
                 .into_iter()
                 .chain(env.class_hierarchy.subtypes(info.type_id))
                 .filter(|&tid| !env.type_status(&tid.full_name(&env.library)).ignored())
-                .map(|tid| format!("[`struct@crate::{}`]", env.library.type_(tid).get_name()))
+                .map(|tid| {
+                    format!(
+                        "[{name}](crate::{name})",
```
MarijnS95, line 340:
```suggestion
                        "[{name}](struct@crate::{name})",
```

src/codegen/doc/mod.rs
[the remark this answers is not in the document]
MarijnS95: Just noticed, `trait@` got lost here too.

src/analysis/object.rs
```diff
@@ -37,6 +43,23 @@ impl Info {
     pub fn has_action_signals(&self) -> bool {
         self.signals.iter().any(|s| s.action_emit_name.is_some())
     }
+
+    // Generate a visible doc name based on whether the passed function and whether the type is final
+    pub fn generate_doc_link_info(&self, fn_info: &functions::Info) -> (String, String) {
+        if self.final_type
+            || fn_info.kind == FunctionKind::Constructor
+            || fn_info.kind == FunctionKind::Function
```
MarijnS95, line 51: `matches!(fn_info.kind, FunctionKind::Constructor | FunctionKind::Function)`?

src/codegen/doc/format.rs
```diff
@@ -191,35 +191,45 @@ fn find_type(name: &str, env: &Env) -> String {
...
         .analysis
         .enumerations
         .iter()
         .find(|e| e.type_(&env.library).c_type == name)
     {
-        symbols.by_tid(enum_.type_id)
-    } else if let Some(flag) = env
+        let sym = symbols.by_tid(enum_info.type_id).unwrap();
+        Some(format!(
+            "[{name}](crate::{parent})",
```
MarijnS95, line 213:
```suggestion
            "[{name}][crate::{parent}]
            ",
```

src/codegen/doc/format.rs
```diff
@@ -191,35 +191,45 @@ fn find_type(name: &str, env: &Env) -> String {
...
+    } else if let Some(flag_info) = env
         .analysis
         .flags
         .iter()
         .find(|f| f.type_(&env.library).c_type == name)
     {
-        symbols.by_tid(flag.type_id)
+        let sym = symbols.by_tid(flag_info.type_id).unwrap();
+        Some(format!(
+            "[{name}](crate::{parent})",
```
MarijnS95, line 225:
```suggestion
            "[{name}][crate::{parent}]",
```

src/codegen/doc/format.rs
```diff
@@ -191,35 +191,45 @@ fn find_type(name: &str, env: &Env) -> String {
...
+            name = flag_info.name,
+            parent = sym.parent().trim_end_matches("::")
+        ))
```
MarijnS95: This is definitely very fishy. Apparently the `type_id` for an _enum type_ gets overridden with the last _enum **variant**_ within the type. That should be solved in the symbols list don't you think?

MarijnS95: Actual fix for this is a one-liner: https://github.com/bilelmoussaoui/gir/pull/3 - you can drop this workaround again :grin:

src/codegen/doc/format.rs
```diff
@@ -54,121 +55,207 @@ fn get_language<'a>(entry: &'a str, out: &mut String) -> &'a str {
...
+    }) {
+        let sym = symbols.by_tid(enum_info.type_id).unwrap();
+        format!(
+            "[`{enum_name}::{member}`][crate::{parent}::{member}]",
+            enum_name = enum_info.name,
+            member = nameutil::enum_member_name(&member_info.name),
+            parent = sym.full_rust_name()
+        )
+    } else {
+        format!("`{}`", symbol)
```
bilelmoussaoui, line 179: this one is not great because now links such as `https://drafts.csswg.org/css-backgrounds-3/#border-radius` will be broken. :/ 
cc @MarijnS95

src/codegen/doc/format.rs
```diff
@@ -54,121 +55,207 @@ fn get_language<'a>(entry: &'a str, out: &mut String) -> &'a str {
...
+            .find(|m| m.c_identifier == symbol)
+            .map(|m| (f, m))
+    }) {
+        let sym = symbols.by_tid(flag_info.type_id).unwrap();
+        format!(
+            "[`{flag_name}::{member_name}`][crate::{parent}::{member_name}]",
+            member_name = nameutil::bitfield_member_name(&member_info.name),
+            flag_name = flag_info.name,
+            parent = sym.full_rust_name()
+        )
```
MarijnS95, line 163: Added an `assert_eq!(flag_info.name, sym.full_rust_name())` here, and it doesn't trigger on any of GStreamer. I don't think we need this extra separation, unless you have found a case where the Rust type linked to should be different from the name of the flag? IMO I'd like to see the namespace (that `full_rust_name()` provides) in that case.

src/codegen/doc/format.rs
```diff
@@ -54,121 +55,207 @@ fn get_language<'a>(entry: &'a str, out: &mut String) -> &'a str {
...
+            .find(|m| m.c_identifier == symbol)
+            .map(|m| (e, m))
+    }) {
+        let sym = symbols.by_tid(enum_info.type_id).unwrap();
+        format!(
+            "[`{enum_name}::{member}`][crate::{parent}::{member}]",
+            enum_name = enum_info.name,
+            member = nameutil::enum_member_name(&member_info.name),
+            parent = sym.full_rust_name()
+        )
```
MarijnS95, line 177: Same here, `enum_info.name` is always equivalent to `sym.full_rust_name()`, do we need this?

src/codegen/doc/format.rs
MarijnS95: I am not sure what this is about? Theoretically this regex only parses `#Foo`, `%SOME_CONSTANT`, `@parameter-or-self-member`, and `some_function()`, and does not deal with links? Do you have documentation source for an example? I don't recognize the link ;)

bilelmoussaoui: The link comes from gtk4

MarijnS95: I need a little more context than that :)

MarijnS95:
```
gtk3-rs/gir-files/Gsk-4.0.gir
2871-The algorithm used for normalizing corner sizes is described in
2872:[the CSS specification](https://drafts.csswg.org/css-backgrounds-3/#border-radius).</doc>
2873-      <field name="bounds" writable="1">

gtk-rs-core/gir-files/Gsk-4.0.gir
2871-The algorithm used for normalizing corner sizes is described in
2872:[the CSS specification](https://drafts.csswg.org/css-backgrounds-3/#border-radius).</doc>
2873-      <field name="bounds" writable="1">

gtk4-rs/gir-files/Gsk-4.0.gir
2871-The algorithm used for normalizing corner sizes is described in
2872:[the CSS specification](https://drafts.csswg.org/css-backgrounds-3/#border-radius).</doc>
2873-      <field name="bounds" writable="1">
```

Those are normal markdown links, and should not be touched by this code IMO.

bilelmoussaoui: Sorry I'm on a phone right now that's why I couldn't provide more context. The regex still catches the link and breaks it by surrounding it with ``

src/codegen/doc/format.rs
bilelmoussaoui: Needed, not really :) I think i did that when the enum members was breaking the symbols list

src/analysis/functions.rs
bilelmoussaoui: The suggestion doesn't quiet work because rsplit returns a `RSplit` which doesn't have a first method :-) first is unique to slices. I could use get but I don't see what would be the differences here? :) at the end you get the same result

src/codegen/doc/format.rs
```diff
@@ -54,121 +55,202 @@ fn get_language<'a>(entry: &'a str, out: &mut String) -> &'a str {
...
+
+/// either an object/interface, record, enum or a flag
+fn find_type(name: &str, env: &Env) -> String {
+    let symbols = env.symbols.borrow();
+
+    let symbol = if let Some(obj) = env
+        .analysis
+        .objects
+        .iter()
+        .map(|(_, o)| o)
```
MarijnS95, line 189: I just realized this is a `BTreeMap`:

```suggestion
        .values()
        
```

src/codegen/doc/format.rs
```diff
@@ -54,121 +55,202 @@ fn get_language<'a>(entry: &'a str, out: &mut String) -> &'a str {
...
+        .find(|o| o.c_type == name)
+    {
+        symbols.by_tid(obj.type_id)
+    } else if let Some(record) = env
+        .analysis
+        .records
+        .iter()
+        .map(|(_, r)| r)
```
MarijnS95, line 197: And here too:

```suggestion
        .values()
```

src/analysis/functions.rs
MarijnS95: Mistyped, `.first()` makes no sense on iterators, that returns the same as `.next()` :wink

src/codegen/doc/format.rs
MarijnS95: I guess we should prevent parsing things inside `[]()` instead... Ugh :)

bilelmoussaoui: without look-ahead/look-behind? :p

src/codegen/doc/format.rs
```diff
@@ -54,121 +60,352 @@ fn get_language<'a>(entry: &'a str, out: &mut String) -> &'a str {
...
+    "g_list_free",
+    "g_strfreev",
+    "printf",
+];
+/// Find a function in all the possible items, if not found return the original name surrounded with backticks
+/// A function can either be a struct/interface/record method, a global function or maybe a virtual function
+///
+/// This function is generic so it can be de-duplicated between a
+/// - "find_method_or_function_by_ctype" where the object/records are looked by their C name
+/// - "find_method_or_function_by_name" where the object/records are looked by their name
```
MarijnS95, line 274: Did `find_method_or_function_by_name` disappear again during some refactor?

MarijnS95: Btw, ``[`find_method_or_function_by_ctype`]`` :wink:

bilelmoussaoui: Nop i refactored it a bit to reuse it for gi-docgen

MarijnS95: Oh, this sits in `gi_docgen`. Yeah, use intradoc-links to clarify that :)

MarijnS95: image

^^

bilelmoussaoui: ah my bad. I forgot we are doing all this work to make docs better... haha

src/codegen/doc/format.rs
bilelmoussaoui: I have fixed this now in the latest commit :)

sdroege:
@bilelmoussaoui @MarijnS95 Is this ready for a review now? Also I assume this still works correctly with the old documentation format?

The new one used by GTK4 is only used by GTK4 and there is no coordination at all with other projects. It's likely to stay a GTK4-only thing for quite a while.

MarijnS95:
@sdroege Ready from my side, I think I've reviewed and adjusted most of the non gi-docgen by now. Still missing support for aliases (needs to be added to `fn find_type`) and dealing with renames of `Gst.MessageType` which is treated as a bitfield with `SNAKE_CASE` but manually implemented as regular enumeration using `CamelCase`. See the most recent documentation delta at https://gitlab.freedesktop.org/MarijnS95/gstreamer-rs/-/commits/docs, last commit.

bilelmoussaoui:
> @bilelmoussaoui @MarijnS95 Is this ready for a review now? Also I assume this still works correctly with the old documentation format?
> 
> The new one used by GTK4 is only used by GTK4 and there is no coordination at all with other projects. It's likely to stay a GTK4-only thing for quite a while.

It's ready from my side as well

src/analysis/object.rs
```diff
@@ -37,6 +52,42 @@ impl Info {
...
+        match self.function_location(fn_info) {
+            LocationInObject::Impl => (self.name.clone(), self.name.clone()),
+            loc @ LocationInObject::Ext | loc @ LocationInObject::ExtManual => {
+                let type_name = if loc == LocationInObject::ExtManual {
+                    format!("{}Manual", self.trait_name)
+                } else {
+                    self.trait_name.clone()
+                };
+                (format!("prelude::{}", type_name), type_name)
+            }
```
GuillaumeGomez, line 85:
```suggestion
            LocationInObject::ExtManual => {
                (format!("prelude::{}Manual", self.trait_name), type_name)
            }
            LocationInObject::Ext => {
                (format!("prelude::{}", self.trait_name), trait_name)
            }
```

src/codegen/doc/format.rs
```diff
@@ -54,121 +79,424 @@ fn get_language<'a>(entry: &'a str, out: &mut String) -> &'a str {
...
+        env,
+        in_type,
+        |f| f.glib_name == name,
+        |o| c_type.map_or(true, |t| o.c_type == t),
+        |r| c_type.map_or(true, |t| r.type_(&env.library).c_type == t),
+        c_type.map_or(false, |t| t.ends_with("Class")),
+    )
+}
+
+/// Find a function in all the possible items, if not found return the original name surrounded with backticks
```
GuillaumeGomez, line 363: You have the right to add dots at the end of your sentences. It's even more important here since it's one paragraph since there is no empty line after. ;)

GuillaumeGomez:
> The new one used by GTK4 is only used by GTK4 and there is no coordination at all with other projects. It's likely to stay a GTK4-only thing for quite a while.

This part is quite worrying. GTK4 is a really small part of the bindings we provide after all...

> Bindings now have to be extra careful with their Gir.toml configurations as that defines what we can link to for the docs or not. Note that this is not super optimal and sometimes you have to duplicate things across multiple Gir.toml configurations. The only way forward to optimize that is by having a way to merge multiple configurations (say use Gio's configuration inside GTK 4 without having to duplicate all the manual/renamed/ and bunch of stuff from there).

Do you have examples for that? I have issues trying to figure out what would need to be changed in `Gir.toml` files...

src/analysis/object.rs
MarijnS95: @GuillaumeGomez I thought about the same, but ended up not doing that because `ExtManual` needs to format the type twice for them to be identical:

```suggestion
            LocationInObject::ExtManual => {
                (format!("prelude::{}Manual", self.trait_name), /* HERE: */ format!("{}Manual", type_name))
            }
            LocationInObject::Ext => {
                (format!("prelude::{}", self.trait_name), trait_name)
            }
```

Your suggestion confirms that that's hard to miss, and that it might have been a good choice to leave this as-is :grimacing:

MarijnS95:
> Do you have examples for that? I have issues trying to figure out what would need to be changed in `Gir.toml` files...

https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/789

sdroege:
> > The new one used by GTK4 is only used by GTK4 and there is no coordination at all with other projects. It's likely to stay a GTK4-only thing for quite a while.
> 
> This part is quite worrying. GTK4 is a really small part of the bindings we provide after all...

It is, but as GTK4 is using that we'll have to support it I guess.

src/analysis/object.rs
GuillaumeGomez: Except that the current code is very hard to read and is cloning unnecessarily. In addition: we match against `LocationInObject::ExtManual`, so it seems pretty obvious. At least much more than having binding and then inner condition which is completely unnecessary.

GuillaumeGomez:
> It is, but as GTK4 is using that we'll have to support it I guess.

My point was that it's a lot of change for handling one exception.



> > Do you have examples for that? I have issues trying to figure out what would need to be changed in `Gir.toml` files...
> 
> https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/789

Ok, so from what I can read, we don't detect functions that can't be generated and we have to ignore them manually. That seems really suboptimal. What prevents us to check that in `gir` directly?

bilelmoussaoui:
> > The new one used by GTK4 is only used by GTK4 and there is no coordination at all with other projects. It's likely to stay a GTK4-only thing for quite a while.
> 
> This part is quite worrying. GTK4 is a really small part of the bindings we provide after all...
> 
> > Bindings now have to be extra careful with their Gir.toml configurations as that defines what we can link to for the docs or not. Note that this is not super optimal and sometimes you have to duplicate things across multiple Gir.toml configurations. The only way forward to optimize that is by having a way to merge multiple configurations (say use Gio's configuration inside GTK 4 without having to duplicate all the manual/renamed/ and bunch of stuff from there).
> 
> Do you have examples for that? I have issues trying to figure out what would need to be changed in `Gir.toml` files...

Not only GTK4 but also Pango(part of gtk-rs-core), GdkPixbuf(part of gtk-rs-core), libadwaita, gexiv2

bilelmoussaoui:
> > It is, but as GTK4 is using that we'll have to support it I guess.
> 
> My point was that it's a lot of change for handling one exception.

No not really, a ton of changes were to fix the broken links generated, the whole gi-docgen can fit in two commits.

bilelmoussaoui:
> Ok, so from what I can read, we don't detect functions that can't be generated and we have to ignore them manually. That seems really suboptimal. What prevents us to check that in `gir` directly?

Not only that but also if you have a method in crate X that's marked as manual, crate Y can't know that it's manual and not auto-generated. So you have to re-define some of the stuff in Y. That's not related to these changes and has been a broken concept in gir since forever, see https://github.com/gtk-rs/gir/issues/1099 for another variant of the same issue.

src/analysis/object.rs
MarijnS95: Yes 2x `loc @` is horrible, and putting the `clone` behind a `Cow` is nasty too. All the `format!`s do allocations too though. FIne to change this then.

GuillaumeGomez:
> Not only GTK4 but also Pango(part of gtk-rs-core), GdkPixbuf(part of gtk-rs-core), libadwaita, gexiv2

Sounds much better.

> > Ok, so from what I can read, we don't detect functions that can't be generated and we have to ignore them manually. That seems really suboptimal. What prevents us to check that in `gir` directly?
> 
> Not only that but also if you have a method in crate X that's marked as manual, crate Y can't know that it's manual and not auto-generated. So you have to re-define some of the stuff in Y. That's not related to these changes and has been a broken concept in gir since forever, see #1099 for another variant of the same issue.

We can go around that limitation by linking to a type method instead of linking to a trait method. :) (might be ugly, but at least it works as long as it's bound XD)

src/analysis/object.rs
GuillaumeGomez: Thanks! :)

MarijnS95: How about doing `Manual` formatting once though:
```
            LocationInObject::ExtManual => {
                let trait_name = format!("{}Manual", self.trait_name);
                (format!("prelude::{}", trait_name), trait_name)
            }
            LocationInObject::Ext => {
                (format!("prelude::{}", self.trait_name), self.trait_name.clone())
            }
```

Your suggestion forget to `clone` in `::Ext` btw.

GuillaumeGomez: You could return a `(String, &str)`, instead, no?

MarijnS95:
> Ok, so from what I can read, we don't detect functions that can't be generated and we have to ignore them manually. That seems really suboptimal. What prevents us to check that in `gir` directly?

Yes it's a bit silly that functions that cannot be autogenerated need to be explicitly marked as `ignore=true` to not show up in docs. OTOH it more easily allows us to:
- Have new functions generated as by magic when new types are supported;
- Use broken link warnings to find these gaps in `Gir.toml` in the first place.

bilelmoussaoui:
> > Ok, so from what I can read, we don't detect functions that can't be generated and we have to ignore them manually. That seems really suboptimal. What prevents us to check that in `gir` directly?
> 
> Yes it's a bit silly that functions that cannot be autogenerated need to be explicitly marked as `ignore=true` to not show up in docs. OTOH it more easily allows us to:
> 
>     * Have new functions generated as by magic when new types are supported;
> 
>     * Use broken link warnings to find these gaps in `Gir.toml` in the first place.

This needs a new GStatus that marks items as "CantBeGenerated", it can be done yes but it's unrelated to this docs PR.

src/analysis/object.rs
MarijnS95: The latter would have to be a `Cow<'static, str>` so that `ExtManual` can pass the _owned_ result of `format!`.

GuillaumeGomez:
It's completely related to this PR. You change a lot of things and "fix links". However, you don't take into account the fact that some items cannot be generated and even if `gir` has this information, you don't use it, forcing instead end users to add a lot of items in their `Gir.toml` file.

bilelmoussaoui:
> It's completely related to this PR. You change a lot of things and "fix links". However, you don't take into account the fact that some items cannot be generated and even if `gir` has this information, you don't use it, forcing instead end users to add a lot of items in their `Gir.toml` file.

They don't have to, the docs links will be just broken for them.

src/analysis/object.rs
MarijnS95: @GuillaumeGomez Would this work for you?:

```rust
    pub fn generate_doc_link_info(
        &self,
        fn_info: &functions::Info,
    ) -> (Cow<'_, str>, Cow<'_, str>) {
        match self.function_location(fn_info) {
            LocationInObject::Impl => (self.name.as_str().into(), self.name.as_str().into()),
            LocationInObject::ExtManual => {
                let trait_name = format!("{}Manual", self.trait_name);
                (format!("prelude::{}", trait_name).into(), trait_name.into())
            }
            LocationInObject::Ext => (
                format!("prelude::{}", self.trait_name).into(),
                self.trait_name.as_str().into(),
            ),
            LocationInObject::Builder => {
                panic!("C documentation is not expected to link to builders (a Rust concept)!")
            }
        }
    }
}
```

GuillaumeGomez: Yep it's fine. :)

MarijnS95: @bilelmoussaoui You know what to copy-paste :wink:

bilelmoussaoui:
I have updated the PR with the requested changes

sdroege:
Let's get it in then once the CI succeeded :)

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks a lot for this!

PR #1162, opened 2021-05-25
codegen/doc: Use `[][]` for intradoc links instead of `[]()`

MarijnS95:
As discussed in #1150, the slightly more complete version. We have #1161 in progress too, but getting this in first (and using the same format there) should make it _much_ easier to do a before-after diff on the generated docs.

---

Rustdoc treats `[foo]` as item path, and by the same rule `[fancy text][foo]` too.  Unlike `[]()`, which generates a dead link if the right-hand side is a valid Rust item path but doesn't exist in scope (common and unfortunately unavoidable with docs for C code auto-ported to Rust), the `[]` and `[][]` do not generate a link at all.

### DIFF
--- a/src/codegen/doc/format.rs
+++ b/src/codegen/doc/format.rs
@@ -100,7 +100,7 @@ fn replace_c_types(entry: &str, symbols: &symbols::Info, in_type: &str) -> Strin
             if sym.owner_name() == Some(in_type) {
                 // `#` or `%` symbols should probably have been `@` to denote
                 // that it is a reference within the current type.
-                format!("[`Self::{}()`]", sym.name())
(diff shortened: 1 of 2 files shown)


bilelmoussaoui:
Do we still need this one?

MarijnS95:
@bilelmoussaoui I guess it would be nice for history purposes, but makes your job rebasing and conflict resolving a little tougher.

bilelmoussaoui:
> @bilelmoussaoui I guess it would be nice for history purposes, but makes your job rebasing and conflict resolving a little tougher.

we already doing that in the other PR, i don't see why you want me to suffer even more :P

MarijnS95:
Let's close this then, should be fine to switch from `[]()` to `[][]` at the same time.

PR #1163, opened 2021-05-26
anlysis: Ignore per-type `ref_mode` override for input transfer-full

MarijnS95:
This override is mainly used to turn common mutability issues for all parameters of a specific type down to immutable borrows when the underlying C source cannot be patched as it uses interior mutability for refcounting etc.

The override is only used in GStreamer, but is unfortunately applied in such a way that the few cases using `transfer-full` parameters also get their argument turned into a borrow, leading to incorrect API and
sometimes failed compilation when trait bounds can (logically) not be satisfied.  This is simply addressed by returning `RefMode::None` whenever an override is provided for an input parameter that happens to
be `transfer-full`.

### DIFF
--- a/src/analysis/child_properties.rs
+++ b/src/analysis/child_properties.rs
@@ -100,7 +100,12 @@ fn analyze_property(
             imports.add_used_types(rust_type.used_types());
         }
 
-        let get_out_ref_mode = RefMode::of(env, typ, library::ParameterDirection::Return);
+        let get_out_ref_mode = RefMode::of(
+            env,
+            typ,
+            library::ParameterDirection::Return,
+            library::Transfer::None,
+        );
         if !is_getter_renamed {
             if let Ok(new_name) = getter_rules::try_rename_getter_suffix(
                 &getter_name,
@@ -110,7 +115,12 @@ fn analyze_property(
             }
         }
 
-        let mut set_in_ref_mode = RefMode::of(env, typ, library::ParameterDirection::In);
+        let mut set_in_ref_mode = RefMode::of(
+            env,
+            typ,
+            library::ParameterDirection::In,
+            library::Transfer::None,
+        );
         if set_in_ref_mode == RefMode::ByRefMut {
             set_in_ref_mode = RefMode::ByRef;
         }
--- a/src/analysis/properties.rs
+++ b/src/analysis/properties.rs
@@ -393,8 +393,18 @@ pub fn get_property_ref_modes(
     env: &Env,
     prop: &library::Property,
 ) -> (RefMode, RefMode, library::Nullable) {
-    let get_out_ref_mode = RefMode::of(env, prop.typ, library::ParameterDirection::Return);
-    let mut set_in_ref_mode = RefMode::of(env, prop.typ, library::ParameterDirection::In);
+    let get_out_ref_mode = RefMode::of(
+        env,
+        prop.typ,
+        library::ParameterDirection::Return,
+        library::Transfer::None,
+    );
+    let mut set_in_ref_mode = RefMode::of(
+        env,
+        prop.typ,
+        library::ParameterDirection::In,
+        library::Transfer::None,
+    );
     if set_in_ref_mode == RefMode::ByRefMut {
         set_in_ref_mode = RefMode::ByRef;
     }
--- a/src/analysis/ref_mode.rs
+++ b/src/analysis/ref_mode.rs
@@ -17,6 +17,7 @@ impl RefMode {
         env: &env::Env,
         tid: library::TypeId,
         direction: library::ParameterDirection,
+        transfer: library::Transfer,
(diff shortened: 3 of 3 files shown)


src/analysis/child_properties.rs
```diff
@@ -100,7 +100,12 @@ fn analyze_property(
             imports.add_used_types(rust_type.used_types());
         }
 
-        let get_out_ref_mode = RefMode::of(env, typ, library::ParameterDirection::Return);
+        let get_out_ref_mode = RefMode::of(
+            env,
+            typ,
+            library::ParameterDirection::Return,
+            library::Transfer::None,
```
MarijnS95, line 107: Not sure if this is a sensible default.

src/analysis/child_properties.rs
```diff
@@ -110,7 +115,12 @@ fn analyze_property(
...
 
-        let mut set_in_ref_mode = RefMode::of(env, typ, library::ParameterDirection::In);
+        let mut set_in_ref_mode = RefMode::of(
+            env,
+            typ,
+            library::ParameterDirection::In,
+            library::Transfer::None,
+        );
         if set_in_ref_mode == RefMode::ByRefMut {
             set_in_ref_mode = RefMode::ByRef;
```
MarijnS95, line 125: Probably not a sensible default.

src/analysis/properties.rs
```diff
@@ -393,8 +393,18 @@ pub fn get_property_ref_modes(
     env: &Env,
     prop: &library::Property,
 ) -> (RefMode, RefMode, library::Nullable) {
-    let get_out_ref_mode = RefMode::of(env, prop.typ, library::ParameterDirection::Return);
-    let mut set_in_ref_mode = RefMode::of(env, prop.typ, library::ParameterDirection::In);
+    let get_out_ref_mode = RefMode::of(
+        env,
+        prop.typ,
+        library::ParameterDirection::Return,
+        library::Transfer::None,
```
MarijnS95, line 400: Not sure if this is a sensible default.

src/analysis/properties.rs
```diff
@@ -393,8 +393,18 @@ pub fn get_property_ref_modes(
...
+    );
+    let mut set_in_ref_mode = RefMode::of(
+        env,
+        prop.typ,
+        library::ParameterDirection::In,
+        library::Transfer::None,
+    );
     if set_in_ref_mode == RefMode::ByRefMut {
         set_in_ref_mode = RefMode::ByRef;
```
MarijnS95, line 409: Probably not a sensible default.

src/analysis/ref_mode.rs
```diff
@@ -25,7 +26,7 @@ impl RefMode {
             ..
         }) = env.config.objects.get(&tid.full_name(library))
         {
-            if direction == library::ParameterDirection::In {
+            if direction == library::ParameterDirection::In && transfer != library::Transfer::Full {
                 return ref_mode;
             } else {
                 return RefMode::None;
```
MarijnS95, line 32: I've been thinking about removing this entirely and letting the code below handle it, but everything that's supposed to be `transfer-full` turns back into `&mut` references. Isn't `transfer-full` handled elsewhere, too, or is it okay to pass these as borrow in the first place??

src/analysis/child_properties.rs
sdroege: As sensible as any other :) There's no good default here.

src/analysis/ref_mode.rs
sdroege: `transfer full` and passing by reference has no relationship really. We pass everything by reference, always. It's only GStreamer's miniobjects that need special handling because the reference count there is connected to the mutability (like in Rust's `Arc`).

sdroege:
I'm not entirely sure what the rationale / reason for this change is, and judging from the GStreamer MR it's not doing the right thing all the time.

src/analysis/ref_mode.rs
MarijnS95: Then this indeed does not make much sense.

MarijnS95:
> I'm not entirely sure what the rationale / reason for this change is, and judging from the GStreamer MR it's not doing the right thing all the time.

In hindsight it seems easier to override the two misbehaving functions, either by changing the refmode on the parameters or the type (within the breaking crate) in its entirety. Indeed, if it's common to pass `transfer-full` by reference in Rust this is the wrong way to go about it.

sdroege:
> Indeed, if it's common to pass `transfer-full` by reference in Rust this is the wrong way to go about it.

That's how it is done everywhere else. It's just the GStreamer miniobjects that are a bit special :)

Ideally we wouldn't do that for other things either but a) for strings/etc we need to create a new copy anyway and b) for other things it's usually just a matter of increasing the refcount.

PR #1164, opened 2021-05-28
enums,flags: Always analyze manual types

MarijnS95:
Docs generation is going to look at analysis data to know what variants and functions are available on enums/flags.  External types (in the `manual = []` list) were previously omitted because of their status.  This changes enums and flags to behave like objects and records which only bail if the type is entirely ignored, and perform a `need_generate()` check at `codegen` instead.

Fixes: 50cee9e ("enums: Move rejection and glib imports from codegen to analysis")
Fixes: e549975 ("Copypaste enum analysis and function generation to bitfield")

---

CC @bilelmoussaoui, we're going to need this for #1161.

### DIFF
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -130,7 +130,7 @@ pub fn analyze<F: Borrow<library::Function>>(
     'func: for func in functions {
         let func = func.borrow();
         let configured_functions = obj.functions.matched(&func.name);
-        let mut status = GStatus::Generate;
+        let mut status = obj.status;
         for f in configured_functions.iter() {
             match f.status {
                 GStatus::Ignore => continue 'func,
--- a/src/analysis/enums.rs
+++ b/src/analysis/enums.rs
@@ -25,14 +25,7 @@ impl Info {
 pub fn new(env: &Env, obj: &GObject, imports: &mut Imports) -> Option<Info> {
     info!("Analyzing enumeration {}", obj.name);
 
-    if !obj.status.need_generate() {
-        return None;
-    }
-
-    if !obj
-        .type_id
-        .map_or(false, |tid| tid.ns_id == namespaces::MAIN)
-    {
+    if obj.status.ignored() {
         return None;
     }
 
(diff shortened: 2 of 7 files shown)


src/analysis/functions.rs
```diff
@@ -130,15 +130,14 @@ pub fn analyze<F: Borrow<library::Function>>(
...
                 GStatus::Ignore => continue 'func,
-                GStatus::Manual => {
-                    status = GStatus::Manual;
+                s => {
+                    status = s;
                     break;
                 }
-                GStatus::Generate => (),
             }
         }
```
MarijnS95, line 142: I'm not really sure what to do here. It is very counter-intuitive to have multiple configurations specifying a different mode. IMO we should assert that all statuses are equal. OTOH that might make it weird when using patterns to apply generic configuration, then later `manual/ignore` a few specific cases?

This now breaks or continues on the first item anyway, and clippy will tell us that this doesn't loop.

Because of using the first item (instead of the first encounter of `ignore = true`) this actually uncovers a little inconsistency in GStreamer's `Element` where the `"seek"` and `"seek_simple"` functions first receive a `bool_return_is_error`, then later down are reconfigured for `manual = true`.

MarijnS95: The best course of action is probably to revert this change as it's not necessary to fix the issue described in the first commit, though semantics change slightly if a parent object would have been set to manual, but a function set to generate (which is not something we support now, but maybe something to recognize in the future).

In addition, the config parser should check and warn/fail on duplicate `Ident::Name`.

bilelmoussaoui:
Can confirm this fixes various warnings on the current docs like
````
[WARN  libgir::codegen::doc::format] Constant/Flag variant/Enum member `GDK_SOURCE_TOUCHSCREEN` was not found
[WARN  libgir::codegen::doc::format] Constant/Flag variant/Enum member `PangoEllipsizeMode` was not found
[WARN  libgir::codegen::doc::format] Constant/Flag variant/Enum member `PANGO_ELLIPSIZE_NONE` was not found
[WARN  libgir::codegen::doc::format] Constant/Flag variant/Enum member `GDK_SOURCE_TABLET_PAD` was not found
```

even if the enums/bitfields are marked as manual.

sdroege (approved):

PR #1168, opened 2021-05-29
codegen/doc: Make extension trait docs link back to the type

MarijnS95:
The name is already there, but it isn't an intradoc-link yet.

### DIFF
--- a/src/codegen/doc/mod.rs
+++ b/src/codegen/doc/mod.rs
@@ -330,10 +330,9 @@ fn create_object_doc(w: &mut dyn Write, env: &Env, info: &analysis::object::Info
 
     if has_trait {
         write_item_doc(w, &ty_ext, |w| {
-            writeln!(w, "Trait containing all `{}` methods.", ty.name)?;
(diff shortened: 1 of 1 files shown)


src/codegen/doc/mod.rs
```diff
@@ -330,10 +330,9 @@ fn create_object_doc(w: &mut dyn Write, env: &Env, info: &analysis::object::Info
 
     if has_trait {
         write_item_doc(w, &ty_ext, |w| {
-            writeln!(w, "Trait containing all `{}` methods.", ty.name)?;
+            writeln!(w, "Trait containing all [`struct@{}`] methods.", ty.name)?;
 
-            let mut implementors = Some(info.type_id)
-                .into_iter()
+            let mut implementors = std::iter::once(info.type_id)
```
MarijnS95, line 335: Just a cleanup :)... But a more pressing question: do we want to repeat the type name right here, as it is already implied by the `Trait containing all {myself} methods` line?

sdroege (approved):

PR #1179, opened 2021-06-01
codegen/doc: Generate links for enum/flag functions/methods

MarijnS95:
Depends on #1178.

Add the missing links (only one in GStreamer) to associated functions and methods on enumerations and flags.

### DIFF
--- a/src/codegen/doc/format.rs
+++ b/src/codegen/doc/format.rs
@@ -356,6 +356,8 @@ fn find_method_or_function_by_ctype(
         |f| f.glib_name == name,
         |o| c_type.map_or(true, |t| o.c_type == t),
         |r| c_type.map_or(true, |t| r.type_(&env.library).c_type == t),
+        |r| c_type.map_or(true, |t| r.type_(&env.library).c_type == t),
+        |r| c_type.map_or(true, |t| r.type_(&env.library).c_type == t),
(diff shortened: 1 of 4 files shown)


bilelmoussaoui:
> Generate links for enum/flag functions/methods

Oh nice! I completely forgot about those as I don't think I have a use case for them in gtk4-rs.

bilelmoussaoui (approved):

src/codegen/doc/format.rs
```diff
@@ -401,12 +401,30 @@ pub(crate) fn find_method_or_function<
...
     {
-        Some(gen_record_fn_doc_link(
+        Some(gen_type_fn_doc_link(
             record_info.type_id,
             fn_info,
             env,
             in_type,
         ))
+    } else if let Some((enum_info, fn_info)) =
+        env.analysis.find_enum_by_function(env, |_| true, search_fn)
```
MarijnS95, line 411: Realized these should probably also get an appropriate type search function. Something for tomorrow to check before merging.

sdroege (approved):

src/codegen/doc/format.rs
MarijnS95: Ready now! Doesn't matter for normal docs where it's just a function name, but relevant for the new GI(?) docs where the type name is separately mentioned.

PR #1180, opened 2021-06-03
enum,flags: Take `version=` attribute in XML for members into account

MarijnS95:
`gir` XML contains `version=` attributes which are currently redefined manually in `Gir.toml`.  Propagate this info through to codegen to save all this redundant manual extra configuration.


MarijnS95:
I have some asserts locally to make sure no manual overrides exist when `gir` annotates it too. Is that something we want to have, perhaps comparing if the version is the same?

sdroege (approved):

PR #1182, opened 2021-06-10
Let cancellable be passed into GioFuture instead of creating it ourselves

sdroege:


sdroege:
See https://github.com/gtk-rs/gtk-rs-core/pull/136

sdroege:
CI here is expected to fail until everything is updated.

GuillaumeGomez (approved):

GuillaumeGomez:
That is really convenient, thanks for that!

PR #1185, opened 2021-06-13
Don't generate duplicated doc aliases

GuillaumeGomez:
You can see the results there: https://github.com/gtk-rs/gtk3-rs/compare/master...GuillaumeGomez:regen?expand=1

### DIFF
--- a/src/codegen/signal.rs
+++ b/src/codegen/signal.rs
@@ -36,12 +36,14 @@ pub fn generate(
     doc_hidden(w, analysis.doc_hidden, comment_prefix, indent)?;
     // Strip the "prefix" from "prefix::prop-name", if any.
     // Ex.: "notify::is-locked".
-    doc_alias(
-        w,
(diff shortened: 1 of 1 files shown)


bilelmoussaoui:
that's actually an expected behaviour

GuillaumeGomez:
Really? How so?

bilelmoussaoui:
> Really? How so?

The doc alias allows search for the `connect_X` by typing the signal name or the property name

GuillaumeGomez:
Yes, but here it is generated on both trait declaration and trait impl. It'll remain on trait declaration.

bilelmoussaoui:
> Yes, but here it is generated on both trait declaration and trait impl. It'll remain on trait declaration.

Ah I see what you mean now. Never mind, i should have been more careful looking at the generated code

bilelmoussaoui (approved):

GuillaumeGomez:
;)

Sending the regens!

PR #1186, opened 2021-06-14
docs: properly handle trait name for manual methods

bilelmoussaoui:
currently if a type has a unique method marked as a manual, has_trait would be set to false
which means the current type to be documented is assumed to be impl Type instead of trait TypeExtManual if the type is not final

This adds a check based on whether the type is final or not which fixes an issue with https://gtk-rs.org/gtk4-rs/git/docs/gdk4_wayland/prelude/trait.WaylandSurfaceExtManual.html

cc @MarijnS95


bilelmoussaoui:
Oh, i found some functions that are not handled correctly after this change. I will give it more thought tomorrow :)

PR #1187, opened 2021-06-15
docs: don't generate for properties with getter/setter

bilelmoussaoui:
Some properties have already a getter/setter already and there's no need to try to document them
based on the property documentation. This removes the extra information dumped in the docs.md file in order to make finding missing/undocumented types easier


sdroege (approved):
You want to run `cargo fmt` but otherwise looks good to me :)

bilelmoussaoui:
Should be fixed now

PR #1189, opened 2021-06-17
Rust 1.53: Use the new nested or_patterns

MarijnS95:
https://blog.rust-lang.org/2021/06/17/Rust-1.53.0.html :shark: 

Previously `|` was only allowed in the top-level of patterns, requiring the circumfering enum variant (usually something wrapped in `Some`, and in `gir` `Type::Fundamental` is common) to be repeated for every inner variant that's matched against.  Not anymore, since 1.53 it is now possible to write `|` within a single variant instead of repeating it many times.

### DIFF
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -167,9 +167,7 @@ impl Bounds {
     pub fn type_for(env: &Env, type_id: TypeId, nullable: Nullable) -> Option<BoundType> {
         use self::BoundType::*;
         match env.library.type_(type_id) {
-            Type::Fundamental(Fundamental::Filename) => Some(AsRef(None)),
-            Type::Fundamental(Fundamental::OsString) => Some(AsRef(None)),
-            Type::Fundamental(Fundamental::Utf8) if *nullable => None,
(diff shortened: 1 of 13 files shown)


src/analysis/bounds.rs
```diff
@@ -167,9 +167,7 @@ impl Bounds {
     pub fn type_for(env: &Env, type_id: TypeId, nullable: Nullable) -> Option<BoundType> {
         use self::BoundType::*;
         match env.library.type_(type_id) {
-            Type::Fundamental(Fundamental::Filename) => Some(AsRef(None)),
-            Type::Fundamental(Fundamental::OsString) => Some(AsRef(None)),
-            Type::Fundamental(Fundamental::Utf8) if *nullable => None,
```
sdroege, line 172: You dropped this branch

sdroege (commented):
Seems good to me otherwise, much easier to read.

src/analysis/bounds.rs
MarijnS95: Yes intentionally, because there's an identical branch below (or so I remember, cannot open unchanged lines on the GH mobile app).

sdroege: Indeed

sdroege (approved):

PR #1191, opened 2021-06-18
Generate warnings for code examples

### LINKED ISSUE #1190: Generate warnings in docs before C code examples

GuillaumeGomez:
Fixes #1190.

### DIFF
--- a/src/codegen/doc/format.rs
+++ b/src/codegen/doc/format.rs
@@ -71,7 +71,10 @@ fn code_blocks_transformation(
 fn get_language<'a>(entry: &'a str, out: &mut String) -> &'a str {
     if let (_, Some(after)) = try_split(entry, LANGUAGE_SEP_BEGIN) {
         if let (before, Some(after)) = try_split(after, LANGUAGE_SEP_END) {
-            out.push_str(&format!("\n```{}", before));
+            if ["text", "rust"].contains(&before) {
+                out.push_str(&format!("\n\n**⚠️ The following code is in {} ⚠️**", before));
+            }
+            out.push_str(&format!("\n\n```{}", before));
             return after;
         }
     }
@@ -104,6 +107,9 @@ fn format(
                 ret.push_str(&replace_symbols(before, env, in_type));
 
                 let (lang, after) = get_markdown_language(after);
+                if ["text", "rust"].contains(&lang) && after.lines().count() > 1 {
+                    ret.push_str(&format!("**⚠️ The following code is in {0} ⚠️**\n\n", lang));
+                }
                 ret.push_str(&format!("```{}\n", lang));
 
                 if let (before, Some(after)) = try_split(after, "```") {


GuillaumeGomez:
Oh that's crazy, my clippy lint is now showing up! :D

src/codegen/doc/format.rs
```diff
@@ -71,7 +71,10 @@ fn code_blocks_transformation(
 fn get_language<'a>(entry: &'a str, out: &mut String) -> &'a str {
     if let (_, Some(after)) = try_split(entry, LANGUAGE_SEP_BEGIN) {
         if let (before, Some(after)) = try_split(after, LANGUAGE_SEP_END) {
-            out.push_str(&format!("\n```{}", before));
+            out.push_str(&format!(
+                "\n\n**⚠️ The following code is in {0} ⚠️**\n\n```{0}",
```
bilelmoussaoui, line 75: I don't think it's needed here. Because often those might be just plain types or functions. Worth trying it locally

GuillaumeGomez: I'll add a check for the language itself, but even so, still needs to be marked as such. ;)

bilelmoussaoui: But those can be sometimes replaced by the docs links no?

GuillaumeGomez: Wait I'm lost now. In which case a \`\`\` code block can be replaced by a link?

src/codegen/doc/format.rs
```diff
@@ -71,7 +71,10 @@ fn code_blocks_transformation(
 fn get_language<'a>(entry: &'a str, out: &mut String) -> &'a str {
     if let (_, Some(after)) = try_split(entry, LANGUAGE_SEP_BEGIN) {
         if let (before, Some(after)) = try_split(after, LANGUAGE_SEP_END) {
-            out.push_str(&format!("\n```{}", before));
+            if before != "text" && before != "rust" {
```
sdroege, line 74:
```suggestion
            if !["text", "rust"].contains(before) {
```

src/codegen/doc/format.rs
```diff
@@ -104,6 +107,9 @@ fn format(
                 ret.push_str(&replace_symbols(before, env, in_type));
 
                 let (lang, after) = get_markdown_language(after);
+                if lang != "text" {
```
sdroege, line 110: Why not check for Rust here?

src/codegen/doc/format.rs
sdroege: I guess when it's a one-line example?

Can you create a gtk-rs PR with this? Then we can check the generated docs :)

GuillaumeGomez: We don't have the docs in the different repositories, so it wouldn't change anything.

sdroege: It's built as part of the github pages

GuillaumeGomez: It would require to merge the PR for that. :)

sdroege: Are you sure? It should be available under https://guillaumegomez.github.io/gtk-rs-core/ and similar.

GuillaumeGomez: Not sure about that. :-/

GuillaumeGomez:
Updated!

sdroege (approved):

SafariMonkey:
@GuillaumeGomez I just looked at this change and it looks like the negation on the language check was dropped? I'm guessing as part of applying [this](https://github.com/gtk-rs/gir/pull/1191#discussion_r654891668) change. Currently it appears to me that it only shows the warning before `text` or `rust` blocks, which I'm pretty sure is the opposite of the intention.

GuillaumeGomez:
From the diff, I don't see how it's possible. :-/

sdroege:
> From the diff, I don't see how it's possible. :-/

Because you forgot the `!` :)

GuillaumeGomez:
Oh crap...

PR #1207, opened 2021-08-05
Added self-linking support to sys mode

xanathar:
The purpose of this is to support a scenario where a single module is half implemented in C and half implemented in Rust (statically linked, resolving circular dependencies at the linker level).
This, while not ideal, is a likely possibility for a codebase which is in the process of migrating from C to Rust. (Note: caveats apply to the execution of `cargo test`, but a stub can be linked in that case to allow unit tests to run).

In this scenario, the sys bindings are generated correctly, but with a few caveats that need to be corrected:
- The `#[link(...)]` attribute should not be emitted as there is no external library to link with... yet (the linker is linking everything in that precise moment)
- The `build.rs` should not be generated for the same reason
- The `tests/` directory does no harm, but is superflous at that point

This adds a `self_linking` configuration key to the `sys` work mode that would skip the above parts, in order to support the self linking scenario I depicted without having to resort to post modification of the generated code.

I'm not sure if this fits the intended purpose of the gir tool, but I think it can be possibly useful to others.


pbor:
We use Marco's patch in our project at work. The use case is the one Marco described, we have library we are incrementally porting to rust and we want to bind some of its API since the parts that we are migrating to rust need to call into the C part, however we want to keep generating a single .so/.dll.

sdroege:
Yeah I meant something public that we can link to for others to understand how this can be used :)

book/src/config_ffi.md
```diff
@@ -35,6 +35,13 @@ work_mode = "sys"
 # and build.rs that generated only if not exists.
 # Defaults to false
 split_build_rs = false
+# If true, then build.rs and tests do not get generated, and
+# no #[link] attribute is emitted. This is supposed to be used
+# when a library consists of an hybrid of C and Rust code, and
```
MarijnS95, line 40:
```suggestion
# when a library consists of a hybrid of C and Rust code, and
```

MarijnS95:
This sounds very similar to what I did for `windows-rs` in https://github.com/microsoft/windows-rs/pull/1863, allowing some Rust code to request `extern "system" fn(...)` functions that are then provided by some other crate/source, also in Rust :tada:

xanathar:
> Yeah I meant something public that we can link to for others to understand how this can be used :)

I'll try to resurrect this and provide the example, sorry for the huge delay.

book/src/config_ffi.md
xanathar: Fixed!

PR #1210, opened 2021-08-23
Remove the `links` annotations

nagisa:
The `links` annotation in `Cargo.toml` is intended to ensure that in the
crate graph there's at most one crate that's an implementation of
some sort concept.

This can make sense in some scenarios, most prominent of which is when
the crate defines `#[no_mangle]` symbols (e.g. by compiling a vendored C
library.) In that situation linking a binary that depends on two
versions of the library cannot work because of colliding symbol names.

There does not appear to be a similar reason to impose such a
restriction on the users of gir-generated bindings, however. All of
the generated crates link to a system library, they do not define any
`#[no_mangle]` symbols nor they vendor/build C libraries as part of
their build process.  Most likely all the different versions of the
bindings will link to the exact same system library too.

### DIFF
--- a/src/codegen/sys/cargo_toml.rs
+++ b/src/codegen/sys/cargo_toml.rs
@@ -36,11 +36,6 @@ fn fill_empty(root: &mut Table, env: &Env, crate_name: &str) {
         let package = upsert_table(root, "package");
         set_string(package, "name", package_name);
         set_string(package, "version", "0.0.1");
-        set_string(
-            package,
-            "links",
-            nameutil::shared_libs_to_links(&env.namespaces.main().shared_libs),
-        );
         set_string(package, "edition", "2018");
     }
 
--- a/src/nameutil.rs
+++ b/src/nameutil.rs
@@ -139,17 +139,6 @@ pub fn shared_lib_name_to_link_name(name: &str) -> &str {
     s
 }
 
-pub fn shared_libs_to_links(shared_libs: &[String]) -> String {
-    // https://github.com/rust-lang/cargo/issues/4533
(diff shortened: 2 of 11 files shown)


sdroege (commented):
I think this only doesn't add it, but when running `gir` on an existing crate it wouldn't remove the `links` annotation automatically.

src/codegen/sys/cargo_toml.rs
```diff
@@ -36,11 +36,6 @@ fn fill_empty(root: &mut Table, env: &Env, crate_name: &str) {
         let package = upsert_table(root, "package");
         set_string(package, "name", package_name);
         set_string(package, "version", "0.0.1");
-        set_string(
-            package,
-            "links",
-            nameutil::shared_libs_to_links(&env.namespaces.main().shared_libs),
-        );
         set_string(package, "edition", "2018");
```
sdroege, line 39: That would require to actually remove it here AFAIU

nagisa: Right, I was thinking that I'd just go through the crates that I care about and submit a PR manually…

GuillaumeGomez: I'm not sure if removing things present in an existing Cargo file is a good idea though...

sdroege: OK, if both of you think that's OK then let's do that :)

GuillaumeGomez:
I'm sending a PR for clippy, just hold on! :)

sdroege:
CI should pass after rebasing.

PR #1217, opened 2021-09-02
Config for nullable callback param

### LINKED ISSUE #1209: Add support to update callback parameters
For now, only signal callback parameters can be update through the `Gir.toml` file. It'd be nice to extend it to function with callbacks.

Something like:

```toml
    [[object.function]]
    name = "function_name"
        [[object.function.parameter]]
        name = "parameter_name"
             [[object.function.parameter.parameter]]
             name = "callback_parameter_name"
             nullable = false # just for the example
```

GuillaumeGomez:
Fixes #1209.

GTK3 PR there: https://github.com/gtk-rs/gtk3-rs/compare/master...GuillaumeGomez:nullable-callback-param?expand=1

### DIFF
--- a/src/analysis/rust_type.rs
+++ b/src/analysis/rust_type.rs
@@ -1,6 +1,7 @@
 use super::conversion_type::ConversionType;
 use crate::{
     analysis::{ref_mode::RefMode, try_from_glib::TryFromGlib},
+    config::functions::{CallbackParameter, CallbackParameters},
     env::Env,
     library::{self, Nullable, ParameterDirection, ParameterScope},
     nameutil::{is_gstring, use_glib_type},
@@ -186,6 +187,7 @@ pub struct RustTypeBuilder<'env> {
     scope: ParameterScope,
     concurrency: library::Concurrency,
     try_from_glib: TryFromGlib,
+    callback_parameters_config: CallbackParameters,
 }
 
 impl<'env> RustTypeBuilder<'env> {
@@ -199,6 +201,7 @@ impl<'env> RustTypeBuilder<'env> {
             scope: ParameterScope::None,
             concurrency: library::Concurrency::None,
             try_from_glib: TryFromGlib::default(),
+            callback_parameters_config: Vec::new(),
         }
     }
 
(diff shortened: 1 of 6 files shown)


GuillaumeGomez:
Note that for some reason, the bound generation isn't using the information from the function parameters but instead re-compute it.

src/analysis/trampoline_parameters.rs
```diff
@@ -35,8 +35,8 @@ impl CParameter {
 
 #[derive(Clone, Debug)]
 pub struct Transformation {
-    pub ind_c: usize,    //index in `Vec<CParameter>`
-    pub ind_rust: usize, //index in `Vec<RustParameter>`
+    pub ind_c: usize,    // index in `Vec<CParameter>`
+    pub ind_rust: usize, // index in `Vec<RustParameter>`
```
MarijnS95, line 39: Make this doc comments? Maybe with intradoc link?

src/analysis/rust_type.rs
```diff
@@ -232,6 +235,14 @@ impl<'env> RustTypeBuilder<'env> {
         self
     }
 
+    pub fn callback_parameters_config(
+        mut self,
+        callback_parameters_config: &[CallbackParameter],
+    ) -> Self {
+        self.callback_parameters_config = callback_parameters_config.to_owned();
```
sdroege, line 242: Why not apply the configuration to the builder here, but instead copy the configuration just to apply it at a later time?

sdroege (commented):
Seems OK to me otherwise

src/analysis/trampoline_parameters.rs
GuillaumeGomez: I simply added a whitespace, we have equivalents a bit everywhere. Don't hesitate to send a PR to turn them into doc comments. :)

src/analysis/rust_type.rs
GuillaumeGomez: What do you mean?

sdroege: You're only storing the configuration in the builder here instead of changing the state of the builder (i.e. the `nullable` field)

sdroege: Oh

PR #1222, opened 2021-09-17
Add default impls for all objects

### LINKED ISSUE #1221: Impl Default trait for all widgets that support the builder pattern
Some widgets like `gtk::Box` or `gtk::ApplicationWindow` don't implement the `Default` trait although you can simple use `Widget::builder().build()` to get a default widget. Currently, this workaround based on the builder pattern is used by Relm4 to get more default configurations for its widget macro with the [`DefaultWidget` trait](https://aaronerhardt.github.io/docs/relm4/relm4/util/default_widgets/trait.DefaultWidget.html).

@Hofer-Julian

AaronErhardt:
Fix #1221

Currently, `Default` is only implemented for objects that don't take parameters in their `new()` function. This PR adds `Default` implementations for all objects that take parameters in their `new()` function. This makes the workaround  using `Object::builder().build()` obsolete.

The generated code looks like this:

```rust
impl Default for Box {
    fn default() -> Self {
        match glib::object::Object::new::<Self>(&[]) {
            Ok(obj) => obj,
            Err(err) => panic!(
                "Can't construct Box object with default parameters: {}",
                err
            ),
        }
    }
}
```

### DIFF
--- a/src/codegen/general.rs
+++ b/src/codegen/general.rs
@@ -744,7 +744,7 @@ pub fn declare_default_from_new(
     env: &Env,
     name: &str,
     functions: &[analysis::functions::Info],
-    is_object: bool,
+    has_builder: bool,
 ) -> Result<()> {
     if let Some(func) = functions.iter().find(|f| {
         !f.visibility.hidden()
@@ -756,23 +756,29 @@ pub fn declare_default_from_new(
         if func.parameters.rust_parameters.is_empty() {
             writeln!(w)?;
             version_condition(w, env, func.version, false, 0)?;
-            writeln!(w, "impl Default for {} {{", name)?;
-            writeln!(w, "    fn default() -> Self {{")?;
-            writeln!(w, "        Self::new()")?;
-            writeln!(w, "    }}")?;
-            writeln!(w, "}}")?;
-        } else if is_object {
+            writeln!(
+                w,
+                "impl Default for {} {{
+                     fn default() -> Self {{
+                         Self::new()
+                     }}
+                 }}",
+                name
+            )?;
+        } else if has_builder {
             // create an alternative default implementation the uses `glib::object::Object::new()`
             writeln!(w)?;
             version_condition(w, env, func.version, false, 0)?;
-            writeln!(w, "impl Default for {} {{", name)?;
-            writeln!(w, "    fn default() -> Self {{")?;
-            writeln!(w, "        match glib::object::Object::new::<Self>(&[]) {{")?;
-            writeln!(w, "            Ok(obj) => obj,")?;
-            writeln!(w, "            Err(err) => panic!(\"Can't construct {} object with default parameters: {{}}\", err),", name)?;
-            writeln!(w, "        }}")?;
-            writeln!(w, "    }}")?;
-            writeln!(w, "}}")?;
+            writeln!(
+                w,
+                "impl Default for {} {{
+                     fn default() -> Self {{
+                         match glib::object::Object::new::<Self>(&[]) {{
+                              Ok(obj) => obj,
+                              Err(err) => panic!(\"Can't construct {} object with default parameters: {{}}\", err),
+                         }}
+                     }}
+                 }}", name, name)?;
         }
     }
 
--- a/src/codegen/general.rs
+++ b/src/codegen/general.rs
@@ -767,7 +767,10 @@ pub fn declare_default_from_new(
             version_condition(w, env, func.version, false, 0)?;
             writeln!(w, "impl Default for {} {{", name)?;
             writeln!(w, "    fn default() -> Self {{")?;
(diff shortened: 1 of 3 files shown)


src/codegen/general.rs
```diff
@@ -744,22 +744,36 @@ pub fn declare_default_from_new(
...
+            writeln!(w, "impl Default for {} {{", name)?;
+            writeln!(w, "    fn default() -> Self {{")?;
+            writeln!(w, "        Self::new()")?;
+            writeln!(w, "    }}")?;
+            writeln!(w, "}}")?;
+        } else if is_object {
+            // create an alternative default implementation the uses `glib::object::Object::new()`
+            writeln!(w)?;
+            version_condition(w, env, func.version, false, 0)?;
+            writeln!(w, "impl Default for {} {{", name)?;
```
GuillaumeGomez, line 768: Please make these multiple `writeln!` into one, it'll make the reading easier and nicer.

src/codegen/general.rs
```diff
@@ -744,22 +744,36 @@ pub fn declare_default_from_new(
...
-        version_condition(w, env, func.version, false, 0)?;
-        writeln!(w, "impl Default for {} {{", name)?;
-        writeln!(w, "    fn default() -> Self {{")?;
-        writeln!(w, "        Self::new()")?;
-        writeln!(w, "    }}")?;
-        writeln!(w, "}}")?;
+        if func.parameters.rust_parameters.is_empty() {
+            writeln!(w)?;
+            version_condition(w, env, func.version, false, 0)?;
+            writeln!(w, "impl Default for {} {{", name)?;
```
GuillaumeGomez, line 759: Same for this one.

GuillaumeGomez:
Can you generate gtk-rs (or any other crate using gir and this feature) so we can see the end result alongside this PR please?

sdroege:
> This PR adds `Default` implementations for all objects that take parameters in their `new()` function.

This should only be done for objects where a `Builder` is generated.

AaronErhardt:
@GuillaumeGomez the result: https://github.com/AaronErhardt/gtk4-rs/tree/default-test
@sdroege should be fixed now

GuillaumeGomez:
The compare URL is [here](https://github.com/gtk-rs/gtk4-rs/compare/master...AaronErhardt:default-test?expand=1). In a lot of cases, we already have `#[derive(Default)]`. I assume that for these cases, implementing `Default` manually is unneeded.

AaronErhardt:
I don't think so. There can only be one trait implementation yet gtk4-rs compiles just fine with those changes. I think the `#[derive(Default)]` is only for the associated builder because otherwise the compiler would throw an error.

GuillaumeGomez:
You're right, sorry, confused the builders and the structs...

src/codegen/general.rs
```diff
@@ -744,22 +744,42 @@ pub fn declare_default_from_new(
...
+            // create an alternative default implementation the uses `glib::object::Object::new()`
+            writeln!(w)?;
+            version_condition(w, env, func.version, false, 0)?;
+            writeln!(
+                w,
+                "impl Default for {} {{
+                     fn default() -> Self {{
+                         match glib::object::Object::new::<Self>(&[]) {{
+                              Ok(obj) => obj,
+                              Err(err) => panic!(\"Can't construct {} object with default parameters: {{}}\", err),
```
sdroege, line 778: Why not just `expect()` here instead? :)

sdroege: This can't really fail FWIW. The only way how this could fail is if there's more broken inside the bindings.

sdroege (commented):
Looks good to me otherwise

src/codegen/general.rs
AaronErhardt: Because I thought I'd be nice to have the underlying glib error printed in the panic message as well. And `expect()` just takes a string.

GuillaumeGomez: I like it like this actually. ^^'

sdroege: `expect()` is also printing the underlying error

AaronErhardt: Alright, should I use expect then or is it the current version okay?

GuillaumeGomez: I'll leave the final decision to @sdroege, I'm fine with both.

sdroege: `expect()` is less code, especially for an error that can't really happen.

AaronErhardt:
Now everything should work as expected :wink: 
New diff: https://github.com/gtk-rs/gtk4-rs/compare/master...AaronErhardt:default-test

sdroege (approved):

sdroege:
@AaronErhardt Can you send PRs for gtk-rs-core, gtk3-rs and gtk4-rs with this update? Thanks!

PR #1225, opened 2021-09-26
[Fix] Implement default for all objects with builders

AaronErhardt:
When updating Relm4 to use the new `Default` implementations in gtk4-rs (see https://github.com/gtk-rs/gir/pull/1222) I've noticed a regression because widgets such as `MessageDialog` still don't implement `Default` despite having a builder. 

This PR fixes this problem by adding a `Default` implementation for all objects with a builder pattern.

### DIFF
--- a/src/codegen/general.rs
+++ b/src/codegen/general.rs
@@ -739,7 +739,7 @@ pub fn write_vec<T: Display>(w: &mut dyn Write, v: &[T]) -> Result<()> {
     Ok(())
 }
 
-pub fn declare_default_from_new(
+pub fn implement_default_from_new_or_builder(
     w: &mut dyn Write,
     env: &Env,
     name: &str,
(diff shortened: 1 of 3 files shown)


AaronErhardt:
Interesting CI results... Is there a reason why `new` and `Default` are implemented manually in the case of `RadioButton`?

sdroege:
> Interesting CI results... Is there a reason why `new` and `Default` are implemented manually in the case of `RadioButton`?

Is the automatic implementation the same? Then it's probably just older than `gir` being able to autogenerate it.

AaronErhardt:
Well, the `Default` implementation added is not the same because gir doesn't know that it can use the manual `new` function. 
Yet I'm pretty sure gir would generate an basically identical `new` function and then an identical `Default` implementation if `RadioButton::new()` wasn't marked as manual.

sdroege:
What's the difference? Maybe it was manual because gir couldn't do it back then?

AaronErhardt:
The difference is that gir would now use `glib::Object::new()` instead of simply `new` because it doesn't know about the manual `new` function.

How does gir know if something is done manually or not? Maybe we can just try out whether gir can now handle `RadioButton`.

AaronErhardt:
Initially, I thought this was some issue with gir. But as it turns out this is a problem with manually implemented `new` functions that don't have a manual `Default` implementation. I think it makes sense to either reject this PR and I'll implement `Default` for the remaining widgets manually or we update gir so it doesn't need the manual parts anymore.

sdroege:
See the `Gir.toml`:

```toml
[[object]]
name = "Gtk.RadioButton"
status = "generate"
generate_builder = true
    [[object.function]]
    pattern = ".+_from_widget"
        [[object.function.parameter]]
        name = "radio_group_member"
        nullable = false
    [[object.function]]
    name = "new"
    manual = true
    [[object.function]]
    name = "new_with_label"
    manual = true
    [[object.function]]
    name = "new_with_mnemonic"
    manual = true
    [[object.function]]
    name = "set_group"
    ignore = true
    [[object.property]]
    name = "group"
    ignore = true
```

They call all be autogenerated if that's removed. That they're ignored comes from https://github.com/gtk-rs/gtk3-rs/commit/580e202fe24 , which comes from https://github.com/gtk-rs/gtk/issues/488 and has all the explanation.

FWIW, gir can't know that the manual `new` implementation actually has no parameters.

sdroege:
> I'll implement `Default` for the remaining widgets manually

Which other widgets are that other than the radio button?

> or we update gir so it doesn't need the manual parts anymore.

In case of the radio button that's not possible.

AaronErhardt:
> Which other widgets are that other than the radio button?

The radio button isn't the problem. It has a manual `Default` implementation. But other widgets such as `MessageDialog` in gtk4-rs don't despite having a builder because they have a manual `new` function and currently gir only adds `Default` implementations if it added a `new` function as well.

sdroege:
So you'll have to check why they have a manual `new()` function and maybe they can be autogenerated, otherwise they need a manual `Default` impl too now :)

AaronErhardt:
I think it's probably better to use a manual `Default` implementation for functions with a manual constructor. It only affects very few widgets anyway.

PR #1228, opened 2021-09-28
Always output git short revisions with 12 characters

sdroege:
Otherwise git will use some heuristics and things might be different for
different environments.


MarijnS95 (approved):

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #1229, opened 2021-09-28
Remove type version heuristics

sdroege:
Previously we would infer the minimum version of a type by the minimum
version of all its functions if not type version information was
available. This causes unnecessarily high minimum version constraints if
the type was available without functions before and later got functions.

Instead this should be fixed in the C library's annotations, and for now
this can also be overridden via Gir.toml if needed.

### DIFF
--- a/src/analysis/info_base.rs
+++ b/src/analysis/info_base.rs
@@ -1,6 +1,5 @@
-use super::{functions::Visibility, imports::Imports, *};
-use crate::{config::gobjects::GObject, env::Env, library, version::Version};
-use std::cmp;
+use super::{imports::Imports, *};
+use crate::{library, version::Version};
 
 #[derive(Debug, Default)]
 pub struct InfoBase {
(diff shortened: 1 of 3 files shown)


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

sdroege:
I'll send PRs for the other repos once this is merged. Nothing's going to change there apart from Gir.toml.

sdroege:
And will also send MRs for the C libraries in question.

MarijnS95:
Awesome! I remember previously having problems with that in at least https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/commit/d596ea45c5a8902a312873243e1c9a7776d2e075, please fix the version annotation for that type too before reverting the commit :)

MarijnS95:
@sdroege I guess the plan for now is to remove/disregard any `version=` override directly on `[[object]]`, see what the diff/fallout is and fix as much as possible upstream in C files, while anntotating the rest of the overrides with a comment (which I conventiently forgot in linked commit, making it harder to find :triumph:

PR #1230, opened 2021-09-28
Handle objects correctly where a supertype has a newer version than the type itself

sdroege:
In that case it is necessary to create multiple glib::wrapper! calls
    with the different sets of the supertypes.

----

Also a fix for the `Version` type in general. This has no effect at all on the gtk-rs crates but in GStreamer has this important diff:

```diff
diff --git a/gstreamer-editing-services/src/auto/uri_clip_asset.rs b/gstreamer-editing-services/src/auto/uri_clip_asset.rs
index 1b55c5b8..28c5757c 100644
--- a/gstreamer-editing-services/src/auto/uri_clip_asset.rs
+++ b/gstreamer-editing-services/src/auto/uri_clip_asset.rs
@@ -1,11 +1,13 @@
 // This file was generated by gir (https://github.com/gtk-rs/gir)
 // from gir-files (https://github.com/gtk-rs/gir-files)
 // from gst-gir-files (https://gitlab.freedesktop.org/gstreamer/gir-files-rs.git)
 // DO NOT EDIT
 
 use crate::Asset;
 use crate::ClipAsset;
 use crate::MetaContainer;
+#[cfg(any(feature = "v1_18", feature = "dox"))]
+#[cfg_attr(feature = "dox", doc(cfg(feature = "v1_18")))]
 use crate::SourceClipAsset;
 use crate::UriSourceAsset;
 use glib::object::Cast;
@@ -21,6 +23,8 @@ use std::boxed::Box as Box_;
 use std::mem::transmute;
 use std::ptr;
 
+#[cfg(any(feature = "v1_18", feature = "dox"))]
+#[cfg_attr(feature = "dox", doc(cfg(feature = "v1_18")))]
 glib::wrapper! {
     #[doc(alias = "GESUriClipAsset")]
     pub struct UriClipAsset(Object<ffi::GESUriClipAsset, ffi::GESUriClipAssetClass>) @extends SourceClipAsset, ClipAsset, Asset, @implements MetaContainer;
@@ -30,6 +34,16 @@ glib::wrapper! {
     }
 }
 
+#[cfg(not(any(feature = "v1_18", feature = "dox")))]
+glib::wrapper! {
+    #[doc(alias = "GESUriClipAsset")]
+    pub struct UriClipAsset(Object<ffi::GESUriClipAsset, ffi::GESUriClipAssetClass>) @extends ClipAsset, Asset, @implements MetaContainer;
+
+    match fn {
+        type_ => || ffi::ges_uri_clip_asset_get_type(),
+    }
+}
+
 impl UriClipAsset {
     //#[cfg(any(feature = "v1_16", feature = "dox"))]
     //#[cfg_attr(feature = "dox", doc(cfg(feature = "v1_16")))]
```

@bilelmoussaoui I remember there was something in GTK4 where an interface impl was added at a later time? You might want to add configuration for that to Gir.toml now. That can then be directly hooked into the place where I added a `TODO` comment :)

### DIFF
--- a/src/analysis/object.rs
+++ b/src/analysis/object.rs
@@ -113,12 +113,15 @@ pub fn class(env: &Env, obj: &GObject, deps: &[library::TypeId]) -> Option<Info>
 
     let klass: &library::Class = type_.maybe_ref()?;
 
+    let version = obj.version.or(klass.version);
+    let deprecated_version = klass.deprecated_version;
+
     let mut imports = Imports::with_defined(&env.library, &name);
     if obj.generate_display_trait {
         imports.add("std::fmt");
     }
 
-    let supertypes = supertypes::analyze(env, class_tid, &mut imports);
+    let supertypes = supertypes::analyze(env, class_tid, version, &mut imports);
 
     let final_type = klass.final_type;
     let trait_name = obj
@@ -176,9 +179,6 @@ pub fn class(env: &Env, obj: &GObject, deps: &[library::TypeId]) -> Option<Info>
 
     let builder_properties =
         class_builder::analyze(env, &klass.properties, class_tid, obj, &mut imports);
-
-    let version = obj.version.or(klass.version);
-    let deprecated_version = klass.deprecated_version;
 
     let child_properties =
         child_properties::analyze(env, obj.child_properties.as_ref(), class_tid, &mut imports);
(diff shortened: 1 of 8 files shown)


bilelmoussaoui:
> @bilelmoussaoui I remember there was something in GTK4 where an interface impl was added at a later time? You might want to add configuration for that to Gir.toml now. That can then be directly hooked into the place where I added a TODO comment :)

Yes, we needed that for GtkStackSidebar that implements Gtk.Orientable in gtk 4.4 iirc. I will have a look at it
cc @elmarco

sdroege:
> Wouldn't it be better to extend the macro first instead? Having the same macro called multiple times seems a bit too much to me...

How would you do that without completely changing how the macro works? I looked at that and considered it an immense amount of work.

Like this the code changes are minimal and at compilation-time at most one of the `glib::wrapper!` calls is going to be evaluated.

GuillaumeGomez:
We could add an optional `$(#[$(attr:attr)+])*` in the macro, it wouldn't change much. Having the same macro called multiple times makes it harder to see which supertype was added.

sdroege:
That doesn't seem easy and will require far more code changes than this. And so far we're aware of two cases where this is needed

sdroege:
Sure, good luck :)

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

Issue #1205, opened 2021-07-27
Special case when a bool is returned instead of a gboolean


bilelmoussaoui:
Currently, gir assumes that a gboolean is used without checking the `c:type`. 

See https://github.com/ebassi/graphene/issues/234#issuecomment-887259734

Ideally, it would check and based on whether it's a gint or not, uses FromGlib/IntoGlib traits

sdroege:
It should be relatively easy to handle in gir: it would be handled like `i32` / `gint`, just that we'd have `bool` instead. The main question is how to detect this reliable from the .gir file, what the heuristic should be.

PR #1232, opened 2021-10-07
Special case C _Bool by treating it like a Rust bool instead of gboolean

sdroege:
Fixes https://github.com/gtk-rs/gir/issues/1205


bilelmoussaoui:
Nice, thanks for looking into it ! Mind submitting a PR against gtk-rs-core to see how the re-generated graphene looks like?

sdroege:
On it

sdroege:
See https://github.com/gtk-rs/gtk-rs-core/pull/282. Also comes with a test that fails before and succeeds after.

GuillaumeGomez:
Thanks!

PR #1234, opened 2021-10-09
Implement support for inline allocated Boxed types

sdroege:
Useful for graphene and code reduction in various other places.

See https://github.com/gtk-rs/gtk-rs-core/issues/211


sdroege:
See https://github.com/gtk-rs/gtk-rs-core/pull/284 for the code making use of this.

PR #1237, opened 2021-10-26
analysis: mark to_str functions as renamed

bilelmoussaoui:
Instead of overriding their original names; this makes the docs work without workarounds
by marking them as renamed manually like in https://github.com/gtk-rs/gtk4-rs/pull/656

### DIFF
--- a/src/analysis/special_functions.rs
+++ b/src/analysis/special_functions.rs
@@ -103,7 +103,8 @@ fn is_stringify(func: &mut FuncInfo, parent_type: &LibType, obj: &GObject) -> bo
 
         if func.name == "to_string" {
             // Rename to to_str to make sure it doesn't clash with ToString::to_string
-            func.name = "to_str".to_owned();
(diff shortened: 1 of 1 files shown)


sdroege:
@bilelmoussaoui This broke gstreamer-rs :)

```
thread 'main' panicked at 'A `to_string` function can't be renamed manually. It's automatically renamed to `to_str`', src/analysis/special_functions.rs:106:13
```

bilelmoussaoui:
@sdroege not really, you just need to drop https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/blob/master/gstreamer-video/Gir.toml#L545

bilelmoussaoui:
as it's already marked as renamed automatically for you

MarijnS95:
We'll have to keep the manual setup though, but the `rename =` can be dropped and maybe the return too?

bilelmoussaoui:
> We'll have to keep the manual setup though, but the `rename =` can be dropped and maybe the return too?

Yes, sorry I wanted to link to the rename only

sdroege:
Oh I see, thanks

PR #1245, opened 2021-10-29
Fix support for nullable paths / OsStrings

sdroege:
See https://github.com/gtk-rs/gtk-rs-core/issues/187

This is not complete yet.

### DIFF
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -10,9 +10,7 @@ use crate::{
     config,
     consts::TYPE_PARAMETERS_START,
     env::Env,
-    library::{
-        Class, Concurrency, Function, Fundamental, Nullable, ParameterDirection, Type, TypeId,
-    },
+    library::{Class, Concurrency, Function, Fundamental, ParameterDirection, Type, TypeId},
     traits::IntoString,
 };
 use std::{collections::vec_deque::VecDeque, slice::Iter};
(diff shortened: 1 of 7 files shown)


sdroege:
Adding strings into the mix here (`AsRef<str>` instead of `&str`) shows up lots of other strange special cases all over the `gir` code for strings that look like they should be handled more centrally.

Taking a step back here, is it actually worthwhile to have `impl AsRef<Path>` parameters instead of `&Path`? Or `impl AsRef<str>` instead of `&str`?

GuillaumeGomez:
Depends. `AsRef<str>` seems a bit weird to me, however `AsRef<Path>` is pretty nice.

sdroege:
How is it any different? :) Why is one nice and the other not, in which cases does it actually simplify/improve any code? Either avoids an `&` at the caller at the cost of hiding that the function does not actually make use of owned values that are passed in but makes a copy anyway.

sdroege:
I guess because `&str` also impls `AsRef<Path>`, so you can directly pass strings as paths?

sdroege:
Trying to somehow fix this whole mess then :)

sdroege:
This is now ready for review @GuillaumeGomez . PR with the generated changes in https://github.com/gtk-rs/gtk-rs-core/pull/318

GuillaumeGomez:
> I guess because `&str` also impls `AsRef<Path>`, so you can directly pass strings as paths?

It's what I meant: `AsRef<Path>` allows to pass strings for paths, however I don't see a case where `AsRef<str>` can be useful...

src/analysis/bounds.rs
```diff
@@ -203,16 +205,21 @@ impl Bounds {
...
             _ => None,
         }
     }
 
-    fn get_to_glib_extra(bound_type: &BoundType) -> String {
+    fn get_to_glib_extra(
+        bound_type: &BoundType,
+        nullable: bool,
+        instance_parameter: bool,
+    ) -> String {
```
GuillaumeGomez, line 217: I wonder why we don't return a `&'static str` instead... We don't format anything after all...

GuillaumeGomez (commented):
Looks good to me apart from the nit I encountered. Feel free to ignore it to merge this PR but if you do, please open an issue so we can come back to it later on. :)

src/analysis/bounds.rs
sdroege: For allowing more complicated things later. We had that here at some point in the past.

Changing the type would require changing it in lots of other places too.

sdroege:
Thanks for the review!

PR #1248, opened 2021-10-31
codgen: generate deprecated cfg attribute for Default trait

bilelmoussaoui:
If the constructor method is deprecated, the builder pattern should have a guard
as well

Needed for gtk4-rs

cc @GuillaumeGomez, @sdroege


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

Issue #1252, opened 2021-11-02
RFC: generate `NONE` constants as associated constants


fengalin:
## NONE constants

For [non "final types"](https://github.com/gtk-rs/gir/blob/8663b33dd5e8c284013c5ff5d306cc04b578d4d4/src/codegen/object.rs#L275), `gir` generates a constant in the form:

``` rust
pub const NONE_PAD: Option<&Pad> = None;
```

This allows passing `NONE_PAD` as an argument instead of the verbose `Option::<&Pad>::None`.

These constants are explicitly re-exported so that they bubble up to the crate's scope ([see in codegen](https://github.com/gtk-rs/gir/blob/8663b33dd5e8c284013c5ff5d306cc04b578d4d4/src/codegen/object.rs#L568)).

## Proposal

Defining an associated constant such as:

``` rust
impl Pad {
    pub const NONE: Option<&'static Pad> = None;
}
```

... would allow passing `Pad::NONE` as an argument.

- This would be more consistent with Rust usage IMHO (see [`Duration::ZERO`](https://doc.rust-lang.org/std/time/struct.Duration.html#associatedconstant.ZERO)).
- No additional re-export is needed since the constant is referred to direclty from its `struct`.
- This is similar to C constants, at least in `gstreamer` (see [`GST_CLOCK_TIME_NONE`](https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/blob/master/gstreamer/sys/src/lib.rs#L432)).

## Options

If we decide to go this route, we could:

1. Generate both `NONE_XXX` and `XXX::NONE`. Existing applications would work unchanged, however this provides two means to achieve the same purpose.
2. Generate both `NONE_XXX` and `XXX::NONE` and add a deprecation attribute to `NONE_XXX`, with a message explaining that `XXX::NONE` should be preferred. Existing applications would compile with warnings. People could migrate their applications when they deem appropriate or silence the compilation warning if they prefer the `NONE_XXX` form. The deprecated form would be kept, just like deprecated items are kept in `std`. I remember we disagreed about deprecation attributes in another proposal.
3. Generate `XXX::NONE` only. Application developers would need to replace all occurrences of `NONE_*` with `*::NONE`, there are probably not many in use though.

sdroege:
I like this and would vote for 3. Personally I think these constants shouldn't really exist anyway but apparently it's making things easier for beginners.

sdroege:
[...]

PR #1253, opened 2021-11-03
codegen: generate NONE constants as associated constants

fengalin:
So that we can use `Pad::NONE` instead of crate global `NONE_PAD`.

Fixes https://github.com/gtk-rs/gir/issues/1252

### DIFF
--- a/src/codegen/object.rs
+++ b/src/codegen/object.rs
@@ -8,7 +8,6 @@ use super::{
 };
 use crate::{
     analysis::{self, ref_mode::RefMode, rust_type::RustType, special_functions::Type},
-    case::CaseExt,
     env::Env,
     library::{self, Nullable},
     nameutil,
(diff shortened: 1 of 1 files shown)


fengalin:
> That's pretty great, but wasn't there already an existing impl block where this constant could be moved into?

I don't think so. The main `impl` block is usually behind a `glib` `macro` and the `NONE` constant generation depends on a condition known to `gir`. Others `impl` blocks generation depend on the generated `struct`. See [buffer_pool](https://gitlab.freedesktop.org/fengalin/gstreamer-rs/-/blob/none-associated-const/gstreamer/src/auto/buffer_pool.rs) as an example.

Besides, I like to separate `impl` blocks by purposes. But I'd be happy to merge with another block if you find a suitable candidate :)

fengalin:
AFAICT, CI failures are expected as it requires the updates to `gtk-rs-core` and co.

MarijnS95:
It may have been bitflags/enums that always have an `impl` block (which this `const` could be conditionally added to), sorry for the confusion!

fengalin:
NP. I applied your suggestion to manually changed code.

sdroege (approved):

sdroege:
@GuillaumeGomez @bilelmoussaoui Ok with you too?

fengalin:
Please ping me before merging so I can update the dependent PRs immediately.

bilelmoussaoui (approved):

PR #1254, opened 2021-11-04
flags/enums: impl FromOwnedValue trait

bilelmoussaoui:
Needed for https://github.com/gtk-rs/gtk-rs-core/pull/325

### DIFF
--- a/src/codegen/enums.rs
+++ b/src/codegen/enums.rs
@@ -396,7 +396,8 @@ impl FromGlib<{sys_crate_name}::{ffi_name}> for {name} {{
     unsafe fn from_value(value: &'a {gvalue}) -> Self {{
         {assert}from_glib({glib}(value.to_glib_none().0))
     }}
-}}",
+}}
+unsafe impl FromOwnedValue for {name} {{}}",
             name = enum_.name,
             glib = use_glib_type(env, "gobject_ffi::g_value_get_enum"),
             gvalue = use_glib_type(env, "Value"),
--- a/src/codegen/flags.rs
+++ b/src/codegen/flags.rs
@@ -256,7 +256,7 @@ impl FromGlib<{sys_crate_name}::{ffi_name}> for {name} {{
     unsafe fn from_value(value: &'a {gvalue}) -> Self {{
         {assert}from_glib({glib}(value.to_glib_none().0))
     }}
-}}",
+}}unsafe impl FromOwnedValue for {name} {{}}",
             name = flags.name,
             glib = use_glib_type(env, "gobject_ffi::g_value_get_flags"),
             gvalue = use_glib_type(env, "Value"),


src/codegen/flags.rs
```diff
@@ -256,7 +256,9 @@ impl FromGlib<{sys_crate_name}::{ffi_name}> for {name} {{
     unsafe fn from_value(value: &'a {gvalue}) -> Self {{
         {assert}from_glib({glib}(value.to_glib_none().0))
     }}
-}}",
+}}
+            unsafe impl FromOwnedValue for {name} {{}}
```
sdroege, line 260: Also for enums

sdroege (approved):

src/codegen/enums.rs
```diff
@@ -396,7 +396,8 @@ impl FromGlib<{sys_crate_name}::{ffi_name}> for {name} {{
     unsafe fn from_value(value: &'a {gvalue}) -> Self {{
         {assert}from_glib({glib}(value.to_glib_none().0))
     }}
-}}",
+}}
+            unsafe impl FromOwnedValue for {name} {{}}",
```
GuillaumeGomez, line 400: The indent is completely broken here. It should be:

```suggestion
unsafe impl FromOwnedValue for {name} {{}}",
```

src/codegen/flags.rs
```diff
@@ -256,7 +256,9 @@ impl FromGlib<{sys_crate_name}::{ffi_name}> for {name} {{
     unsafe fn from_value(value: &'a {gvalue}) -> Self {{
         {assert}from_glib({glib}(value.to_glib_none().0))
     }}
-}}",
+}}
```
GuillaumeGomez, line 259: Same for the line below this one (gh doesn't let me create a new conversation on this line :'( ).

src/codegen/enums.rs
bilelmoussaoui: Does it really matter ? At the end we call rustfmt to fix the broken formatting

GuillaumeGomez: It does, the code didn't make sense when I was reading it because I thought it was an impl inside an impl (because of the indent).

bilelmoussaoui: Fair, updated

src/codegen/flags.rs
```diff
@@ -256,7 +256,7 @@ impl FromGlib<{sys_crate_name}::{ffi_name}> for {name} {{
     unsafe fn from_value(value: &'a {gvalue}) -> Self {{
         {assert}from_glib({glib}(value.to_glib_none().0))
     }}
-}}",
+}}unsafe impl FromOwnedValue for {name} {{}}",
```
GuillaumeGomez, line 259: Please put it on its own line. Also, 
please add an empty line before it, makes it easier to spot different items.

src/codegen/enums.rs
```diff
@@ -396,7 +396,8 @@ impl FromGlib<{sys_crate_name}::{ffi_name}> for {name} {{
     unsafe fn from_value(value: &'a {gvalue}) -> Self {{
         {assert}from_glib({glib}(value.to_glib_none().0))
     }}
-}}",
+}}
+unsafe impl FromOwnedValue for {name} {{}}",
```
GuillaumeGomez, line 400: Please add an empty line before it, makes it easier to spot different items.

bilelmoussaoui:
This one is no longer needed after adding a blanket impl around the old trait

PR #1255, opened 2021-11-06
codgen: simplify properties getters/setters

bilelmoussaoui:
This PR simplifies the properties getters/setters generation code so we don't go through ffi and use ObjectExt trait directly. This drops a huge amount of code.

It also adapts emit_signal_by_name usages so they are "compatible" with changes in https://github.com/gtk-rs/gtk-rs-core/pull/323

The generated code looks fine, it needs ObjectExt trait to be in-scope, something I have to figure out first before getting this one merged.

Note, this merge request, at least the first commit of it is required for https://github.com/gtk-rs/gtk-rs-core/pull/323 to be merged. As otherwise almost all the usages of the emit_$signal in generated code will panic.

### DIFF
--- a/src/codegen/property_body.rs
+++ b/src/codegen/property_body.rs
@@ -76,13 +76,51 @@ impl<'a> Builder<'a> {
     }
 
     fn chunks_for_get(&self) -> Vec<Chunk> {
-        let mut params = Vec::new();
-
-        let cast_target = if self.is_child_property {
-            use_gtk_type(self.env, "ffi::GtkContainer")
+        if self.is_child_property {
+            // TODO: make use of safe bindings for child properties setter
+            self.child_property_setter()
         } else {
-            use_glib_type(self.env, "gobject_ffi::GObject")
-        };
+            let self_ = if self.in_trait {
+                "self.as_ref()"
+            } else {
+                "self"
+            };
+
+            vec![Chunk::Custom(format!(
+                "{}::property({}, \"{}\")",
+                use_glib_type(self.env, "ObjectExt"),
+                self_,
+                self.name
+            ))]
+        }
+    }
+
+    fn chunks_for_set(&self) -> Vec<Chunk> {
+        if self.is_child_property {
+            // TODO: make use of safe bindings for child properties getter
+            self.child_property_getter()
+        } else {
+            let self_ = if self.in_trait {
+                "self.as_ref()"
+            } else {
+                "self"
+            };
+
+            vec![Chunk::Custom(format!(
+                "{}::set_property({},\"{}\", &{})",
+                use_glib_type(self.env, "ObjectExt"),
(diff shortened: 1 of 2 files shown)


src/codegen/property_body.rs
```diff
@@ -76,125 +76,123 @@ impl<'a> Builder<'a> {
     }
 
     fn chunks_for_get(&self) -> Vec<Chunk> {
-        let mut params = Vec::new();
+        if self.is_child_property {
+            let mut body = Vec::new();
+
```
GuillaumeGomez, line 81: Weird coding style. ;)

bilelmoussaoui:
gtk4-rs PR: https://github.com/gtk-rs/gtk4-rs/pull/685

Will do the gtk3-rs/gtk-rs-core/gstreamer-rs later unless someone wants to help with that :)

src/codegen/property_body.rs
bilelmoussaoui: Which parts of it? :) I mean, my code fixes simplifies the properties getters/setters; it doesn't do the same for the GTK 3 child properties

sdroege: Maybe you can move the whole child property generation into a generate_getter/setter function so it doesn't pollute the main code path? Who cares about GTK child properties after all :)

src/codegen/signal.rs
```diff
@@ -134,14 +133,13 @@ pub fn generate(
 
             writeln!(
                 w,
-                "{}let {} = unsafe {{ glib::Object::from_glib_borrow(self.as_ptr() as *mut {}).emit_by_name(\"{}\", &[{}]).unwrap() }};",
+                "{}let {} = self.emit_by_name(\"{}\", &[{}]);",
```
sdroege, line 136: And one day we also move the signal connection functions over to `ObjectExt::connect()` :)

sdroege (commented):
Looks good to me, but I'd ideally like to have the child property code out of the main code path into a separate function because that's completely irrelevant code for everything but a single crate :)

bilelmoussaoui:
> Looks good to me, but I'd ideally like to have the child property code out of the main code path into a separate function because that's completely irrelevant code for everything but a single crate :)

sure, can do that :)

src/codegen/signal.rs
bilelmoussaoui: indeed, a task for another day

sdroege: Definitely. It's not so easy :)

sdroege:
@bilelmoussaoui Can you update accordingly, then I'll merge this one already and we can get the other PRs in line :)

bilelmoussaoui:
Should be updated, I also added todo comments for gtk3-rs simplifications in the future

Issue #1231, opened 2021-10-03
Missing #cfg guard for multiple glib::wrapper!


A6GibKm:
Using the latest gir in libhandy-rs it generates multiple wrappers for a struct, e.g.

```
glib::wrapper! {
    #[doc(alias = "HdyApplicationWindow")]
    pub struct ApplicationWindow(Object<ffi::HdyApplicationWindow, ffi::HdyApplicationWindowClass>) @extends gtk::ApplicationWindow, gtk::Window, gtk::Bin, gtk::Container, gtk::Widget, @implements gtk::Buildable, gio::ActionGroup, gio::ActionMap;

    match fn {
        type_ => || ffi::hdy_application_window_get_type(),
    }
}

glib::wrapper! {
    #[doc(alias = "HdyApplicationWindow")]
    pub struct ApplicationWindow(Object<ffi::HdyApplicationWindow, ffi::HdyApplicationWindowClass>) @extends gtk::ApplicationWindow, gtk::Window, gtk::Bin, gtk::Container, gtk::Widget, @implements gtk::Buildable, gio::ActionGroup;

    match fn {
        type_ => || ffi::hdy_application_window_get_type(),
    }
}
```
Both should have a #cfg guard with different version requirements.

Reproducer:  Update libhandy-rs's gir submodule and regen bindings.

Related: https://github.com/gtk-rs/gir/pull/1230

sdroege:
```diff
diff --git a/libhandy/src/auto/action_row.rs b/libhandy/src/auto/action_row.rs
index cc6fa1f..c578431 100644
--- a/libhandy/src/auto/action_row.rs
+++ b/libhandy/src/auto/action_row.rs
@@ -15,6 +15,8 @@ use std::boxed::Box as Box_;
 use std::fmt;
 use std::mem::transmute;
 
+#[cfg(any(feature = "v3_4", feature = "dox"))]
+#[cfg_attr(feature = "dox", doc(cfg(feature = "v3_4")))]
 glib::wrapper! {
     #[doc(alias = "HdyActionRow")]
     pub struct ActionRow(Object<ffi::HdyActionRow, ffi::HdyActionRowClass>) @extends PreferencesRow, gtk::ListBoxRow, gtk::Bin, gtk::Container, gtk::Widget, @implements 
gtk::Buildable, gtk::Actionable;
@@ -24,6 +26,16 @@ glib::wrapper! {
     }
 }
 
[...]

PR #1256, opened 2021-11-09
per crate min version

bilelmoussaoui:
Fixes https://github.com/gtk-rs/gir/issues/1231
Fixes https://github.com/gtk-rs/gir/issues/1099

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ authors = [
 ]
 build = "build.rs"
 exclude = ["Gir*.toml", "tests/**/*", "*.md", "generator.py"]
-edition = "2021"
+edition = "2018"
 
 [dependencies]
 bitflags = "1.0"
--- a/src/config/external_libraries.rs
+++ b/src/config/external_libraries.rs
@@ -1,10 +1,13 @@
 use super::error::*;
 use crate::nameutil::crate_name;
+use crate::version::Version;
+use std::str::FromStr;
 
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct ExternalLibrary {
     pub namespace: String,
     pub crate_name: String,
+    pub min_version: Option<Version>,
 }
 
 pub fn read_external_libraries(toml: &toml::Value) -> Result<Vec<ExternalLibrary>, String> {
@@ -15,6 +18,7 @@ pub fn read_external_libraries(toml: &toml::Value) -> Result<Vec<ExternalLibrary
             .filter_map(|v| v.as_str().map(String::from))
             .map(|namespace| ExternalLibrary {
                 crate_name: crate_name(&namespace),
+                min_version: None,
                 namespace,
             })
             .collect(),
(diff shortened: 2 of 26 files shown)


Cargo.toml
```diff
@@ -8,7 +8,7 @@ authors = [
 ]
 build = "build.rs"
 exclude = ["Gir*.toml", "tests/**/*", "*.md", "generator.py"]
-edition = "2021"
+edition = "2018"
```
GuillaumeGomez, line 11: Why reverting this?

GuillaumeGomez, line 11: Why reverting this?

src/config/external_libraries.rs
```diff
@@ -25,17 +29,35 @@ pub fn read_external_libraries(toml: &toml::Value) -> Result<Vec<ExternalLibrary
...
         for custom_lib in custom_libs {
-            let crate_name = custom_lib.0;
-            if let Some(namespace) = custom_lib.1.as_str() {
+            if let Some(info) = custom_lib.1.as_table() {
+                println!("{:#?}", info);
+                let namespace = custom_lib.0.as_str();
+                let crate_name_ = info
+                    .get("crate")
+                    .map(|c| c.as_str().expect("crate name must be a string").to_string())
+                    .unwrap_or_else(|| crate_name(&namespace));
```
GuillaumeGomez, line 38:
```suggestion
                    .unwrap_or_else(|| crate_name(namespace));
```

sophie-h:
Can't reproduce that adding 

```
[external_libraries]
Gio = {min_version = "2.32"}
```

fixes libadwaita

Cargo.toml
bilelmoussaoui: For rust analyzer, I wasn't meaning to push it

sdroege: Why? For me it works with 2021 :)

GuillaumeGomez: Question remains. :)

GuillaumeGomez: Still waiting.

bilelmoussaoui:
I will re-generate the various crates tomorrow and add some docs to the book to explain the new options.

sdroege:
Why does this need changes in the existing crates?

bilelmoussaoui:
> Why does this need changes in the existing crates?

Because gtk3/gtk4 needs to set the new gio features or define a min version and regen

sdroege (approved):
Looks good to me

sdroege:
@bilelmoussaoui Did you check if this works correctly with gstreamer-rs? In theory nothing should be needed there except for maybe adding a minimum GLib/GIO version.

bilelmoussaoui:
> @bilelmoussaoui Did you check if this works correctly with gstreamer-rs? In theory nothing should be needed there except for maybe adding a minimum GLib/GIO version.

Yes, I will be sending a regen PR in a minute. Doing the libadwaita/libhandy ones

sdroege:
@bilelmoussaoui Ok then let's get this one here in so you can do the other PRs properly :)

PR #1262, opened 2021-11-12
Correctly generate opaque types

GuillaumeGomez:
Based on <https://doc.rust-lang.org/nomicon/ffi.html#representing-opaque-structs>.

You can see the result [here](https://github.com/gtk-rs/gtk3-rs/pull/670).

### DIFF
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -376,6 +376,19 @@ fn generate_classes_structs(w: &mut dyn Write, env: &Env, classes: &[&Class]) ->
     Ok(())
 }
 
+fn generate_opaque_type(w: &mut dyn Write, name: &str) -> Result<()> {
+    writeln!(
+        w,
+        r#"#[repr(C)]
+pub struct {} {{
+    _data: [u8; 0],
+    _marker: core::marker::PhantomData<(*mut u8, core::marker::PhantomPinned)>,
+}}
+"#,
+        name
+    )
+}
+
 fn generate_interfaces_structs(
     w: &mut dyn Write,
     env: &Env,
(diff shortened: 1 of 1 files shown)


src/codegen/sys/lib_.rs
```diff
@@ -389,7 +397,8 @@ fn generate_interfaces_structs(
         if !env.type_status_sys(&full_name).need_generate() {
             continue;
         }
-        writeln!(w, "#[repr(C)]\npub struct {}(c_void);\n", interface.c_type)?;
+        writeln!(w, "#[repr(C)]")?;
```
sdroege, line 400: Why not move the `#[repr(C)]` into the `generate_opaque_type` function?

src/codegen/sys/lib_.rs
```diff
@@ -489,7 +497,7 @@ fn generate_from_fields(
         // It would be nice to represent those using extern types
         // from RFC 1861, once they are available in stable Rust.
         // https://github.com/rust-lang/rust/issues/43467
-        writeln!(w, "pub struct {name}(c_void);", name = &fields.name)?;
```
sdroege, line 492: Do we still use `c_void` for something? If not we can remove it from the imports

sdroege (commented):
Looks good to me otherwise

src/codegen/sys/lib_.rs
GuillaumeGomez: I did, but that creates quite the diff haha.

src/codegen/sys/lib_.rs
GuillaumeGomez: I don't think we do. Let's try removing it. :)

GuillaumeGomez:
```
error[E0412]: cannot find type `c_void` in this scope
  --> gdk/sys/src/lib.rs:35:22
   |
35 | pub type GdkXEvent = c_void;
   |                      ^^^^^^ not found in this scope
```

So not yet!

GuillaumeGomez:
Are you ok with the change of position of `repr(C)`?

src/codegen/sys/lib_.rs
sdroege: Why?

sdroege (approved):

src/codegen/sys/lib_.rs
sdroege: I see :man_shrugging:

PR #1263, opened 2021-11-15
codgen: don't re-export builder types on crate root

bilelmoussaoui:
they are available nowadays through Type::builder method

cc @sdroege

### DIFF
--- a/src/codegen/object.rs
+++ b/src/codegen/object.rs
@@ -439,6 +439,7 @@ impl {name}Builder {{
         "
     // rustdoc-stripper-ignore-next
     /// Build the [`{name}`].
+    #[must_use = \"The builder must be built to be used\"]
     pub fn build(self) -> {name} {{
         let mut properties: Vec<(&str, &dyn ToValue)> = vec![];",
         name = analysis.name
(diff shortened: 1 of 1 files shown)


sdroege (approved):

bilelmoussaoui:
@sdroege this can be merged, it needs examples changes on both gtk3-rs/gtk4-rs. I will get those done :)

GuillaumeGomez (approved):

GuillaumeGomez:
Should we wait for the examples fix first?

bilelmoussaoui:
> Should we wait for the examples fix first?

Let's wait until I add a `#[must_use]` to the build method of the builder

GuillaumeGomez:
Ok!

sdroege:
> Should we wait for the examples fix first?

First we need to merge this and then we can regenerate and fix the code, or not?

bilelmoussaoui:
> > Should we wait for the examples fix first?
> 
> First we need to merge this and then we can regenerate and fix the code, or not?

We could but that requires regenerating everything once merged again. There are very few cases that needs to be fixed in the examples so it's not worth it

PR #1264, opened 2021-11-15
codgen: generate builders inside a builders module

bilelmoussaoui:
This is a workaround to make sure the builders are usable from outside
without making the modules public nor clutering the main root docs

### DIFF
--- a/src/codegen/mod.rs
+++ b/src/codegen/mod.rs
@@ -71,15 +71,20 @@ pub fn generate_mod_rs(
         general::start_comments(w, &env.config)?;
         general::write_vec(w, mod_rs)?;
         writeln!(w)?;
-        writeln!(w, "#[doc(hidden)]")?;
-        writeln!(w, "pub mod traits {{")?;
-        general::write_vec(w, traits)?;
-        writeln!(w, "}}")?;
+        if !traits.is_empty() {
+            writeln!(w, "#[doc(hidden)]")?;
+            writeln!(w, "pub mod traits {{")?;
+            general::write_vec(w, traits)?;
+            writeln!(w, "}}")?;
+        }
 
-        writeln!(w, "#[doc(hidden)]")?;
-        writeln!(w, "pub mod builders {{")?;
-        general::write_vec(w, builders)?;
-        writeln!(w, "}}")
+        if !builders.is_empty() {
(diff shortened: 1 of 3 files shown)


src/codegen/object.rs
```diff
@@ -607,4 +608,14 @@ pub fn generate_reexports(
             module_name, analysis.trait_name
         ));
     }
+
+    if has_builder_properties(&analysis.builder_properties) {
+        for cfg in &cfgs {
+            builders.push(format!("\t{}", cfg));
+        }
+        builders.push(format!(
+            "\t pub use super::{}::{}Builder;",
```
GuillaumeGomez, line 617: Please don't use `\t` and write four whitespace directly. In this case it's even more weird considering that there is a whitespace just after...

GuillaumeGomez:
Also, please provide a regen PR so we can see the changes.

src/codegen/object.rs
bilelmoussaoui: I copied whatever used few lines above

bilelmoussaoui:
> Also, please provide a regen PR so we can see the changes.

here you go https://github.com/gtk-rs/gtk4-rs/pull/719/files

bilelmoussaoui:
This does what we do for the automatically generated traits; we will need to a module for builders that re-exports the ones from auto & the manual ones.

src/codegen/object.rs
bilelmoussaoui: oh never mind; i didn't intend to add the whitespace after \t

src/codegen/mod.rs
```diff
@@ -67,6 +74,11 @@ pub fn generate_mod_rs(env: &Env, root_path: &Path, mod_rs: &[String], traits: &
         writeln!(w, "#[doc(hidden)]")?;
         writeln!(w, "pub mod traits {{")?;
         general::write_vec(w, traits)?;
+        writeln!(w, "}}")?;
+
+        writeln!(w, "#[doc(hidden)]")?;
+        writeln!(w, "pub mod builders {{")?;
+        general::write_vec(w, builders)?;
```
GuillaumeGomez, line 81: Can you check if `builders` is empty before adding this new module? (Would be nice to do it for traits too :smile: ).

sdroege (approved):

PR #1266, opened 2021-11-16
codgen: don't generate empty impl blocks

### LINKED ISSUE #976: When an impl block is empty, it shouldn't be generated
As seen in https://github.com/gtk-rs/gdkx11/pull/7.

bilelmoussaoui:
Fixes #976

gtk4-rs PR: https://github.com/gtk-rs/gtk4-rs/pull/722

### DIFF
--- a/src/analysis/object.rs
+++ b/src/analysis/object.rs
@@ -56,12 +56,23 @@ impl Info {
     pub fn should_generate_impl_block(&self) -> bool {
         self.has_constructors
             || has_builder_properties(&self.builder_properties)
-            || !self.generate_trait
+            || !self.need_generate_trait()
                 && !(self.methods().is_empty()
                     && self.properties.is_empty()
                     && self.child_properties.is_empty()
                     && self.signals.is_empty())
             || self.has_functions
+    }
+
+    pub fn need_generate_inherent(&self) -> bool {
+        self.has_constructors
+            || self.has_functions
+            || !self.need_generate_trait()
+            || has_builder_properties(&self.builder_properties)
+    }
+
+    pub fn need_generate_trait(&self) -> bool {
+        self.generate_trait
     }
 
     pub fn has_action_signals(&self) -> bool {
--- a/src/analysis/object.rs
+++ b/src/analysis/object.rs
@@ -49,6 +49,21 @@ impl Info {
             || self.notify_signals.iter().any(|s| s.trampoline.is_ok())
     }
 
+    /// Whether we should generate an impl block for this object
+    /// We don't generate an impl block if the type doesn't have any of the followings:
+    /// - Constructors / Functions / Builder properties (no build function)
+    /// - Is a final type & doesn't have either methods / properties / child properties / signals
+    pub fn should_generate_impl_block(&self) -> bool {
+        self.has_constructors
+            || has_builder_properties(&self.builder_properties)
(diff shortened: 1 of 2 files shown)


src/codegen/object.rs
```diff
@@ -141,7 +141,22 @@ pub fn generate(
 
     if need_generate_inherent(analysis) {
         writeln!(w)?;
-        write!(w, "impl {} {{", analysis.name)?;
+        // We don't generate an impl block if the type doesn't have any of the followings:
+        // - Constructors / Functions / Builder properties (no build function)
+        // - Is a final type & doesn't have either methods / properties / child properties / signals
+        let should_generate_impl_block = !analysis.constructors().is_empty()
```
GuillaumeGomez, line 147: That seems overly complicated... Is it really worth it?

bilelmoussaoui: Well, it's complicated yes, but that's how things are :) Whether it's worth it or not; that's up to you as you opened the issue :p

sdroege: You're usually the one complaining about such things in autogenerated code :)

bilelmoussaoui: I could move the logic to analysis and it would be just a func call here

bilelmoussaoui:
I opened a gtk3-rs PR as well to ensure it does what it's supposed to https://github.com/gtk-rs/gtk3-rs/pull/677. I could follow up with a gst/gtk-rs-core ones if we think it fine to include this change

src/codegen/object.rs
bilelmoussaoui: I moved things around so it looks "simpler"

src/codegen/object.rs
```diff
@@ -141,7 +141,10 @@ pub fn generate(
 
     if need_generate_inherent(analysis) {
         writeln!(w)?;
-        write!(w, "impl {} {{", analysis.name)?;
+        if analysis.should_generate_impl_block() {
```
sdroege, line 144:
```suggestion
        if !analysis.should_generate_impl_block() {
            return Ok(());
        }
```

would be even easier or not? Or to otherwise skip the whole block below

bilelmoussaoui: Ah Indeed we can simplify it a bit more, will update it in a bit

src/analysis/object.rs
```diff
@@ -49,6 +49,32 @@ impl Info {
...
+    }
+
+    pub fn need_generate_inherent(&self) -> bool {
+        self.has_constructors
+            || self.has_functions
+            || !self.need_generate_trait()
+            || has_builder_properties(&self.builder_properties)
+    }
+
+    pub fn need_generate_trait(&self) -> bool {
```
bilelmoussaoui, line 74: Tbh I don't think this should exist as the `generate_trait` is public; or do we want stuff to be private & create getters for them when needed?

sdroege (approved):

src/analysis/object.rs
sdroege: Sometimes using fields, sometimes getters seems a bit confusing. That looks like it needs a bit more cleanup than would be in the scope of this PR.

bilelmoussaoui: Indeed, just wondering in general in case I get bored someday :-)

PR #1268, opened 2021-11-18
codgen: don't append () to gi-docgen not found functions

### LINKED ISSUE #1200: Generated documentation for gtk4::Widget is broken
The first error seems to be in the fourth paragraph under [this](https://gtk-rs.org/gtk4-rs/stable/latest/docs/gtk4/struct.Widget.html#building-composite-widgets-from-template-xml) heading, there are two backticks in the end to close an inline code block.
Edit: I investigated a bit further and the problem seems to be a ` ``<template>`` ` that is enclosed in two backticks instead of just one. And then rustdoc interprets the <template> tag as html markup?

So, apart from making gir not enclose something with two backticks instead of one, one possible fix could be to translate `<` and `>` to `&lt;` and `&gt;`. At least with my limited understanding of what git already does...

bilelmoussaoui:
Otherwise, the gtk-doc style parsing code, will detect those as functions which ends up breaking
the whole docs, such as issues seen in #1200. Ideally, we would have more robust code
that detects whether the detected item is already a rust item or an item we couldn't found the first time with gi-docgen
to avoid breaking things. But such features are not possible with the currently use regex rust crate

cc @MarijnS95 , what do you think? This only affects gtk4/pango & other crates that uses gi-docgen

Fixes #1200

### DIFF
--- a/src/codegen/doc/gi_docgen.rs
+++ b/src/codegen/doc/gi_docgen.rs
@@ -317,7 +317,7 @@ impl GiDocgen {
                     |f| f.name == mangle_keywords(name),
                 )
                 .map_or_else(
-                    || format!("`{}::{}()`", ns_type_to_doc(namespace, type_), name),
+                    || format!("`{}::{}`", ns_type_to_doc(namespace, type_), name),
                     |(obj_info, fn_info)| {
                         gen_object_fn_doc_link(obj_info, fn_info, env, in_type, type_)
                     },
@@ -329,9 +329,9 @@ impl GiDocgen {
             } => find_method_or_function_by_name(type_.as_deref(), name, env, in_type, false)
                 .unwrap_or_else(|| {
                     if let Some(ty) = type_ {
-                        format!("`{}::{}()`", ns_type_to_doc(namespace, ty), name)
+                        format!("`{}::{}`", ns_type_to_doc(namespace, ty), name)
                     } else {
-                        format!("`{}()`", name)
+                        format!("`{}`", name)
                     }
                 }),
             GiDocgen::Alias(alias) => gen_alias_doc_link(alias),
@@ -341,7 +341,7 @@ impl GiDocgen {
                 name,
                 is_class_method,
             } => find_method_or_function_by_name(Some(type_), name, env, in_type, *is_class_method)
-                .unwrap_or_else(|| format!("`{}::{}()`", ns_type_to_doc(namespace, type_), name)),
+                .unwrap_or_else(|| format!("`{}::{}`", ns_type_to_doc(namespace, type_), name)),
             GiDocgen::Callback { namespace, name } => {
                 gen_callback_doc_link(&ns_type_to_doc(namespace, name))
             }


sdroege (approved):

MarijnS95:
@bilelmoussaoui Are you sure the `()` is at fault? What code is running after this and replacing it with:

```rust
    /// To create composite widgets with [`Builder`][crate::Builder] XML, one must associate
    /// the interface description with the widget class at class initialization
    /// time using `Gtk::`WidgetClass::set_template()``.
```
? IMO it shouldn't replace symbols that are already inside `` ` `` or have a `::` namespace prefix.

Can you show the relevant bit of "gtk-doc style parsing code" that does this?

MarijnS95:
Also there's more weird stuff on that page such as a space being on the wrong side of a backtick:
```rust
/// `GtkWidget `exposes
```

And gtk-doc is also breaking normal links:

```rust
/// created with [``SignalAction::new()``][crate::`SignalAction::new()`].
```

So this PR won't work around that issue - gtk-doc needs to be fixed instead.

bilelmoussaoui:
> @bilelmoussaoui Are you sure the `()` is at fault? What code is running after this and replacing it with:
> 
> ```rust
>     /// To create composite widgets with [`Builder`][crate::Builder] XML, one must associate
>     /// the interface description with the widget class at class initialization
>     /// time using `Gtk::`WidgetClass::set_template()``.
> ```

It's, In general we are supposed to generate `Gtk::WidgetClass::set_template()` from gi-docgen (which is GTK4 specific). But then, the we look for anything that might look like a function, find `set_template()`  -> look for a function like it -> don't find it -> we end up adding more "`" around it which causes the whole breakage you see in the issue I mentioned
> 

> ? IMO it shouldn't replace symbols that are already inside `` ` `` or have a `::` namespace prefix.

Ideally yes, but we already do that to find random symbols in the docs. If we drop that, a bunch of items in Gstreamer's docs won't be clickable anymore

> 
> EDIT: Or rather, why is some extra tool touching our carefully generated docs from `gir` and messing it all up after the fact?

There's no extra-tool; we run gi-docgen's parser & replace all the links then we run the gtk-doc one

> 
> Can you show the relevant bit of "gtk-doc style parsing code" that does this?

https://github.com/gtk-rs/gir/blob/master/src/codegen/doc/format.rs#L138-L144

MarijnS95:
> There's no extra-tool; we run gi-docgen's parser & replace all the links then we run the gtk-doc one
> 
> > Can you show the relevant bit of "gtk-doc style parsing code" that does this?
> 
> https://github.com/gtk-rs/gir/blob/master/src/codegen/doc/format.rs#L138-L144

We should fix/remove that then, as it really makes no sense to replace random regex matches when we have a nice system.

> but we already do that to find random symbols in the docs. If we drop that, a bunch of items in Gstreamer's docs won't be clickable anymore

Why is the "picky" documentation parser not picking this up then? We made such a new system to parse and improve the docs, only to have a rogue regex ruin it again?

bilelmoussaoui:
> We should fix/remove that then, as it really makes no sense to replace random regex matches when we have a nice system.

That would be fine for me.  What system you're talking about here? :) 

> Why is the "picky" documentation parser not picking this up then? We made such a new system to parse and improve the docs, only to have a rogue regex ruin it again?

I'm not sure again which system you're talking about

MarijnS95:
The "system" here are the `replace_c_types()` functions you wrote so carefully, based on gir introspection.

> Ideally yes, but we already do that to find random symbols in the docs. If we drop that, a bunch of items in Gstreamer's docs won't be clickable anymore.

Not sure what you mean. I removed the offending regex and nothing changes in GStreamer docs. Can you just remove it for gtk4 too?

bilelmoussaoui:
> The "system" here are the `replace_c_types()` functions you wrote so carefully, based on gir introspection.

Well, the `replace_c_types` https://github.com/gtk-rs/gir/blob/master/src/codegen/doc/format.rs#L160, is problematic as well. It does look for functions in https://github.com/gtk-rs/gir/blob/master/src/codegen/doc/format.rs#L165 which is bad because it ends up over-doing everything that was already handled by gi-docgen (gi-docgen, is only specific to gtk4/pango; gstreamer doesn't make use of it) and ends up adding extra "`" around functions links.

There's also https://github.com/gtk-rs/gir/blob/master/src/codegen/doc/format.rs#L184 that breaks links with a target but that one is less problematic. 


>  I removed the offending regex and nothing changes in GStreamer docs. Can you test that removal on gtk4 too?

Which regex you tried removing? i will give it a try

MarijnS95:
> Which regex you tried removing? i will give it a try

Never mind this - it is indeed `replace_c_types` overdoing the work `gi_docgen` has already done. We should figure out a way to combine them (they after all do the exact same thing, except parsing a different input format) or make sure they don't touch the same code _somehow_.

MarijnS95:
> > Which regex you tried removing? i will give it a try
> 
> Never mind this - it is indeed `replace_c_types` overdoing the work `gi_docgen` has already done. We should figure out a way to combine them (they after all do the exact same thing, except parsing a different input format) or make sure they don't touch the same code _somehow_.

EDIT: On that note it seems safe to remove the regex either way: it doesn't change the docs on gtk4 nor gstreamer - == dead code.

bilelmoussaoui:
> Never mind this - it is indeed `replace_c_types` overdoing the work `gi_docgen` has already done. We should figure out a way to combine them (they after all do the exact same thing, except parsing a different input format) or make sure they don't touch the same code _somehow_.

That would be the ideal solution, but I don't have any ideas on how to achieve that. 

> On that note it seems safe to remove the regex either way: it doesn't change the docs on gtk4 nor gstreamer - if it doesn't change in any other crate this is just dead code.

Sure, but that regex is not the problematic one; the code I linked above was meant to explain that one "parser" is run after the other one and that the later overrides changes made by the first one

MarijnS95:
> > Never mind this - it is indeed `replace_c_types` overdoing the work `gi_docgen` has > That would be the ideal solution, but I don't have any ideas on how to achieve that.

Might be a tricky refactor - but as you can see it's botching the links as well (but you don't see that until hovering over or clicking them) so the current workaround won't fix this - and makes the docs _**IMO**_ uglier as a result.

> Sure, but that regex is not the problematic one; the code I linked above was meant to explain that one "parser" is run after the other one and that the later overrides changes made by the first one

Then we are _still_ saying the exact same thing :grimacing:

bilelmoussaoui:
I think i'm going to make the gi-docgen explicitly opt-in and if you opt for it, the other replace_c_types function won't be used. Then make gtk4 opt for it.

This way i'm sure that whatever expected to work links will work and the rest will behave as is until a better solution for merging both parsers is found

PR #1270, opened 2021-11-18
Remove "auto" folder when regenarating non-sys crate

### LINKED ISSUE #1269: generator script should remove src/auto before regen
Otherwise removed files (no longer generated ones) are kept there and you might end up forgetting to drop them

GuillaumeGomez:
Fixes  #1269.

### DIFF
--- a/generator.py
+++ b/generator.py
@@ -5,6 +5,7 @@ import argparse
 import subprocess
 import sys
 import asyncio
+import shutil
 
 DEFAULT_GIR_FILES_DIRECTORY = Path("./gir-files")
 DEFAULT_GIR_DIRECTORY = Path("./gir/")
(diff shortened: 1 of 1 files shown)


generator.py
```diff
@@ -136,6 +137,8 @@ def regen_crates(path, conf):
         else:
             if is_sys_crate:
                 args.extend(["-m", "sys"])
+            else:
+                shutil.rmtree(path.parent / "src/auto")
```
sdroege, line 141: Does this follow symlinks? I hope not

MarijnS95:
Can't we do this inside `gir` instead?

GuillaumeGomez:
I guess we could. What do you think about it @sdroege ?

sdroege:
I prefer in the script a bit over the other solution, but whatever you prefer

GuillaumeGomez:
And you @bilelmoussaoui ?

bilelmoussaoui:
In gir it would make more sense I guess

PR #1274, opened 2021-11-18
Mark gi-docgen as optional

### LINKED ISSUE #1200: Generated documentation for gtk4::Widget is broken

bilelmoussaoui:
The PR moves gi-docgen from always being run along with gtk-doc parser to being optional and opt-in. This way only one of the two is run and avoids the whole issue of what I tried to workaround in https://github.com/gtk-rs/gir/pull/1268

Closes both #1268 & Fixes #1200

### DIFF
--- a/book/src/config_api.md
+++ b/book/src/config_api.md
@@ -12,6 +12,8 @@ target_path = "."
 # Path where objects generated (defaults to <target_path>/src/auto)
 # auto_path = "src/auto"
 work_mode = "normal"
+# Whether the library uses https://gitlab.gnome.org/GNOME/gi-docgen for its documentation
+use_gi_docgen = false
 generate_safety_asserts = true
 deprecate_by_min_version = true
 # With this option enabled, versions for gir and gir-files saved only to one file to minimize noise,
--- a/src/codegen/doc/format.rs
+++ b/src/codegen/doc/format.rs
@@ -165,8 +165,10 @@ static FUNCTION: Lazy<Regex> =
 // **note**
 // The optional . at the end is to make the regex more relaxed for some weird broken cases on gtk3's docs
 // it doesn't hurt other docs so please don't drop it
-static GDK_GTK: Lazy<Regex> =
-    Lazy::new(|| Regex::new(r"`([^\(:])?((G[dts]k|Pango)\w+\b)(\.)?`").unwrap());
+static GDK_GTK: Lazy<Regex> = Lazy::new(|| {
+    Regex::new(r"`([^\(:])?((G[dts]k|Pango|cairo_|graphene_|Adw|Hdy|GtkSource)\w+\b)(\.)?`")
+        .unwrap()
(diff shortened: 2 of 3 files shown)


bilelmoussaoui:
cc @MarijnS95

book/src/config_api.md
```diff
@@ -12,6 +12,8 @@ target_path = "."
 # Path where objects generated (defaults to <target_path>/src/auto)
 # auto_path = "src/auto"
 work_mode = "normal"
+# Whether the library uses https://gitlab.gnome.org/GNOME/gi-docgen for its documentation
+use_gi_docgen = true
```
MarijnS95, line 16: The default is false, no?

src/codegen/doc/format.rs
```diff
@@ -135,18 +135,34 @@ fn replace_symbols(
...
-        .replace_all(&out, |caps: &Captures<'_>| format!("`{}`", &caps[2]))
-        .to_string()
+    if env.config.use_gi_docgen {
+        let out = gi_docgen::replace_c_types(input, env, in_type);
+        let out = GI_DOCGEN_SYMBOL.replace_all(&out, |caps: &Captures<'_>| match &caps[2] {
+            "TRUE" => "[`true`]".to_string(),
+            "FALSE" => "[`false`]".to_string(),
+            "NULL" => "[`None`]".to_string(),
+            symbol_name => match &caps[1] {
+                // Opt-in only for the %SYMBOLS, @/# causes breakages
```
MarijnS95, line 145: Can we factor this out?

MarijnS95 (commented):
Github doesn't detect "closes both" as valid prefix to close the other PR automatically when this is merged.

src/codegen/doc/format.rs
bilelmoussaoui: we can't because you use a different regex, but we can merge some parts of it somehow

book/src/config_api.md
bilelmoussaoui: it's, yes; will update it to use the default value

src/codegen/doc/format.rs
bilelmoussaoui: you here as in the gtk-doc parser

sdroege (approved):

PR #1277, opened 2021-12-02
Remove useless `#[must_use]`

GuillaumeGomez:
Added in #1263.

What would be nice would be to panic if the `builder` wasn't "built" in a `Drop` implementation.

### DIFF
--- a/src/codegen/object.rs
+++ b/src/codegen/object.rs
@@ -440,7 +440,6 @@ impl {name}Builder {{
         "
     // rustdoc-stripper-ignore-next
     /// Build the [`{name}`].
-    #[must_use = \"The builder must be built to be used\"]
     pub fn build(self) -> {name} {{
         let mut properties: Vec<(&str, &dyn ToValue)> = vec![];",
         name = analysis.name


GuillaumeGomez:
False alert, nevermind.

src/codegen/object.rs
```diff
@@ -440,7 +440,6 @@ impl {name}Builder {{
         "
     // rustdoc-stripper-ignore-next
     /// Build the [`{name}`].
-    #[must_use = \"The builder must be built to be used\"]
     pub fn build(self) -> {name} {{
```
sdroege, line 443: This is actually semi-useful: if you don't do anything with the return value of `build()` then you're probably doing something wrong.

src/codegen/object.rs
```diff
@@ -440,7 +440,6 @@ impl {name}Builder {{
...
         let mut properties: Vec<(&str, &dyn ToValue)> = vec![];",
```
sdroege, line 444: But it could also make sense to mark the builder type as `#[must_use]`, see the [docs](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute):

> When used on user-defined composite types, if the expression of an expression statement has that type, then the unused_must_use lint is violated.

I.e. if you forget to call `build()` then you have an expression statement that has the type of the builder.

sdroege: See https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d1080d58a2abe179ca6483ef685bdd64

Issue #1279, opened 2021-12-02
codegen: assert that TRUE is returned XOR the GError is set


lucab:
From the [glib guide](https://docs.gtk.org/glib/error-reporting.html):

> If FALSE is returned, the error must be set to a non-NULL value.
> One exception to this is that in situations that are already considered to be undefined behaviour (such as when a g_return_val_if_fail() check fails), the error need not be set.
> Instead of checking separately whether the error is set, callers should ensure that they do not provoke undefined behaviour, then assume that the error will be set on failure.

Right now `gir` generates function wrappers like these:
```
let mut error = ptr::null_mut();
let _ = ffi::wrapped_fn(..., &mut error);
if error.is_null() { Ok(...) } else { Err(from_glib_full(error)) }
```

So the auto-generated Rust logic can totally miss those "undefined behaviour" cases where `FALSE` is returned but no GError is set (hopefully it never gets into those cases, but ...).
Worse, those cases will end up in the `Ok(...)` happy path from the point of view of the consumer.
The status quo is at least sketchy if not actually unsound; it would be great to have an auto-generated `assert!()` to make sure the `Ok(...)` result in not actually coming as the result of stepping into glib undefined behaviour.

sdroege:
Yes, go for it :)

cgwalters:
We had a big chat about this on Matrix (and also some between luca and me on a private chat system) and I'm just writing down my 2¢ which is:

In modern times, `g_return_if_fail()` is almost always a bad idea and should be replaced with one of:

- `g_assert()`
- Returning a GError
- Rust

In particular for ostree because the whole thing is transactional/restartable/stateless, we can always cleanly abort.  And doing so is better than potentially introducing memory corruption or undefined behavior.

lucab:
Yes for context I spot this on `ostree-rs`, where the underlying C library accumulated ~200 of those `g_return_val_if_fail (..., FALSE)` over time. And I'm guilty of adding several of them quite recently as I was unaware of this potential undefined corner case.
We can/should clean up libostree, but I think this pitfall does not only apply to this library and me.

PR #1281, opened 2021-12-03
codegen/function: assert sane GError behavior

lucab:
This introduces an assertion in generated code for FFI logic
making use of GError.
When the wrapped function returns an error code, there are some
idioms like `g_return_val_if_fail (..., FALSE)` where the GError
does not get properly set and the two mechanisms end up disagreeing.
Such case is considered undefined behavior by glib, and may indeed
result in buggy execution flow not detected as such.
In order to avoid that, gir now cross-checks the returned code and
the GError value before proceeding further.

Closes: https://github.com/gtk-rs/gir/issues/1279

### DIFF
--- a/src/writer/to_code.rs
+++ b/src/writer/to_code.rs
@@ -116,12 +116,21 @@ impl ToCode for Chunk {
                 let s = format_block_one_line(&prefix, suffix, &value_strings, "", "");
                 vec![s]
             }
-            ErrorResultReturn { ref value } => {
+            AssertErrorSanity => {
+                let assert = "assert_eq!(is_ok == 0, !error.is_null());";
+                vec![assert.to_string()]
+            }
+            ErrorResultReturn { ref ret, ref value } => {
+                let mut lines = match ret {
+                    Some(r) => r.to_code(env),
+                    None => vec![],
+                };
                 let value_strings = value.to_code(env);
                 let prefix = "if error.is_null() { Ok(";
                 let suffix = ") } else { Err(from_glib_full(error)) }";
                 let s = format_block_one_line(prefix, suffix, &value_strings, "", "");
-                vec![s]
+                lines.push(s);
+                lines
             }
             AssertInit(x) => vec![safety_assertion_mode_to_str(x).to_owned()],
             Connect {
--- a/src/chunk/chunk.rs
+++ b/src/chunk/chunk.rs
@@ -51,7 +51,9 @@ pub enum Chunk {
         condition: String,
         value: Box<Chunk>,
     },
+    AssertErrorSanity,
     ErrorResultReturn {
+        ret: Option<Box<Chunk>>,
         value: Box<Chunk>,
     },
     AssertInit(SafetyAssertionMode),
--- a/src/codegen/function_body_chunk.rs
+++ b/src/codegen/function_body_chunk.rs
@@ -864,6 +864,7 @@ impl Builder {
             name: "result".to_string(),
             is_mut: false,
             value: Box::new(Chunk::ErrorResultReturn {
+                ret: None,
                 value: Box::new(result),
             }),
             type_: None,
@@ -1190,7 +1191,9 @@ impl Builder {
                     panic!("Call without Chunk::FfiCallConversion")
                 };
                 self.remove_extra_assume_init(&array_length_name, uninitialized_vars);
+                let assert_safe_ret;
                 let call = if use_ret {
+                    assert_safe_ret = Option::None;
                     Chunk::Let {
                         name: "ret".into(),
                         is_mut: false,
@@ -1198,8 +1201,9 @@ impl Builder {
                         type_: Option::None,
                     }
                 } else {
+                    assert_safe_ret = Some(Box::new(Chunk::AssertErrorSanity));
                     Chunk::Let {
-                        name: "_".into(),
+                        name: "is_ok".into(),
                         is_mut: false,
                         value: boxed_call,
                         type_: Option::None,
@@ -1221,6 +1225,7 @@ impl Builder {
                     panic!("Return is not Tuple")
                 }
                 ret = Chunk::ErrorResultReturn {
+                    ret: assert_safe_ret,
                     value: Box::new(ret),
                 };
                 (call, Some(ret))


lucab:
I honestly unsure if I'm missing some corner cases that I'm not seeing, but I tried regenerating ostree-rs with this (on 0.14) and it keeps building+testing fine.

sdroege:
Can you create draft PRs for gtk-rs-core/gtk3-rs/gtk4-rs so we can see if it does the right thing there everywhere?

src/writer/to_code.rs
```diff
@@ -116,12 +116,23 @@ impl ToCode for Chunk {
                 let s = format_block_one_line(&prefix, suffix, &value_strings, "", "");
                 vec![s]
             }
-            ErrorResultReturn { ref value } => {
+            AssertErrorSanity => {
+                let condition = r#"(is_ok == 0) ^ error.is_null()"#;
```
GuillaumeGomez, line 120: Wut. O.o

sdroege: What's wrong with that line? :) You don't like the `r#` or the xor?

GuillaumeGomez: The xor.

sdroege: How would you express that more readable? Either one or the other must be true, but never both at once or none of the two.

GuillaumeGomez: Something like this:

```
(is_ok == 0) != error.is_null()
```

?

lucab: Yes there are certainly many other ways of writing that same condition.
If you prefer an equivalence check, perhaps you'd like an `assert_eq!()` instead so that it will also show the values on failures?

GuillaumeGomez: As long as it's easier to read, it's fine.

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me now, thanks! Please open a draft using this update so we can check the results too.

sdroege:
The results in gtk-rs-core look good to me, but I found the version with the xor more readable :shrug:

sdroege (approved):

GuillaumeGomez:
> The results in gtk-rs-core look good to me, but I found the version with the xor more readable 🤷

You're too used to C code. It took me a few seconds to understand what was going on, so xor wasn't great for me. :)

sdroege:
I would've never used that in C because there it's actually a footgun on "bool" values :)

sdroege:
This breaks in gtk3-rs:

```rust
            let is_ok =
                ffi::gtk_print_operation_get_error(self.as_ref().to_glib_none().0, &mut error);
            assert_eq!(is_ok == 0, !error.is_null());
```

That functions returns `()`.

sdroege:
That's the only function though and it was wrongly handled before already. In GStreamer I have similar functions and those got manual implementations instead.

This function should return an `Option<glib::Error>` and not a `Result<(), glib::Error>`.

@lucab Can you submit a PR for gtk3-rs to manually implement that function accordingly, and then we can get this PR here in :)

sdroege:
Ah another one in `glib`: `g_clear_error()`. Why are we generating that function at all? :) Someone was a bit too optimistic with whitelisting the global functions. That function will also always panic the way how it is generated right now.

lucab:
I'm happy this is turning out mostly fine, AND also unearthing a few odd cases.
I started looking into draft bumps by approaching gtk-rs-cores; but I realized I first need to get familiar with the repos and flows there, so I'm deferring the rest of the work to (hopefully) next week. Please keep this PR pending for the moment.

sdroege:
For the autogenerated global functions in glib/gio, I'm also going to take a look over all of them and get rid of the useless / broken ones. There are quite a few.

src/writer/to_code.rs
```diff
@@ -116,12 +116,21 @@ impl ToCode for Chunk {
                 let s = format_block_one_line(&prefix, suffix, &value_strings, "", "");
                 vec![s]
             }
-            ErrorResultReturn { ref value } => {
+            AssertErrorSanity => {
+                let assert = "assert_eq!(is_ok == 0, !error.is_null());";
```
bilelmoussaoui, line 120: Why not use the ffi value the gboolean here? It would be more readable that way

lucab: I'm not sure I'm correctly understanding the suggestion.
Isn't gboolean a type alias to an int, in Rust?
Or are you focusing on the `0` and want to replace that with `glib_sys::GFALSE`?

bilelmoussaoui:
Yes I was talking about the 0 specifically and replace it with glib::ffi::GFalse

src/writer/to_code.rs
```diff
@@ -116,12 +117,24 @@ impl ToCode for Chunk {
                 let s = format_block_one_line(&prefix, suffix, &value_strings, "", "");
                 vec![s]
             }
-            ErrorResultReturn { ref value } => {
+            AssertErrorSanity => {
+                let assert = format!(
+                    "assert_eq!(is_ok == {}, !error.is_null());",
+                    use_glib_type(env, "ffi::GFALSE")
+                );
```
sdroege, line 124: Maybe `assert_eq!(<bool as glib::translate::FromGlib>::from_glib(is_ok), error_is_null())` or so?

src/writer/to_code.rs
```diff
@@ -116,12 +116,21 @@ impl ToCode for Chunk {
                 let s = format_block_one_line(&prefix, suffix, &value_strings, "", "");
                 vec![s]
             }
-            ErrorResultReturn { ref value } => {
+            AssertErrorSanity => {
+                let assert = "assert_eq!(is_ok == glib::ffi::GFALSE, !error.is_null());";
```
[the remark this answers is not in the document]
bilelmoussaoui: Yes it's called use_glib iirc

src/writer/to_code.rs
lucab: I'm unconvinced.
This worked fine on ostree-rs and the CI is failing on a sign mismatch (`i32` vs `u32`).
I doubt further type hints would help, as it seems to be a case of `is_ok` not actually being a `gboolean`.
I think this hit some breakage further down in the line of consumers: for example `gtk_builder_add_from_resource` returns a `u32`, which is not the same type as a `gboolean`. From what I can see `FromGlib` do not (currently) cover that.

In which direction should we fix this?

sdroege: I think that function is not following the API guidelines by returning a `guint`, so having the implemented manually seems fine.

lucab:
Current checkpoint for today:
 * got rid of all the wrong autogenerated bits in gtk-rs-core / gtk3-rs / gtk4-rs / gstreamer-rs
 * found out that there some libraries which use their own "homemade booleans" (e.g. `gtk_builder_add_from_resource`)

sdroege:
Alternatively let's just work with `== 0` as before and keep this complication for the future.

lucab:
Ok, reverted to use plain zero comparison and noted down the GtkBuilder mismatch in https://github.com/gtk-rs/gtk3-rs/issues/690 (I'm not planning to tackle that right now).
This is now ready to merge, and all the relevant downstream WIPs are rebased and their CIs green too.

Post-merge, if this can be cherry-picked to the `0.14` branch, that'll make me happy :)

sdroege (approved):
Good for me. @bilelmoussaoui @GuillaumeGomez ?

bilelmoussaoui (approved):

sdroege:
@lucab Please update the other PRs now then :)

cgwalters:
Nice work on this!

PR #1285, opened 2021-12-08
docs: allow to set doc_struct_name for global functions

bilelmoussaoui:
Some of those are moved to a different struct in some cases
like gdk::Key and would be nice to have them documented

### DIFF
--- a/src/codegen/doc/mod.rs
+++ b/src/codegen/doc/mod.rs
@@ -415,9 +415,18 @@ fn create_object_doc(w: &mut dyn Write, env: &Env, info: &analysis::object::Info
             && function.parameters.iter().any(|p| p.instance_parameter)
             && !info.final_type
         {
+            if let Some(struct_name) = configured_functions
+                .iter()
+                .find_map(|f| f.doc_struct_name.as_ref())
+            {
+                (
+                    TypeStruct::new(SType::Impl, struct_name),
+                    Some(LocationInObject::Impl),
+                )
+            }
             // We use "original_name" here to be sure to get the correct object since the "name"
             // field could have been renamed.
-            if let Some(trait_name) = configured_functions
+            else if let Some(trait_name) = configured_functions
                 .iter()
                 .find_map(|f| f.doc_trait_name.as_ref())
             {
--- a/src/codegen/doc/mod.rs
+++ b/src/codegen/doc/mod.rs
@@ -170,18 +170,12 @@ fn generate_doc(w: &mut dyn Write, env: &Env) -> Result<()> {
 
         for function in functions {
             if let Some(ref c_identifier) = function.c_identifier {
-                let fn_new_name = (&global_functions.functions)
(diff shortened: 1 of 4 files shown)


src/codegen/doc/mod.rs
```diff
@@ -179,7 +179,18 @@ fn generate_doc(w: &mut dyn Write, env: &Env) -> Result<()> {
...
                     .flatten();
-                let parent = doc_trait_name.map(|p| Box::new(TypeStruct::new(SType::Trait, p)));
+                let doc_struct_name = (&global_functions.functions)
+                    .iter()
+                    .find(move |f| &f.glib_name == c_identifier)
+                    .map(|f| f.doc_struct_name.as_ref())
+                    .flatten();
+                let parent = if let Some(doc_trait_name) = doc_trait_name {
+                    Some(Box::new(TypeStruct::new(SType::Trait, doc_trait_name)))
+                } else if let Some(doc_struct_name) = doc_struct_name {
```
sdroege, line 189: clippy is angry

sdroege:
Seems good to me otherwise

src/codegen/doc/mod.rs
bilelmoussaoui: Fixed

src/codegen/doc/mod.rs
```diff
@@ -179,7 +179,24 @@ fn generate_doc(w: &mut dyn Write, env: &Env) -> Result<()> {
                     .find(move |f| &f.glib_name == c_identifier)
                     .map(|f| f.doc_trait_name.as_ref())
                     .flatten();
-                let parent = doc_trait_name.map(|p| Box::new(TypeStruct::new(SType::Trait, p)));
+                let doc_struct_name = (&global_functions.functions)
+                    .iter()
+                    .find(move |f| &f.glib_name == c_identifier)
+                    .map(|f| f.doc_struct_name.as_ref())
+                    .flatten();
```
sdroege, line 186:
```suggestion
                    .and_then(|f| f.doc_struct_name.as_ref());
```

?

bilelmoussaoui: ah, for whatever reason i didn't notice we have the same thing a few lines above that uses and_then :)

sdroege (approved):

bilelmoussaoui:
I pushed another commit to not generate docs for functions that are not generated, this reduces the size of the generated docs.txt (to hopefully completely remove it which means all the generated docs gets embedded properly)

bilelmoussaoui:
Don't merge yet, i noticed part of 9bb9c111cb945130ac2814b2996377f9d427f7fc that doesn't behave as it should

bilelmoussaoui:
Should be good to go now, there are more stuff to cleanup but are unrelated to those changes

sdroege (approved):
@GuillaumeGomez ?

GuillaumeGomez (approved):

PR #1286, opened 2021-12-08
codgen: trim extra _async in future variants

### LINKED ISSUE #1208: Improve naming of native async methods
Whenever (I am looking at you gtk-rs/gtk-rs-core#203) there are bindings to async function `fn f` in gio, there is also a `async fn f_future`. This leads to excruciatingly long function names like `replace_contents_async_future`, or even worse the not-yet-implemented `replace_contents_bytes_async_future`.

It could make more sense to just do a `s/async/future/`, e.g. `replace_contents_future` instead of `replace_contents_async_future`.

bilelmoussaoui:
Fixes #1208

### DIFF
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -1037,7 +1037,7 @@ fn analyze_async(
         if !no_future {
             *async_future = Some(AsyncFuture {
                 is_method,
-                name: format!("{}_future", codegen_name),
+                name: format!("{}_future", codegen_name.trim_end_matches("_async")),
                 success_parameters,
                 error_parameters,
                 assertion: match SafetyAssertionMode::of(env, is_method, parameters) {


bilelmoussaoui:
It needs manual changes for the various crates as a bunch of the future variants are manually implemented

sdroege:
You might also want to do something about `_sync` then (GDbus) in the opposite direction. Also what does the generated code here look like afterwards?

bilelmoussaoui:
> You might also want to do something about `_sync` then (GDbus) in the opposite direction. Also what does the generated code here look like afterwards?

I will handle that one later today and send a regen PR to gt-rs-core & others

bilelmoussaoui:
Looking at the various `_sync` functions, here's what I found
image

Replacing those with dropping the `_sync` means renaming the variants without the `_sync` to `_async` or something else. As there are only very few functions, we should probably do the renames on Gir's config instead of automatically handling it here

bilelmoussaoui:
Before we merge this, can't we just mark the functions as `async` and drop the whole future prefix ?

sdroege:
> Before we merge this, can't we just mark the functions as `async` and drop the whole future prefix ?

How would that solve anything? `async fn foo()` and `fn foo()` can't co-exist in the some place as they have the same name :)

> Replacing those with dropping the `_sync` means renaming the variants without the `_sync` to `_async` or something else. As there are only very few functions, we should probably do the renames on Gir's config instead of automatically handling it here

Makes sense for now, yes

bilelmoussaoui:
> > Before we merge this, can't we just mark the functions as `async` and drop the whole future prefix ?
> 
> How would that solve anything? `async fn foo()` and `fn foo()` can't co-exist in the some place as they have the same name :)

Yeah, well that can't work neither in traits so it's not a viable approach for now. 

> > Replacing those with dropping the `_sync` means renaming the variants without the `_sync` to `_async` or something else. As there are only very few functions, we should probably do the renames on Gir's config instead of automatically handling it here
> 
> Makes sense for now, yes

I suppose this can land as is

PR #1287, opened 2021-12-14
writer: check return code as gboolean

lucab:
This makes the check on return code stricter, by using the actual
gboolean value.

Ref: https://github.com/gtk-rs/gtk3-rs/pull/691#issuecomment-992355228


sdroege (approved):

bilelmoussaoui (approved):

sdroege:
@lucab Please update all the PRs then and then let's get it in :) Thanks!

PR #1290, opened 2021-12-17
Update some dependencies

sdroege:


GuillaumeGomez (approved):

MarijnS95:
Any reason for `gir` to use a lock file and not just pick whatever is latest based on (hopefully proper) `Cargo.toml` and dependencies that adhere to semver?

GuillaumeGomez:
Because it's a binary, not a lib.

MarijnS95:
Sure enough `Cargo.lock` only works for binaries (and local building/testing of library crates), but the question is whether we absolutely need it to get reproducible results? I guess yes, that's why it's here?

sdroege:
I think it's useless for our purposes (and also many others: `cargo install` from crates.io ignores `Cargo.lock` FWIW), but it's there so I updated it ;)

sdroege:
Not even sure if the `Cargo.lock` is going to be used if `gir` is used as a submodule of a workspace.

MarijnS95:
> Not even sure if the `Cargo.lock` is going to be used if `gir` is used as a submodule of a workspace.

Seems like it is explicitly excluded from the top-level workspace:

https://github.com/gtk-rs/gtk-rs-core/blob/f710e5999dad4a656df0dac3b36fb1a348e69bf2/Cargo.toml#L22

PR #1292, opened 2021-12-26
Allow tweaking visibility 

### LINKED ISSUE #1238: Support custom visibility on generated functions
This is a new-feature request to support generating specific functions with custom visibility.

Right now auto-generated functions get a `pub` visibility modifier by default. In some cases, it would be useful to auto-generate some entries with `pub(crate)` visibility (for example), so that they can be used to build higher-level ergonomic Rust patterns in the same crate without exposing the low-level primitives in the public API.

For a concrete example, `libostree` has a few `ostree_sysroot_<foo>` [helpers](https://ostreedev.github.io/ostree/reference/reference.html#api-index-full) which rely on state (e.g. loaded, ro-locked, rw-locked, etc.) that could be encoded in a more ergonomic way in the Rust type-system.
The auto-generated functions are useful as building blocks for the higher-level types, but they should not end up in the final public surface of the crate. A `pub(crate)` modifier on those would suit better in this case.

bilelmoussaoui:
This PR allows to tweak the visibility of functions/enums/flags & also objects/records. For global functions that are marked as non-public, they also have a `#[allow(dead_code)]` guard added to avoid any clippy complaints.

Fixes #1238 
cc @lucab

### DIFF
--- a/src/analysis/functions.rs
+++ b/src/analysis/functions.rs
@@ -36,24 +36,6 @@ use std::{
 
 use super::special_functions;
 
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub enum FuncVisibility {
-    Public,
-    Comment,
-    Private,
-    Hidden,
-}
-
-impl FuncVisibility {
-    pub fn hidden(self) -> bool {
-        self == Self::Hidden
-    }
-
-    pub fn code_visible(self) -> bool {
-        matches!(self, Self::Private | Self::Public)
-    }
-}
-
 #[derive(Clone, Debug)]
 pub struct AsyncTrampoline {
     pub is_method: bool,
@@ -82,7 +64,6 @@ pub struct Info {
     pub glib_name: String,
     pub status: GStatus,
     pub kind: library::FunctionKind,
-    pub func_visibility: FuncVisibility,
     pub visibility: Visibility,
     pub type_name: Result,
     pub parameters: Parameters,
@@ -105,6 +86,11 @@ pub struct Info {
     pub destroys: Vec<Trampoline>,
     pub remove_params: Vec<usize>,
     pub async_future: Option<AsyncFuture>,
+    /// Whether the function is hidden (an implementation detail)
+    /// Like the ref/unref/copy/free functions
+    pub hidden: bool,
+    /// Whether the function can't be generated
+    pub commented: bool,
 }
 
 impl Info {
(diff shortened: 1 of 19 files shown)


src/analysis/functions.rs
```diff
@@ -36,14 +37,14 @@ use std::{
 use super::special_functions;
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub enum Visibility {
+pub enum FuncVisibility {
```
bilelmoussaoui, line 40: I'm not really sure how this enum type is fit for the rest of the code. It confuses two things, a function visibility and whether gir is capable of generating the code or not (when it's set to Comment). I suppose we should ideally drop it and move the Comment to a new Status mode for "not capable of generating"

bilelmoussaoui:
You can see the glib crate re-generated with the changes of this PR at https://github.com/gtk-rs/gtk-rs-core/pull/412/commits/27e611dc7163ad1974c74433266a385e4e957c10

src/codegen/visibility.rs
```diff
@@ -0,0 +1,75 @@
...
+        Self::Public
+    }
+}
+
+impl fmt::Display for Visibility {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Public => write!(f, "pub"),
+            Self::Crate => write!(f, "pub(crate)"),
+            Self::Mod => write!(f, "mod"),
```
lucab, line 41: This one seems weird, where is it coming from? (Or, in which context does this come up?)

Was this maybe meant to be the entry for `pub(self)`?

bilelmoussaoui: Ah I actually don't think we want to have a "mod" visibility modifier here as I don't see where it could be used

src/analysis/functions.rs
bilelmoussaoui: I have dropped the enum and replaced it with two variants of GStatus, ideally the later (GStatus::Hidden) should be a different attribute of the analysis::Info, any opinions?

bilelmoussaoui: with this change, it gets pretty interesting; we can finally drop all the ignore=true we had in Gir.toml in order to not have a doc link because the function can't be generated for whatever reasons

src/codegen/visibility.rs
sdroege: Yes that seems useless

sdroege (commented):
Looks good to me otherwise

bilelmoussaoui:
Updated & added documentation to the book

sdroege (approved):

lucab:
Thanks for landing this!

PR #1293, opened 2021-12-28
parser: ignore private records

bilelmoussaoui:
This should fix the generation of gtk3 using the latest gir files
As some private structs lack a type on their fields, so better ignore them

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -265,6 +265,11 @@ impl Library {
         parent_ctype_prefix: Option<&str>,
     ) -> Result<Option<Type>, String> {
         let record_name = elem.attr_required("name")?;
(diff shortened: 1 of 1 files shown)


src/parser.rs
```diff
@@ -265,6 +265,11 @@ impl Library {
         parent_ctype_prefix: Option<&str>,
     ) -> Result<Option<Type>, String> {
         let record_name = elem.attr_required("name")?;
+        // Records starting with `_` are intended to be privat and should not be bound
```
sdroege, line 268:
```suggestion
        // Records starting with `_` are intended to be private and should not be bound
```

sdroege (approved):

PR #1294, opened 2021-12-29
support fundamental types

### LINKED ISSUE #1069: Don't use IsA for fundamental types
Currently gir, uses `IsA` for generated code for any fundamental type. What would be nice is to have gir know the difference and only use `IsA` for objects.

bilelmoussaoui:
This merge request allows generating a portion of the code needed for fundamental types. It still needs manual code to create a common trait & implement it for the various sub-types along with the casting functions.

The PR as is, allows to drop a huge portion of the manual code in gtk4-rs. 

The next step before merging this is to automatically use `AsRef<T>` instead of `IsA<T>` where needed. 

Fixes #1069


MarijnS95:
Seeing the title, is this something I could leverage for https://github.com/gtk-rs/gir/pull/1134?

bilelmoussaoui:
> Seeing the title, is this something I could leverage for #1134?

Not really, the ones this PR attempts to fix are fundamental types in "glib terms", so they do provide a ref/unref/get_type functions

MarijnS95:
That's unfortunate, I thought this added a way to specify fundamental types (somehow) in `.gir` and generate them here.

sdroege:
Can you create a WIP PR for gtk4-rs?

bilelmoussaoui:
> Can you create a WIP PR for gtk4-rs?

Will do once it's +/- compiling locally

src/codegen/general.rs
```diff
@@ -113,6 +113,91 @@ fn format_parent_name(env: &Env, p: &StatusedTypeId) -> String {
...
+                }
+            })
+            .next()
+            .unwrap();
+        // otherwise get that information from the parent
+        (
+            ref_fn,
+            unref_fn,
+            format!("ptr as *mut {}::{}", sys_crate_name, parent_c_type),
+        )
```
bilelmoussaoui, line 166: Just noticed thats the sys_crate_name here is wrong, it should get the one of the parent type in case it comes from a different crate

bilelmoussaoui:
It should be ready for a review :) It can properly generate things like `ParamSpecExpression`

src/parser.rs
```diff
@@ -152,6 +152,16 @@ impl Library {
...
+        let is_fundamental = elem.attr("fundamental").map(|x| x == "1").unwrap_or(false);
+        let (ref_fn, unref_fn) = if is_fundamental {
+            (
+                elem.attr("ref-func").map(ToOwned::to_owned),
+                elem.attr("unref-func").map(ToOwned::to_owned),
+            )
+        } else {
+            (None, None)
+        };
+
```
sdroege, line 164: There's also `get-value-func` / `set-value-func` which you can use to implement the `glib::Value` traits automatically here

sdroege:
Looks good to me otherwise. Less code than I feared :)

src/parser.rs
bilelmoussaoui: The thing I noticed for those is only GtkExpression & GskRenderNode that have a `get-value-func` / `set-value-func` and for GskRenderNode they are behind a feature guard (were only added in v4_6),  not sure if it's worth the change in gir for now

sdroege: Ideally they would also be added for `GdkEvent` so that the way how they're currently passed through signals in C is less of a disaster but OK with me to do that later. Please create an issue about that then :)

sdroege:
Good to merge then?

src/parser.rs
bilelmoussaoui: Done https://github.com/gtk-rs/gir/issues/1299

bilelmoussaoui:
Looks good to me yes

PR #1296, opened 2021-12-29
analysis: don't take a slice of copy types by ref

### LINKED ISSUE #1247: Handle copy types for slice parameters
Currently, if a function takes a slice as a parameter, it uses the `&[&SomeType]` by default even if the type is Copy. Instead it should use `&[SomeType]`.

This will allow us to drop a bunch of manual code where a `&[gdk::RGBA]` is used for example.

bilelmoussaoui:
Currently, gir generates the following &[&SomeCopyType] instead of &[SomeCopyType]

Fixes #1247

### DIFF
--- a/src/analysis/rust_type.rs
+++ b/src/analysis/rust_type.rs
@@ -1,6 +1,6 @@
 use super::conversion_type::ConversionType;
 use crate::{
-    analysis::{ref_mode::RefMode, try_from_glib::TryFromGlib},
+    analysis::{record_type::RecordType, ref_mode::RefMode, try_from_glib::TryFromGlib},
     config::functions::{CallbackParameter, CallbackParameters},
     env::Env,
     library::{self, Nullable, ParameterDirection, ParameterScope},
@@ -345,8 +345,19 @@ impl<'env> RustTypeBuilder<'env> {
                 if ConversionType::of(self.env, inner_tid) == ConversionType::Pointer =>
             {
                 skip_option = true;
-                let inner_ref_mode = match self.env.library.type_(inner_tid) {
+                let inner_ref_mode = match self.env.type_(inner_tid) {
                     Class(..) | Interface(..) => RefMode::None,
+                    Record(record) => match RecordType::of(record) {
+                        RecordType::Boxed => RefMode::None,
+                        RecordType::AutoBoxed => {
+                            if record.has_copy() {
+                                RefMode::None
+                            } else {
+                                self.ref_mode
+                            }
+                        }
+                        _ => self.ref_mode,
+                    },
                     _ => self.ref_mode,
                 };
                 RustType::builder(self.env, inner_tid)
@@ -367,7 +378,7 @@ impl<'env> RustTypeBuilder<'env> {
             CArray(inner_tid)
                 if ConversionType::of(self.env, inner_tid) == ConversionType::Direct =>
             {
-                if let Fundamental(fund) = self.env.library.type_(inner_tid) {
+                if let Fundamental(fund) = self.env.type_(inner_tid) {
                     let array_type = match fund {
                         Int8 => Some("i8"),
                         UInt8 => Some("u8"),
--- a/src/library.rs
+++ b/src/library.rs
@@ -445,6 +445,30 @@ pub struct Record {
     pub disguised: bool,
 }
 
+impl Record {
+    pub fn has_free(&self) -> bool {
+        self.functions.iter().any(|f| f.name == "free") || (self.has_copy() && self.has_destroy())
+    }
+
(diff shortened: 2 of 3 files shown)


bilelmoussaoui:
Hmm, seems like GdkAtom needs some trait to be implemented

bilelmoussaoui:
I fixed the gtk3-rs issue on https://github.com/gtk-rs/gtk3-rs/pull/695, it requires this being merged first & then the gtk3-rs change after.

src/analysis/rust_type.rs
```diff
@@ -347,6 +347,12 @@ impl<'env> RustTypeBuilder<'env> {
                 skip_option = true;
                 let inner_ref_mode = match self.env.library.type_(inner_tid) {
                     Class(..) | Interface(..) => RefMode::None,
+                    Record(..) => {
+                        match RecordType::of(self.env.type_(inner_tid).maybe_ref().unwrap()) {
+                            RecordType::AutoBoxed | RecordType::Boxed => RefMode::None,
+                            _ => self.ref_mode,
```
sdroege, line 353: I think this can use some more auto-detection here. If we generate an `BoxedInline` then it should do this automatically, otherwise based on `ref_mode` configuration or fall back to the default.

sdroege: Hmm but not all `BoxedInline` are `Copy`. Do we have a way to detect this better?

bilelmoussaoui: > Hmm but not all `BoxedInline` are `Copy`. Do we have a way to detect this better?

I updated the code to handle that. 

> I think this can use some more auto-detection here. If we generate an BoxedInline then it should do this automatically, otherwise based on ref_mode configuration or fall back to the default.

Not sure what kind of auto-detection that would be needed here?

src/library.rs
```diff
@@ -445,6 +445,28 @@ pub struct Record {
     pub disguised: bool,
 }
 
+impl Record {
+    pub fn has_free(&self) -> bool {
+        self.functions.iter().any(|f| f.name == "free") || (self.is_copy() && self.has_destroy())
+    }
+
+    pub fn is_copy(&self) -> bool {
```
sdroege, line 453:
```suggestion
    pub fn has_copy(&self) -> bool {
```

If it has a `copy` function then the type does not implement the `Copy` trait but only `Clone`.

There's also the `copy_into` function btw

bilelmoussaoui: > There's also the copy_into function btw

Should we check for that as well when checking the `has_copy` ?

sdroege: Yes. If either exists then the type can't be `Copy` (as copying is more than a plain `memcpy`)

bilelmoussaoui: Done, and after adding the check for the copy type. It no longer generates `&[gdk::Atom]` but instead it generates `&[&gdk::Atom]` which makes the gtk3-rs change unneeded.

sdroege: Ah why? `gdk::Atom` is `Copy`

bilelmoussaoui: Because it doesn't provide a `copy` function, maybe we need to check for something else as well?

sdroege: Yes, because it does **not** provide a `copy` / `copy_into` function it implements `Copy` (i.e. copying/cloning it is just a `memcpy` and nothing else)

src/analysis/rust_type.rs
sdroege: Seems fine for now I guess. If we need more then we can add that later

sdroege (approved):
This can be merged then, and the fundamental type PR is also ready AFAICS so let's do all this together @bilelmoussaoui ?

sdroege:
This just needs a `cargo fmt` run first:

```diff
 Diff in /home/runner/work/gir/gir/src/library.rs at line 451:
     }
 
     pub fn has_copy(&self) -> bool {
-        self.functions.iter().any(|f| f.name == "copy" || f.name == "copy_into")
+        self.functions
+            .iter()
+            .any(|f| f.name == "copy" || f.name == "copy_into")
     }
 
     pub fn has_destroy(&self) -> bool {
```

bilelmoussaoui:
Done, thanks!

bilelmoussaoui:
Wait before merging this, it still produces wrong code for GdkAtom btw :)

sdroege:
> Wait before merging this, it still produces wrong code for GdkAtom btw :)

Bah, too late. What wrong code does it produce?

bilelmoussaoui:
Ha, it still produces `&[&gdk::Atom]` instead of `&[gdk::Atom]`

bilelmoussaoui:
No worries, will send a fix

Issue #1295, opened 2021-12-29
Panic on GIR not_bound mode for GModule


wash2:
I'm working on generating bindings to GModule but it panics in not_bound work_mode. I added some debug statements to the point where it was unwrapping, and I've attached the output of each of the failing types:
```
[WARN  libgir::analysis::functions] Function g_module_symbol has unsupported outs
[WARN  libgir::analysis::functions] Function g_module_open_full has unsupported outs
[ERROR libgir::analysis::record] Missing memory management functions for GModule.Module
[src/library.rs:1091] &tid = TypeId {
    ns_id: 1,
    id: 0,
}
[src/library.rs:1092] &full_name = "GModule.Module"
[src/library.rs:1093] env.library.find_type(0, "GObject.Object") = None
[src/library.rs:1091] &tid = TypeId {
    ns_id: 1,
    id: 1,
}
[src/library.rs:1092] &full_name = "GModule.ModuleFlags"
[src/library.rs:1093] env.library.find_type(0, "GObject.Object") = None
[src/library.rs:1091] &tid = TypeId {
    ns_id: 1,
    id: 2,
}
[src/library.rs:1092] &full_name = "GModule.ModuleCheckInit"
[src/library.rs:1093] env.library.find_type(0, "GObject.Object") = None
[src/library.rs:1091] &tid = TypeId {
    ns_id: 1,
    id: 3,
}
[src/library.rs:1092] &full_name = "GModule.ModuleError"
[src/library.rs:1093] env.library.find_type(0, "GObject.Object") = None
[src/library.rs:1091] &tid = TypeId {
    ns_id: 1,
    id: 4,
}
[src/library.rs:1092] &full_name = "GModule.ModuleUnload"
[src/library.rs:1093] env.library.find_type(0, "GObject.Object") = None

```

Also, it seems like the generation is failing, partially bc of missing memory management functions and unsupported output.. Does this mean that the safe bindings would all have to be implemented manually, or is it maybe an issue with the GIR file?
```
[WARN  libgir::analysis::functions] Function g_module_symbol has unsupported outs
[WARN  libgir::analysis::functions] Function g_module_open_full has unsupported outs
[ERROR libgir::analysis::record] Missing memory management functions for GModule.Module
thread 'main' panicked at 'Record Module has record_boxed=true but don't have glib:get_type function', src/codegen/record.rs:31:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

[...]

Issue #1236, opened 2021-10-20
Support BoxedInline without a get_type function


bilelmoussaoui:
currently gir panics when it encounters a type that doesn't have a get_type function
```
thread 'main' panicked at 'Record ToplevelSize has record_boxed=true but don't have glib:get_type function', src/codegen/record.rs:31:13
```
cc @sdroege

bilelmoussaoui:
Closing as a duplicate of #1297

PR #1297, opened 2021-12-29
Support BoxedInline without a get_type function #1295 #1236

wash2:
This is my attempt at supporting BoxedInline without a get_type function. Generating the bindings for gmodule does not panic and now outputs:
```
[WARN  libgir::analysis::functions] Function g_module_symbol has unsupported outs
[WARN  libgir::analysis::functions] Function g_module_open_full has unsupported outs
[ERROR libgir::analysis::record] Missing memory management functions for GModule.Module
```

### DIFF
--- a/src/codegen/general.rs
+++ b/src/codegen/general.rs
@@ -377,7 +377,7 @@ pub fn define_auto_boxed_type(
     init_function_expression: &Option<String>,
     copy_into_function_expression: &Option<String>,
     clear_function_expression: &Option<String>,
-    get_type_fn: Option<(String, Option<Version>)>,
+    get_type_fn: &Option<(String, Option<Version>)>,
     derive: &[Derive],
 ) -> Result<()> {
     let sys_crate_name = env.main_sys_crate_name();
@@ -394,13 +394,13 @@ pub fn define_auto_boxed_type(
     )?;
     writeln!(w)?;
     writeln!(w, "\tmatch fn {{")?;
-    if let Some(get_type_fn) = get_type_fn.clone() {
+    if let Some((ref get_type_fn, _get_type_version)) = get_type_fn {
         writeln!(
             w,
             "\t\tcopy => |ptr| {}({}::{}(), ptr as *mut _) as *mut {}::{},",
             use_glib_type(env, "gobject_ffi::g_boxed_copy"),
             sys_crate_name,
-            get_type_fn.0,
+            get_type_fn,
             sys_crate_name,
             glib_name
         )?;
@@ -409,7 +409,7 @@ pub fn define_auto_boxed_type(
             "\t\tfree => |ptr| {}({}::{}(), ptr as *mut _),",
             use_glib_type(env, "gobject_ffi::g_boxed_free"),
             sys_crate_name,
-            get_type_fn.0
+            get_type_fn
         )?;
     }
 
@@ -426,11 +426,11 @@ pub fn define_auto_boxed_type(
         writeln!(w, "\t\tcopy_into => {},", copy_into_function_expression,)?;
         writeln!(w, "\t\tclear => {},", clear_function_expression,)?;
     }
-    if let Some(get_type_fn) = get_type_fn {
+    if let Some((ref get_type_fn, _get_type_version)) = get_type_fn {
         writeln!(
             w,
             "\t\ttype_ => || {}::{}(),",
-            sys_crate_name, get_type_fn.0
+            sys_crate_name, get_type_fn
         )?;
     }
     writeln!(w, "\t}}")?;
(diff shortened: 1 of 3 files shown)


src/codegen/general.rs
```diff
@@ -394,22 +394,24 @@ pub fn define_auto_boxed_type(
...
-        glib_name
-    )?;
-    writeln!(
-        w,
-        "\t\tfree => |ptr| {}({}::{}(), ptr as *mut _),",
-        use_glib_type(env, "gobject_ffi::g_boxed_free"),
-        sys_crate_name,
-        get_type_fn
-    )?;
+    if let Some((ref get_type_fn, _get_type_version)) = get_type_fn {
```
sdroege, line 397: This requires a `get_type` function, or otherwise more changes in `glib` are needed

sdroege: For the case of not `BoxedInline` but normal `Boxed`

src/library.rs
```diff
@@ -1085,18 +1085,19 @@ impl Library {
                 }
             }
             if let Some(tid) = env.library.find_type(0, &full_name) {
-                let gobject_id = env.library.find_type(0, "GObject.Object").unwrap();
+                let gobject_id = env.library.find_type(0, "GObject.Object");
```
sdroege, line 1088: If you don't have `GObject` in your dependencies then there can't be any class hierarchy at all, so the code below should simply not run

sdroege (commented):
Otherwise makes sense.

Why do you want to create bindings for `GModule` though? I think that's going to require writing the whole bindings manually to end up with anything usable and somewhat safe. The autogenerated bindings won't be usable.

src/codegen/record.rs
```diff
@@ -14,25 +14,18 @@ pub fn generate(w: &mut dyn Write, env: &Env, analysis: &analysis::record::Info)
...
+        general::define_auto_boxed_type(
+            w,
+            env,
+            &analysis.name,
+            &type_.c_type,
+            analysis.boxed_inline,
+            &analysis.init_function_expression,
+            &analysis.copy_into_function_expression,
+            &analysis.clear_function_expression,
+            &analysis.glib_get_type,
```
bilelmoussaoui, line 26: maybe `analysis.glib_get_type.map(|ref a,_| a)` or something like that to avoid passing the version that's unneeded here ?

src/library.rs
wash2: Ok gotcha

src/codegen/general.rs
wash2: So the get_type_fn here can't be None then?

sdroege: Well, there won't be a copy/free function here otherwise and that won't compile without further changes in glib

src/codegen/record.rs
wash2: Ok

wash2:
> Why do you want to create bindings for `GModule` though? I think that's going to require writing the whole bindings manually to end up with anything usable and somewhat safe. The autogenerated bindings won't be usable.

I want to use GModule to support "plug-ins"
I was not aware that the auto-generated bindings would not be usable, but I could try manually writing safe bindings if necessary.

sdroege:
> I was not aware that the auto-generated bindings would not be usable

The `-sys` bindings should be fine, the non-`sys` ones probably only have very few functions generated and you'll have to be very careful here with memory safety. Take a look at all the complications in the [`libloading`](https://docs.rs/libloading/latest/libloading/index.html) API for example.

wash2:
> > I was not aware that the auto-generated bindings would not be usable
> 
> The `-sys` bindings should be fine, the non-`sys` ones probably only have very few functions generated and you'll have to be very careful here with memory safety. Take a look at all the complications in the [`libloading`](https://docs.rs/libloading/latest/libloading/index.html) API for example.

Ok, thanks, I'll check that out

src/codegen/general.rs
```diff
@@ -377,7 +377,7 @@ pub fn define_auto_boxed_type(
     init_function_expression: &Option<String>,
     copy_into_function_expression: &Option<String>,
     clear_function_expression: &Option<String>,
-    get_type_fn: &str,
+    get_type_fn: Option<&String>,
```
bilelmoussaoui, line 380:
```suggestion
    get_type_fn: Option<&str>,
```

src/codegen/record.rs
```diff
@@ -14,25 +14,18 @@ pub fn generate(w: &mut dyn Write, env: &Env, analysis: &analysis::record::Info)
...
+        general::define_auto_boxed_type(
+            w,
+            env,
+            &analysis.name,
+            &type_.c_type,
+            analysis.boxed_inline,
+            &analysis.init_function_expression,
+            &analysis.copy_into_function_expression,
+            &analysis.clear_function_expression,
+            analysis.glib_get_type.as_ref().map(|(ref a, _)| a),
```
bilelmoussaoui, line 26:
```suggestion
            analysis.glib_get_type.as_ref().map(|(ref a, _)| a).as_deref(),
```

bilelmoussaoui (commented):
Wouldn't this work ?

PR #1301, opened 2022-01-01
Generate #[must_use] where needed

GuillaumeGomez:
You can see the result [here](https://github.com/gtk-rs/gtk-rs-core/compare/master...GuillaumeGomez:regen-must-use?expand=1).

It is the second step of https://github.com/gtk-rs/gtk-rs-core/pull/445.

### DIFF
--- a/src/codegen/function.rs
+++ b/src/codegen/function.rs
@@ -5,14 +5,14 @@ use super::{
         version_condition,
     },
     parameter::ToParameter,
-    return_value::{out_parameters_as_return, ToReturnValue},
+    return_value::{out_parameter_types, out_parameters_as_return, ToReturnValue},
     special_functions,
 };
 use crate::{
     analysis::{self, bounds::Bounds, namespaces, try_from_glib::TryFromGlib},
     chunk::{ffi_function_todo, Chunk},
     env::Env,
-    library,
+    library::{self, TypeId},
     version::Version,
     writer::{primitives::tabs, safety_assertion_mode_to_str, ToCode},
 };
@@ -23,9 +23,38 @@ use std::{
     result::Result as StdResult,
 };
 
+// We follow the rules of the `return_self_not_must_use` clippy lint:
+//
+// If `Self` is returned (so `-> Self`) in a method (whatever the form of the `self`), then
+// the `#[must_use]` attribute must be added.
+pub fn get_must_use_if_needed(
+    parent_type_id: Option<TypeId>,
+    analysis: &analysis::functions::Info,
+    in_trait: bool,
+    only_declaration: bool,
+    comment_prefix: &str,
+) -> Option<String> {
+    // If there is no parent, it means it's not a (trait) method so we're not interested.
+    if let Some(parent_type_id) = parent_type_id {
+        // Check it's a trait declaration or a method declaration (outside of a trait
+        // implementation).
+        if analysis.kind == library::FunctionKind::Method && (!in_trait || only_declaration) {
+            // We now get the list of the returned types.
+            let outs = out_parameter_types(analysis);
+            // If there is only one type returned, we check if it's the same type as `self` (stored
+            // in `parent_type_id`).
+            if [parent_type_id] == *outs.as_slice() {
+                return Some(format!("{}#[must_use]\n", comment_prefix));
+            }
+        }
+    }
+    None
+}
+
 pub fn generate(
     w: &mut dyn Write,
     env: &Env,
+    parent_type_id: Option<TypeId>,
     analysis: &analysis::functions::Info,
     special_functions: Option<&analysis::special_functions::Infos>,
     scope_version: Option<Version>,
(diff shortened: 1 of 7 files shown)


src/codegen/function.rs
```diff
@@ -23,9 +25,92 @@ use std::{
...
+        // that into account. If the opposite, it means that there are two arguments in any case so
+        // we need "()" too.
+        let num_outs = std::cmp::max(num_out_args, num_out_sizes);
+        match analysis.outs.mode {
+            Mode::Normal | Mode::Combined if num_outs == 1 => {
+                let array_lengths: Vec<_> = analysis
+                    .outs
+                    .iter()
+                    .filter_map(|out| out.lib_par.array_length)
+                    .collect();
```
sdroege, line 63: clippy doesn't like this `collect()`

src/codegen/function.rs
```diff
@@ -23,9 +25,92 @@ use std::{
...
+    only_declaration: bool,
+    comment_prefix: &str,
+) -> Option<String> {
+    if analysis.ret.bool_return_is_error.is_some()
+        || analysis.ret.nullable_return_is_error.is_some()
+    {
+        return None;
+    }
+
+    let return_type = if !analysis.outs.is_empty() {
```
sdroege, line 41: Don't we have exactly this code somewhere else already?

src/codegen/function.rs
```diff
@@ -23,9 +25,92 @@ use std::{
...
+    if let Some(ref return_type) = return_type {
+        if let Some(parent_type_id) = parent_type_id {
+            if analysis.kind == library::FunctionKind::Method
+                && (!in_trait || only_declaration)
+                && parent_type_id == *return_type
+            {
+                return Some(format!("{}#[must_use]\n", comment_prefix));
+            }
+        }
+    }
```
bilelmoussaoui, line 106: Completely unrelated to what you're working on, but this code could probably be re-used to check whether we should return Self or SomeTypeName; that should allow us to drop yet another `#[allow(some_clippy_warning)]`

src/codegen/function.rs
```diff
@@ -23,9 +25,92 @@ use std::{
...
+            _ => None,
+        }
+    } else {
+        analysis.ret.parameter.as_ref().map(|out| out.lib_par.typ)
+    };
+    if let Some(ref return_type) = return_type {
+        if let Some(parent_type_id) = parent_type_id {
+            if analysis.kind == library::FunctionKind::Method
+                && (!in_trait || only_declaration)
+                && parent_type_id == *return_type
```
sdroege, line 101: This should get some comments to explain what it actually does. AFAIU you add the annotation on every method that returns the type itself?

src/codegen/function.rs
GuillaumeGomez: Kinda but not the same. :-/

src/codegen/function.rs
GuillaumeGomez: Ah true.

src/codegen/function.rs
GuillaumeGomez: Good idea.

src/codegen/function.rs
sdroege: That's surprising :) We're just building the return type as a string here or not?

GuillaumeGomez: Here we just check if there is one returned type and if it's the same as `self`. However I could check if we could somehow merge both functions.

src/codegen/function.rs
GuillaumeGomez: Ah right, need to ignore it here as well.

GuillaumeGomez:
I reworked the code a bit to make it clearer. @bilelmoussaoui's suggestion (returning `Self` when possible) could be a nice follow-up.

src/codegen/function.rs
sdroege: And can we? :)

src/codegen/function.rs
sdroege: You didn't add a comment to the code yet

src/codegen/function.rs
```diff
@@ -23,9 +23,30 @@ use std::{
...
+    analysis: &analysis::functions::Info,
+    in_trait: bool,
+    only_declaration: bool,
+    comment_prefix: &str,
+) -> Option<String> {
+    if let Some(parent_type_id) = parent_type_id {
+        if analysis.kind == library::FunctionKind::Method && (!in_trait || only_declaration) {
+            let outs = out_parameter_types(analysis);
+            if let [return_type] = *outs.as_slice() {
+                if parent_type_id == return_type {
```
sdroege, line 37:
```suggestion
            if [parent_type_id] == *outs.as_slice() {
```

Maybe?

src/codegen/function.rs
GuillaumeGomez: No. :'(

src/codegen/function.rs
GuillaumeGomez: Wait that works? O.o

GuillaumeGomez: No it doesn't work. You can check it with:

```rust
fn check(s: &[String], matcher: &String) {
    if let [matcher] = s {
        eprintln!("yolo!");
    }
}

fn main() {
    let matcher = "b".to_owned();
    check(&vec!["a".to_owned()], &matcher);
    check(&vec!["b".to_owned()], &matcher);
}
```

GuillaumeGomez:
I added code comments. :)

src/codegen/function.rs
GuillaumeGomez: Ah indeed, updating then!

sdroege (approved):

PR #1303, opened 2022-01-02
codegen/special_functions: Add missing space between `pub` and `fn`

MarijnS95:
Originally the `visibility` modifier contained this space ("pub ") to keep indentation in check.  A recent change however formats the visibility modifier from configuration directly into the file, without inserting the necessary trailing space, resulting in the text `pubfn` in `special_functions`, which is invalid syntax.

Since output is almost always formatted with rustfmt anyway, this indentation mismatch is irrelevant to allocate a string for here.

Fixes: e3bd76a3 ("replace FunctionVisibility with hidden/commented attributes"), cc @bilelmoussaoui

### DIFF
--- a/src/codegen/special_functions.rs
+++ b/src/codegen/special_functions.rs
@@ -40,7 +40,7 @@ pub(super) fn generate_static_to_str(
     writeln!(
         w,
         "\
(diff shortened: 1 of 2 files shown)


bilelmoussaoui:
Thanks for the fixes!

sdroege (approved):

PR #1311, opened 2022-01-19
sys: run ABI tests only under linux

bilelmoussaoui:
they need more work to work on other platforms for now

### DIFF
--- a/src/codegen/signal.rs
+++ b/src/codegen/signal.rs
@@ -160,9 +160,7 @@ fn function_type_string(
     analysis: &analysis::signals::Info,
     closure: bool,
 ) -> Option<String> {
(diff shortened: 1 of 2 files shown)


sdroege:
Can't this be put directly into the `Cargo.toml`?

bilelmoussaoui:
No idea, it would also mean writing more code to override more stuff in the -sys Cargo.toml

sdroege:
Ok then :)

bilelmoussaoui:
It seems like there is a `autotests = false` config but it can't be set per target_os

sdroege (approved):

sdroege (approved):

bilelmoussaoui:
Let us try those new super powers!

Issue #1305, opened 2022-01-09
GIO-style async functions should not require `Send` for the callback


sdroege:
Instead they should enforce that the calling thread (of the async function) owns the main context and add an additional check that the callback is actually called from the same thread as where it was created.

The main context to check for would be the thread default main context, and if there's none then the global default main context.

The way how the functions are currently generated it is harder to use and allows (if you try hard enough) to send non-`Send` values to a different thread.

sdroege:
The main context check would be basically [this](https://github.com/gtk-rs/gtk-rs-core/blob/6373fe1b5a850ce06f94135c37f9b07129e500cf/gio/src/gio_future.rs#L62-L66). And that the callback is called from the correct thread could be done via [this](https://github.com/gtk-rs/gtk-rs-core/blob/6373fe1b5a850ce06f94135c37f9b07129e500cf/gio/src/gio_future.rs#L135-L146).

sdroege:
The change would basically look like this.

We should probably make `ThreadGuard` public in `glib` as part of all this.

```diff
diff --git a/gio/src/gio_future.rs b/gio/src/gio_future.rs
index 425dd77f77..8b2ec511a2 100644
--- a/gio/src/gio_future.rs
+++ b/gio/src/gio_future.rs
@@ -132,20 +132,38 @@ fn thread_id() -> usize {
 }
 
 // Taken from glib-rs, but we don't want this to be public API
-struct ThreadGuard<T> {
+pub(crate) struct ThreadGuard<T> {
     thread_id: usize,
     value: Option<T>,
 }
 
 impl<T> ThreadGuard<T> {
-    fn new(value: T) -> Self {
+    pub(crate) fn new(value: T) -> Self {
         Self {
             thread_id: thread_id(),
             value: Some(value),
         }
     }
 
-    fn into_inner(mut self) -> T {
+    pub(crate) fn as_ref(&self) -> &T {
+        assert!(
+            self.thread_id == thread_id(),
+            "Value accessed from different thread than where it was created"
+        );
+
+        self.value.as_ref().expect("no value anymore called twice")
+    }
+
+    pub(crate) fn as_mut(&mut self) -> &mut T {
+        assert!(
+            self.thread_id == thread_id(),
+            "Value accessed from different thread than where it was created"
+        );
+
+        self.value.as_mut().expect("no value anymore called twice")
+    }
+
+    pub(crate) fn into_inner(mut self) -> T {
         assert!(
[...]

PR #1312, opened 2022-01-24
Don't require `Send` closures for GIO-style async functions

sdroege:
Instead make sure that they're only ever called from the thread that
owns the thread default main context, and check at runtime that they're
really only called from the correct thread.

While this sounds like adding new constraints at runtime, anything else
was already not allowed before but not checked.

Fixes https://github.com/gtk-rs/gir/issues/1305

----

CC @pbor


bilelmoussaoui:
Nice! we can finally drop a bunch of manual functions from gtk4-rs :) This also fixes https://github.com/gtk-rs/gtk4-rs/issues/804 normally

bilelmoussaoui:
> > we can finally drop a bunch of manual functions from gtk4-rs
> 
> Oh, why? You had some manual implementations just to get rid of `Send`?

Basically all of this https://github.com/gtk-rs/gtk4-rs/blob/master/gdk4/src/clipboard.rs and probably others I don't remember, I will have a closer look later

> > This also fixes [gtk-rs/gtk4-rs#804](https://github.com/gtk-rs/gtk4-rs/issues/804) normally
> 
> How? That's not about an GIO-style async function or is it? I didn't really understand the issue so didn't comment on it before :)

Never mind, it is not actually related. It is about making the PrintOperation Sync/Send

sdroege:
> Basically all of this https://github.com/gtk-rs/gtk4-rs/blob/master/gdk4/src/clipboard.rs and probably others I don't remember, I will have a closer look later

Thanks, I'll leave gtk4-rs for you then and take care of the other crates.

PR #1313, opened 2022-01-24
codgen: run the visitor on global functions as well

bilelmoussaoui:
otherwise glib::Priority is not replaced for them and we have to manually implement them in such case
The same change is needed for Builder pattern

### DIFF
--- a/src/codegen/object.rs
+++ b/src/codegen/object.rs
@@ -348,9 +348,29 @@ fn generate_builder(w: &mut dyn Write, env: &Env, analysis: &analysis::object::I
     )?;
     writeln!(w, "#[must_use = \"The builder must be built to be used\"]")?;
     writeln!(w, "pub struct {}Builder {{", analysis.name)?;
+
+    let tid_int = env
+        .library
+        .find_type(0, "*.gint")
+        .expect("No fundamental type *.gint");
+
+    let glib_ns_id = env
+        .library
+        .find_namespace("GLib")
+        .expect("Missing `GLib` namespace in add_glib_priority!");
+    let priority_id = env
+        .library
+        .find_type(glib_ns_id, "Priority")
+        .expect("No Priority type");
+
     for (builder_props, super_tid) in &analysis.builder_properties {
         for property in builder_props {
-            match RustType::try_new(env, property.typ) {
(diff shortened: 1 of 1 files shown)


bilelmoussaoui:
Do not merge yet, I will handle other cases as well first :)

src/custom_type_glib_priority.rs
```diff
@@ -43,14 +43,16 @@ struct ReplaceToPriority {
 
 impl FunctionsMutVisitor for ReplaceToPriority {
     fn visit_function_mut(&mut self, func: &mut Function) -> bool {
-        if !func.name.ends_with("_async") {
```
bilelmoussaoui, line 46: Sadly this change is not great because it ends up using the glib::Priority where it doesn't make sense.  Well there was only one case in gtk3-rs and another in gtk4-rs. E.g, https://docs.gtk.org/gtk3/method.TextTag.set_priority.html. Although, it drops a bunch of manual code from gtk4-rs, could be a fair trade, not sure :)

bilelmoussaoui: will test it with gstreamer once the send on callbacks changes lands

sdroege:
Let me know when this is good for a review. I think changing every int that is called something with priority is not great, it will have many false positives. Maybe the functions have some other heuristic you could use? Or maybe we can have a configuration to change a parameter type to something else (don't we have that already?)?

bilelmoussaoui:
> Let me know when this is good for a review. I think changing every int that is called something with priority is not great, it will have many false positives. Maybe the functions have some other heuristic you could use? Or maybe we can have a configuration to change a parameter type to something else (don't we have that already?)?

I think the best way going forward is to keep the detection automatic with a way to configure it per parameter/function return. It needs a bit more work for now

bilelmoussaoui:
@sdroege can you think of any other potential conversions we can configure like the priority one? Just thinking if we need some generic configuration like "disable_preprocess_visitor = false` or have something per type (probably unneeded)

sdroege:
I don't know, sorry

bilelmoussaoui:
We also have conversions from an int to `gdk::Key` type in gtk4-rs, all of those functions have to be manually implemented sadly. So maybe we need a generic solution here

PR #1318, opened 2022-02-20
docs: Handle gi-docgen namespaces when looking for types to link

bilelmoussaoui:
Otherwise causes issue when looking for GtkApplication and you end up with a GApplication link

### DIFF
--- a/src/analysis/enums.rs
+++ b/src/analysis/enums.rs
@@ -65,7 +65,7 @@ pub fn new(env: &Env, obj: &GObject, imports: &mut Imports) -> Option<Info> {
     let mut functions = functions::analyze(
         env,
         &enumeration.functions,
-        enumeration_tid,
+        Some(enumeration_tid),
         false,
         false,
         obj,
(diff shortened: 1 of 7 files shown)


bilelmoussaoui:
not tested yet, doing that right now

bilelmoussaoui:
Before: 
image

After
image

GuillaumeGomez (approved):

GuillaumeGomez:
Looks good to me, thanks!

MarijnS95:
@bilelmoussaoui in the second screenshot the `core::gtk::Application` link broke, is that the intended change of this PR?

bilelmoussaoui:
> @bilelmoussaoui in the second screenshot the `core::gtk::Application` link broke (but `Value::init()` got fixed), is that the intended change of this PR?

Because I built the docs locally with `--no-deps`, it needs the ` extern-html-root-url` to work but also shows that it is actually using gtk::Application instead of adw::Application there

PR #1321, opened 2022-02-28
book: Pass output path to gir in console examples

philn:
Seems like this has to be explicitly supplied.


GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #1325, opened 2022-03-06
Add "default_value" parameter

### LINKED ISSUE #1324: Add a way to generate `Default` for enums
It would be great to annotate the default value in `Gir.toml` and have gir generate the corresponding `impl Default`.

GuillaumeGomez:
Fixes #1324.

For now this is a draft because:

 * Is the parameter name ok? If not what should we use?
 * I chose to let the user put whatever they wanted as default value. Did you have something else in mind?
 * I need to add documentation.

### DIFF
--- a/src/codegen/enums.rs
+++ b/src/codegen/enums.rs
@@ -6,6 +6,7 @@ use crate::{
         self, cfg_condition, cfg_condition_no_doc, cfg_condition_string, cfg_deprecated, derives,
         doc_alias, version_condition, version_condition_no_doc, version_condition_string,
     },
+    codegen::generate_default_impl,
     config::gobjects::GObject,
     env::Env,
     file_saver,
@@ -449,6 +450,25 @@ impl FromGlib<{sys_crate_name}::{ffi_name}> for {name} {{
         )?;
         writeln!(w)?;
     }
+
+    generate_default_impl(
+        w,
+        env,
+        config,
+        &enum_.name,
+        enum_.version,
+        enum_.members.iter(),
+        |member| {
+            let e_member = members.iter().find(|m| m.c_name == member.c_identifier)?;
+            let member_config = config.members.matched(&member.name);
+            let version = member_config
+                .iter()
+                .find_map(|m| m.version)
+                .or(e_member.version);
+            let cfg_condition = member_config.iter().find_map(|m| m.cfg_condition.as_ref());
+            Some((version, cfg_condition, e_member.name.as_str()))
+        },
+    )?;
 
     Ok(())
 }
--- a/src/codegen/mod.rs
+++ b/src/codegen/mod.rs
@@ -1,5 +1,10 @@
-use crate::{config::WorkMode, env::Env, file_saver::*};
+use crate::config::gobjects::GObject;
+use crate::{config::WorkMode, env::Env, file_saver::*, library::Member, version::Version};
+use std::fmt::Display;
+use std::io::{Result, Write};
 use std::path::Path;
+
+use general::{cfg_condition, version_condition};
 
 mod alias;
 mod bound;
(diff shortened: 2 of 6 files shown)


src/codegen/enums.rs
```diff
@@ -6,6 +6,7 @@ use crate::{
         self, cfg_condition, cfg_condition_no_doc, cfg_condition_string, cfg_deprecated, derives,
         doc_alias, version_condition, version_condition_no_doc, version_condition_string,
     },
+    codegen::object::generate_default_impl,
```
sdroege, line 9: Why do you move it to the `object` module? Has nothing to do with objects :)

src/config/gobjects.rs
```diff
@@ -417,6 +420,10 @@ fn parse_object(
         .lookup("clear_function_expression")
         .and_then(Value::as_str)
         .map(ToOwned::to_owned);
+    let default_impl_value = toml_object
+        .lookup("default_impl_value")
```
sdroege, line 424:
```suggestion
        .lookup("default_value")
```

Seems more descriptive

sdroege (commented):
Looks good to me

src/config/gobjects.rs
bilelmoussaoui: default_value looks good to me as well

src/codegen/enums.rs
GuillaumeGomez: Where should I put it then?

sdroege: `codegen/mod.rs`? And maybe call it `generate_enum_flags_default_impl`

GuillaumeGomez: Oh you only want it for enums and flags? Not for structs?

sdroege: How would this work properly for structs?

GuillaumeGomez: Exactly the same. It's expecting an expression to put into the `default` function.

bilelmoussaoui: Ah that is not really what I was thinking of, it should take a member name and check if it exists or error out. Then return `Self::MemberName` in the default impl.

GuillaumeGomez: Hum ok. I'll send an error if objects have this attribute then.

sdroege (approved):

src/codegen/mod.rs
```diff
@@ -97,3 +100,40 @@ pub fn generate_single_version_file(env: &Env) {
...
+
+pub fn generate_default_impl<'a>(
+    w: &mut dyn Write,
+    env: &Env,
+    config: &crate::config::gobjects::GObject,
+    name: &str,
+    version: Option<crate::version::Version>,
+    mut members: impl Iterator<Item = &'a str>,
+) -> Result<()> {
+    if let Some(ref default_value) = config.default_value {
```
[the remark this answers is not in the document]
bilelmoussaoui: I would rather be consistent with the other configurations we have :)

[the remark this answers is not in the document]
GuillaumeGomez: So for `gtk::Align::Fill` it'd be the C name `GTK_ALIGN_FILL`? It's a const after all...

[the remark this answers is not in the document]
bilelmoussaoui: the name is `fill`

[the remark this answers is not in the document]
GuillaumeGomez: @sdroege An opinion maybe?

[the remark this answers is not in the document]
bilelmoussaoui: It confused me as well the first time i had to set a specific version of an enum member but at least it is consistent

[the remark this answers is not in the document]
sdroege: What @bilelmoussaoui said. That would be consistent with how other configurations for enums/flags/signals/etc work.

bilelmoussaoui (approved):

Issue #1331, opened 2022-03-21
gir v0.15 inserts assertion to check return value even if throw function returns void


takaswie:
In API mode, `assert_eq!()` macro is inserted to evaluate `is_ok` variable
which should have return value of function call even if the function returns
void in C API.

For example of [alsa-gobject](https://github.com/alsa-project/alsa-gobject/blob/master/src/ctl/card.c#L191),

```
/**
 * alsactl_card_open:
 * @self: A #ALSACtlCard.
 * @card_id: The numerical ID of sound card.
 * @open_flag: The flag of open(2) system call. O_RDONLY is forced to fulfil internally.
 * @error: A #GError. Error is generated with two domains; #g_file_error_quark() and
 *         #alsactl_card_error_quark().
 */
void alsactl_card_open(ALSACtlCard *self, guint card_id, gint open_flag,
                       GError **error)
{
    ...
}
```

The generated FFI crate has [below API](https://github.com/alsa-project/alsa-gobject-rs/blob/master/alsactl-sys/src/lib.rs#L356) for the above:

```
    pub fn alsactl_card_open(
        self_: *mut ALSACtlCard,
        card_id: c_uint,
        open_flag: c_int,
        error: *mut *mut glib::GError,
    );
```

The gir v0.15 generates below codes for the above, which includes issued assertion and variable:

```
pub trait CardExt: 'static {
    ...
    #[doc(alias = "alsactl_card_open")]
    fn open(&self, card_id: u32, open_flag: i32) -> Result<(), glib::Error>;
    ...
}
...
impl<O: IsA<Card>> CardExt for O {
    ...
    fn open(&self, card_id: u32, open_flag: i32) -> Result<(), glib::Error> {
        unsafe {
            let mut error = ptr::null_mut();
            let is_ok = ffi::alsactl_card_open(
                self.as_ref().to_glib_none().0,
                card_id,
                open_flag,
                &mut error,
            );
            assert_eq!(is_ok == glib::ffi::GFALSE, !error.is_null());
            if error.is_null() {
                Ok(())
            } else {
                Err(from_glib_full(error))
            }
        }
    }
...
```

The assertion brings compilation error:

```
$ cargo build -p alsactl
   Compiling alsactl-sys v0.2.0 (/home/mocchi/git/alsa-gobject-rs/alsactl-sys)
   Compiling alsactl v0.2.0 (/home/mocchi/git/alsa-gobject-rs/alsactl)
...
error[E0308]: mismatched types
[...]

PR #1333, opened 2022-03-26
suppress insertion of assertion for result of throw function to return void

takaswie:
Hi,

As I filed #1331 , gir v0.15 or later inserts assertion for result of throw function even if the function returns void. It's based on GNOME convention that throw function should return boolean value to express whether GError argument is set or not. However we have some libraries which don't follow to the convention yet.

This patchset adds enumeration to express the strategy of return value for throw function,
then suppress the insertion for throw function returning void. As a result, the generated code is almost the same as the one generated gir v0.14 or before in regards of throw function to return void.

### DIFF
--- a/src/analysis/out_parameters.rs
+++ b/src/analysis/out_parameters.rs
@@ -202,12 +202,10 @@ fn decide_throw_function_return_strategy(
         .unwrap_or_default();
     if env.type_(typ).eq(&Type::Fundamental(Fundamental::None)) {
         ThrowFunctionReturnStrategy::Void
+    } else if use_function_return_for_result(env, typ, func_name, configured_functions) {
+        ThrowFunctionReturnStrategy::ReturnResult
     } else {
-        if use_function_return_for_result(env, typ, func_name, configured_functions) {
-            ThrowFunctionReturnStrategy::ReturnResult
-        } else {
-            ThrowFunctionReturnStrategy::CheckError
-        }
+        ThrowFunctionReturnStrategy::CheckError
     }
 }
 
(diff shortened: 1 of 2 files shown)


takaswie:
Cargo clippy generates warning of `clippy::collapsible_else_if`:

```
$ cargo clippy
   Compiling gir v0.0.1 (/home/mocchi/git/gtk-rs/gir)
warning: this `else { if .. }` block can be collapsed
   --> src/analysis/out_parameters.rs:205:12
    |
205 |       } else {
    |  ____________^
206 | |         if use_function_return_for_result(env, typ, func_name, configured_functions) {
207 | |             ThrowFunctionReturnStrategy::ReturnResult
208 | |         } else {
209 | |             ThrowFunctionReturnStrategy::CheckError
210 | |         }
211 | |     }
    | |_____^
    |
    = note: `#[warn(clippy::collapsible_else_if)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#collapsible_else_if
help: collapse nested if block
    |
205 ~     } else if use_function_return_for_result(env, typ, func_name, configured_functions) {
206 +         ThrowFunctionReturnStrategy::ReturnResult
207 +     } else {
208 +         ThrowFunctionReturnStrategy::CheckError
209 +     }
    |

warning: `gir` (lib) generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 8.64s
```

Is it preferrable to follow the advice or not in this project?

sdroege:
Thanks, I'll take a look at this in the next days.

Also CC @pbor

sdroege (approved):

PR #1337, opened 2022-04-06
sys, statics: Add "use" libc in6_addr and stat

qinqon:
When using gir to generate NetworkManager sys bindings a pair of libc
artifacts are missing "in6_addr" and "stat". This change add them to
the generated "use" directive.

### DIFF
--- a/src/codegen/sys/statics.rs
+++ b/src/codegen/sys/statics.rs
@@ -19,7 +19,7 @@ pub fn after_extern_crates(w: &mut dyn Write) -> Result<()> {
         "#[allow(unused_imports)]",
         "use libc::{c_int, c_char, c_uchar, c_float, c_uint, c_double,",
         "    c_short, c_ushort, c_long, c_ulong,",
-        "    c_void, size_t, ssize_t, intptr_t, uintptr_t, FILE};",
+        "    c_void, size_t, ssize_t, intptr_t, uintptr_t, FILE, in6_addr, stat};",
     ];
 
     write_vec(w, &v)


bilelmoussaoui (approved):

src/codegen/sys/statics.rs
```diff
@@ -19,7 +19,7 @@ pub fn after_extern_crates(w: &mut dyn Write) -> Result<()> {
         "#[allow(unused_imports)]",
         "use libc::{c_int, c_char, c_uchar, c_float, c_uint, c_double,",
         "    c_short, c_ushort, c_long, c_ulong,",
-        "    c_void, size_t, ssize_t, intptr_t, uintptr_t, FILE};",
+        "    c_void, size_t, ssize_t, intptr_t, uintptr_t, FILE, in6_addr, stat};",
```
GuillaumeGomez, line 22: `in6_addr` doesn't exist on windows' libc. You need to add it there first.

qinqon: You mean add it to rust libc library ?

GuillaumeGomez: Yes, I'm creating a PR for it because it has unions which is a bit tricky.

qinqon: Thanks!

GuillaumeGomez: I opened https://github.com/rust-lang/libc/pull/2741.

qinqon: Do you know how much time usually those PR takes to land ?

GuillaumeGomez: Considering it's a small one, I think a few days.

qinqon:
@GuillaumeGomez I am still having issues https://github.com/gtk-rs/gir/issues/1340, can you help there ?

fmartinsons:
Hello @qinqon 

I'm currently trying to generate rust bindings for NetworkManager and I hit the same issue about missing in6_addr and stat, what was the solution in the end since the submitted patch have not been merged ?

Thanks

PR #1341, opened 2022-04-06
Add generate_doc

### LINKED ISSUE #1310: Add a way to disable automatic documentation embedding per type/function
There are cases where the C documentation doesn't make much sense for Rust code, especially for manually implemented items. So it would be great to be able to disable those.

GuillaumeGomez:
Fixes #1310.


sdroege (approved):

sdroege (approved):
Conflicts resolved

PR #1345, opened 2022-04-10
Remove is_windows_utf8 function

jim4067:
see #1342


sdroege (approved):
Thanks, looks good to me. Can you squash all commits into a single one with a useful commit message, and then update `gtk-rs-core` and `gtk3-rs` accordingly?

jim4067:
on It.

PR #1347, opened 2022-04-11
Fixed the command to generate the docs

pentamassiv:


book/src/generate_docs.md
```diff
@@ -3,7 +3,7 @@
 And finally the last feature! Just run the following command (note the `-m doc` at the end):
 
 ```console
-> cargo run --release -- -c YourGirFile.toml -d ../gir-files --doc-target-path the-output-file-name -m doc
+> gir -c Gir.toml -d ../gir-files --doc-target-path docs.md -m doc
```
GuillaumeGomez, line 6:
```suggestion
cargo run --release -- -c Gir.toml -d ../gir-files --doc-target-path docs.md -m doc
```

bilelmoussaoui:
It should probably also mention the `generator.py` script with `--embed-docs` / `--strip-docs` options.

book/src/generate_docs.md
pentamassiv: If I run your suggested command, I get ``error: a bin target must be available for `cargo run` ``. In the previous commands of the tutorial, gir was also used instead of using the `cargo run` command. That is why I suggested the above change.

pentamassiv:
I believe the "Generating documentation" should also not be it's own chapter but be part 3.4 of the tutorial.

pentamassiv:
The generator.py script is for regenerating the bindings and the wrapper one you've created the Git.toml files and all that and there is a new .gir file, right? I could write a few sentences about it if you'd like me to.

bilelmoussaoui:
> The generator.py script is for regenerating the bindings and the wrapper one you've created the Git.toml files and all that and there is a new .gir file, right? I could write a few sentences about it if you'd like me to.

Yes but it can also be used to generate the docs making usage of the gir directly unnecessary in general, but both should be documented. 

There is already a small section about the generator.py script in the first page of the book https://gtk-rs.org/gir/book/introduction.html#helpers

PR #1349, opened 2022-04-21
Use Error out params in a new sections for functions docs

### LINKED ISSUE #1158: Use Error out params in a new sections for functions docs
Currently `GError**` 's docs are completely ignored, instead maybe we can add a new section after ` # Return` like `# Error?` that would print the error's docs so that we have a way to know when the function fails

cc @MarijnS95 @sdroege

Rakielle:
fixes #1158


Rakielle:
@bilelmoussaoui PTAL

src/codegen/doc/mod.rs
```diff
@@ -870,6 +870,32 @@ where
                 )?;
             }
         }
+        // document function's Error
+        let gerror_out_parameter = no_array_length_params
+            .iter()
+            .filter(|param| {
```
bilelmoussaoui, line 876: Please use find instead, it is basically the same except it will return the first found item.

src/codegen/doc/mod.rs
```diff
@@ -870,6 +870,32 @@ where
...
+                param.direction == ParameterDirection::Out
+                    && !doc_ignored_parameters.contains(&param.name)
+                    && param.c_type == "GError**"
+            })
+            .next();
+            
+        if fn_.ret().doc.is_some() || !gerror_out_parameter.name.is_empty() {
```
bilelmoussaoui, line 883: that won't work, the `gerror_out_parameter` is an `Option<T>` which means you can't access `T` until you are sure it is `Some(T)` and not `None`. You want something like 
```suggestion
        if fn_.ret().doc.is_some() || gerror_out_parameter.is_some() {
```
instead

src/codegen/doc/mod.rs
```diff
@@ -870,6 +870,32 @@ where
...
+            writeln!(w, "\n# Errors\n")?;
+        }
+
+        if let Some(ref doc) = gerror_out_parameter.doc {
```
bilelmoussaoui, line 887: Same as above, this won't work.

src/codegen/doc/mod.rs
```diff
@@ -870,6 +870,32 @@ where
...
+            writeln!(
```
bilelmoussaoui, line 888: There is no need to display the error parameter name, it is useless.

src/codegen/doc/mod.rs
Rakielle: 🙈

PR #1350, opened 2022-04-22
Use Error out params in a new sections for functions docs

### LINKED ISSUE #1158: Use Error out params in a new sections for functions docs

Rakielle:
Fixes #1158

### DIFF
--- a/src/codegen/doc/mod.rs
+++ b/src/codegen/doc/mod.rs
@@ -870,6 +870,23 @@ where
                 )?;
             }
         }
+        // document function's Error
+        let gerror_out_parameter = no_array_length_params.iter().find(|param| {
+            param.direction == ParameterDirection::Out
+                && !doc_ignored_parameters.contains(&param.name)
+                && param.c_type == "GError**"
+        });
+
+        if let Some(param) = gerror_out_parameter {
+            if let Some(ref doc) = param.doc {
+                writeln!(w, "\n# Errors\n")?;
+                writeln!(
+                    w,
+                    "{}",
+                    reformat_doc(&fix_param_names(doc, &self_name), env, in_type),
+                )?;
+            }
+        }
         Ok(())
     })
 }


Rakielle:
@bilelmoussaoui PTAL. Made a mistake trying to squash my commits in the previous PR

bilelmoussaoui:
The code changes looks good to me, could you try it out in a project like gtk4-rs? For that you will have to 
- Clone that repo
- Apply the changes to the local gir git submodule
- Run `python3 generator.py --embed-docs`
- Run `cargo doc -p gtk4` for example
- Check the output docs how they do looks like.

Rakielle:
@bilelmoussaoui 
Can I finalize this contribution on outreachy?
Deadline is in an hour and it's taking some time to clone the gtk4-rs repo recursively

Rakielle:
A921432B-9392-4D45-8772-193971F037CA

bilelmoussaoui:
It is already finalized, if you mean getting it merged then I will have to test it locally when I have got the time before merging it. Submitting the fix should be already enough though.

bilelmoussaoui:
> A921432B-9392-4D45-8772-193971F037CA
> 

Use --depth=1

Rakielle:
29461833-6297-4160-8BB4-A1CBF2DB6CC2

I get this error after running 
`python3 generate.py --embed-docs`

GuillaumeGomez:
You know that could can copy/paste output instead of putting screenshots right?

Also your last screenshot doesn't show any error.

Rakielle:
This is the error 

 "Error: [Errno 2] No such file or directory; 'rustdoc-stripper'"

bilelmoussaoui:
You need to install it with `cargo install rustdoc-stripper`

Rakielle:
@bilelmoussaoui I updated the code and it works after testing it on the gtk4-rs repo. One thing though, the doc written under the `Errors` section doesn't seem to be about errors. I can't tell if it's correct or not, could you please check it when you have the time?

bilelmoussaoui:
Do you have an example where you noticed that?

src/codegen/doc/mod.rs
```diff
@@ -870,6 +870,28 @@ where
...
+            param.direction == ParameterDirection::Out
+                && !doc_ignored_parameters.contains(&param.name)
+                && param.c_type == "GError**"
+        });
+
+        if gerror_out_parameter.is_some() {
+            writeln!(w, "\n# Errors\n")?;
+        }
+
+        if let Some(param) = gerror_out_parameter {
```
bilelmoussaoui, line 884:
```suggestion
        if let Some(param) = gerror_out_parameter {
            writeln!(w, "\n# Errors\n")?;
```

src/codegen/doc/mod.rs
```diff
@@ -870,6 +870,28 @@ where
...
+            writeln!(
+                w,
+                "{}",
+                reformat_doc(
+                    &fix_param_names(param.doc.as_ref().unwrap_or(&"".to_string()), &self_name),
```
bilelmoussaoui, line 889: I would do a `if let Some(doc) = param.doc` here instead.

bilelmoussaoui:
> Do you have an example where you noticed that?

? :)

Rakielle:
It's not showing up after the current modification to the code. I reverted the change and realised the `Errors` section comes in between the doc in the `Returns` section. With the current change though, the `Error` section is not being documented, I guess that's because there's no error doc in the `gtk4` module?

bilelmoussaoui:
Looking at various functions, it seems like the documentation is always `Return location for ....` which is not useful at all. @sdroege do you have an idea if gstreamer does better here?

sdroege:
No, it's (usually) the same. Very few errors contain more information in the docs like which domains are actually valid for them.

Rakielle:
@sdroege, @bilelmoussaoui Is there other modules other than `gtk4-rs` with error docs we can test on?

sdroege:
`gio` should have lots of such functions

Issue #1346, opened 2022-04-11
-sys crates not found


pentamassiv:
TLDR:

I believe I stumbled upon three issues:

- [ ] Ensure that glib, gdk and gtk are present as dependencies if a Cargo.toml file already exists
- [x] Change link for gdk and gtk crates to https://github.com/gtk-rs/gtk3-rs
- [x] Rename the sys packages to the "regular"/safe wraper crate

Hi,
thank you for gir and gtk-rs. Its a great project! I believe I found some minor issues though.

I wanted to update my wrapper for gtk-layer-shell. To make sure everything is up to date because I made some manual edits to generated files last time, I started from scratch again. I followed the tutorial in the book again to create the wrapper for gtk-layer-shell. I get an error when building the sys crate for gtk-layer-shell.

[This](https://github.com/pentamassiv/gtk-layer-shell-gir/commit/6251467178d71721479b92cdb08b61d0fc0b954d) is what the repo looks like before running gir.
If I run `gir -o .` and the old Cargo.toml from creating the new lib crates still exists, I get the following errors:
```
use of undeclared crate or module `glib`
use of undeclared crate or module `gtk`
use of undeclared crate or module `gdk`
```
I guess this is because the old Cargo.toml was present and for some reason the dependencies for gtk, gdk and glib were not added. You can find the resulting code [here](https://github.com/pentamassiv/gtk-layer-shell-gir/commit/a04a17dea6d368cf5542325ae46b86f9f71a04ed).

If I then run `rm Cargo.*` and `gir -o .` for a second time, I get the following error

```
error: no matching package found
searched package name: `gdk-sys`
perhaps you meant:      gio-sys or glib-sys
location searched: https://github.com/gtk-rs/gtk-rs-core
required by package `gtk-layer-shell-sys v0.0.1
```
You can find the resulting code [here](https://github.com/pentamassiv/gtk-layer-shell-gir/commit/6cbfd17093364bfe85c2a8010215a324c6668364).
I assume this is because in the newly created Cargo.toml the dependencies for gtk, gdk and glib were successfully added, but the link for the gdk-sys crate is incorrect. GDK and GTK can actually be found under https://github.com/gtk-rs/gtk3-rs. When I changed the link in the Cargo.toml, I get it to build a bit more. Until I get the next error.

Now I am get the following errors again:
```
use of undeclared crate or module `glib`
use of undeclared crate or module `gtk`
use of undeclared crate or module `gdk`
```
[...]

PR #1352, opened 2022-04-24
Fix finding the deps for the sys crate

pentamassiv:
This fixes the the second and the third issue mentioned in https://github.com/gtk-rs/gir/issues/1346. The links for the dependencies should now be correct and not always use the gtk-rs-core repo. Also the dependencies are renamed so they can be found. The gdk-sys dependency for example is renamed to gdk.

This pull request does not fix the first problem mentioned in https://github.com/gtk-rs/gir/issues/1346. I'll look into how to fix that tomorrow.


src/codegen/sys/cargo_toml.rs
```diff
@@ -46,15 +46,31 @@ fn fill_empty(root: &mut Table, env: &Env, crate_name: &str) {
...
         } else if ext_lib.crate_name == "gdk_pixbuf" {
             "gdk-pixbuf-sys".into()
         } else {
             format!("{}-sys", ext_lib.crate_name)
         };
-        let dep = upsert_table(deps, &*ext_package);
-        set_string(dep, "git", "https://github.com/gtk-rs/gtk-rs-core");
+        let repo_url = match ext_package.as_str() {
+            // Not sure if glib-macros and gobject-sys are needed here
+            "cairo-sys-rs" | "gdk-pixbuf-sys" | "gio-sys" | "gobject-sys" | "glib-macros"
```
GuillaumeGomez, line 59: That's pretty bad because it means gir can only generate crates it knows about. Which is not what gir is aiming for...

pentamassiv: Maybe I am mistaken, but I think the current code always adds `git = "https://github.com/gtk-rs/gtk-rs-core"` to all dependencies in the Cargo.toml. This is only correct for the crates included in that repo. For all other dependencies (including e.g. gdk, gtk3) the build will fail because the url is wrong and the user needs to look up the url under which the crate can be found and correct it.
With this code change, at least for the current dependencies from the gtk-rs-core, gtk3-rs and gtk4-rs repos, it uses the correct links. I guess the `&_ => "Unknown dependency, please create an issue",` branch could be changed to return the gtk-rs-core url. Since I believe I added all dependencies included in the repos, I don't see how that would be better though since the dependency would not be found under that url.
I don't know how gir would be able to determine the correct url for arbitrary dependencies. If it is a dependency that is not part of gtk-rs repos, the user can correct the url. If it is a new crate that is part of any of the gtk-rs repos, the first user depending on it would get a build error. They would see the compiler complaining about the dependency not being found and read the part asking them to open an issue. Ideally they would quickly create an issue, the dependency is added to the match statement and it's fixed. If they don't open an issue, they could easily fix the broken url in the Cargo.toml file themselves and somebody else would raise the issue.

GuillaumeGomez: I guess we should write "ADD GIT REPOSITORY URL HERE".

pentamassiv: Here is an example:
Before my proposed code change this would be added to the Cargo.toml of the sys crate:

[dependencies.gtk-sys]
git = "https://github.com/gtk-rs/gtk-rs-core"

With the proposed code change this is added to the Cargo.toml of the sys crate:

[dependencies.gtk]
package = "gtk-sys"
git = "https://github.com/gtk-rs/gtk3-rs"

GuillaumeGomez: And with my suggestion it would give:

```
[dependencies.gtk]
package = "gtk-sys"
git = "ADD GIT REPOSITORY URL HERE"
```

bilelmoussaoui: > That's pretty bad because it means gir can only generate crates it knows about. Which is not what gir is aiming for...

Is it ? In general most of the libraries are based on glib/gobject and then either gstreamer/gtk3/gtk4. but for other cases it is better to write some fallback like
```
[dependencies.gtk]
package = "gtk-sys"
git = "ADD GIT REPOSITORY URL HERE"
```

GuillaumeGomez: Then maybe we should only list `gtk-rs-core` crates?

pentamassiv: I added the gstreamer crates and its url and changed the message for unknown dependencies. This minimizes the manual changes users need to do.

src/codegen/sys/cargo_toml.rs
```diff
@@ -46,15 +46,52 @@ fn fill_empty(root: &mut Table, env: &Env, crate_name: &str) {
...
             format!("{}-sys", ext_lib.crate_name)
         };
-        let dep = upsert_table(deps, &*ext_package);
-        set_string(dep, "git", "https://github.com/gtk-rs/gtk-rs-core");
+        let repo_url = match ext_package.as_str() {
+            "cairo-sys-rs" | "gdk-pixbuf-sys" | "gio-sys" | "gobject-sys" | "glib-sys"
+            | "graphene-sys" | "pango-sys" | "pangocairo-sys" => {
+                "https://github.com/gtk-rs/gtk-rs-core"
+            }
+            "atk-sys" | "gdk-sys" | "gdkwayland-sys" | "gdkx11-sys" | "gtk-sys" | "gtk-macros" => {
```
bilelmoussaoui, line 62: gtk-macros is not needed

pentamassiv: Thanks, I removed it

src/codegen/sys/cargo_toml.rs
```diff
@@ -46,15 +46,52 @@ fn fill_empty(root: &mut Table, env: &Env, crate_name: &str) {
...
+            | "gstreamer-app-sys"
+            | "gstreamer-audio-sys"
+            | "gstreamer-base-sys"
+            | "gstreamer-check-sys"
+            | "gstreamer-controller-sys"
+            | "gstreamer-editing-services-sys"
+            | "gstreamer-gl-sys"
+            | "gstreamer-gl/egl-sys"
+            | "gstreamer-gl/wayland-sys"
+            | "gstreamer-gl/x11-sys"
```
sdroege, line 78:
```suggestion
            | "gstreamer-gl-egl-sys"
            | "gstreamer-gl-wayland-sys"
            | "gstreamer-gl-x11-sys"
```

Or not?

sdroege (commented):
Seems good to me otherwise

src/codegen/sys/cargo_toml.rs
pentamassiv: Good catch! I missed those but its fixed now

sdroege (approved):
Seems good to me then. @bilelmoussaoui @GuillaumeGomez ?

bilelmoussaoui (approved):

Issue #1353, opened 2022-04-24
Book needs some love


pentamassiv:
The book and the tutorial need some love. The GtkSource gir file no longer exists, some of the links are outdated and some information is missing. This can make it difficult to follow. Also at some points it can be simplified and be more noob friendly.
Since I am one myself but successfully went through the process of generating bindings and a wrapper, I believe I am a good candidate to update the book. I already started working on it. I hope that's fine :-)

pentamassiv:
Which library should I use as an example, now that SourceView is no longer an option?

bilelmoussaoui:
That would be very appreciated, the book was put together from the old docs and wasn't necessarily updated. Feel free to send PRs for improving specific parts of it

> Which library should I use as an example, now that SourceView is no longer an option?

I would use something like pango maybe

pentamassiv:
I created a PR for it https://github.com/gtk-rs/gir/pull/1379

bilelmoussaoui:
Are there things you still want to work on or the book is good enough for now and we could open individual issues for separate tasks?

pentamassiv:
I don't think I can improve it much more. I am not that familiar with gir. I believe it is relatively beginner friendly now but advanced users might find more things to improve. Individual issues are probably the right way to go from here

bilelmoussaoui:
Thanks for your contribution!!

PR #1354, opened 2022-04-27
Book overhaul

pentamassiv:
The improvements include:
- Restructuring the book a little bit to make it easier to find information
- Adding description on what gir is and when it can be used
- Explanation on how to set up the project folder
- Explanation on how to find .gir files
- Mentioning of various other foot guns that a beginner might encounter

I marked it as a draft, because there are still a few things that need to be done, but I need some help to get it across the finishing line.

- [ ] The pango example is incomplete
- [ ] A native speaker should proofread it

As suggested by @bilelmoussaoui in https://github.com/gtk-rs/gir/issues/1353, I used pango as an example. This worked fine for the -sys crate, but for the safe wrapper, I get a lot of errors and when I look at https://github.com/gtk-rs/gtk-rs-core/blob/master/pango/Gir.toml, there are a lot "advanced" things such as implementing types or ignoring stuff and unfortunately I am not knowledgeable enough to understand and explain it. The part that is missing can be found in `high_level_rust_api.md` and you can find it by searching for "TODO: Add steps of example".
I guess there are three options on how to proceed:
- somebody else can finish that part
- I finish it but only add a few objects to the "generate" and "manual" lists and the example does not go through the full process of generating the wrapper for pango
- I use a different example (but from looking at the other bindings, all of them seem to have these complications)

Thank you for taking the time to look at the suggested changes :)


bilelmoussaoui:
I don't think there are bindings that works out of the box for all the possible types, you will always have to add overrides. There are libraries that requires more changes than others, but maybe Pango wasn't a great suggestion after all...

pentamassiv:
Do you have a suggestion how we can proceed with this? Should I just add "TODOs" to the text, where more explanations about the overrides are needed or can somebody help with finishing it for pango? I believe this draft contains all the information of the current tutorial while being more clear & simple and fixing some outdated things. Merging it soonish would be beneficial for new users

bilelmoussaoui:
> Do you have a suggestion how we can proceed with this? Should I just add "TODOs" to the text, where more explanations about the overrides are needed or can somebody help with finishing it for pango? I believe this draft contains all the information of the current tutorial while being more clear & simple and fixing some outdated things. Merging it soonish would be beneficial for new users

If you have specific questions I would be more than happy to answer them

pentamassiv:
I created a pull request that is not a draft here https://github.com/gtk-rs/gir/pull/1379

Issue #1356, opened 2022-05-16
Some generated parse functions do not pass clippy


filnet:
This generated code:

```
    #[doc(alias = "gst_play_message_parse_buffering_percent")]
    pub fn parse_buffering_percent(msg: &gst::Message) -> u32 {
        assert_initialized_main_thread!();
        unsafe {
            let mut percent = mem::MaybeUninit::uninit();
            ffi::gst_play_message_parse_buffering_percent(
                msg.to_glib_none().0,
                percent.as_mut_ptr(),
            );
            let percent = percent.assume_init();
            percent
        }
    }
```

Generates this clippy error:

```
error: returning the result of a `let` binding from a block
   --> gstreamer-play/src/auto/enums.rs:282:13
    |
281 |             let percent = percent.assume_init();
    |             ------------------------------------ unnecessary `let` binding
282 |             percent
    |             ^^^^^^^
    |
    = note: `-D clippy::let-and-return` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#let_and_return
help: return the expression directly
    |
281 ~             
282 ~             percent.assume_init()
```

filnet:
This issue blocks https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/1036

bilelmoussaoui:
I don't think it is a blocker, you could `#[allow(clippy::let_and_return)]` before your `mod auto;`. Fixing that would be great though

filnet:
It is not simple to fix as the code handles a lot of situations (need_from_glib, tuples, ...).

sdroege:
> This issue blocks https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/1036

@filnet FWIW, I'm fine with adding `#![allow(clippy::let_and_return)]` in gstreamer-rs for that for the time being. Fixing this doesn't seem to be too simple so let's not block on that.

filnet:
> @filnet FWIW, I'm fine with adding `#![allow(clippy::let_and_return)]` in gstreamer-rs for that for the time being. Fixing this doesn't seem to be too simple so let's not block on that.

I am working on fix and it looks promising. I was able to fold the rhs of the `Let` into the `Tuple`.
Works for the new gstreamer-play project. I need to test on the whole gstreamer-rs now.

filnet:
The fix looks good too when applied to the whole gstreamer-rs repo.
[...]

PR #1357, opened 2022-05-19
Fix UB and clippy::let-and-return in function out variables handling

filnet:
fixes https://github.com/gtk-rs/gir/issues/1356

### DIFF
--- a/src/codegen/function_body_chunk.rs
+++ b/src/codegen/function_body_chunk.rs
@@ -208,7 +208,6 @@ impl Builder {
         let (call, ret) = self.apply_outs_mode(call, ret, &mut uninitialized_vars);
 
         body.push(call);
-        self.write_out_uninitialized(&mut body, uninitialized_vars);
         if let Some(chunk) = ret {
             body.push(chunk);
         }
@@ -335,26 +334,6 @@ impl Builder {
         Chunk::BlockHalf(chunks)
     }
 
-    fn write_out_uninitialized(
-        &self,
-        body: &mut Vec<Chunk>,
-        uninitialized_vars: Vec<(String, bool)>,
-    ) {
-        for (uninitialized_var, need_from_glib) in uninitialized_vars {
-            body.push(Chunk::Let {
-                name: uninitialized_var.clone(),
-                is_mut: false,
-                value: Box::new(Chunk::Custom(format!(
-                    "{}{}.assume_init(){}",
-                    if need_from_glib { "from_glib(" } else { "" },
-                    uninitialized_var,
-                    if need_from_glib { ")" } else { "" },
-                ))),
-                type_: None,
-            });
-        }
-    }
-
     fn remove_extra_assume_init(
         &self,
         array_length_name: &Option<String>,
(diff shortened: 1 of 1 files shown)


src/codegen/function_body_chunk.rs
```diff
@@ -367,6 +346,18 @@ impl Builder {
         }
     }
 
+    fn is_uninitialized_var(
+        &self,
+        name: &String,
+        uninitialized_vars: &mut Vec<(String, bool)>,
```
sdroege, line 352:
```
 error: writing `&mut Vec` instead of `&mut [_]` involves a new object where a slice will do
Error:    --> src/codegen/function_body_chunk.rs:352:29
    |
352 |         uninitialized_vars: &mut Vec<(String, bool)>,
    |                             ^^^^^^^^^^^^^^^^^^^^^^^^ help: change this to: `&mut [(String, bool)]`
```

src/codegen/function_body_chunk.rs
```diff
@@ -367,6 +346,18 @@ impl Builder {
...
+    ) -> Option<bool> {
+        if let Some((_, need_from_glib)) = uninitialized_vars.iter().find(|(n, _)| n.eq(name)) {
+            Some(*need_from_glib)
+        } else {
+            None
+        }
```
GuillaumeGomez, line 358:
```suggestion
        uninitialized_vars.iter().find(|(n, _)| n.eq(name)).map(|(_, need_from_glib)| need_from_glib)
```

src/codegen/function_body_chunk.rs
```diff
@@ -367,6 +346,17 @@ impl Builder {
         }
     }
 
+    fn is_uninitialized_var(
+        &self,
+        name: &String,
```
sdroege, line 351:
```suggestion
        name: &str,
```

I'm surprised that clippy didn't complain about this one

src/codegen/function_body_chunk.rs
```diff
@@ -367,6 +346,17 @@ impl Builder {
...
+        uninitialized_vars: &mut [(String, bool)],
```
sdroege, line 352:
```suggestion
        uninitialized_vars: &[(String, bool)],
```

Doesn't have to be mutable

sdroege (commented):

src/codegen/function_body_chunk.rs
```diff
@@ -1159,7 +1146,18 @@ impl Builder {
         mem_mode: &OutMemMode,
         uninitialized_vars: &mut Vec<(String, bool)>,
     ) -> Chunk {
-        let value = Chunk::Custom(parameter.name.clone());
+        let value = if let Some(need_from_glib) =
+            self.is_uninitialized_var(&parameter.name, uninitialized_vars)
+        {
+            Chunk::Custom(format!(
```
sdroege, line 1152: This seems to fail for bool out parameters

```
 error[E0277]: the trait bound `bool: glib::translate::FromGlib<MaybeUninit<i32>>` is not satisfied
Error:     --> gtk/src/auto/icon_info.rs:152:72
     |
152  |             let result = if error.is_null() { Ok((from_glib_full(ret), from_glib(was_symbolic))) } else { Err(from_glib_full(error)) };
     |                                                                        ^^^^^^^^^ the trait `glib::translate::FromGlib<MaybeUninit<i32>>` is not implemented for `bool`
     |
```

filnet: Issue was in async trampoline.

sdroege (approved):

filnet:
Last commit needed fixup :(

sdroege:
> Last commit needed fixup :(

Oh no, I thought you did that already with the force-push right afterwards. Should've checked more closely. Oh well, too late now :)

gtk-rs-core PR: https://github.com/gtk-rs/gtk-rs-core/pull/699

filnet:
> Oh no, I thought you did that already with the force-push right afterwards. Should've checked more closely. Oh well, too late now :)

I'll stick to force push from now on :)

sdroege:
Thanks for caring so much about commit history, that's unfortunately not very common :)

PR #1358, opened 2022-05-21
docs: don't generate implements links for fundamental types

bilelmoussaoui:
as those don't have Ext traits, see https://gtk-rs.org/gtk4-rs/git/docs/gdk4/struct.DNDEvent.html#implements

Before
image

image

After

image

### DIFF
--- a/src/library.rs
+++ b/src/library.rs
@@ -896,6 +896,13 @@ impl Type {
         }
     }
 
+    pub fn is_fundamental(&self) -> bool {
+        match *self {
+            Type::Class(Class { is_fundamental, .. }) => is_fundamental,
+            _ => false,
+        }
+    }
+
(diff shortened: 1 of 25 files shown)


src/library.rs
```diff
@@ -848,16 +848,16 @@ impl Type {
         }
     }
 
-    pub fn is_fundamental(&self) -> bool {
+    pub fn is_basic(&self) -> bool {
         matches!(*self, Type::Fundamental(_))
```
bilelmoussaoui, line 852: I wonder if I should rename the enum variant from Type::Fundamental to Type::Basic

GuillaumeGomez: That would make it coherent with the function name at least.

bilelmoussaoui: Done

sdroege (approved):

PR #1291, opened 2021-12-18
parser: support properties getters/setters

bilelmoussaoui:
The change requires the Gir files to be generated with gobject-introspection 1.70. With the current changes, the number of generated but unneeded docs for properties getters/setters 

cc @fengalin

PR #1359, opened 2022-05-21
docs: Generate properties/signals docs

### LINKED ISSUE #1076: Add properties to docs
Gir currently adds an "Implements" section to the docs (e.g. see [here](https://gtk-rs.org/gtk4-rs/gtk4/struct.Switch.html))

To me it would make sense to add a "Properties" section as well.
After a quick look it seems like most pieces already seem to be in place for that.

Unless anything speaks against it, I'd like to work on it.

bilelmoussaoui:
Fixes #1076

Current status:

Signals
image

Properties
image

### DIFF
--- a/src/codegen/doc/format.rs
+++ b/src/codegen/doc/format.rs
@@ -542,7 +542,7 @@ pub(crate) fn gen_signal_doc_link(symbol: &str, signal: &str) -> String {
 }
 
 pub(crate) fn gen_property_doc_link(symbol: &str, property: &str) -> String {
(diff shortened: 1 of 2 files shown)


bilelmoussaoui:
Depends on https://github.com/rust-lang/rust/pull/97249 & still needs some fixes.

bilelmoussaoui:
Ideally we would want https://github.com/gtk-rs/gir/pull/1291 to land first so we can add a link to the getter/setter with the property documentation.

Issue #1360, opened 2022-05-21
gir fails to compile on Windows


filnet:
Fails with :
```
$ cargo build
   Compiling gir v0.0.1 (C:\msys64\home\Philippe\rust\gstreamer-rs\gir)
error[E0433]: failed to resolve: could not find `unix` in `os`
 --> src\git.rs:3:14
  |
3 | use std::os::unix::prelude::OsStringExt;
  |              ^^^^ could not find `unix` in `os`

error[E0599]: no function or associated item named `from_vec` found for struct `OsString` in the current scope
  --> src\git.rs:70:30
   |
70 |     let toplevel = OsString::from_vec(output.stdout);
   |                              ^^^^^^^^ function or associated item not found in `OsString`
```

GuillaumeGomez:
Interested into sending a fix? It should be pretty easy to do. :)

filnet:
Not that simple...

Use of `*::unix::*` was introduced here https://github.com/gtk-rs/gir/commit/958e1b7c613e543e3d30cc69f1d1ff957f2ce8ac

It uses `OsStringExt::from_vec()` to parse the output of a git command and turn it into an `OsString` (and later to a `PathBuff`).
Relevant line is `let toplevel = OsString::from_vec(output.stdout);`.

On Windows, it is another story : https://users.rust-lang.org/t/is-there-a-way-to-convert-vec-u8-to-osstring/49970
The usual mess...

Turning that one line into `let toplevel = String::from_utf8(output.stdout).ok()?;` _works on my PC_ but will probably fail on more exotic Windows setups (?).

filnet:
Relevant question about stdout encoding https://users.rust-lang.org/t/stdin-stdout-stderr-and-encoding/5354

PR #1361, opened 2022-05-21
git: add non unix fallback for creating PathBuf from Vec<u8>

filnet:
not sure the new Windows related code is 100% correct
we will see...

fixes https://github.com/gtk-rs/gir/issues/1360

### DIFF
--- a/src/git.rs
+++ b/src/git.rs
@@ -1,6 +1,4 @@
-use std::ffi::OsString;
 use std::io::Result;
-use std::os::unix::prelude::OsStringExt;
 use std::path::{Path, PathBuf};
 use std::process::{Command, Output};
 
@@ -67,8 +65,19 @@ fn path_command(path: impl AsRef<Path>, subcommand: &[&str]) -> Option<PathBuf> 
             .and_then(std::char::from_u32),
         Some('\n')
     );
-    let toplevel = OsString::from_vec(output.stdout);
-    Some(toplevel.into())
+    Some(path_from_output(output.stdout))
+}
+
+#[cfg(unix)]
+fn path_from_output(output: Vec<u8>) -> PathBuf {
+    use std::ffi::OsString;
+    use std::os::unix::prelude::OsStringExt;
+    OsString::from_vec(output).into()
+}
+
+#[cfg(not(unix))]
+fn path_from_output(output: Vec<u8>) -> PathBuf {
+    String::from_utf8(output).unwrap().into()
 }
 
 pub fn toplevel(path: impl AsRef<Path>) -> Option<PathBuf> {


src/git.rs
```diff
@@ -67,8 +65,19 @@ fn path_command(path: impl AsRef<Path>, subcommand: &[&str]) -> Option<PathBuf>
...
+}
+
+#[cfg(unix)]
+fn path_from_output(output: Vec<u8>) -> PathBuf {
+    use std::ffi::OsString;
+    use std::os::unix::prelude::OsStringExt;
+    OsString::from_vec(output).into()
+}
+
+#[cfg(all(not(unix)))]
```
MarijnS95, line 78: Drop the all().

MarijnS95 (commented):
Title nit: you're not "fixing" things but merely extending gir to run on Windows.

src/git.rs
```diff
@@ -67,8 +65,19 @@ fn path_command(path: impl AsRef<Path>, subcommand: &[&str]) -> Option<PathBuf>
...
+fn path_from_output(output: Vec<u8>) -> PathBuf {
+    String::from_utf8(output).ok().unwrap().into()
```
bilelmoussaoui, line 80: Is the ok really useful if you are unwrapping anyway?

filnet:
> Title nit: you're not "fixing" things but merely extending gir to run on Windows.

Writing a good commit log is often harder than making the code changes...

filnet:
> Title nit: you're not "fixing" things but merely extending gir to run on Windows.

What about : "Make git command wrapper compile on Windows"

MarijnS95:
> Writing a good commit log is often harder than making the code changes...

Correct, one can always learn and improve, if reviewers are kind enough in pointing this out / suggesting improvements.

You can still edit/amend it :wink: 

> What about : "Make git command wrapper compile on Windows"

I'd have written here ``git: Provide `PathBuf` conversion [fallback] that works on Windows``.

But it's not really a "conversion", just the way it is transformed to a sensible and safe Rust type.

`[fallback]` between square brackets here means "optional"; ie. I may or may not add it depending on how I feel that day).

src/git.rs
MarijnS95: You won't see the error context this way, so I'd rather see the `ok()` dropped as well :)

src/git.rs
MarijnS95: Does `gir` already use `cfg_if!`? Might have been nice to use within this function so that you don't have to repeat the function body nor negate the cfg clause, at the expense of two extra indent levels :/

filnet: It does not. Should I add it ?

MarijnS95 (approved):

bilelmoussaoui (approved):

filnet:
It is quite ironic that we jump through hoops on unix where it is guaranteed that stdout is utf8, while on ambiguous platforms we do a plain `from_utf8`.

filnet:
From what I have gathered, `std::process::Command` hands out a `Vec<u8>` to support "binary streams" (gzip can do that for instance). Thus, I guess, using `ffi` (when available) makes sense.

sdroege:
I assume https://doc.rust-lang.org/std/os/windows/ffi/trait.OsStringExt.html is also not the right approach on Windows because usually the output is not actually UCS2 but ASCII/UTF-8 and it's all completely undefined?

I would go with assuming UTF-8 for now and if it fails for someone we can fix that then. What do you think?

MarijnS95:
> I would go with assuming UTF-8 for now and if it fails for someone we can fix that then. What do you think?

Yeah that was my thought as well when approving this :)

filnet:
That same module also does this:
`let hash = String::from_utf8(output.stdout).ok()?;`

sdroege (approved):
So let's get this in for now and reconsider once we notice that it fails for someone.

MarijnS95:
@filnet Yes any binary stream should be supported, so it just depends on what `git` outputs. Assuming UTF-8 and panicking otherwise is probably fool-proof / safe enough for now.

PR #1363, opened 2022-06-02
Basic types

filnet:
This PR is a follow up of https://github.com/gtk-rs/gir/pull/1134. Not sure what is the accepted way of contributing to someone else's PR.

Changes include:
* Renamed the new `Basic::Vulkan` enum to `Basic::Typedef`.
  Vulkan-1.0.gir currently contains only what can be considered as `typedef` records.
  See https://gitlab.gnome.org/GNOME/gobject-introspection/-/blob/main/misc/update-vulkan-gir.py#L17
* Hack the parser to produce `Basic::Typedef`s when parsing the `Vulkan` namespace.

So a bit more generic and the `const VULKAN: &[&str] = &[...]` is gone. :)


filnet:
A next step would be to address the Vulkan specific hack.

We need a cleaner way to trigger the "special" handling of Vulkan records:
1/ could be based on a new flag in the Record itself (or based on the state of the Record)
2/ could be globally enabled in the Vulkan-1.0.gir file (via meta data or something else)
3/ could be globally enabled in consumer Gir.toml file when declaring the external library
4/ could be enabled in gir itself (like now but in a "better" way via extension/plugin/what not).

src/parser.rs
```diff
@@ -264,6 +265,13 @@ impl Library {
     ) -> Result<(), String> {
         if let Some(typ) = self.read_record(parser, ns_id, elem, None, None)? {
             let name = typ.get_name();
+            // TODO: Don't hardcode to "Vulkan"
+            let typ = if self.namespace(ns_id).name.eq("Vulkan") {
+                assert!(typ.is_incomplete(self));
```
filnet, line 270: We probably won't keep this assert but it shows that Vulkan record are incomplete.
Unfortunately, non Vulkan record can be incomplete too so we can't use `is_incomplete` as trigger for the new behavior.

And then there is the problem of `ash::vk::{}` format string: where should it be specified if we want to externalize it (to make `gir` Vulkan agnostic again).

filnet: Likewise, instead of tweaking the Vulkan namespace on the fly, we could introduce a postprocess hook that tweaks the namespace after it was parsed.

filnet:
I went with post processing the Vulkan namespace.
And found `src\custom_type_glib_priority.rs` as example to follow.

bilelmoussaoui:
What is the status of this one?

PR #1365, opened 2022-06-13
Remove leading '>' from tutorial

gwutz:
This change removes all leading '>' characters of console commands in order to be able to copy them from the tutorial.


MarijnS95:
This was done to make markdown render `console` blocks directly, but it seems the markdown renderer used here (just like GitHub 's) wasn't fixed to prevent the `>` from being selected.

If you're serious about doing this, at the very least _demote_ the type to `sh` or `bash`, as `console` now becomes entirely meaningless and broken.

PR #1366, opened 2022-06-13
book: Unbreak syntax highlighting on `console` codeblocks, address concerns consistently

MarijnS95:
[#1365] was blatantly merged without considering review, breaking syntax highlighting on these codeblocks in the process.  As suggested, when removing the "effective UID" prefix `console` codeblocks loose their formatting and must instead be converted to regular `sh` or `bash` types.

At the same time, make this change consistently across the book, instead of considering `sys_library.md` to be the only page.  Note that it has been retained in places where a `console` block is intented, because the lines after a console command represent program output and are formatted accordingly when no "effective UID" prefix is detected.

[#1365]: https://github.com/gtk-rs/gir/pull/1365


bilelmoussaoui:
thanks!

GuillaumeGomez:
@MarijnS95 Sorry, I forgot to comment after your review: even though it wasn't complete, it was still an improvement over the existing so I just decided so it was good enough as is.

However, I'm glad you sent the complete fix and thank you for doing so!

MarijnS95:
Hmh, I don't think something is an improvement if it goes one step forward in one area, but a step backwards in another - whilst that's trivially solvable to make it a step forward without taking a step backwards. Alas, that's why I opened this PR, thanks for approving of it :heart:

GuillaumeGomez:
Sorry if that felt like I ignored your comment (I was working on a LLVM patch and let's just say their tool is suboptimal). So again: thanks a lot for your contribution. :)

PR #1368, opened 2022-06-20
don't assume GObject.Object is always available

gwutz:


bilelmoussaoui (approved):

src/library.rs
```diff
@@ -1118,9 +1118,10 @@ impl Library {
                     }
                 }
             }
-            if let Some(tid) = env.library.find_type(0, &full_name) {
-                let gobject_id = env.library.find_type(0, "GObject.Object").unwrap();
-
+            if let (Some(tid), Some(gobject_id)) = (
+                env.library.find_type(0, &full_name),
+                env.library.find_type(0, "GObject.Object"),
+            ) {
```
sdroege, line 1124:
```suggestion
            if let Some((tid, object_id)) = Option::zip(
                env.library.find_type(0, &full_name),
                env.library.find_type(0, "GObject.Object"),
            ) {
```

If you feel like it

sdroege:
I'm surprised this is the only place where the existence of `GObject` is assumed :)

PR #1374, opened 2022-08-11
codgen: Allow renaming objects

bilelmoussaoui:
Needed for https://github.com/gtk-rs/gtk-rs-core/issues/731

Not tested on that issue yet, but the code seems correct.


bilelmoussaoui:
Marked as a draft as handling the docs links side seems to be more complicated than I thought.

sdroege:
Any news on this?

bilelmoussaoui:
> Any news on this?

not much because it is more complicated than I expected. Currently the docs links are generated from a `analysis::symbol::Symbol` which has no idea about a potentially renamed objects. We will need to pass the new name to it somehow...

PR #1376, opened 2022-08-14
config: Add a move configuration for functions parameters

bilelmoussaoui:
To make use of the new IntoGlibPtr trait
Fixes the remaining of https://github.com/gtk-rs/gtk-rs-core/issues/50

### DIFF
--- a/book/src/config_api.md
+++ b/book/src/config_api.md
@@ -161,6 +161,8 @@ generate_doc = false
         name = "website_label"
         # allow to remove/add Option<>
         nullable = true
+        # Take the parameter by value instead of by ref
+        move = true
         # allow to make parameter immutable
         const = true
         # parameter is calculated as length of string or array and removed from function declaration
--- a/src/analysis/function_parameters.rs
+++ b/src/analysis/function_parameters.rs
@@ -72,6 +72,7 @@ pub struct CParameter {
     //analysis fields
     pub ref_mode: RefMode,
     pub try_from_glib: TryFromGlib,
+    pub move_: bool,
 }
 
 #[derive(Clone, Debug)]
@@ -95,6 +96,7 @@ pub enum TransformationType {
         pointer_cast: String,
         in_trait: bool,
         nullable: bool,
+        move_: bool,
     },
     ToGlibBorrow,
     ToGlibUnknown {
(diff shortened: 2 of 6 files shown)


GuillaumeGomez (approved):

book/src/config_api.md
```diff
@@ -161,6 +161,8 @@ generate_doc = false
         name = "website_label"
         # allow to remove/add Option<>
         nullable = true
+        # Take the parameter by value instead of by ref, only applies to the instance parameter
+        move = true
```
sdroege, line 165: Why? This should also work for other parameters, that would be mostly useless otherwise :) In GStreamer I need this for other parameters in almost all cases.

bilelmoussaoui: Ah ok, I will change that then

bilelmoussaoui: Should work for all the params now, do you mind trying it on some places in gstreamer to see if it behaves correctly?

sdroege:
Also this should probably be the default for transfer full parameters of pointer types that are not refcounted (I.e. also for strings and arrays)

bilelmoussaoui:
> Also this should probably be the default for transfer full parameters of pointer types that are not refcounted (I.e. also for strings and arrays)

I rather handle that in a separate PR

sdroege:
Works for me. Is it difficult?

bilelmoussaoui:
We don't have the necessary information as part of `TransformationType::ToGlibPointer` to know if it is a refcounted type or not. Same for arrays/strings.

bilelmoussaoui:
Ah, I think it might be doable. I will check in a bit.

sdroege:
Any news on this? I'm fine with merging this as-is for now but it seems to be only minimally useful then :)

bilelmoussaoui:
> Any news on this? I'm fine with merging this as-is for now but it seems to be only minimally useful then :)

We can land this as is and open an issue for only making use of it in more cases. The changes shouldn't be too complicated but I see that I have accumulated plenty of WIP PRs here that I should look at landing instead of adding more to that list :P

sdroege:
Ok with me. This is good to go then?

> Fixes the remaining of https://github.com/gtk-rs/gtk-rs-core/issues/50

This is of course not the case anymore then. Let's keep that issue open.

bilelmoussaoui:
> This is of course not the case anymore then. Let's keep that issue open.

Yes

> Ok with me. This is good to go then?

Should be yes

PR #1384, opened 2022-10-08
analysis: Use the new move trait for in transfer-full params

bilelmoussaoui:
Needed for https://github.com/gtk-rs/gtk-rs-core/issues/50

### DIFF
--- a/src/analysis/function_parameters.rs
+++ b/src/analysis/function_parameters.rs
@@ -7,7 +7,7 @@ use crate::{
     analysis::{self, bounds::Bounds},
     config::{self, parameter_matchable::ParameterMatchable},
     env::Env,
-    library::{self, Nullable, ParameterScope, TypeId},
+    library::{self, Nullable, ParameterScope, Transfer, TypeId},
     nameutil,
     traits::IntoString,
 };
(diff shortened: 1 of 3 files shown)


src/analysis/function_parameters.rs
```diff
@@ -297,7 +297,11 @@ pub fn analyze(
             user_data_index: par.closure,
             destroy_index: par.destroy,
             try_from_glib: try_from_glib.clone(),
-            move_: configured_parameters.iter().any(|p| p.move_),
+            move_: if transfer == Transfer::Full && par.direction.is_in() {
+                true
+            } else {
+                configured_parameters.iter().any(|p| p.move_)
+            },
```
sdroege, line 304:
```suggestion
            move_: configured_parameters.iter().find_map(|p| p.move).unwrap_or(if transfer == Transfer::Full && par.direction.is_in());
```

Otherwise you can't disable it explictly via the configuration. Also means that `p.move` needs to become `Option<bool>`

bilelmoussaoui: ah right, let me fix that.

sdroege (approved):

sdroege (approved):

PR #1385, opened 2022-10-11
analysis: Handle move configuration in Ext trait

bilelmoussaoui:
By calling to_owned() on the result of as_ref() as the into_glib_ptr requires Self and AsRef returns &Self

### DIFF
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -102,6 +102,7 @@ impl Bounds {
                     &bound_type,
                     *par.nullable,
                     par.instance_parameter,
+                    par.move_,
                 ));
                 if r#async && (par.name == "callback" || par.name.ends_with("_callback")) {
                     let func_name = func.c_identifier.as_ref().unwrap();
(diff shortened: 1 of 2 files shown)


sdroege:
> By calling to_owned() on the result of as_ref() as the into_glib_ptr requires Self and AsRef returns &Self

Instead of using `impl AsRef<T>` parameters you could use `impl IsA<T>`, and once my comment above is implemented you would be able to just pass in an owned object.

bilelmoussaoui:
> The problem with `impl AsRef<T>` is also that you can pass in a `T` and a `&T`, but the latter we actually want to avoid here or not?

Probably yes, but the real problem is we don't decide whether to pass `IsA<T>` or `AsRef<T>`, it is whatever the `Ext`  trait is implemented for. At least the change on https://github.com/gtk-rs/gtk-rs-core/pull/791#pullrequestreview-1138556561, could make it work I think without the `as_ref()` call at all. Let us see

sdroege:
Ah, you mean that the function has to be implemented for `self` and not `&self` if it's the self-parameter that is `transfer full`? Yes that's true

sdroege:
So I think there are two cases here for GObjects:
  - `self` parameter is `transfer full`: Use `self` instead of `&self` and use `self.into_glib_ptr()`
  - other parameter is `transfer full`: Use `impl IsA<T>` instead of `impl AsRef<T>` and use `other.into_glib_ptr()`

sdroege (approved):

PR #1394, opened 2022-10-26
nameutil: correct dll link name

elmarco:


MarijnS95:
This code could make much better use of [`strip_prefix()`](https://doc.rust-lang.org/std/primitive.str.html#method.strip_prefix)/[`strip_suffix()`](https://doc.rust-lang.org/std/primitive.str.html#method.strip_suffix) :grimacing:

elmarco:
> This code could make much better use of [`strip_prefix()`](https://doc.rust-lang.org/std/primitive.str.html#method.strip_prefix)/[`strip_suffix()`](https://doc.rust-lang.org/std/primitive.str.html#method.strip_suffix) grimacing
> 

Not really, it needs to strip `.so.X` for example

> EDIT: And [`rsplit_once()`](https://doc.rust-lang.org/std/primitive.str.html#method.rsplit_once) for the `'-'`, I think.

Could be, it does more than just taking the first substring.

Feel free to submit a patch :)

MarijnS95:
> Feel free to submit a patch :)

It looks like you're submitting a patch :wink:

sdroege (approved):

PR #1395, opened 2022-10-26
Add `#[allow(clippy::should_implement_trait)]` if method is named `default`

GuillaumeGomez:
Needed for https://github.com/gtk-rs/gtk-rs-core/pull/816


sdroege (commented):
It's sufficient to put it on the function instead of the type? Sure, go ahead

GuillaumeGomez:
Apparently. :)

PR #1397, opened 2022-11-04
codegen: Handle _finish functions not taking a GError inout param

bilelmoussaoui:


src/analysis/bounds.rs
```diff
@@ -355,19 +359,21 @@ fn format_out_parameters(parameters: &[String]) -> String {
...
 }
 
-fn find_error_type(env: &Env, function: &Function) -> String {
+fn find_error_type(env: &Env, function: &Function) -> Option<String> {
     let error_param = function
         .parameters
         .iter()
-        .find(|param| param.direction == ParameterDirection::Out && param.name == "error")
-        .expect("error type");
+        .find(|param| param.direction == ParameterDirection::Out && param.name == "error")?;
```
sdroege, line 366: Wat. How does this not fail with any existing code.

This should really check for `throws=1` and then find an out parameter of type `GError**`. I'm sure you can find some examples that use `err` instead of `error` for the parameter name :)

sdroege (commented):
looks good otherwise

bilelmoussaoui:
This is blocked by a change that is needed on gio side, as `GioFuture` always returns a `Result<T, E>` which is no longer true with this change.

src/analysis/bounds.rs
bilelmoussaoui: Fixed

sdroege (approved):

Issue #1400, opened 2022-11-14
Incorrect warnings when generating in `sys` mode


Megadash452:
When trying to ignore some symbols, the gir program will falsely flag those symbols as "missing from the library".

For example, in my case i have this ignore in my `Gir.toml`:
```toml
ignore = [
    "GIRepository.MICRO_VERSION",
    "GIRepository.MINOR_VERSION",
]
```
and running gir with this gives the following output:
```
[WARN  libgir::config::gobjects] Configured object `GIRepository.MICRO_VERSION` missing from the library
[WARN  libgir::config::gobjects] Configured object `GIRepository.MINOR_VERSION` missing from the library
```

but when I build and test, those constants are not present in `test/abi.rs`, so the ignore was successful.

@bilelmoussaoui suggested i open this issue to get this fixed.

Megadash452:
I think the problem has to do with these being constants, but somewhere they are being seen as GObject types. In [`src/config/gobjects.rs::resolve_type_ids`](https://github.com/gtk-rs/gir/blob/e94fdc6499e45d89bada123f4d2e63ceb6af997d/src/config/gobjects.rs#L580), the objects arg contains these constants, which then this function tries to find a Gobject type with this name (i think this is what this part does: `library.find_type(0, name);`), but it returns none because there is no type with those names since those are constants.

PR #1405, opened 2022-11-21
Fix incorrect 'missing from library' warning when generating in sys mode

Megadash452:
Resolves [this issue](https://github.com/gtk-rs/gir/issues/1400).


sdroege (approved):

sdroege:
Thanks!

PR #1409, opened 2022-11-30
fix the layout tests

jf2048:
Part of https://github.com/gtk-rs/gtk-rs-core/issues/64, only really needed to pick up `gio-unix-2.0`. Also fixes those warnings about `PACKAGES`, etc being unused on windows


sdroege (approved):

bilelmoussaoui (approved):

jf2048:
BTW I think this means the tests will still just be broken on windows due to not having `gio-unix`? Related to #1052

src/codegen/sys/tests.rs
```diff
@@ -145,7 +145,11 @@ fn prepare_cconsts(env: &Env) -> Vec<CConstant> {
                     // see the GValue machinery around them for example
                     constants.push(CConstant {
                         name: format!("(guint) {}", member.c_identifier),
-                        value: member.value.clone(),
+                        value: member
+                            .value
+                            .parse::<i32>()
```
sdroege, line 150: Are you sure these can never be an actual `u32` that is bigger than `i32::MAX`?

jf2048: The `unwrap_or_else` should cover that?

sdroege: True

PR #1410, opened 2022-12-07
Fix generation without any shared libraries

vlinkz:
For example with [gsettings-desktop-schemas-rs](https://gitlab.gnome.org/vlinkz/gsettings-desktop-schemas-rs), there is no library bound. At the moment `sys/lib.rs` generates with the following at the end:
```rust
#[link(name = "")]
extern "C" {

}
```
which leads to the error:
```
error[E0454]: link name must not be empty
   --> gsettings-desktop-schemas/sys/src/lib.rs:199:15
    |
199 | #[link(name = "")]
    |               ^^ empty link name
```

### DIFF
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -1,6 +1,6 @@
 use std::{
     fs,
-    io::{Result, Write},
+    io::{Error, ErrorKind, Result, Write},
 };
 
 use log::info;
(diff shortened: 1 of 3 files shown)


sdroege:
How does it work without a shared library?

vlinkz:
For example with [gsettings-desktop-schemas](https://gitlab.gnome.org/GNOME/gsettings-desktop-schemas), the `.gir` file has the following namespace element
```xml
  <namespace name="GDesktopEnums"
             version="3.0"
             shared-library=""
             c:identifier-prefixes="GDesktop"
             c:symbol-prefixes="g_desktop">
```
since there isn't an actual shared library, only some enums being defined in a header file.

sdroege:
Ah it's a header-only library. That makes sense.

sdroege:
You probably still want a shared library that provides `GType`s for the enums though, but from a Rust point of view that doesn't matter much. Just means you can't pass these enums through `GValue`s (in any language), only as integers.

src/codegen/sys/lib_.rs
```diff
@@ -71,17 +71,19 @@ fn generate_lib(w: &mut dyn Write, env: &Env) -> Result<()> {
...
-    functions::generate_enums_funcs(w, env, &enums)?;
-    functions::generate_bitfields_funcs(w, env, &bitfields)?;
-    functions::generate_unions_funcs(w, env, &unions)?;
-    functions::generate_records_funcs(w, env, &records)?;
-    functions::generate_classes_funcs(w, env, &classes)?;
-    functions::generate_interfaces_funcs(w, env, &interfaces)?;
-    functions::generate_other_funcs(w, env, &ns.functions)?;
-
-    writeln!(w, "\n}}")?;
+    if !env.namespaces.main().shared_libs.is_empty() {
```
sdroege, line 74: You probably want to add something here that makes things fail if there are any functions but no library name

vlinkz: Something like this?

```rust
    if !ns.functions.is_empty() && env.namespaces.main().shared_libs.is_empty() {
        return Err(Error::new(ErrorKind::Other, "No shared library found, but functions found"));
    }
```

sdroege: Not sure what the easiest. `ns.functions` only contains the global functions and not the methods IIRC.

vlinkz: Maybe something like this:
```rust
    if !ns.package_names.is_empty() && env.namespaces.main().shared_libs.is_empty() {
        return Err(Error::new(ErrorKind::Other, "No shared library found, but package names found"));
    }
```
That way `.gir`s with libraries defined have to have a package name, but headers like this one need to not have one. That way you can't have the combination of having a shared library with functions, but not defining the package.

sdroege: Having a package name seems to be orthogonal from have shared libraries. Your header-only library will have a package name (and as such a `pkg-config` file) or not?

I think what just needs checking here is whether there are global functions, or enums/object/flags/records have functions

vlinkz: Pushed with it modified to this, which checks all the function lengths. Adding more than just `classes` and `interfaces` might be a bit overkill though.

```rust
    if env.namespaces.main().shared_libs.is_empty()
        && !(records.iter().all(|x| x.functions.is_empty())
            && classes.iter().all(|x| x.functions.is_empty())
            && interfaces.iter().all(|x| x.functions.is_empty())
            && bitfields.iter().all(|x| x.functions.is_empty())
            && enums.iter().all(|x| x.functions.is_empty())
            && unions.iter().all(|x| x.functions.is_empty()))
    {
        return Err(Error::new(
            ErrorKind::Other,
            "No shared library found, but functions were found",
        ));
    }
```

sdroege (approved):

sdroege:
Thanks!

PR #1411, opened 2022-12-07
Backport: codegen: Handle _finish functions not taking a GError inout param

bilelmoussaoui:
Backport version of https://github.com/gtk-rs/gir/pull/1397/ that makes uses of GioInfaliableFuture

### DIFF
--- a/src/codegen/function.rs
+++ b/src/codegen/function.rs
@@ -409,10 +409,21 @@ pub fn body_chunk_futures(
 
     let mut body = String::new();
 
+    let future_name = analysis
+        .async_future
+        .as_ref()
+        .map(|a| {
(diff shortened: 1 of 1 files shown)


src/codegen/function.rs
```diff
@@ -402,10 +409,21 @@ pub fn body_chunk_futures(
...
     let mut body = String::new();
 
+    let future_name = analysis
+        .async_future
+        .as_ref()
+        .map(|a| {
+            if a.error_parameters.is_some() {
+                "GioFuture"
+            } else {
+                "GioInfaliableFuture"
```
sdroege, line 419:
```suggestion
                "GioInfallibleFuture"
```

PR #1413, opened 2022-12-09
analysis: Generate correct bounds when move is used for nullable types

bilelmoussaoui:


src/analysis/bounds.rs
```diff
@@ -228,9 +228,11 @@ impl Bounds {
     ) -> String {
         use self::BoundType::*;
         match bound_type {
+            AsRef(_) if move_ && nullable => ".map(|p| p.as_ref().clone().upcast())".to_owned(),
```
sdroege, line 231: This seems strange. We shouldn't really make `AsRef` for the `move` case. But I guess that's for another time to fix?

Seems fine to me otherwise

bilelmoussaoui: We can't do much here and this is only for fundamental types where you would get `Option<impl AsRef<gtk::Expression>>` as param. Otherwise everything would have to be generated manually or we would need a new config option to define a trait to use for that specific fundamental type and it sub-types

sdroege: So would need some more thought. Maybe create an issue for it for some later time

bilelmoussaoui: Opened https://github.com/gtk-rs/gir/issues/1414 for now

Issue #1391, opened 2022-10-22
Builders should store a SmallVec<[(&'static str, Value)]>


jf2048:
This would significantly reduce the stack size of the builder, and would not cause any extra overhead since `Object::new_internal` has to do this anyway. The builders can also unsafely call `new_internal` to avoid checks, since we know they only use valid properties.

After that, the builders could be made generic and the property setters could be moved into extension traits, which would cut down on a lot of the duplicated generated code.

sdroege:
> The builders can also unsafely call `new_internal` to avoid checks, since we know they only use valid properties.

That's not actually what it does. The only thing it doesn't do is checking whether the type implements `GInitable` or `GAsyncInitable`.

> This would significantly reduce the stack size of the builder

It would actually increase the stack size or not? Right now it's just allocated on the heap in a `Vec`. But it gets rid of a useless heap allocation in many cases.

Actually, I was thinking that we could make use of the generic `ObjectBuilder` for the specific builders. But that would go against your idea of omitting the property validity checks as that one always would have to do that.

PR #1417, opened 2022-12-11
Replace builders with generated traits on ObjectBuilder

jf2048:
Alternate solution to #1391. Tested with gtk4 it removes around 40,000 lines of code.

I had to remove `builder_postprocess` too. For that I would like to somehow come up with a way to fold the setup into `FromGlibPtr` and `FromValue`.


src/codegen/object.rs
```diff
@@ -636,8 +561,9 @@ pub fn generate_reexports(
             builders.push(format!("\t{}", cfg));
         }
         builders.push(format!(
-            "\tpub use super::{}::{}Builder;",
-            module_name, analysis.name
+            "\tpub use super::{}::{}BuilderExt;",
+            module_name,
+            analysis.builder_trait_prefix()
```
bilelmoussaoui, line 566: Can't this return the actual builder trait name instead?

sdroege (commented):
Looks generally good to me, nice cleanup :)
Can you create a PR to show the changes before we merge this here?

jf2048:
I think something needs to be done in gtk-rs-core about the missing `builder_postprocess` first

jf2048:
Sorry meant gtk3-rs and gtk4-rs, those seem to be the only places using it

sdroege:
@jf2048 @bilelmoussaoui What's the plan/status here?

jf2048:
It is a real pain to modify that wrapper macro to do what we need to do here

jf2048:
https://github.com/gtk-rs/gtk-rs-core/issues/916

PR #1418, opened 2022-12-11
add a flag for generating variant traits for flags/enums

### LINKED ISSUE #1355: Generate `FromVariant` and `ToVariant` for enums
Currently introspected enums do not easily convert to variants as they just expose their `u32` representation.

For instance this would be useful when storing enums in `gsettings`

jf2048:
GTK will probably not use this, but other libraries can enable it if they want

Fixes #1355


src/config/config.rs
```diff
@@ -88,6 +88,43 @@ impl GirVersion {
     }
 }
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum VariantTraitMode {
+    None,
+    Repr,
+    String,
+}
```
bilelmoussaoui, line 96:
```suggestion
#[derive(Default, Clone, Copy, Debug, Eq, PartialEq)]
pub enum VariantTraitMode {
    #[default]
    None,
    Repr,
    String,
}
```
instead of a manual default impl?

src/config/config.rs
```diff
@@ -88,6 +88,43 @@ impl GirVersion {
...
+impl VariantTraitMode {
+    pub fn none(self) -> bool {
+        self == Self::None
+    }
+    pub fn repr(self) -> bool {
+        self == Self::Repr
+    }
+    pub fn string(self) -> bool {
+        self == Self::String
+    }
```
bilelmoussaoui, line 107:
```suggestion
    pub fn is_none(self) -> bool {
        self == Self::None
    }
    pub fn is_repr(self) -> bool {
        self == Self::Repr
    }
    pub fn is_string(self) -> bool {
        self == Self::String
    }
```

src/config/config.rs
jf2048: Oh right, I forgot that was a thing now

jf2048:
cc @pbor

pbor:
I have skimmed through the patch (I am not familiar enough with gir code to properly review)... can you show an example of how the syntax in `Gir.toml` would look like?

src/config/config.rs
```diff
@@ -255,6 +288,13 @@ impl Config {
             None => true,
         };
 
+        let generate_variant_traits = match toml.lookup("options.generate_variant_traits") {
```
bilelmoussaoui, line 291: The new option should be added to the book

jf2048:
It's just `generate_variant_traits = "repr"` or `generate_variant_traits = "string"` on the gir options (or on the type itself) and then you get traits for either serializing to int or string variant.

PR #1421, opened 2022-12-19
Inline various enum/flags functions

sdroege:
CC @jf2048

### DIFF
--- a/src/codegen/enums.rs
+++ b/src/codegen/enums.rs
@@ -357,25 +357,16 @@ impl FromGlib<{sys_crate_name}::{ffi_name}> for {name} {{
         self.into_glib()
     }}
 
+    #[allow(clippy::match_single_binding)]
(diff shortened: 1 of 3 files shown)


src/codegen/enums.rs
```diff
@@ -236,6 +236,13 @@ fn generate_enum(
         )?;
     }
 
+    // Only inline from_glib / into_glib implementations if there are not many enums members
+    let maybe_inline = if members.len() <= 12 {
```
sdroege, line 240: 12 is just a random number

sdroege: That's always 4 bytes :) The relevant metric here is the number of conditions in the match, which is equivalent to the number of enum variants.

GuillaumeGomez: Oh I see.

GuillaumeGomez (approved):

PR #1424, opened 2023-01-06
Generate string constants as `&'static GStr` / `[u8]`

sdroege:
This gets rid of all the `once_cell` `Lazy` usage and allows using them as NUL-terminated strings in more places.

### DIFF
--- a/src/codegen/constants.rs
+++ b/src/codegen/constants.rs
@@ -18,7 +18,7 @@ pub fn generate(env: &Env, root_path: &Path, mod_rs: &mut Vec<String>) {
     }
 
     let sys_crate_name = env.main_sys_crate_name();
-    imports.add("std::ffi::CStr");
+    imports.add("glib::GStr");
 
     file_saver::save_to_file(path, env.config.make_backup, |w| {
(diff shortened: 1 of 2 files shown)


sdroege:
See https://github.com/gtk-rs/gtk-rs-core/pull/885 for generated code

src/codegen/constants.rs
```diff
@@ -1,28 +1,25 @@
...
 pub fn generate(env: &Env, root_path: &Path, mod_rs: &mut Vec<String>) {
     let path = root_path.join("constants.rs");
-    let mut imports = Imports::new(&env.library);
 
     if env.analysis.constants.is_empty() {
         return;
     }
 
     let sys_crate_name = env.main_sys_crate_name();
-    imports.add("std::ffi::CStr");
```
bilelmoussaoui, line 21: I would keep this and import GStr here, that would drop various lines of code.

sdroege: Sure, done :)

bilelmoussaoui (approved):

PR #1429, opened 2023-01-13
Handle user_data renaming

### LINKED ISSUE #1428: Can't autogenerate `g_file_copy()` and others with latest gir files anymore
This comes from https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3111 . Need to fix that before the next release or we won't be able to update gir files anymore.

CC @GuillaumeGomez who implemented support for callbacks at some point.

GuillaumeGomez:
Fixes #1428.

@sdroege Can you confirm it fixes the issue please? Don't know when I'll have the time otherwise.

### DIFF
--- a/src/analysis/function_parameters.rs
+++ b/src/analysis/function_parameters.rs
@@ -198,12 +198,30 @@ pub fn analyze(
         .filter_map(|p| p.array_length.map(|pos| (pos, p)))
         .collect();
 
+    let mut to_remove = Vec::new();
+    let mut correction_instance = 0;
+
+    for par in function_parameters.iter() {
+        if par.scope.is_none() {
+            continue;
+        }
+        if let Some(index) = par.closure {
+            to_remove.push(index);
+        }
+        if let Some(index) = par.destroy {
+            to_remove.push(index);
+        }
+    }
+
     for (pos, par) in function_parameters.iter().enumerate() {
         let name = if par.instance_parameter {
             par.name.clone()
         } else {
             nameutil::mangle_keywords(&*par.name).into_owned()
         };
+        if par.instance_parameter {
+            correction_instance = 1;
+        }
 
         let configured_parameters = configured_functions.matched_parameters(&name);
 
@@ -222,7 +240,7 @@ pub fn analyze(
             }
         };
 
-        if async_func && async_param_to_remove(&par.name) {
+        if async_func && to_remove.contains(&(pos - correction_instance)) {
             add_rust_parameter = false;
         }
         let mut transfer = par.transfer;
@@ -483,9 +501,4 @@ fn has_length(env: &Env, typ: TypeId) -> bool {
         Type::Alias(alias) => has_length(env, alias.typ),
         _ => false,
     }
-}
-
-pub fn async_param_to_remove(name: &str) -> bool {
-    name == "user_data" || name.ends_with("data") // FIXME: use async indexes
-                                                  // instead
 }
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -539,7 +539,7 @@ impl Library {
                 if typ.is_some() {
                     return Err(parser.fail("Too many <type> elements"));
                 }
-                self.read_function(parser, ns_id, elem.name(), elem)
+                self.read_function(parser, ns_id, elem.name(), elem, true)
                     .map(|f| {
                         typ = Some((Type::function(self, f), None, None));
                     })
@@ -570,7 +570,7 @@ impl Library {
         ns_id: u16,
         elem: &Element,
     ) -> Result<(), String> {
-        self.read_function_if_not_moved(parser, ns_id, elem.name(), elem)?
+        self.read_function_if_not_moved(parser, ns_id, elem.name(), elem, true)?
             .map(|func| {
                 let name = func.name.clone();
                 self.add_type(ns_id, &name, Type::Function(func))
@@ -749,7 +749,7 @@ impl Library {
         ns_id: u16,
         elem: &Element,
     ) -> Result<(), String> {
-        self.read_function_if_not_moved(parser, ns_id, "global", elem)
+        self.read_function_if_not_moved(parser, ns_id, "global", elem, false)
             .map(|func| {
                 if let Some(func) = func {
                     self.add_function(ns_id, func);
@@ -918,6 +918,7 @@ impl Library {
         ns_id: u16,
         kind_str: &str,
         elem: &Element,
+        is_callback: bool,
     ) -> Result<Function, String> {
         let fn_name = elem.attr_required("name")?;
         let c_identifier = elem.attr("identifier").or_else(|| elem.attr("type"));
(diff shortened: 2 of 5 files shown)


sdroege:
Yes but it breaks other things

```diff
diff --git a/gdk-pixbuf/src/auto/pixbuf.rs b/gdk-pixbuf/src/auto/pixbuf.rs
index c2fc58abdc..0359f02759 100644
--- a/gdk-pixbuf/src/auto/pixbuf.rs
+++ b/gdk-pixbuf/src/auto/pixbuf.rs
@@ -48,7 +48,7 @@ impl Pixbuf {
     ) -> Pixbuf {
         unsafe {
             from_glib_full(ffi::gdk_pixbuf_new_from_bytes(
-                data.to_glib_none().0,
+                Box_::into_raw(data) as *mut _,
                 colorspace.into_glib(),
                 has_alpha.into_glib(),
                 bits_per_sample,
@@ -217,7 +217,11 @@ impl Pixbuf {
     #[doc(alias = "gdk_pixbuf_new_from_xpm_data")]
     #[doc(alias = "new_from_xpm_data")]
     pub fn from_xpm_data(data: &[&str]) -> Pixbuf {
-        unsafe { from_glib_full(ffi::gdk_pixbuf_new_from_xpm_data(data.to_glib_none().0)) }
+        unsafe {
+            from_glib_full(ffi::gdk_pixbuf_new_from_xpm_data(
+                Box_::into_raw(data) as *mut _
+            ))
+        }
     }
 
     #[doc(alias = "gdk_pixbuf_add_alpha")]
```

src/analysis/function_parameters.rs
```diff
@@ -198,12 +198,36 @@ pub fn analyze(
...
 
+    let mut to_remove = Vec::new();
+    let mut correction_instance = 0;
+
+    for par in function_parameters.iter() {
+        if par.scope.is_none() {
+            continue;
+        }
+        if let Some(index) = par.closure {
+            to_remove.push(index);
```
sdroege, line 209: You could take a look at how this is handled for `array` attributes. See e.g. `find_index_to_ignore` in `src/analysis/functions.rs`. Not sure if that improves anything though

sdroege, line 209: You could take a look at how this is handled for `array` attributes. See e.g. `find_index_to_ignore` in `src/analysis/functions.rs`. Not sure if that improves anything though

src/analysis/function_parameters.rs
```diff
@@ -198,12 +198,36 @@ pub fn analyze(
...
         .collect();
 
+    let mut to_remove = Vec::new();
+    let mut correction_instance = 0;
+
+    for par in function_parameters.iter() {
+        if par.scope.is_none() {
+            continue;
+        }
+        if let Some(index) = par.closure {
```
sdroege, line 208: Unfortunately the attribute is also (wrongly) on the `user_data` sometimes. When using it you should first check if the parameter is callback-typed, and if not just ignore the existence of the attribute.

src/analysis/function_parameters.rs
sdroege: Also if `scope = notified` you would additionally take the `GDestroyNotify` at `index+1` and ignore that.

That seems to be handled kind of in `src/analysis/functions.rs:analyze_callbacks()`, so maybe that should go both into the same place?

sdroege: And there's e.g. `Trampoline::user_data_index` and `destroy_index`, so it seems like there kind of is infrastructure for this stuff already.

sdroege: Aaaand in this file there is `async_param_to_remove()`, which is completely wrong (it matches the parameter name too).

src/analysis/function_parameters.rs
GuillaumeGomez: Hence why I added above `par.scope.is_none()` to ignore this case. ;)

src/analysis/function_parameters.rs
GuillaumeGomez: Which is why I removed this function. ;)

src/analysis/function_parameters.rs
sdroege: Ah clever. That works of course

src/analysis/function_parameters.rs
sdroege: Right, forgot about that sorry :) The other things still apply though

GuillaumeGomez: Absolutely, but first I'd like to understand why these changes still don't work. Then I'll clean up.

src/parser.rs
```diff
@@ -952,6 +953,29 @@ impl Library {
             "attribute" => parser.ignore_element(),
             _ => Err(parser.unexpected_element(elem)),
         })?;
+        let mut fix_pos = 0;
+        // Fixing the closure attributes in case they're badly set.
+        for index in 0..params.len() {
```
GuillaumeGomez, line 958: Most of the fix is here, the rest is "cleanup".

src/parser.rs
```diff
@@ -952,6 +953,29 @@ impl Library {
...
+                fix_pos += 1;
+            } else if let Some(pos) = params[index].closure {
+                if params[pos - fix_pos].closure.is_none() {
+                    params[pos - fix_pos].closure = Some(pos - fix_pos);
+                }
+            }
+        }
+        // Adding closure attribute for callback parameters in case there is something completely
+        // wrong.
+        if is_callback && params.last().map(|x| x.closure.is_none()).unwrap_or(false) {
```
sdroege, line 969: The attribute is not supposed to be on the callback types, at all. The last parameter is always going to be the one we're looking for here

src/parser.rs
```diff
@@ -952,6 +953,29 @@ impl Library {
             "attribute" => parser.ignore_element(),
             _ => Err(parser.unexpected_element(elem)),
         })?;
+        let mut fix_pos = 0;
+        // Fixing the closure attributes in case they're badly set.
+        for index in 0..params.len() {
+            if params[index].instance_parameter {
+                fix_pos += 1;
+            } else if let Some(pos) = params[index].closure {
+                if params[pos - fix_pos].closure.is_none() {
```
sdroege, line 962: This is setting it so that callback points at user data and the other way around? The attribute is only supposed to be on the callback

src/parser.rs
GuillaumeGomez: It's how `gir` knows unfortunately. We use this attribute everywhere.

src/parser.rs
```diff
@@ -952,6 +953,11 @@ impl Library {
             "attribute" => parser.ignore_element(),
             _ => Err(parser.unexpected_element(elem)),
         })?;
+        // The last argument of a callback is ALWAYS user data, so it has to be marked as such
+        // in case it's missing.
```
sdroege, line 957: It will always be missing unless the annotations are wrong :)

GuillaumeGomez: Well, apparently it was always present until now...

sdroege (approved):

PR #1431, opened 2023-01-21
Implement object builders around `glib::object::ObjectBuilder`

sdroege:
This removes quite a bit of unnecessary code and reduces stack memory usage when using builders.

### DIFF
--- a/src/codegen/object.rs
+++ b/src/codegen/object.rs
@@ -217,7 +217,7 @@ pub fn generate(
             ///
             /// This method returns an instance of [`{builder_name}`](crate::builders::{builder_name}) which can be used to create [`{name}`] objects.
             pub fn builder() -> {builder_name} {{
-                {builder_name}::default()
+                {builder_name}::new()
             }}
         ",
                 name = analysis.name,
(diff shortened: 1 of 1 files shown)


sdroege:
Regenerated code in https://github.com/gtk-rs/gtk-rs-core/pull/914

src/codegen/object.rs
```diff
@@ -360,171 +356,108 @@ fn generate_builder(w: &mut dyn Write, env: &Env, analysis: &analysis::object::I
         analysis.name,
     )?;
     writeln!(w, "#[must_use = \"The builder must be built to be used\"]")?;
-    writeln!(w, "pub struct {}Builder {{", analysis.name)?;
+    writeln!(
+        w,
+        "pub struct {name}Builder {{
+            builder: {glib_name}::object::ObjectBuilder<'static, {name}>,
+        }}
```
bilelmoussaoui, line 363:
```suggestion
        "pub struct {name}Builder({glib_name}::object::ObjectBuilder<'static, {name}>);
```

maybe?

sdroege: I have plans to allow for other fields there :)

PR #1379, opened 2022-09-21
Reworked the book/tutorial

pentamassiv:
A while ago I created a draft PR ( https://github.com/gtk-rs/gir/pull/1354) to improve the book. A good number of commits were done since then. I am not that familiar with git and I think now the PR is ready, so I created this PR.
A few months have passed and I didn't find the time to add the missing information to explain the whole process of how to create the bindings for pango. The overhaul contains all the information of the current version and adds some to it. I believe this version can already be merged and is already an improvement.
To recap, the improvements include:

- Restructuring the book a little bit to make it easier to find information
- Adding description on what gir is and when it can be used
- Explanation on how to set up the project folder
- Explanation on how to find .gir files
- Mentioning of various other foot guns that a beginner might encounter

A native speaker should probably proofread it.

PR #1439, opened 2023-02-01
docs: Update Cargo.toml glib dependency

davidmhewitt:
I'm just following through the book and found it wasn't possible to continue without changing this.


bilelmoussaoui:
Ideally we would land #1379

bilelmoussaoui:
No longer needed

PR #1446, opened 2023-02-26
Refactor `impl Default` generation

pbor:
Move the detection of a default constructor function in the analysis. Move the logic about using Object::new into object.

### DIFF
--- a/src/analysis/info_base.rs
+++ b/src/analysis/info_base.rs
@@ -1,5 +1,5 @@
 use super::{imports::Imports, *};
-use crate::{codegen::Visibility, library, version::Version};
+use crate::{codegen::Visibility, config::gobjects::GStatus, library, version::Version};
 
 #[derive(Debug, Default)]
 pub struct InfoBase {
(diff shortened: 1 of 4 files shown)


pbor:
I know a gir rework is ongoing, so I am not sure if refactoring PRs makes sense.

The reason for this is to tackle https://github.com/gtk-rs/gtk-rs-core/issues/975 by adding a new `generate_default = false` flag to Gir.toml, but before adding it I figured I would submit the refactoring

pbor:
gtk3 fails because it has

```
impl Default for RadioButton {
    fn default() -> Self {
        Self::new()
    }
}
```

which is now generated... on the one hand I guess it is a good thing we can drop manual code, on the other it is not clear to me why this was not happening  before :)

src/codegen/object.rs
```diff
@@ -340,6 +333,27 @@ pub fn generate(
     Ok(())
 }
 
+fn generate_default(w: &mut dyn Write, env: &Env, analysis: &analysis::object::Info) -> Result<()> {
+    let name = &analysis.name;
+    if let Some(func) = analysis.default_constructor() {
+        general::declare_default(w, env, name, func)
+    } else if has_builder_properties(&analysis.builder_properties) {
```
pbor, line 340: The reason why with this patch RadioButton etc get picked up is because before we were falling back to object::new only if there was a `new()` function, but it did not have 0 params.

That seems rather odd and with the new code I did not get any false positive in core and gtk3

src/analysis/info_base.rs
```diff
@@ -38,4 +38,18 @@ impl InfoBase {
...
+    pub fn default_constructor(&self) -> Option<&functions::Info> {
+        self.functions.iter().find(|f| {
+            !f.hidden
+                && f.status.need_generate()
+                && f.kind == library::FunctionKind::Constructor
+                && f.status == GStatus::Generate
+                // For now we only look for new() with no params
+                && f.name == "new"
+                && f.parameters.rust_parameters.is_empty()
+                // Cannot generate Default implementation for Option<>
```
bilelmoussaoui, line 51: Should also do the same for Result<T>?

bilelmoussaoui:
Would be nice to land this. But it needs PRs in at least 2 out of gtk-rs-core/gstreamer-rs/gtk4-rs with the resulting changes.

pbor:
I looked at this again, and I do not think the PR is correct: the old code generated the fallback `Object::new()` constructor only if `new` was found (and was not hidden etc).

To make it correct we would have to remove the "0 args" check in the generic code, but that would defeat the point of centralizing this logic.

PR #1447, opened 2023-03-01
Use collection types for return values

jf2048:
Is there a better way to get the actual type of a C array from the builder? The best I could do is adding the `c_type` like this.

I left the `Vec<OsString>` and `Vec<PathBuf>` return as is, since those require allocations and they are opt-in.


sdroege:
Can you make a draft PR with the changes this causes in the generated code?

src/analysis/function_parameters.rs
```diff
@@ -240,7 +240,12 @@ pub fn analyze(
             }
         };
 
-        if async_func && to_remove.contains(&(pos - correction_instance)) {
+        if async_func
+            && pos
+                .checked_sub(correction_instance)
+                .map(|p| to_remove.contains(&p))
+                .unwrap_or(false)
+        {
```
pbor, line 248: Are these `checked_sub` related to https://github.com/gtk-rs/gir/pull/1427/files ?

jf2048: Seems to be a similar issue, but I was getting panics just running against gtk-rs-core

sdroege:
What also might be worth considering here is parameters.
  - transfer-none array parameters can be handled directly as `&[T]` for many types (shared, boxed, object) as long as the C function takes a length (i.e. does not require a NULL-terminated array)
  - transfer-full collection parameters can be handled as exactly the underlying type
  - For NULL-terminated string arrays there's `impl IntoStrV` (transfer-none) and `StrV`

This might also be something to handle in a separate PR though.

jf2048:
Those should be handled in #1448

src/analysis/function_parameters.rs
jf2048: Resolving since that one is merged now

src/analysis/rust_type.rs
```diff
@@ -373,7 +381,57 @@ impl<'env> RustTypeBuilder<'env> {
...
+                                    ),
+                                    SList(_) if is_ptr => {
+                                        format!("{}<{typ}>", use_glib_type(self.env, "SList"))
+                                    }
+                                    SList(_) if is_string => format!(
+                                        "{}<{}>",
+                                        use_glib_type(self.env, "SList"),
+                                        use_glib_type(self.env, "GStringPtr")
+                                    ),
+                                    _ => format!("Vec<{typ}>"),
```
sdroege, line 433: Which cases are left here? Only `PtrArray`?

sdroege:
> Those should be handled in https://github.com/gtk-rs/gir/pull/1448

The first two fit more here or in a separate PR, or not? #1448 is string stuff, this here is collections :) The third one fits better in #1448 though, that's true.

src/analysis/rust_type.rs
jf2048: I suppose `GArray` as well? But we need wrappers with type parameters

sdroege: Can you add a comment for listing the remaining cases?

sdroege:
Looks good to me otherwise, if that wasn't clear. How should we proceed? :)

jf2048:
Sorry for delay, I did some more testing and this is looking good for me too. But I think this situation should be resolved first: https://github.com/gtk-rs/gtk-rs-core/pull/1078

PR #1448, opened 2023-03-01
Change &str input args into impl IntoGStr

jf2048:
Also changes strings with transfer=full to `impl Into<GString>`

I added new `Bound`s but not a new `Transformation`, if anything else requires this stack-callback approach then one could be added


sdroege:
Same question here, can you create a draft PR with the changes this causes? :)

sdroege:
There's also `impl Into<GString>` for the transfer-full case and `impl IntoStrV` for the "array of strings" case that would be useful to consider here.

sdroege:
>  analysis: Don't use IntoGStr for string params with length 

Why not?

jf2048:
I added `IntoStrV` but I think we want to do https://github.com/gtk-rs/gtk-rs-core/issues/978 first, I don't want to make gir subtract one from `len()` only to remove that later

jf2048:
> Why not?

Those don't need to be nul terminated, it's fine for them to just be `&str`. I mean, I guess it could be `AsRef<str>` if we wanted

sdroege:
I think the futures change would be worth its own PR. That can be merged already :)

PR #1455, opened 2023-04-06
Replace dox feature with docsrs attribute

### LINKED ISSUE #1454: Suggestion: Remove dox feature in favor of rustc args
Cargo features are supposed to be additive, but the "dox" feature isn't. Compiling with "dox" skips compile time checks for the underlying C libraries and therefore should not be enabled during normal development.

A better solution would be to use a "docsrs" or "dox" config argument for rustc directly. This seems to be what most crates do and it's also what [docs.rs recommends](https://docs.rs/about/builds#cross-compiling). This would make possible to use `--all-features` again and removes the need to pass the "dox" feature to the dependencies.

The code for conditional compilation would only change from `#[cfg(feature = "dox")]` to `#[cfg(dox)]`.

This doesn't only affect the gtk4 crate, so feel free to move the issue if there's a better repo for this topic.

AaronErhardt:
Fixes #1454 

I think this should be all from the gir side.
I've tested this branch on gtk4-rs and everything looked and worked great so far.

Let me know if I should submit PRs for various gtk-rs crates once this is merged.

### DIFF
--- a/book/src/tutorial/generate_docs.md
+++ b/book/src/tutorial/generate_docs.md
@@ -30,10 +30,22 @@ rustdoc-stripper -s -n
 rustdoc-stripper -g -o docs.md
 ```
 
-Try building the documentation and also try it with the various features you might have
+Try building the documentation and also try it with the various features you might have:
 
 ```sh
 cargo doc
+```
+
+For building the docs on systems without the required dependencies installed (for example docs.rs or CI jobs), you can use the `docsrs` attribute.
+This will disable the linking of the underlying C libraries and won't throw an error if those libraries are not available.
+
+On top of that, the `docsrs` attribute will enable the [doc_cfg feature](https://doc.rust-lang.org/beta/unstable-book/language-features/doc-cfg.html) which is only available in the nightly compiler toolchain at the time of writing.
+This is mostly useful for clarifying feature requirements through the docs.
+
+To build the docs with the `docsrs` attribute, you can use the following command:
+
+```sh
+RUSTDOCFLAGS='--cfg docsrs' cargo +nightly doc
 ```
 
 Congratulations, we are done.
--- a/book/src/tutorial/generate_docs.md
+++ b/book/src/tutorial/generate_docs.md
@@ -47,6 +47,6 @@ Gir will not override them when you re-generate bindings.
 Easy, right.
 If this is your first time publishing a crate, you can find a detailed guide [here](https://doc.rust-lang.org/cargo/reference/publishing.html).
 
-Before you publish the crate, please ensure docs.rs will activate the dox feature and the dox feature of the safe wrapper crate also activates the feature of its dependencies and the unsafe FFI bindings you created.
+Before you publish the crate, please ensure docs.rs will activate the `docsrs` attribute.
 Feel free to go back to the chapter about the [Cargo.toml file of the safe wrapper](high_level_rust_api.md#the-cargotoml-file) to read more about it.
 If you skip this step, your crate and all crates depending on it will not have documentation available on docs.rs.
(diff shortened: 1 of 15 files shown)


book/src/tutorial/high_level_rust_api.md
```diff
@@ -115,7 +113,7 @@ We need to include all `auto` files in our library.
 To do so, let's update the `src/lib.rs` file as follows:
 
 ```rust
-#![cfg_attr(feature = "dox", feature(doc_cfg))]
+#![cfg_attr(docsrs, feature(doc_cfg))]
```
bilelmoussaoui, line 116: Can't this be named dox?

bilelmoussaoui:
I would also suggest documenting how cargo doc should be ran in the gir book if possible

book/src/tutorial/high_level_rust_api.md
AaronErhardt: Sure, we can change this. "docsrs" is just what seems to be used in other crates, I haven't seen "dox" in the wild yet. Just let me know if you prefer "dox".

sdroege: Which other crates are that?

sdroege: I guess that decides it then. Let's use the same

book/src/tutorial/generate_docs.md
```diff
@@ -47,6 +47,6 @@ Gir will not override them when you re-generate bindings.
 Easy, right.
 If this is your first time publishing a crate, you can find a detailed guide [here](https://doc.rust-lang.org/cargo/reference/publishing.html).
 
-Before you publish the crate, please ensure docs.rs will activate the dox feature and the dox feature of the safe wrapper crate also activates the feature of its dependencies and the unsafe FFI bindings you created.
+Before you publish the crate, please ensure docs.rs will activate the `docsrs` attribute.
```
bilelmoussaoui, line 50: What about when you run cargo doc like we do on the CI side of things?

AaronErhardt: If you have the gtk4-devel package installed, you should be fine without any extra effort.

For annotation of required features with `doc_attr` (e. g. for CI) you need to use nightly as usual and set one env variable for rustdoc, for example: `RUSTDOCFLAGS='--cfg docsrs' cargo +nightly doc`.

Should I mention that in the docs as well?

bilelmoussaoui: Yes please

GuillaumeGomez:
So sadly it will require to add a `--feature` argument when running `cargo doc`. Not much we can do about it... Really annoyed that `cargo` still doesn't allow to have `doc` profile for that.

bilelmoussaoui:
> So sadly it will require to add a `--feature` argument when running `cargo doc`. Not much we can do about it... Really annoyed that `cargo` still doesn't allow to have `doc` profile for that.

why a --feature is needed? i thought everything would be handled by the RUSTFLAGS?

GuillaumeGomez:
Sorry, you're right. Wrote too quickly. `RUSTFLAGS` is what I meant.

sdroege:
And `generator.py` would set that accordingly before calling `cargo doc`.

AaronErhardt:
Speaking of `generator.py`, I haven't changed anything on this file so far. As far as I understand it doesn't run `cargo doc` but expects the user to do so after following all required steps. Should I change something about that?

Everything else should be in place by now, I've pushed an update for the book earlier today.

sdroege:
Oh right. But we set `RUSTFLAGS` for that already anyway so adding just one more thing is not a problem. Currently we have to set `RUSTFLAGS=-Z unstable-options` at least.

AaronErhardt:
Is there still anything missing in this PR?

bilelmoussaoui (approved):

sdroege:
@AaronErhardt Once merged, will you create PRs for gtk-rs-core and gtk4-rs at least to update accordingly?

AaronErhardt:
Sure

sdroege:
Great, then let's get this in :) I can take care of gstreamer-rs once the others are in.

PR #1459, opened 2023-04-15
codegen: Drop useless any on a single feature

bilelmoussaoui:
any was used at the time we had the dox feature. Nowadays it is only a single feature so the usage of any is useless


bilelmoussaoui:
Going to go ahead and land this so we can regen gtk4-rs with the latest doc cfg changes

PR #1461, opened 2023-04-28
codegen: only generate trait signatures once

jf2048:
Is there some reason this is not done already? It seems to reduce code size in gtk4-rs by about 5-6%.

### DIFF
--- a/src/codegen/object.rs
+++ b/src/codegen/object.rs
@@ -484,41 +484,10 @@ fn generate_builder(w: &mut dyn Write, env: &Env, analysis: &analysis::object::I
 }
 
 fn generate_trait(w: &mut dyn Write, env: &Env, analysis: &analysis::object::Info) -> Result<()> {
-    write!(w, "pub trait {}: 'static {{", analysis.trait_name)?;
-
-    for func_analysis in &analysis.methods() {
-        function::generate(
-            w,
-            env,
-            Some(analysis.type_id),
-            func_analysis,
-            Some(&analysis.specials),
-            analysis.version,
-            true,
-            true,
-            1,
-        )?;
-    }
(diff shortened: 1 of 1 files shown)


sdroege:
I don't know why it is done this way. It was like that when I first looked at gtk-rs :) @GuillaumeGomez ?

bilelmoussaoui (approved):

GuillaumeGomez (approved):

GuillaumeGomez:
I don't remember at all. It looks like a failed rebase. If it doesn't change anything, then it's a great improvement!

sdroege:
What do you mean with failed rebase?

Also when regenerating, this will require changes to all the manual traits too.

GuillaumeGomez:
Then I clearly misunderstood what it was doing. Oh well.

sdroege:
It removes the ext trait function implementations and moves them into the trait definition as default implementations for the functions. So the ext trait implementation is now only a single line, and the signatures for the functions have to provided only a single time.

The main difference here is that if it's possible to also implement the trait on other types then they would automatically get the default implementations of the functions now instead of having to provide some. The default implementations would likely be wrong for any other implementation. But can there be other implementations of those traits or would all possible implementations overlap with the blanket implementation?

GuillaumeGomez:
So the `show` method for `WidgetExt` default impl would be working only for `Widget` and not for any other types implementing `WidgetExt` is what you meant?

sdroege:
It would work for `T: IsA<Widget>` but not for `i32`. But can someone implement the trait for `i32`?

jf2048:
I think it makes sense to say we don't want anybody implementing `WidgetExt` etc for random types, maybe even seal those traits

Issue #1319, opened 2022-02-20
Implement basic vfunc support


bilelmoussaoui:
Currently, the parsing part ignores the virtual functions definitions. Ideally we would need:

- [x] Parse the XML for all the various vfuncs
- [x] Have a `[object.subclass]` configuration that would allow you to

    * [x] Modify the subclassing trait, e.g it would be by default called `TypeImpl` & `TypeImplExt` for chaining up
    * [x] Modify the `[object.subclass.vfunc]` by marking at either `ignore = true` or `rename = "some_different_vfunc_name"`
    * In a far feature, configure the vfuncs once we can auto generate them
- [x] Embed the docs for the `TypeImpl::some_vfunc` from the corresponding `TypeClass::some_vfunc`
- [x] Generate working docs links for the vfuncs

hohilwik:
Hi, I would like to work on this. It would take me at least a week to go through the codebase before I can start contributing though, is that okay?

bilelmoussaoui:
Sure, make sure to check https://github.com/gtk-rs/gir/pull/604 for inspiration as well.

hohilwik:
Could you give me some examples of the kinds of parsing for virtual-method parser.rs needs to support? I am not entirely sure of the inputs

bilelmoussaoui:
For example
``` xml
      <virtual-method name="get_action_name" invoker="get_action_name">
        <attribute name="org.gtk.Property.get" value="action-name"/>
        <doc xml:space="preserve">Gets the action name for @actionable.</doc>
        <return-value transfer-ownership="none" nullable="1">
          <doc xml:space="preserve">the action name</doc>
          <type name="utf8" c:type="const char*"/>
        </return-value>
        <parameters>
          <instance-parameter name="actionable" transfer-ownership="none">
            <doc xml:space="preserve">a `GtkActionable` widget</doc>
            <type name="Actionable" c:type="GtkActionable*"/>
          </instance-parameter>
        </parameters>
      </virtual-method>
```

Look at the files in the gir-files repository.

hohilwik:
Thank you. I will try to get the first task done by 16th

hohilwik:
Hello, I have slowly realized that this issue would require hundreds of lines of code so, if possible could you give me some guidance on the ways you want all this implemented? I am not very familiar with the codebase so don't know how gobject and gir internals work. 

[...]

PR #1463, opened 2023-04-29
Implement basic virtual methods support

bilelmoussaoui:
The intention behind this PR is not to add code generation for subclassing code but it does a good step forward. 

What the PR does currently:
- Parse virtual-methods from the gir file 
- Add configuration options for `[object.virtual_method]`
- Embed the docs for virtual methods
- Working docs links witth gi-docgen

Before

image

image

After

image

image

TODO: 
- [x] Add more tests for parsing config options
- [x] Document the new config options
- [x] Test it with gtk-rs-core/gstreamer-rs

Parts of #1319

### DIFF
--- a/src/analysis/object.rs
+++ b/src/analysis/object.rs
@@ -20,6 +20,7 @@ pub enum LocationInObject {
     Impl,
     VirtualExt,
     ClassExt,
+    ClassExtManual,
     Ext,
     ExtManual,
     Builder,
@@ -94,6 +95,7 @@ impl Info {
     /// Returns the location of the function within this object
     pub fn function_location(&self, fn_info: &functions::Info) -> LocationInObject {
         if fn_info.kind == FunctionKind::ClassMethod {
+            // TODO: Fix location here once we can auto generate virtual methods
             LocationInObject::ClassExt
         } else if fn_info.kind == FunctionKind::VirtualMethod {
             // TODO: Fix location here once we can auto generate virtual methods
@@ -138,10 +140,13 @@ impl Info {
                     trait_name.into(),
                 )
             }
-            LocationInObject::ClassExt => (
-                format!("subclass::prelude::{}", self.trait_name).into(),
-                self.trait_name.as_str().into(),
-            ),
+            LocationInObject::ClassExt | LocationInObject::ClassExtManual => {
+                let trait_name = format!("{}Ext", self.trait_name);
+                (
+                    format!("subclass::prelude::{}", trait_name).into(),
+                    trait_name.into(),
+                )
+            }
             LocationInObject::Builder => {
                 panic!("C documentation is not expected to link to builders (a Rust concept)!")
             }
--- a/src/analysis/object.rs
+++ b/src/analysis/object.rs
@@ -18,6 +18,7 @@ use crate::{
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum LocationInObject {
     Impl,
+    VirtualExt,
     Ext,
     ExtManual,
     Builder,
@@ -91,7 +92,10 @@ impl Info {
 
     /// Returns the location of the function within this object
     pub fn function_location(&self, fn_info: &functions::Info) -> LocationInObject {
-        if self.final_type
+        if fn_info.kind == FunctionKind::VirtualMethod {
+            // TODO: Fix location here once we can auto generate virtual methods
+            LocationInObject::VirtualExt
+        } else if self.final_type
             || self.is_fundamental
             || matches!(
                 fn_info.kind,
@@ -123,6 +127,14 @@ impl Info {
                 format!("prelude::{}", self.trait_name).into(),
                 self.trait_name.as_str().into(),
             ),
+            LocationInObject::VirtualExt => {
+                // TODO: maybe a different config for subclass trait name?
+                let trait_name = format!("{}Impl", self.trait_name.trim_end_matches("Ext"));
+                (
+                    format!("subclass::prelude::{trait_name}").into(),
+                    trait_name.into(),
+                )
+            }
             LocationInObject::Builder => {
                 panic!("C documentation is not expected to link to builders (a Rust concept)!")
             }
--- a/src/analysis/object.rs
+++ b/src/analysis/object.rs
@@ -37,6 +37,7 @@ pub struct Info {
     pub trait_name: String,
     pub has_constructors: bool,
     pub has_functions: bool,
+    pub virtual_methods: Vec<functions::Info>,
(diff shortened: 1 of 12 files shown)


bilelmoussaoui:
In the latest commits, I went ahead and added docs support for class methods.

Before
image

image


After

image

image

src/codegen/doc/mod.rs
```diff
@@ -178,16 +178,9 @@ fn generate_doc(w: &mut dyn Write, env: &Env) -> Result<()> {
                     .find(move |f| &f.glib_name == c_identifier);
                 let fn_new_name = f_info.and_then(|analysed_f| analysed_f.new_name.clone());
                 let doc_trait_name = f_info.and_then(|f| f.doc_trait_name.as_ref());
-                let doc_struct_name = f_info.and_then(|f| f.doc_struct_name.as_ref());
```
bilelmoussaoui, line 181: Not sure why i dropped this, this is wrong

bilelmoussaoui:
This should be ready for a proper review now. Can't do much about testing with gtk-rs-core as it is still blocked by the gir rustdoc feature changes.

bilelmoussaoui:
Tested with gstreamer, works great.

src/analysis/object.rs
```diff
@@ -122,6 +133,21 @@ impl Info {
                 format!("prelude::{}", self.trait_name).into(),
                 self.trait_name.as_str().into(),
             ),
+            LocationInObject::VirtualExt | LocationInObject::VirtualExtManual => {
+                // TODO: maybe a different config for subclass trait name?
+                let trait_name = format!("{}Impl", self.trait_name.trim_end_matches("Ext"));
```
sdroege, line 138: We call it `ImplExt` and not `ExtImpl`

src/analysis/object.rs
```diff
@@ -122,6 +133,21 @@ impl Info {
...
+                (
+                    format!("subclass::prelude::{trait_name}").into(),
+                    trait_name.into(),
+                )
+            }
+            LocationInObject::ClassExt | LocationInObject::ClassExtManual => {
+                let trait_name = format!("{}SubclassExt", self.trait_name.trim_end_matches("Ext"));
```
sdroege, line 145: Do we have examples of this other than `ObjectSubclassExt`? What exactly is this good for?

Is this basically like https://gstreamer.pages.freedesktop.org/gstreamer-rs/stable/latest/docs/gstreamer/prelude/trait.ElementClassExt.html , i.e. for class methods? Then the naming here would be wrong

src/codegen/doc/gi_docgen.rs
```diff
@@ -196,6 +196,36 @@ fn ns_type_to_doc(namespace: &Option<String>, type_: &str) -> String {
     }
 }
 
+fn find_virtual_method(
```
sdroege, line 199:
```suggestion
fn find_virtual_method_by_name(
```

for consistency with the function just below?

src/analysis/object.rs
bilelmoussaoui: we have https://gtk-rs.org/gtk4-rs/git/docs/gtk4/subclass/widget/trait.WidgetClassSubclassExt.html  & https://gtk-rs.org/gtk4-rs/git/docs/gtk4/subclass/cell_area/trait.CellAreaClassSubclassExt.html. So probably the gstreamer one has to be renamed?

src/codegen/doc/gi_docgen.rs
```diff
@@ -196,6 +196,36 @@ fn ns_type_to_doc(namespace: &Option<String>, type_: &str) -> String {
...
+        in_type,
+        |f| {
+            f.name == mangle_keywords(name)
+                && namespace.as_ref().map_or(f.ns_id == MAIN_NAMESPACE, |n| {
+                    &env.library.namespaces[f.ns_id as usize].name == n
+                })
+        },
+        |o| {
+            type_.map_or(true, |t| {
+                o.name.trim_end_matches("Class") == t
```
sdroege, line 217: Isn't there something in the .gir that connects class and object?

sdroege: Also this would then need some additional strings for handling interfaces. Those are often called `XXXIface` or `XXXInterface`

src/codegen/doc/mod.rs
```diff
@@ -555,6 +558,62 @@ fn create_object_doc(w: &mut dyn Write, env: &Env, info: &analysis::object::Info
...
+                TypeStruct::new(SType::Trait, trait_name),
+                // Because we cannot sensibly deduce where the docs end up,
+                // assume they're outside the docs so that no `Self::` links
+                // are generated.  It is currently quite uncommon to specify
+                // the `{}Manual` trait, which would be ObjectLocation::ExtManual.
+                None,
+            )
+        } else if is_manual {
+            (
+                TypeStruct::new(SType::Trait, &format!("{}ImplManual", info.name)),
```
sdroege, line 579: What's this? :)

src/analysis/object.rs
sdroege: The GTK4 ones are wrong though. Not everything there is subclass specific and can also be called by other code on existing, external classes. E.g. `gtk_widget_class_query_action()`. So the naming seems wrong and the module it's placed in.

That's the same function pattern as for the GStreamer one. Those functions are called by normal application code.

All the setters in yours are a bit more tricky. Those should probably be in the subclass module, but then we need to somehow distinguish these two cases in gir. OTOH they could as well be outside the subclass module: you can only get a mutable class reference inside `class_init()` so nothing else can call it anyway.

sdroege:
Looks good to me apart from the naming things

src/codegen/doc/mod.rs
bilelmoussaoui: This is unused so far, maybe I should drop it until we get code generations

sdroege: Yeah. This is exactly the case that must not happen when we generate these things.

src/codegen/doc/gi_docgen.rs
bilelmoussaoui: > Isn't there something in the .gir that connects class and object?

The gir files contains this information yes, but gir itself doesn't seem to do anything with it yet.

bilelmoussaoui: Actually the trim here wasn't needed at all. not sure why i thought it was useful the time i wrote the code :)

sdroege: If the .gir file contains the information maybe we can use it here instead of heuristics?

PR #1464, opened 2023-04-30
docs/gi-docgen: quote arg identifiers with backticks

jf2048:
Looks like this was missed on the transition from gtk-doc. This is for things like `@self` and `@name` in https://gtk-rs.org/gtk4-rs/git/docs/gtk4/prelude/trait.WidgetExt.html#method.activate_action

And also avoid a copy when converting from `Cow<str>`

### DIFF
--- a/src/codegen/doc/format.rs
+++ b/src/codegen/doc/format.rs
@@ -151,16 +151,20 @@ fn replace_symbols(
                 s => panic!("Unknown symbol prefix `{s}`"),
             },
         });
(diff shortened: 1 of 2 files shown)


GuillaumeGomez (approved):

bilelmoussaoui:
Please don't merge this. There was a reason why I didn't made this a default behaviour of gi-docgen docs and it would need more thinking

jf2048:
After matrix discussion, let's prefer to replace these upstream

jf2048:
Upstream PR here https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/5897

PR #1469, opened 2023-05-08
analysis: Prefer prelude/exported crates

bilelmoussaoui:


src/analysis/enums.rs
```diff
@@ -46,15 +46,13 @@ pub fn new(env: &Env, obj: &GObject, imports: &mut Imports) -> Option<Info> {
...
             imports.add("glib::Quark");
-            imports.add("glib::error::ErrorDomain");
+            imports.add("glib::prelude::*");
         }
 
         let has_get_type = enumeration.glib_get_type.is_some();
         if has_get_type {
             imports.add("glib::Type");
-            imports.add("glib::StaticType");
-            imports.add("glib::value::FromValue");
```
sdroege, line 56: `FromValue` is not in the prelude IIRC

bilelmoussaoui: I have a following up PR for gtk-rs-core :)

sdroege: Ok, but don't add it to the prelude :) That trait is not supposed to be called by users and as such shouldn't be in the prelude.

Also, generally this here seems like the wrong approach. Ideally we'd want to have `user Trait as _` in the prelude (see also the issue about that). `gir` should simply provide the whole path to the traits.

bilelmoussaoui: > gir should simply provide the whole path to the traits

I switched to making gir generates the full path for the FromValue trait.

bilelmoussaoui:
The CI would fail until we update all the relevant crates

PR #1474, opened 2023-06-09
Update to bitflags 2.2

sdroege:


sdroege:
See https://github.com/gtk-rs/gtk-rs-core/pull/1052

GuillaumeGomez (approved):

sdroege:
Thanks for the review, will merge this once someone took care of https://github.com/gtk-rs/gtk-rs-core/pull/1052#discussion_r1223919994 (which involves more manual work than I have time for right now).

bilelmoussaoui:
I would love if we could merge this simultaneously with https://github.com/gtk-rs/gir/pull/1469 so we can drop the bitflags direct dependency from all the crates except glib.

sdroege:
Sounds good to me

PR #1475, opened 2023-06-09
Revert "Specify link attribute for Windows to be happy"

thiblahute:
This reverts commit a8dc21319bf6edaba19b2046ffd60c80fd91a2d7.

Linking flags should be provided by the system-deps crate. Specifying linking with the attribute creates issue when statically linking system dependencies.

### DIFF
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -25,18 +25,6 @@ pub fn generate(env: &Env) {
     save_to_file(&path, env.config.make_backup, |w| generate_lib(w, env));
 }
 
-fn write_link_attr(w: &mut dyn Write, shared_libs: &[String]) -> Result<()> {
-    for it in shared_libs {
-        writeln!(
-            w,
-            "#[link(name = \"{}\")]",
-            shared_lib_name_to_link_name(it)
-        )?;
-    }
-
-    Ok(())
-}
-
 fn generate_lib(w: &mut dyn Write, env: &Env) -> Result<()> {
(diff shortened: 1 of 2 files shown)


sdroege:
This will likely break things on Windows CC @seungha-yang.

Also after this it will be required to change some manual cases, e.g. in `glib/src/param_spec.rs`

sdroege:
Unfortunately none of us can remember the details of this and the PR description was not very... descriptive.

Can you try removing all these, and then try building
  1. gst-plugins-rs in MSVC
  2. Some Rust application using the bindings in MSVC

If that all works fine it's probably all OK.

thiblahute:
I update gst-rs with that commit and CI passes: https://gitlab.freedesktop.org/thiblahute/gstreamer-rs/-/pipelines/904700 (well apart from regen-check) .

I updated gst-plugins-rs on top of that and ci looks all good https://gitlab.freedesktop.org/thiblahute/gst-plugins-rs/-/pipelines/904754 

Do you think that is enough?

sdroege:
Please compile gst-plugins-rs with gtk-rs-core, gtk4 and gstreamer-rs regenerated with this *and* the manual `#[link]` attributes removed from the code too. That would be enough to convince me that this is not needed anymore :)

sdroege:
https://github.com/gtk-rs/gtk-rs-core/commit/d14883cbb9f56a5b63af0ab6a600a76d5c6a49c2 has some more background why this was added, and also https://github.com/gtk-rs/gtk-rs-core/pull/1199 is related.

PR #1482, opened 2023-07-09
codegen: Use async keyword for not in traits functions

bilelmoussaoui:
Keep the Pin<Box<T>> only for in trait async funcitons. Reduces the generated code a little bit :)


src/codegen/function.rs
```diff
@@ -224,19 +224,32 @@ pub fn declaration(env: &Env, analysis: &analysis::functions::Info) -> String {
...
+) -> String {
     let async_future = analysis.async_future.as_ref().unwrap();
 
     let return_str = if let Some(ref error_parameters) = async_future.error_parameters {
-        format!(
-            " -> Pin<Box_<dyn std::future::Future<Output = Result<{}, {}>> + 'static>>",
-            async_future.success_parameters, error_parameters
-        )
-    } else {
+        if in_trait {
```
sdroege, line 235: https://github.com/gtk-rs/gtk4-rs/pull/1421#discussion_r1257719379 FWIW. This will now create differently behaving API depending on whether it's in a trait or not

sdroege (commented):
Generally looks good but I don't like that the API behaves different between traits and non-traits, and also that it's "shape" is different.

Maybe for traits we can make use of the `async-trait` attribute macro for the time being then?

bilelmoussaoui:
> Maybe for traits we can make use of the async-trait attribute macro for the time being then?

That is a possibility yes, I can find some time for that later.

sdroege:
> That is a possibility yes, I can find some time for that later.

Should make *this* PR only simpler, I guess :)

bilelmoussaoui:
Now that async traits are stable, should we re-consider this and bump the minimal required rust version?

PR #1487, opened 2023-08-05
Fix docsrs

pentamassiv:
This PR includes a few improvements/cleanups regarding building docs on docs.rs:

- The `docs_rs_features` option in the `Gir.toml` was removed. I searched on Github if anybody was using it and I couldn't find anyone. It was also not documented so I don't think anyone used it. It would have complicated with the last point.
- Since the removal of the dox feature, a few rustc args need to get set in order to build the docs without linking the libraries. They are always the same so we can set them by default in the `Cargo.toml` of the sys crate.
- Automatically activate all features of the sys crate so that they all get documented.

### DIFF
--- a/src/codegen/sys/cargo_toml.rs
+++ b/src/codegen/sys/cargo_toml.rs
@@ -188,16 +188,10 @@ fn fill_in(root: &mut Table, env: &Env) {
             ])
         });
 
-        let docs_rs_features: Vec<String> = vec![];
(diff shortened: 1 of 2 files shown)


bilelmoussaoui (approved):

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

Issue #1488, opened 2023-08-06
Question: How to rename Object


pentamassiv:
I am currently generating bindings for GOA. It worked fine for the -sys crate, but I have an issue with the wrapper. Goa has a Goa.Object. When the code is generated, I have a naming conflict with the ObjectExt trait. There is one for the Goa.Object and one from glib. You can see the repo with the generated file [here](https://github.com/pentamassiv/goa-gir/blob/main/goa/src/auto/object.rs).

I thought the easiest would be to rename Object to GoaObject or something like that but I don't know how to rename it. I looked at the options in the Gir book and tried with 
```
[[object]]
name = "Goa.Object"
status = "generate"
doc_trait_name = "Goa.GOAObject"
```
but I get a `[ERROR libgir::config::error] "object Goa.Object": Unknown key: ["rename"]` when running gir. Also, if renaming is the way to go, are there any conventions on how to rename it?

bilelmoussaoui:
We don't support renaming objects, but you could rename you trait, similar to https://github.com/gtk-rs/gtk4-rs/blob/master/gtk4/Gir.toml#L1617

pentamassiv:
Awesome, that worked. Thank you

bilelmoussaoui:
Have you seen https://gitlab.gnome.org/World/Rust/libgoa-rs btw?

pentamassiv:
Yes, it is the reason why I want to generate the bindings. I want to open a PR to update them. They are very outdated

PR #1489, opened 2023-08-06
Add trait_name to API docs

pentamassiv:
I had a naming conflict with a generated trait and could not find a solution in the docs. Better to add it to the book to avoid the same [question](https://github.com/gtk-rs/gir/issues/1488) in the future


bilelmoussaoui:
Thank you

PR #1491, opened 2023-08-07
Exclude tests folder and Gir.toml, versions.txt files

pentamassiv:
Similar to https://github.com/gtk-rs/gtk-rs-core/pull/1158

The size of the crates can be reduced by excluding the `tests` folder and the `Gir.toml` + `versions.txt` files from being published to crates.io. All crates that are generated by gir have the two files and most will have a `tests` folder. They never have to get published to crates.io so gir can automatically generate that line for us. If for whatever reason they do need to be published, the entries can be removed from the list. The entry is never overwritten if it exists. There can also be an `include` list. If that is the case, no `exclude` is added.

While this in most cases will only save a few KB, the gtk, gstreamer and related crates do get a lot of downloads so I believe it is beneficial.

You can see which files will be included without publishing the crates by running cargo package --list.

### DIFF
--- a/src/codegen/sys/cargo_toml.rs
+++ b/src/codegen/sys/cargo_toml.rs
@@ -102,6 +102,16 @@ fn fill_in(root: &mut Table, env: &Env) {
         let package = upsert_table(root, "package");
         set_string(package, "build", "build.rs");
         // set_string(package, "version", "0.2.0");
+
+        if let toml::map::Entry::Vacant(_) = package.entry("include") {
+            package.entry("exclude").or_insert_with(|| {
+                Value::Array(vec![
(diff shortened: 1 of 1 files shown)


pentamassiv:
Another option would be to use
```toml
include = ["src/*", "build.rs", "Cargo.toml", "LICENSE", "COPYRIGHT"]
```
but that would probably is more error prone and would lead to more questions.

sdroege:
Same questions as for https://github.com/gtk-rs/gtk-rs-core/pull/1158#pullrequestreview-1565363425

pentamassiv:
I did not know packagers use the tests from crates.io and the other included files

PR #1502, opened 2023-08-16
config/codegen: Stop generating useless Display impls

bilelmoussaoui:
- objects: It uses to print the type name: not useful at all
- enums: It does print what a debug impl would print
- flags: It does print what a debug impl would print

So just get rid of that generated code and let the bindings generate a display impl manually when it makes sense

### DIFF
--- a/book/src/config/api.md
+++ b/book/src/config/api.md
@@ -21,10 +21,6 @@ deprecate_by_min_version = true
 # can also take path to the directory for saving "versions.txt" or filename with extension.
 # Relative to target_path
 single_version_file = true
-# Generation of Display trait enabled for all enums, classes, etc.,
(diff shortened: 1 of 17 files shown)


bilelmoussaoui:
Note that the special display implementations are kept, if the type has a `to_string` / `name` or similarly named methods.  I won't touch these

sdroege:
Yes keeping those is a good idea.

pbor:
could it be worth to have a `display_method = ... ` in Gir.toml instead of manually overriding?
Eg for `GFile` it would point to `parse_name`

(This can also be a follow up)

bilelmoussaoui:
should be good to go :)

sdroege (approved):

PR #1528, opened 2023-11-29
Remove unnecessary casts of function pointers through usize

sdroege:


sdroege:
Probably also relevant with regards to preserving provenance

GuillaumeGomez (approved):

PR #1530, opened 2023-12-13
record: Differentiate disguised vs pointer types

ocrete:
Disguised types are not pointers, they're just typedefs to a private struct, while pointer types are typedef to a pointer to a private struct. Generate the right code for each.

In C, there are 2 types of "disguised" typdefs:
`typedef struct _Blob *Blob`
and
`typedef struct _Blob Blob`

In Gir files, they're both called "disguised", but the first kind also has the "pointer" property set. The current generator only knows about the pointer type.

Currently, it generates 
`pub type Blob = *mut _Blob;`

But , it should do `pub type Blob = _Blob` for the non-pointer type.

The effect of this bug is that the rest of the code gen will think that the type is a pointer to a pointer.

So it will generate functions with a signature of `fn func(*mut *mut blob)` instead of `fn func(*mut blob)`

### DIFF
--- a/src/analysis/types.rs
+++ b/src/analysis/types.rs
@@ -108,7 +108,7 @@ impl IsIncomplete for Class {
 
 impl IsIncomplete for Record {
     fn is_incomplete(&self, lib: &Library) -> bool {
-        if self.c_type == "GHookList" || self.disguised {
+        if self.c_type == "GHookList" || self.disguised || self.pointer {
             // Search for GHookList in sys codegen for rationale.
             false
         } else {
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -452,7 +452,7 @@ fn generate_records(w: &mut dyn Write, env: &Env, records: &[&Record]) -> Result
             // 5. Thus, we use custom generated GHookList.
             //    Hopefully someone will profit from all this.
             generate_ghooklist(w)?;
-        } else if record.disguised {
+        } else if record.disguised || record.pointer {
             generate_disguised(w, env, record)?;
         } else {
             let align = config.and_then(|c| c.align);
@@ -499,7 +499,11 @@ fn generate_disguised(w: &mut dyn Write, env: &Env, record: &Record) -> Result<(
     cfg_condition(w, cfg_condition_, false, 0)?;
     generate_opaque_type(w, &format!("_{}", record.c_type))?;
     cfg_condition(w, cfg_condition_, false, 0)?;
-    writeln!(w, "pub type {name} = *mut _{name};", name = record.c_type)?;
+    if record.pointer {
+        writeln!(w, "pub type {name} = *mut _{name};", name = record.c_type)?;
+    } else {
+        writeln!(w, "pub type {name} = _{name};", name = record.c_type)?;
+    }
     writeln!(w)
 }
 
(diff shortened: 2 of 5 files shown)


sdroege (approved):

MarijnS95:
Thinking out loud, I'm curious if this would help out with mapping Vulkan types in `gir`.

bilelmoussaoui:
Was this actually tested? it makes the gtk4-rs build fail with the following
```
error[E0277]: the trait bound `_GtkConstraintTargetInterface: std::marker::Copy` is not satisfied
   --> gtk4/src/subclass/constraint_target.rs:11:38
    |
11  | unsafe impl<T: ConstraintTargetImpl> IsImplementable<T> for ConstraintTarget {
    |                                      ^^^^^^^^^^^^^^^^^^ the trait `std::marker::Copy` is not implemented for `_GtkConstraintTargetInterface`
    |
note: required by a bound in `glib::subclass::types::IsImplementable`
   --> /home/bilalelmoussaoui/.cargo/git/checkouts/gtk-rs-core-7be42ca38bd6361c/38f4ba1/glib/src/subclass/types.rs:287:42
    |
285 | pub unsafe trait IsImplementable<T: ObjectSubclass>: IsInterface
    |                  --------------- required by a bound in this trait
286 | where
287 |     <Self as ObjectType>::GlibClassType: Copy,
    |                                          ^^^^ required by this bound in `IsImplementable`

error[E0277]: the trait bound `_GtkNativeInterface: std::marker::Copy` is not satisfied
   --> gtk4/src/subclass/native.rs:10:28
    |
10  | unsafe impl<T: NativeImpl> IsImplementable<T> for Native {}
    |                            ^^^^^^^^^^^^^^^^^^ the trait `std::marker::Copy` is not implemented for `_GtkNativeInterface`
    |
note: required by a bound in `glib::subclass::types::IsImplementable`
   --> /home/bilalelmoussaoui/.cargo/git/checkouts/gtk-rs-core-7be42ca38bd6361c/38f4ba1/glib/src/subclass/types.rs:287:42
    |
285 | pub unsafe trait IsImplementable<T: ObjectSubclass>: IsInterface
    |                  --------------- required by a bound in this trait
286 | where
287 |     <Self as ObjectType>::GlibClassType: Copy,
    |                                          ^^^^ required by this bound in `IsImplementable`

error[E0277]: the trait bound `_GtkRootInterface: std::marker::Copy` is not satisfied
   --> gtk4/src/subclass/root.rs:10:26
    |
10  | unsafe impl<T: RootImpl> IsImplementable<T> for Root {}
    |                          ^^^^^^^^^^^^^^^^^^ the trait `std::marker::Copy` is not implemented for `_GtkRootInterface`
    |
note: required by a bound in `glib::subclass::types::IsImplementable`
   --> /home/bilalelmoussaoui/.cargo/git/checkouts/gtk-rs-core-7be42ca38bd6361c/38f4ba1/glib/src/subclass/types.rs:287:42
    |
285 | pub unsafe trait IsImplementable<T: ObjectSubclass>: IsInterface
    |                  --------------- required by a bound in this trait
286 | where
287 |     <Self as ObjectType>::GlibClassType: Copy,
    |                                          ^^^^ required by this bound in `IsImplementable`
```

sdroege:
Yes, the failures look correct. These are all interfaces that can't be implemented outside of GTK itself.

PR #1539, opened 2024-01-26
CI/tests: Remove gtk3 related bits

bilelmoussaoui:
As the crate is no longer maintained; don't use it as a blocker in our CI/tests


sdroege (approved):
Might want to add gtk4 or gst at a later time but this is already a good improvement :)

sdroege:
Can you add an issue for adding gtk4 and/or gst?

bilelmoussaoui:
> Might want to add gtk4 or gst at a later time but this is already a good improvement :)

It already regen & builds gtk4. But I didn't add it to the windows tests and didn't add a sys regen from scratch like we currently do for glib, I think we should also remove that :)

bilelmoussaoui:
I pushed another commit to fully phase out those old ffi regen tests

sdroege (approved):

PR #1540, opened 2024-01-28
Automatically assume that `win32_` and `unix_` should use the related cfg

GuillaumeGomez:
You can take a look at the result [here](https://github.com/gtk-rs/gtk-rs-core/compare/master...GuillaumeGomez:gtk-rs-core:regen-with-cfg?expand=1) (I didn't remove all unneeded `cfg_condition` yet).

### DIFF
--- a/src/update_cfgs.rs
+++ b/src/update_cfgs.rs
@@ -1,0 +1,101 @@
+use log::warn;
+
+pub fn get_cfg_condition(
+    object_name: &str,
+    cfg_condition: &Option<String>,
+    identifier_prefixes: &[String],
+) -> Option<String> {
+    let sub_object_name = identifier_prefixes
+        .iter()
+        .filter_map(|prefix| object_name.strip_prefix(prefix))
+        .find_map(|name| name.strip_prefix('_'))
+        .unwrap_or(object_name);
+
+    if sub_object_name.starts_with("win32_") {
+        match cfg_condition.as_deref() {
+            Some("windows") => {
+                warn!("\"object {object_name}\": No need to set `cfg_condition` to `windows` if name starts with `win32_`");
+                Some("windows".to_string())
+            }
+            None => Some("windows".to_string()),
+            Some(cfg) => Some(format!("{cfg},windows")),
+        }
+    } else if sub_object_name.starts_with("unix_") {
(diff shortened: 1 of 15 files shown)


src/update_cfgs.rs
```diff
@@ -0,0 +1,85 @@
+use log::warn;
+
+pub fn get_cfg_condition(object_name: &str, cfg_condition: &Option<String>) -> Option<String> {
```
sdroege, line 3: I'm surprised clippy does not warn about that. You usually want `Option<&str>`

sdroege: (for the parameter)

src/update_cfgs.rs
```diff
@@ -0,0 +1,85 @@
...
+    if sub_object_name.starts_with("Win32") {
+        match cfg_condition.as_deref() {
+            Some("windows") => {
+                warn!("\"object {object_name}\": No need to set `cfg_condition` to `windows` if object name starts with `Win32`");
+                Some("windows".to_string())
+            }
+            None => Some("windows".to_string()),
+            Some(cfg) => Some(format!("{cfg},windows")),
+        }
+    } else if sub_object_name.starts_with("Unix") {
```
sdroege, line 44: Maybe also UNIX (all-uppercase)?

sdroege:
Generated code seems fine to me

sdroege:
@GuillaumeGomez Are the are Windows/UNIX cfgs left in gtk-rs-core `Gir.toml`s after this?

GuillaumeGomez:
Yes. I only did it for gio iirc.

src/update_cfgs.rs
GuillaumeGomez: Well, in this case it mostly depends on what we have. And we do have `Option<String>`, so I think it's better to pass a reference to t rather than calling `.as_deref()` on every instance.

GuillaumeGomez:
Updated: it now takes into account the namespace the type comes from. Nothing to do for constants: their (non-c) name doesn't have a prefix apparently.

sdroege (approved):

GuillaumeGomez:
I'll send PRs to other projects next.

PR #1542, opened 2024-01-31
codegen/sys: Don't expect the crates are renamed

bilelmoussaoui:
In the past, this seemed like a decent idea but the switch to using cargo workspace proves it wasn't as we now have conflicting crates with the same names

### DIFF
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -41,6 +41,7 @@ fn generate_lib(w: &mut dyn Write, env: &Env) -> Result<()> {
     general::start_comments(w, &env.config)?;
     statics::begin(w)?;
 
+    generate_extern_crates(w, env)?;
     include_custom_modules(w, env)?;
     statics::after_extern_crates(w)?;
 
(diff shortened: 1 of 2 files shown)


src/codegen/sys/lib_.rs
```diff
@@ -104,6 +105,21 @@ fn generate_lib(w: &mut dyn Write, env: &Env) -> Result<()> {
     Ok(())
 }
 
+fn generate_extern_crates(w: &mut dyn Write, env: &Env) -> Result<()> {
+    for library in &env.config.external_libraries {
+        w.write_all(
+            format!(
+                "use {}_sys as {};\n",
```
sdroege, line 112: Why not just use `foo_sys` everywhere directly instead of going via the rename?

bilelmoussaoui: I tried to do that, but the whole code is a mess. So for now, I just want a single file to update whenever I do a release

sdroege: Ok, so this is just a hack for now to make things work and at a later time the whole renaming should be removed?

Maybe something @GuillaumeGomez wants to look at?

bilelmoussaoui:
The CI is expected to fail, the crates have to be adapted

src/codegen/sys/lib_.rs
bilelmoussaoui: > Ok, so this is just a hack for now to make things work and at a later time the whole renaming should be removed?

Yeah, just going back to the original thing we had

bilelmoussaoui:
Associated gtk4-rs PR https://github.com/gtk-rs/gtk4-rs/pull/1587

PR #1547, opened 2024-02-09
Revert Automatically assume that win32_ and unix_ should use the related cfg

GuillaumeGomez:
Reverts #1540, #1545, #1546.


sdroege:
@bilelmoussaoui What should we do here?

bilelmoussaoui:
@sdroege I think we should land this. The glib patch for splitting the API in separate gir files has landed so we would need to adapt our bindings anyways.

PR #1551, opened 2024-02-27
codegen: Fix nullable array in function params

### LINKED ISSUE #1133: `data` parameter for `gio::content_type_guess` should be an `Option`

bilelmoussaoui:
Co-authored-by: Andy Russell <arussell123@gmail.com>
Fixes #1133

Take over of https://github.com/gtk-rs/gir/pull/1136 except it only generates Option<T> if it is not a return type. To avoid generating `Option<Vec<T>>`

### DIFF
--- a/src/analysis/function_parameters.rs
+++ b/src/analysis/function_parameters.rs
@@ -107,6 +107,7 @@ pub enum TransformationType {
         array_name: String,
         array_length_name: String,
         array_length_type: String,
+        nullable: library::Nullable,
     },
     IntoRaw(String),
     ToSome(String),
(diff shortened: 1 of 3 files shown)


bilelmoussaoui:
Ok seems like gir doesn't give us a way to know if the array is zero-terminated or not...that makes it more

sdroege:
> To avoid generating Option<Vec<T>>

Why? That seems like the correct thing to do, even if it's inconvenient.

> Ok seems like gir doesn't give us a way to know if the array is zero-terminated or not...that makes it more difficult

It's an attribute in the XML though.

bilelmoussaoui:
> It's an attribute in the XML though.

I know, but it is not parsed nor passed around...

sdroege:
> I know, but it is not parsed nor passed around...

Uh... what?! :rofl: 

We assume always that it's zero-terminated unless there is an array length?

djmaze:
Any chance of getting forward on this? This prevents certain features from being used at all.

(For example: user scripts & stylesheets in the webkit6 crate. I needed to make local changes to the generated code in order to use those.)

PR #1563, opened 2024-04-01
Fix nightly clippy warnings

bilelmoussaoui:


src/analysis/functions.rs
```diff
@@ -512,7 +512,9 @@ fn analyze_callbacks(
         for (pos, typ) in to_replace {
             let ty = env.library.type_(typ);
             params[pos].typ = typ;
-            params[pos].c_type = ty.get_glib_name().unwrap().to_owned();
+            ty.get_glib_name()
+                .unwrap()
+                .clone_into(&mut params[pos].c_type);
```
sdroege, line 517: These `clone_into()` are not really more readable IMHO. Let's maybe not do that unless this lint actually becomes stable (which it probably won't anytime soon).

sdroege (commented):
Looks good otherwise

bilelmoussaoui:
Replaced with https://github.com/gtk-rs/gir/pull/1566

PR #1566, opened 2024-05-12
Codegen clippy fixes

bilelmoussaoui:


src/writer/to_code.rs
```diff
@@ -163,7 +163,7 @@ impl ToCode for Chunk {
                 }
                 let self_str = if in_trait { "Self, " } else { "" };
                 v.push(format!(
-                    "\tSome(std::mem::transmute::<_, unsafe extern \"C\" fn()>({trampoline}::<{self_str}F> as *const ())), Box_::into_raw(f))"
+                    "\tSome(std::mem::transmute::<*const (), unsafe extern \"C\" fn()>({trampoline}::<{self_str}F> as *const ())), Box_::into_raw(f))"
```
sdroege, line 166: We should change this API so that this transmute is not needed...

sdroege (approved):
I'm not a fan of the `clone_into()` but if it still builds everything...

bilelmoussaoui:
> I'm not a fan of the `clone_into()` but if it still builds everything...

Not a fan either, just applied `cargo clippy --fix` here, could silence the lint if that is your preference

bilelmoussaoui:
> I'm not a fan of the `clone_into()` but if it still builds everything...

Reverted and silenced the warnings instead

PR #1569, opened 2024-06-01
codegen: Stop renaming ffi crate

bilelmoussaoui:
As we would like to define those crates in the main workspace file to simplify the release process

Needs a gtk4-rs/gtk-rs-core regen first, to ensure everything is still working as expected.


sdroege (approved):
Good for me once you made it work for gtk-rs-core/gtk4-rs/gstreamer-rs and there are no bad surprises :)

Issue #1561, opened 2024-04-01
Add support for `glib:finish-func` and related attributes for GIO async APIs


sdroege:
Allows us to get rid of some heuristics

PR #1570, opened 2024-06-01
Support finish-func annotation

A6GibKm:
To get the correct name of the finish function if any.

### DIFF
--- a/src/analysis/bounds.rs
+++ b/src/analysis/bounds.rs
@@ -107,7 +107,11 @@ impl Bounds {
                 ));
                 if r#async && (par.name == "callback" || par.name.ends_with("_callback")) {
                     let func_name = func.c_identifier.as_ref().unwrap();
-                    let finish_func_name = finish_function_name(func_name);
+                    let finish_func_name = if let Some(finish_func_name) = &func.finish_func {
+                        finish_func_name.to_string()
+                    } else {
+                        finish_function_name(func_name)
+                    };
                     if let Some(function) = find_function(env, &finish_func_name) {
                         // FIXME: This should work completely based on the analysis of the finish()
                         // function but that a) happens afterwards and b) is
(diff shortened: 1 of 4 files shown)


src/analysis/bounds.rs
```diff
@@ -107,7 +107,11 @@ impl Bounds {
                 ));
                 if r#async && (par.name == "callback" || par.name.ends_with("_callback")) {
                     let func_name = func.c_identifier.as_ref().unwrap();
-                    let finish_func_name = finish_function_name(func_name);
+                    let finish_func_name = if let Some(finish_func_name) = &func.finish_func {
```
sdroege, line 110: In addition, if there is a finish function you can know that the function is async and get around the "is this function async" heuristics

A6GibKm: Done in a new commit.

A6GibKm:
Found this in the process https://gitlab.gnome.org/GNOME/glib/-/merge_requests/4099.

sdroege (approved):
Looks good to me but we probably have to wait until that ---^--- bug is fixed and we imported that gir?

bilelmoussaoui:
> Looks good to me but we probably have to wait until that ---^--- bug is fixed and we imported that gir?

Right, at worst we could land this after #1571 and not regen gtk-rs-core until the new gir file lands. So we can test it in gtk4-rs/gstreamer-sr

bilelmoussaoui:
The commit mentions it fixes https://github.com/gtk-rs/gir/issues/1561 but it doesn't cover all the other annotations. Even if we don't have an actual use of them now, it would be nice to parse the data so we can find potential breakages in the GIR files

A6GibKm:
> The commit mentions it fixes #1561 but it doesn't cover all the other annotations. Even if we don't have an actual use of them now, it would be nice to parse the data so we can find potential breakages in the GIR files

Added a commit I had stashed to parse them, but atm we do nothing with them.

bilelmoussaoui:
Needs a rebase

A6GibKm:
> Needs a rebase

rebased

PR #1571, opened 2024-06-01
Use getter and setter annotations

A6GibKm:


A6GibKm:
CI fails because the setter for `GtkImage:paintable` actually is `set_from_paintable`

sdroege (approved):
Looks good to me if @bilelmoussaoui is happy with the renegenerated gtk4-rs too

PR #1573, opened 2024-06-03
Various properties fixes

bilelmoussaoui:


bilelmoussaoui:
Can land as is but we wouldn't be able to make use of it until we get new gir files from gtk with fixed annotaitons

sdroege (approved):

PR #1574, opened 2024-06-03
members: drop useless alias config

bilelmoussaoui:
See commits.


bilelmoussaoui:
tested with gtk4-rs, works as expected

sdroege (approved):

PR #1577, opened 2024-06-09
Use gir-parser

bilelmoussaoui:
Switch from the in-house gir files parsing to a separate crate I built. It uses quick-xml internally. 

For now this is capable of generating almost all gtk bindings, most of the failures are arrays related.


src/parser.rs
```diff
@@ -112,145 +112,216 @@ impl Library {
...
+                ..Field::default()
+            });
+            union_count += 1;
+        }
 
-        let parent = elem
-            .attr("parent")
-            .map(|s| self.find_or_stub_type(ns_id, s));
+        let parent = elem.parent().map(|s| self.find_or_stub_type(ns_id, s));
         let typ = Type::Class(Class {
```
sdroege, line 319: Instead of using custom types for these things, could we move over to just using directly the types provided by your parser?

bilelmoussaoui: That would require a huge amount of changes to be done before doing the switch. I rather do that as a second step and my intention for now is to make the gir types wrappers of the ones from the types provided by the parser + TypeId + fields for overrides the type sets

sdroege: That makes sense

bilelmoussaoui:
> I have a branch at https://github.com/a6gibkm/gir/commits/bilelmoussaoui/gir-parser where I added a small test that showcases the difference between both parsers and it fails after the port, the problem is located in `read_array()`'s specifically, the argument passed to `c_array`, but I cannot figure out whats wrong.
> 
> The branch above can generate graphene with `python generator.py graphene` and can generate most of pango, I suspect the issue above is the only issue remaining here.

can you turn your branch into a PR so we can close this one and continue the conversation there?

A6GibKm:
Done at https://github.com/gtk-rs/gir/pull/1715.

PR #1597, opened 2024-09-05
generator.py: Use cargo run instead of guessing path

sophie-h:
Currently, the script just guesses a target path based on the default config. This fails if the project or the user has configured a different target directory.

This code is shorter and more correct.

### DIFF
--- a/generator.py
+++ b/generator.py
@@ -8,7 +8,6 @@ import asyncio
 
 DEFAULT_GIR_FILES_DIRECTORY = Path("./gir-files")
 DEFAULT_GIR_DIRECTORY = Path("./gir/")
-DEFAULT_GIR_PATH = DEFAULT_GIR_DIRECTORY / "target/release/gir"
 
 
 def run_command(command, folder=None):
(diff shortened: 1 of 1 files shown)


generator.py
```diff
@@ -30,8 +30,13 @@ async def spawn_process(exe, args):
     return stdout, stderr
 
 
-async def spawn_gir(gir_exe, args):
-    stdout, stderr = await spawn_process(gir_exe, args)
+async def spawn_gir(conf, args):
+    if conf.gir_path:
+        stdout, stderr = await spawn_process(conf.gir_path, args)
+    else:
+        cargo_args = ["run", "--release", "--quiet", "--manifest-path=gir/Cargo.toml", "--"]
```
MarijnS95, line 37: Shouldn't that be `DEFAULT_GIR_DIRECTORY`? Now you're assuming that custom projects always have a submodule named `gir`? Perhaps we better fold this in with the configuration parameter `gir_path` somehow?

sophie-h: > Shouldn't that be `DEFAULT_GIR_DIRECTORY`? 

Could make sense to change this. But I'm currently following the previous code where it was hardcoded.

> Now you're assuming that custom projects always have a submodule named `gir`? 

That's nothing new, it's the same as with the previous code:

```
def update_workspace():
    return run_command(["cargo", "build", "--release"], "gir")
```

> Perhaps we better fold this in with the configuration parameter `gir_path` somehow?

Well, that path is a binary path. Not sure how you would couple that with a source directory for gir.

generator.py
```diff
@@ -228,10 +222,6 @@ async def main():
     if not conf.gir_files_paths:
         update_submodule(DEFAULT_GIR_FILES_DIRECTORY, conf)
 
-    if conf.gir_path == DEFAULT_GIR_PATH:
-        update_submodule(DEFAULT_GIR_DIRECTORY, conf)
```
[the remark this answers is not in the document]
sophie-h: Thanks! I confused this with the weirdly named old function called `update_workspace`.

sdroege (approved):
Looks good to me. Any other comments?

generator.py
MarijnS95: By adding a "defaulted" "gir directory" path argument, and making it mutually exclusive with `--gir-path` perhaps? That way the default (hardcoded) "gir" suddenly becomes visible too.

MarijnS95 (approved):

MarijnS95:
One downside of this, found now that I started using this, is that `run_command("cargo build ...")` which was used previously _does not_ capture `stdout`/`stderr`. The newly used `stdout, stderr = await spawn_process(...)` obviously does,meaning that the first `` => Looking in path ``some-crate` `` now takes very long without any progress indication.

sophie-h:
I don't remember how the old solution solved that, but you could bring back a separate `cargo build` call and use `cargo run` after that. With this, the run shouldn't have to do build anymore.

MarijnS95:
Finally tackled in https://github.com/gtk-rs/gir/pull/1613.

PR #1598, opened 2024-09-07
ffi: Use std::ffi types instead of libc

A6GibKm:
These aliases were added in rust 1.64. Since we cannot replace all libc types at once it might make little sense to do this.


bilelmoussaoui:
What is the impact of this change on generated code?

A6GibKm:
> What is the impact of this change on generated code?

Replaces the uses of the libc types with std's, e.g.

```diff
modified   gtk4/src/auto/about_dialog.rs
@@ -346,7 +346,7 @@ impl AboutDialog {
             F: Fn(&AboutDialog, &str) -> glib::Propagation + 'static,
         >(
             this: *mut ffi::GtkAboutDialog,
-            uri: *mut libc::c_char,
+            uri: *mut std::ffi::c_char,
             f: glib::ffi::gpointer,
         ) -> glib::ffi::gboolean {
             let f: &F = &*(f as *const F);
modified   gtk4/src/auto/app_chooser_button.rs
@@ -238,7 +238,7 @@ impl AppChooserButton {
             F: Fn(&AppChooserButton, &str) + 'static,
         >(
             this: *mut ffi::GtkAppChooserButton,
-            item_name: *mut libc::c_char,
+            item_name: *mut std::ffi::c_char,
             f: glib::ffi::gpointer,
         ) {
             let f: &F = &*(f as *const F);
 ```

**Comment by A6GibKm** [comment]
> ```
> src/codegen/sys/statics.rs
> 21:        "use libc::{c_int, c_char, c_uchar, c_float, c_uint, c_double,",
> ```
> 
> You forgot to change this, otherwise good idea :)

Done, this is what you meant?

File: src/codegen/sys/statics.rs [hunk 1]
```diff
@@ -18,9 +18,9 @@ pub fn after_extern_crates(w: &mut dyn Write) -> Result<()> {
     let v = vec![
         "",
         "#[allow(unused_imports)]",
-        "use libc::{c_int, c_char, c_uchar, c_float, c_uint, c_double,",
-        "    c_short, c_ushort, c_long, c_ulong,",
-        "    c_void, size_t, ssize_t, time_t, off_t, intptr_t, uintptr_t, FILE};",
+        "use std::ffi::{c_int, c_char, c_uchar, c_float, c_uint, c_double,",
+        "    c_short, c_ushort, c_long, c_ulong, c_void};",
+        "use libc::{size_t, ssize_t, time_t, off_t, intptr_t, uintptr_t, FILE};",
```
sdroege, line 23:
```suggestion
        "#[allow(unused_imports)]",
        "use libc::{size_t, ssize_t, time_t, off_t, intptr_t, uintptr_t, FILE};",
```

PR #1600, opened 2024-09-15
Remove link attribute generation

eerii:
Follow up to #1475. Motivated by work at `gtk-rs-core` and `gstreamer-rs` to override library linking.

At the moment a `#[link]` attribute is being generated for every `extern "C"` code block. This is not generally needed since `system-deps` already handles linking libraries, and it makes overriding dependencies difficult from higher up crates.

The only exception is one edge case when using Windows and static constants, but that is a limitation of the compiler. To handle this case, allow to use custom `build.rs` scripts by not overriding the package field.

All of the context for this change and all of the alternatives we considered are explained in gtk-rs/gtk-rs-core#1508.

## Changes

- Remove `#[link]` attribute generation.
- Allow custom build scripts. If the `package.build` key doesn't exist in `Cargo.toml`, generate it to be `build.rs`. Otherwise keep the existing value.

## Caveats

Something to consider is whether removing the `#[link]` generation could affect other crates. From an overview of other crates relying on `gir`, it seems that they are not linking static constants, just functions, which do not explicitly need dllimport. If this is expected to be a problem, maybe a command line flag like `--link` could be added to the generator script that allows customizing this behaviour.

### DIFF
--- a/src/codegen/sys/cargo_toml.rs
+++ b/src/codegen/sys/cargo_toml.rs
@@ -99,7 +99,9 @@ fn fill_empty(root: &mut Table, env: &Env, crate_name: &str) {
 fn fill_in(root: &mut Table, env: &Env) {
     {
         let package = upsert_table(root, "package");
-        set_string(package, "build", "build.rs");
+        package
+            .entry("build")
+            .or_insert_with(|| Value::String("build.rs".into()));
         // set_string(package, "version", "0.2.0");
     }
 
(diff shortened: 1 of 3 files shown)


pbor:
See also https://github.com/gtk-rs/gir/pull/1207

eerii:
> See also #1207

I missed that, but yes, a toggle like that would be very useful indeed, thank you for pointing it out! The custom build script part is still relevant because in some very specific cases you need to do some work to make Windows happy.

sdroege (approved):

PR #1601, opened 2024-09-19
[book] Add note to use ffi

jreidinger:
Hi,
at first let me thanks you for this nice project that is quite useful. I use it to setup some first draft for using gobject API from rust for my project and when I follow instructions in book and failed to compile it. After debugging I found that `use ffi;` in lib.rs solve my issue with generated code and I think it should be part of book so others do not need to face it.
Thanks


bilelmoussaoui:
there are probably other things that should be added there but this seems important enough to merge it for now
thanks!

jreidinger:
Thanks for merging it. I also consider if I should add `pub use auto::traits::*;` which I also needs for interfaces, but it is partly covered in handling errors and there is suggested solution to use prelude, so I go with minimal patch which is critical to make basically any generated code works.

PR #1602, opened 2024-09-24
Add underscore to Enum/Flag names starting with digit

antunonline:
This had to be done in order to enable of genration of NetworkManager library bindings (not all bindings)

### DIFF
--- a/src/analysis/imports.rs
+++ b/src/analysis/imports.rs
@@ -8,6 +8,7 @@ use std::{
 
 use super::namespaces;
 use crate::{library::Library, nameutil::crate_name, version::Version};
+use crate::nameutil::mangle_crate;
 
 fn is_first_char_up(s: &str) -> bool {
     s.chars().next().unwrap().is_uppercase()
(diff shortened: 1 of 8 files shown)


bilelmoussaoui:
That is an issue on the C api, it should be fixed instead.

antunonline:
This is a part of NetworkManager API which has been in a such state for a long time. 
I'm not sure that I can influence them to change it, nor I have the time to analyze how NM works and prepare pull requests for it. 

I would rather add ability to circumvent this issues on gir side and create pull requests here. 

I know that the pull requests are not perfect, but if we can find some compromise without me spending a lot of time, enabling other uses to also create NetworkManager bindings, I would rather to that.

bilelmoussaoui:
Sure, as long as it is documented as such

> I know that the pull requests are not perfect, but if we can find some compromise without me spending a lot of time, enabling other uses to also create NetworkManager bindings, I would rather to that.

You can also change the gir file to fix it without changing the API upstream

antunonline:
This is quite a complicated procedure and I'm not sure that it's the proper way to do it, because then every user should
understand how it all works, and repeat the procedure from the start. 

It's not a copy and run experience. 

I think that I'll have to take some time to dive deeper into the project and prepare proper patches to support it 
without adding technical debt.

antunonline:
I'll close it for now, until I prepare better code

Issue #1590, opened 2024-07-23
gtk-4 generated builders should panic if GTK hasn't been initialized


anna-hope:
Currently, trying to use some builders before initializing GTK can cause a segfault. Instead, the generated builder should check if GTK has been initialized, and panic if not.

<details>
<summary>Minimal example</summary>

```rust
use gtk4::prelude::*;
use gtk4::{Application, ApplicationWindow, glib};

const APP_ID: &str = "me.annahope.gtk4-rs-segfault";

fn build_ui(app: &Application) {
    let window = ApplicationWindow::builder()
        .application(app)
        .title("Segfault")
        .build();  // Segfaults if GTK hasn't been initialized.

    window.present();
}

fn main() -> glib::ExitCode {
    let app = Application::builder().application_id(APP_ID).build();
    // app.connect_activate(build_ui);
    build_ui(&app); // Segfaults.
    app.run()
}
```

</details>

I tried fixing this naively by adding a check to the builder codegen code https://github.com/anna-hope/gir/commit/a3ee2522917a49578fb654cf7a9523d5396ae9a3, but then realized that we definitely can't do this for all builders because:

- Some builders, like https://github.com/gtk-rs/gtk4-rs/blob/8850a124cccded5f9b7483c88b1891cafe30e5f9/gtk4/src/auto/application.rs#L103 actually can/need to be called before GTK is initialized
- Others, like https://github.com/gtk-rs/gtk4-rs/blob/8850a124cccded5f9b7483c88b1891cafe30e5f9/gdk4/src/auto/cursor.rs#L140 aren't even part of gtk4 itself, so it doesn't make sense for them to have this check (and they can't do it anyway)

I briefly investigated the `builder_properties` to see if we could set some field there, but couldn't find an easy way to add one. I could look into it further, but I am brand new to this project and GTK overall, so perhaps there is a better way to accomplish this that I don't know about.

bilelmoussaoui:
The solution is actually pretty easy, doesn't require anything like you did. Just add `assert_initialized_main_thread!();` macro call there.

PR #1607, opened 2024-10-18
codegen/builder: Assert the main thread is initalized

bilelmoussaoui:
Fixes https://github.com/gtk-rs/gir/issues/1590

### DIFF
--- a/src/codegen/object.rs
+++ b/src/codegen/object.rs
@@ -14,12 +14,13 @@ use super::{
 use crate::{
     analysis::{
         self, bounds::BoundType, object::has_builder_properties, record_type::RecordType,
-        ref_mode::RefMode, rust_type::RustType,
+        ref_mode::RefMode, rust_type::RustType, safety_assertion_mode::SafetyAssertionMode,
     },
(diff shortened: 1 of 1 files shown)


src/codegen/object.rs
```diff
@@ -456,6 +457,11 @@ fn generate_builder(w: &mut dyn Write, env: &Env, analysis: &analysis::object::I
     pub fn build(self) -> {name} {{",
         name = analysis.name,
     )?;
+    writeln!(
+        w,
+        "{}",
+        safety_assertion_mode_to_str(SafetyAssertionMode::InMainThread)
```
bilelmoussaoui, line 463: Don't ask me why this was done like this...

sdroege: What exactly?

sdroege (approved):

src/codegen/object.rs
bilelmoussaoui: having to call safety_assertion_mode_to_str instead of just implementing Display or a method in the trait.

PR #1613, opened 2024-11-05
generator.py: Restore pre-building `gir` with build/progress output

MarijnS95:
In commit c6ba79ed ("generator.py: Use cargo run instead of guessing path") the `cargo build` command on the default (hardcoded) `"gir"` directory was replaced with a single `cargo run` for correctly finding the right executable.

This isn't an issue given that `cargo run` also builds the tool, and above all makes it easier / more correct to find the binary as the primary reason for that commit, but it does make us lose the build log as a progress indicator (thanks to `--quiet` and capturing `stderr`). In addition, it causes the first `` => Looking in path `xxx` `` item to take a very long time without aforementioned progres indication, making the user wonder what is going on.

By building `gir` again with a separate `cargo build` invocation before running, this issue is alleviated somewhat.

### DIFF
--- a/generator.py
+++ b/generator.py
@@ -7,7 +7,7 @@ import sys
 import asyncio
 
 DEFAULT_GIR_FILES_DIRECTORY = Path("./gir-files")
-DEFAULT_GIR_DIRECTORY = Path("./gir/")
+DEFAULT_GIR_DIRECTORY = Path("./gir/") # TODO: This is typically the directory _of this Python script_ (which external projects symlink to)
 
 
 def run_command(command, folder=None):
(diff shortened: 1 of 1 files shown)


sdroege:
@sophie-h What do you think?

generator.py
```diff
@@ -7,7 +7,7 @@
 import asyncio
 
 DEFAULT_GIR_FILES_DIRECTORY = Path("./gir-files")
-DEFAULT_GIR_DIRECTORY = Path("./gir/")
+DEFAULT_GIR_DIRECTORY = Path("./gir/") # TODO: This is typically the directory _of this Python script_ (which external projects symlink to)
```
MarijnS95, line 10: The TODO here means: should we come up with a different way, instead of "guessing/hardcoding" this _relative directory_?

MarijnS95:
@sophie-h what are your thoughts on this?

sophie-h:
Sorry for the delayed response. Looks like the solution I originally suggested. So :+1: from my side

PR #1617, opened 2024-11-13
Fix reference / non-reference mismatches for functions with multiple …

sdroege:
…closures that are not scope-call


sdroege:
Doesn't cause any changes in gtk-rs-core / gtk4-rs, and fixes on mis-generated function in GStreamer.

PR #1620, opened 2024-12-02
Generate `Into<Option<_>>` in argument position where applicable

fengalin:
The [`TryFromGlib` PR](https://github.com/gtk-rs/gir/pull/1092) used `impl Into<Option<_>>` for `TryFromGlib` arguments which resolve to an `Option`. This improved usability, e.g.:

```rust
// without `impl Into<Option<_>>`, we had to:
pipeline.set_latency(Some(5.seconds()));

// with `impl Into<Option<_>>`, we can:
pipeline.set_latency(5.seconds());
```

However, other types were not supported, so we still needed to write e.g.:

```rust
pipeline.use_clock(Some(clock));
```

# Re-generation of the main code bases

This turned out to impact many parts of the generator. I split the result into 3 commits:

* The first two are reworks which address limitations around `RustType`, `Bound` & `to_glib_extra` and the way they were used (or not used in some code paths). These reworks only impact the commented out generated functions (and fix a shortcoming in `gstreamer-rs`).
* The 3d commit is the actual implementation of this PR.

The result can be observed as 2 re-generations in:

* gtk-rs-core: https://github.com/gtk-rs/gtk-rs-core/pull/1590
* gtk-4: https://github.com/gtk-rs/gtk4-rs/pull/1929
* gtstreamer-rs: https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/-/merge_requests/1606

See individual commits and specific repository PRs/MR.


bilelmoussaoui:
I haven't had the time to look into in details, but we would need the same treatment for the generated Builders as well :)

fengalin:
Builder generation updated.

fengalin:
Postponing these changes to a future Rust bindings generator rewrite

PR #1622, opened 2024-12-07
Don't skip `Option` for nullable collections

### LINKED ISSUE #1133: `data` parameter for `gio::content_type_guess` should be an `Option`

fengalin:
Nullable collection were represented as a `Vec<_>` or `[_]`, but an empty collection is not the same as an undefined collection. Rust code can express this using an `Option<_>`.

Don't do this for return and output values though: in practice the C implementations return NULL when the size is 0. So it doesn't make much sense using an `Option<_>` in that case.

Finally, Gir now parses the `zero-terminated` annotation which allows detecting functions returning an array for which the length is not defined by a length argument. Those functions used to require an explicit `manual` declaration in the `Gir.toml` and are automatically commented out now.

Sample regen results (no changes for GStreamer):

* `gtk-rs-core`: https://github.com/fengalin/gtk-rs-core/commit/4f0786b2c3b09f08762c7f482f888046a7f1d9b3
* `gtk4-rs`: https://github.com/fengalin/gtk4-rs/commit/4bf3298cacb6654b6834c50153c4095d72a2648f

Fixes: https://github.com/gtk-rs/gir/issues/1133
See also: https://github.com/gtk-rs/gtk-rs-core/issues/1257


sdroege:
> [ ] Add support for `glib::value` `Option<Vec<_>>` conversions.

I think this also involves untangling our existing implementations a bit as they currently map `NULL` to empty arrays, and possibly also the other way around?

sdroege (approved):
This looks too easy, I didn't expect that :)

sdroege:
There's also https://github.com/gtk-rs/gir/pull/1551 doing more or less the same (but a bit more?). I think the main problem here now is fixing `glib/translate.rs`

sdroege:
@fengalin What's the plan to continue here, especially with the translation traits?

fengalin:
My plan was to start over from the PR you linked and implement the changes in glib. But I got involved in other stuffs. I'll try to resume this later this week.

fengalin:
Status on this:

- [x] Rebased
- [x] Started over from the commit from #1551
- [x] Added basic handling of the `zero-terminated` annotation. Note: I also considered handling `optional` & `allow-none` annotations, but they seem to actually indicate whether the API allows the caller to pass `NULL` for an in-out argument.

Quoting the last commit:

>    Nullable collection were represented as a `Vec<_>` or `[_]`, but an empty
>    collection is not the same as an undefined collection.
>
>    Don't do this for return and output values though: in practice the C
>    implementations return NULL when the size is 0. So it doesn't make much
>    sense using an `Option<_>` in that case.
>
>    Finally, Gir now parses the `zero-terminated` annotation which allows
>    detecting functions returning an array for which the length is not defined
>    by a length argument. Those functions used to require an explicit `manual`
>    declaration in the `Gir.toml` and are automatically commented out now.

Sample regen results (no update for GStreamer):

* `gtk-rs-core`: https://github.com/fengalin/gtk-rs-core/commit/4f0786b2c3b09f08762c7f482f888046a7f1d9b3
* `gtk4-rs`: https://github.com/fengalin/gtk4-rs/commit/4bf3298cacb6654b6834c50153c4095d72a2648f

The outcome is that [`content_type_guess()`](https://github.com/fengalin/gtk-rs-core/blob/4f0786b2c3b09f08762c7f482f888046a7f1d9b3/gio/src/auto/functions.rs#L182) is generated as expected by #1133:

```rust
pub fn content_type_guess(
    filename: Option<impl AsRef<std::path::Path>>,
    data: Option<&[u8]>,
) -> (glib::GString, bool) { .. }
```

`data` is optional, which could makes sense. Do we want this as an API instead of passing an empty array? I don't really know.

Because similar annotations don't necessarily mean that the argument can be an `Option`, I'm unsure we want this applied systematically anyway. Take [`gsk4::GskStroke::set_dash`](https://github.com/fengalin/gtk4-rs/blob/4bf3298cacb6654b6834c50153c4095d72a2648f/gsk4/src/auto/stroke.rs#L70) for instance:

```rust
pub fn set_dash(&self, dash: Option<&[f32]>) { .. }
```

We could think passing `None` would discard a previous dash for current `GskStroke`. But the [documentation](https://docs.gtk.org/gsk4/method.Stroke.set_dash.html) doesn't mention this possibility.

Another example is [`gdk4::ContentFormats::new()`](https://github.com/fengalin/gtk4-rs/blob/4bf3298cacb6654b6834c50153c4095d72a2648f/gdk4/src/auto/content_formats.rs#L21).

```rust
pub fn new(mime_types: Option<&[&str]>) -> ContentFormats { .. }
```

The [doc](https://docs.gtk.org/gdk4/ctor.ContentFormats.new.html) doesn't say anything about a difference between an empty array and an undefined argument.

See the [`gtk4-rs` regen commit](https://github.com/fengalin/gtk4-rs/commit/4bf3298cacb6654b6834c50153c4095d72a2648f) for more examples.

Another approach could be to not generate the `Option` for these use cases and let the bindings override this (using `optional = true` or `nullable = true`) in their `Gir.toml`. @sdroege @bilelmoussaoui what do you think?

src/xmlparser.rs
```diff
@@ -242,7 +242,7 @@ impl XmlParser<'_> {
 
     fn end_document(&mut self) -> Result<(), String> {
         match self.next_event()? {
-            XmlEvent::EndDocument { .. } => Ok(()),
+            XmlEvent::EndDocument => Ok(()),
```
bilelmoussaoui, line 247: can be split into a separate PR

src/parser.rs
```diff
@@ -13,6 +13,14 @@ use crate::{
 
 const EMPTY_CTYPE: &str = "/*EMPTY*/";
 
+#[derive(Debug, Default)]
+struct ParsedType {
```
bilelmoussaoui, line 17: The naming here confused me a bit, could use a small comment on what it is.

fengalin:
@bilelmoussaoui: when you get a chance, I'd be interested by your opinion on the question in this comment https://github.com/gtk-rs/gir/pull/1622#issuecomment-2897061677

bilelmoussaoui:
> Another approach could be to not generate the Option for these use cases and let the bindings override this (using optional = true or nullable = true) in their Gir.toml. @sdroege @bilelmoussaoui what do you think?

I would generate the Option<T> and let the user override if they want to disable that. As we to tend to trust the nullability annotations

src/parser.rs
fengalin: Updated

fengalin:
> can you open PRs for the changes for gtk4-rs/gtk-rs-core? i will review and merge this one afterwards.

Done!

* https://github.com/gtk-rs/gtk-rs-core/pull/1728
* https://github.com/gtk-rs/gtk4-rs/pull/2064

src/analysis/return_value.rs
```diff
@@ -120,19 +120,32 @@ pub fn analyze(
...
+            && par.lib_par.array_length.is_none()
         {
-            used_types.extend(rust_type.into_used_types());
-        }
+            // The array length for the return value of this function can not be inferred
+            // from any argument. How the length should be inferred is usually specified
+            // in the doc of the function, so there's nothing we can do automatically.
+            // Currently the functions matching this condition are explicitly marked
+            // `manual = true` in the matching Gir.toml.
+            // FIXME: `FixedArray`s could be handled automatically
```
sdroege, line 132: Is that hard to add? Also I thought we have some heuristics here somewhere if there's a parameter called `len`, `size` or so

src/codegen/properties.rs
```diff
@@ -103,7 +103,7 @@ fn declaration(env: &Env, prop: &Property) -> String {
         let dir = library::ParameterDirection::Return;
         let ret_type = RustType::builder(env, prop.typ)
             .direction(dir)
-            .nullable(prop.nullable)
+            .nullable(prop.nullable & !prop.is_array)
```
sdroege, line 106: Which APIs is this relevant with?

src/parser.rs
```diff
@@ -1311,20 +1324,45 @@ impl Library {
...
+        if let Some(mut elem_type) = typ {
+            if elem_type.is_array {
+                // Note: these could be handled in the `analysis` part, however that would be quite
+                //       annoying because many code places check the parsed `nullable` instead of
+                //       the `analysis` result. Gir necessitates a rewrite...
+                use ParameterDirection::*;
+                match direction {
+                    Return | Out | InOut => {
+                        // Some arrays in return position come with nullable=true & zero-terminated=false
+                        // annotations. In practice, the C implementations return NULL when the size is 0.
```
sdroege, line 1336: You mean with `nullable=false` and `zero-terminated=false`? That's actually an interesting case because not returning NULL in that situation is actually hard for C code (`malloc(0)` is implementation defined behaviour IIRC).

For `nullable=false` and `zero-terminated=true` it's IMHO a bug in the C code and it should return a pointer that dereferences to `NULL`.

src/parser.rs
```diff
@@ -1311,20 +1324,45 @@ impl Library {
...
+                        // So it doesn't make much sense using an `Option<_>` by default for nullable arrays.
+                        nullable = false
+                    }
+                    _ => {
+                        // We can only generate a Rust `Option` from a nullable array annotation
+                        // if the array is not zero terminated. Otherwise we would need an extra
+                        // indication to differentiate an empty array from an undefined value,
+                        // which should be indicated in the doc and would need manual implementation.
```
sdroege, line 1344: I don't understand this part. With `zero-terminated=true` it actually seems clear how to handle nullability but not otherwise (see above comment)

sdroege:
This looks like fewer changes and cleaner than I expected

sdroege:
Do we have example API somewhere with nullable array return values?

src/analysis/return_value.rs
fengalin: It should not be too hard, but that's a different issue from the one this PR is trying to fix imho.
https://github.com/gtk-rs/gir/issues/1655

src/parser.rs
fengalin: Yes, I'm not sure what led me to this comment & condition... I must have been too sleepy :sweat_smile: 
Updated and also updated the [`gtk-rs-core`](https://github.com/gtk-rs/gtk-rs-core/pull/1728) & [`gtk4`](https://github.com/gtk-rs/gtk4-rs/pull/2064) PRs.

sdroege:
Do we have examples of array returns that are *not* nullable btw?

src/parser.rs
sdroege: * In parameters: `Option` if nullable, otherwise not
* Return/out parameters: `Option` if nullable and zero-terminated, no `Option` if nullable and not zero-terminated (and obviously no `Option` if not nullable)

fengalin:
> Do we have examples of array returns that are _not_ nullable btw?

Samples from `gtk-rs-core`:

* [`gdk_pixbuf_save_to_buffer`](https://docs.gtk.org/gdk-pixbuf/method.Pixbuf.save_to_buffer.html)
  > Note that the buffer is not `NUL`-terminated and may contain embedded `NUL` characters.
  > The length of the array is specified in the `buffer_size` argument.
* [`g_buffered_input_stream_peek_buffer`](https://docs.gtk.org/gio/method.BufferedInputStream.peek.html)
  > Peeks in the buffer, copying data of size `count` into `buffer`, offset `offset` bytes.

The following functions don't come with an explicit `nullable=true`, but it seems the type `char**` / `const gchar* const *` implies it:

* [`g_action_group_list_actions`](https://docs.gtk.org/gio/method.ActionGroup.list_actions.html)
  > A `NULL`-terminated array of the names of the actions in the group.
* [`g_app_info_get_supported_types`](https://docs.gtk.org/gio/method.AppInfo.get_supported_types.html)
  > If this information is not provided by the environment, this function will return `NULL`.
* [`g_app_launch_context_get_environment`](https://docs.gtk.org/gio/method.AppLaunchContext.get_environment.html)
  > This is a `NULL`-terminated array of strings, where each string has the form `KEY=VALUE`
* [`g_application_command_line_get_arguments`](https://docs.gtk.org/gio/method.ApplicationCommandLine.get_arguments.html)
  > The return value is `NULL`-terminated
* [`g_application_command_line_get_environ`](https://docs.gtk.org/gio/method.ApplicationCommandLine.get_environ.html)
  > [...] as would be returned by `g_get_environ()`, ie as a `NULL`-terminated list of strings in the form ‘NAME=VALUE’.
* [`g_desktop_app_info_get_keywords`](https://docs.gtk.org/gio-unix/type_func.DesktopAppInfo.get_keywords.html)
  > An array of `utf8`
* [`g_desktop_app_info_get_string_list`](https://docs.gtk.org/gio/method.DesktopAppInfo.get_string_list.html)
  > An array of `utf8`
  > a `NULL`-terminated string array or `NULL` if the specified key cannot be found.

The following function return a `guchar*`:

* [`g_dbus_message_to_blob`](https://docs.gtk.org/gio/method.DBusMessage.to_blob.html)
  > an array of `guchar` [...] `NULL` if `error` is set.

The following functions return a `GSList*`:

* [`gdk_pixbuf_get_formats`](https://docs.gtk.org/gdk-pixbuf/type_func.Pixbuf.get_formats.html)
* [`g_app_info_get_all`](https://docs.gtk.org/gio/type_func.AppInfo.get_all.html)
* [`g_app_info_get_all_for_type`](https://docs.gtk.org/gio/type_func.AppInfo.get_all_for_type.html)
* [`g_app_info_get_fallback_for_type`](https://docs.gtk.org/gio/type_func.AppInfo.get_fallback_for_type.html)
* [`g_app_info_get_recommended_for_type`](https://docs.gtk.org/gio/type_func.AppInfo.get_recommended_for_type.html)
* [`g_dbus_interface_skeleton_get_connections`](https://docs.gtk.org/gio/method.DBusInterfaceSkeleton.get_connections.html)
* [`g_dbus_object_get_interfaces`](https://gnome.pages.gitlab.gnome.org/libsoup/gio/GDBusObject.html#g-dbus-object-get-interfaces)
* `g_dbus_object_manager_get_objects`

sdroege:
> Samples from `gtk-rs-core`:
> [...]

These are just byte arrays though, they can't really be NUL-terminated (and have an explicit size).

`gdk_pixbuf_save_to_buffer()` has an `GError**`. Functions that have `GError**`s have an implicit nullable return value (ugly but that's how it is). If you explicitly annotate them with `nullable` that means that they can *also* return `NULL` without an error.

`g_buffered_input_stream_peek()` is an in-parameter and not a return value so irrelevant in that regard. And passing `NULL` here is invalid so shouldn't be an `Option`.

----

> The following functions don't come with an explicit `nullable=true`, but it seems the type `char**` / `const gchar* const *` implies it:
> [...]

So these are interesting because they have some complications around `zero-terminated=1`. Let's discuss those first.

See e.g. 978ca4731cfd4d562e884d2dcf27c9af22b99628 in GLib. The annotation was recently added.

For most of the others there is an annotation in the code but it is not showing up in the .gir file. That seems like a bug. It should probably a) show up, b) it seems to be the default for `char**` so gir should also add it by default if the annotation is not specified. Can you create an issue for that against gobject-introspection?

For `g_application_command_line_get_arguments()`, that's missing the annotation but also it has an explicit length. It's marked as not zero-terminated in the gir file (wrong), it's probably added in this case because of the explicit length parameter. That needs fixing of some sort. Similar `g_desktop_app_info_get_string_list()` is annotated correctly and has both of the attributes in the gir file, so that should be how `g_application_command_line_get_arguments()` should be handled.

`g_desktop_app_info_get_keywords()` is missing the annotation.

----

Now for the nullability of those functions.  `g_app_info_get_supported_types()` is documented to return NULL so needs the annotation. Same for `g_application_command_line_get_environ()` and `g_desktop_app_info_get_string_list()`. Can you create an MR for that?

The others are not documented to return `NULL` so don't need the annotation. If they can actually return `NULL` then they need the annotation.

---- 

> The following function returns a `guchar*`:

That should be treated as a byte array. Has explicit length parameter, can't be NUL-terminated.

That returns NULL if an error happens, which is fine. Functions that have `GError**`s have an implicit nullable return value (ugly but that's how it is). If you explicitly annotate them with `nullable` that means that they can *also* return `NULL` without an error.

So that seems fine.

----

> The following functions return a `GSList*`:

`GList` and `GSList` are special. `NULL` is the empty list and always allowed for this type.

----

> Do we have example API somewhere with nullable array return values?

`gsk_stroke_get_dash()` returns a float array with explicit size, and returns `NULL` if empty. That shouldn't become an `Option` as outlined above. Same for `pango_font_face_list_sizes()`.

`gdk_event_get_history()` is the same but slightly more complicated because it returns a "boxed inline" type (in our terminology) instead of a fundamental type like an int or float.

fengalin:
Postponing these changes to a future Rust bindings generator rewrite

PR #1632, opened 2025-02-09
Improved the documentation for derive clauses in API configuration

xanathar:
As very briefly discussed on `#rust:gnome.org`, this updates the documentation for specifying explicit `#[derive(...)]` clauses in `Gir.toml`.

As anticipated on the chat, the behavior of derives is a bit inconsistent depending on the type. This documentation should reflect the current reality (to the best of my knowledge).

### DIFF
--- a/book/src/config/api.md
+++ b/book/src/config/api.md
@@ -298,8 +298,24 @@ version = "3.12"
     version = "3.18"
 ```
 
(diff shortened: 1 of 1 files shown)


sdroege (approved):

sdroege:
A bit confusing, thanks for documenting. We should probably clean up the behaviour at some point to be more consistent between types.

PR #1635, opened 2025-02-21
Changes needed for rust 2024 edition

A6GibKm:
These were tested with https://gitlab.gnome.org/World/Rust/libadwaita-rs/-/merge_requests/94 and it solves the remaining issues. There are more changes that are likely needed. This would benefit from https://github.com/gtk-rs/gtk-rs-core/issues/1663.


sdroege:
Does this continue to compile without warnings (and clippy lints) with 1.80?

A6GibKm:
Libadwaita-rs: Yes, gtk4-rs: Not even close.

sdroege:
Then we can't really merge this until we update our MSRV to 1.85

bilelmoussaoui:
Can we get this merged?

bilelmoussaoui:
> If all applications need to be updated to 2024 edition (to compile, or to compile without warnings) that's a lot of churn quite late in the cycle. I don't know if that's the case and @A6GibKm's answers to my question were not really conclusive.

if all that has to be fixed is mostly unsafe calls, there wouldn't be that many in apps anyway

PR #1637, opened 2025-02-22
Adds FromStr and TryFrom<&str> traits to enums implementations

xanathar:
This adds the generation of `FromStr` and `TryFrom<&str>` trait impls for enums.

- Contrary to FromGlib / IntoGlib, I removed the `doc(hidden)` attribute so that the traits get documented by cargo doc
- As discussed on matrix, the error type is `BoolError`
- The parsing is based on the enum member's nick; for members with no nick specified, the original "name" is used

This does **not** use a configuration parameter. After implementing it, I'm leaning on the opinion that a configuration flag should be provided, as it's not entirely unlikely for somebody to have implemented FromStr or TryFrom in their `/manual/` module, and if that implementation does not use the same nicks it can be a little painful to migrate.

## Not part of this PR
These are not in the PR, might be desired in future PRs:
- Same support for bitfields (might conflict with bitfield's default parser)
- `impl Display for $enum` for the reverse operation

## gtk-rs-core
Tested gtk-core-rs with this and it seems to work. Also a couple of personal projects build just fine.
If/when this gets merged, I can follow-up with a PR on gtk-core-rs.

### DIFF
--- a/src/codegen/enums.rs
+++ b/src/codegen/enums.rs
@@ -82,6 +82,7 @@ fn generate_enum(
     struct Member<'a> {
         name: String,
         c_name: String,
+        nick: String,
         version: Option<Version>,
         deprecated_version: Option<Version>,
         cfg_condition: Option<&'a String>,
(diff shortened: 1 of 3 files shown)


bilelmoussaoui:
Can you do a patch for gtk-rs-core/gtk4-rs already so we can see if the changes looks good and work?

xanathar:
Of course, just give me a few days.

PR #1640, opened 2025-03-07
Address clippy warning

Ytsejam76:
NUL-terminated  strings are  currently  generated  as b"<string>\0"  for signal names rather than as c"<string>" as recommended in:

https://rust-lang.github.io/rust-clippy/master/index.html#manual_c_str_literals

### DIFF
--- a/src/writer/to_code.rs
+++ b/src/writer/to_code.rs
@@ -150,7 +150,7 @@ impl ToCode for Chunk {
                         r#"let detailed_signal_name = detail.map(|name| {{ format!("{signal}::{{name}}\0") }});"#
                     ));
                     v.push(format!(
-                        r#"let signal_name: &[u8] = detailed_signal_name.as_ref().map_or(&b"{signal}\0"[..], |n| n.as_bytes());"#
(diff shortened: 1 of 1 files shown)


src/writer/to_code.rs
```diff
@@ -150,7 +150,7 @@ impl ToCode for Chunk {
                         r#"let detailed_signal_name = detail.map(|name| {{ format!("{signal}::{{name}}\0") }});"#
```
bilelmoussaoui, line 150: You forgot this one, no?

Ytsejam76: Indeed. Fixed it, thanks.

Ytsejam76:
I rebased it but checks are still failing (outside the changes though)... Should I do something?

bilelmoussaoui:
> I rebased it but checks are still failing (outside the changes though)... Should I do something?

The checks are failing for the generated code using those changes, so yes. Something has to be fixed.

sdroege:
Seems not very useful or consistent, but also doesn't hurt :)

Ytsejam76:
> Seems not very useful or consistent, but also doesn't hurt :)

Indeed, but it's handy for me since it removes quite a few clippy warnings from my code.

PR #1641, opened 2025-03-10
parser: Add doc format support

bilelmoussaoui:
Replaces the current use_gi_docgen key

We would need https://github.com/gtk-rs/gir-files/pull/237 to land first but that is blocked on more core libraries defining the doc:format properly.

### DIFF
--- a/book/src/config/api.md
+++ b/book/src/config/api.md
@@ -14,7 +14,6 @@ target_path = "."
 # auto_path = "src/auto"
 work_mode = "normal"
 # Whether the library uses https://gitlab.gnome.org/GNOME/gi-docgen for its documentation
-use_gi_docgen = false
 generate_safety_asserts = true
 deprecate_by_min_version = true
 # With this option enabled, versions for gir and gir-files saved only to one file to minimize noise,
@@ -123,7 +122,7 @@ visibility = "pub" # or 'crate' / 'private' / 'super'
 # works for flags and enums. You have to pass the "GIR" member name.
 default_value = "fill"
 # Change the name of the generated trait to e.g avoid naming conflicts
-trait_name = "TraitnameExt" 
+trait_name = "TraitnameExt"
 # In case you don't want to generate the documentation for this type.
 generate_doc = false
     # define overrides for function
@@ -336,7 +335,7 @@ status = "generate"
     name = "stock_list_ids"
     # allows to ignore global functions
     ignore = true
-    # allows to define if the function was moved to a trait 
+    # allows to define if the function was moved to a trait
     doc_trait_name = "StockExt"
     # allows to define if the function was moved to a struct
     doc_struct_name = "Stock"
--- a/src/codegen/doc/format.rs
+++ b/src/codegen/doc/format.rs
@@ -8,7 +8,9 @@ use super::{gi_docgen, LocationInObject};
 use crate::{
     analysis::functions::Info,
     library::{FunctionKind, TypeId},
-    nameutil, Env,
+    nameutil,
+    parser::DocFormat,
+    Env,
 };
 
 const LANGUAGE_SEP_BEGIN: &str = "<!-- language=\"";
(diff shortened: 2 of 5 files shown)


bilelmoussaoui:
Obviously this was untested, so please don't merge.

src/codegen/doc/format.rs
```diff
@@ -138,7 +140,7 @@ fn replace_symbols(
     env: &Env,
     in_type: Option<(&TypeId, Option<LocationInObject>)>,
 ) -> String {
-    if env.config.use_gi_docgen {
+    if env.library.doc_format == DocFormat::GiDocgen {
```
A6GibKm, line 143: Perhaps for now this could be a `if library.doc_format == DocFormat::GiDocgen || config.use_gi_docgen` for the moment.

bilelmoussaoui: that is an option yes, although, it will end up staying there given no one will remember to remove it. As this option only impacts docs generation, i rather have bindings manually amend the format into the gir file then gir keeping this old "hack"

A6GibKm: My fear is that this is a bit of a chicken and egg situation. Breaking gir generation is a discouragement for setting doc_format and this MR depends on libraries using it. 

I suppose an alternative is to have a prequel for this PR that just reads the doc_format and does nothing with it (instead of crashing) to unblock the situation.

bilelmoussaoui: It is always been like that when it comes to new annotations so i rather fix things properly.

book/src/config/api.md
```diff
@@ -336,7 +335,7 @@ status = "generate"
     name = "stock_list_ids"
     # allows to ignore global functions
     ignore = true
-    # allows to define if the function was moved to a trait 
+    # allows to define if the function was moved to a trait
```
A6GibKm, line 338: These could go in a separate commit.

A6GibKm:
Please link https://gitlab.gnome.org/GNOME/gobject-introspection/-/merge_requests/401 in the commit message.

src/parser.rs
```diff
@@ -11,6 +11,31 @@ use crate::{
     xmlparser::{Element, XmlParser},
 };
 
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub enum DocFormat {
+    GtkDocMarkdown,
+    GtkDocDocbook,
+    GiDocgen,
+    HotDoc,
```
A6GibKm, line 19: This should be Hotdoc if the key is not hot-doc.

book/src/config/api.md
```diff
@@ -14,7 +14,6 @@ target_path = "."
 # auto_path = "src/auto"
 work_mode = "normal"
 # Whether the library uses https://gitlab.gnome.org/GNOME/gi-docgen for its documentation
-use_gi_docgen = false
```
sdroege, line 17: It's going to take a while until everything uses this so for the time being maybe change `use_gi_docgen` to `doc_format` and allow the same values there, and don't make the attribute in the XML mandatory yet?

bilelmoussaoui: What is everything? It is GTK, GLib, Pango and libadwaita (merged already https://gitlab.gnome.org/GNOME/libadwaita/-/merge_requests/1428).

The remaining bindings don't get updated often at all. 

If you just want to ignore the element for now, it could be done in a separate patch by someone else.

sdroege: If everything using gi-docgen is updated then that's fine with me. Just default to "unknown" if the attribute is not in the XML.

werdahias:
I build gir 0.20.5 with this MR added as patch; though I still get the Unexpected element <format>" error.
Any ideas as to why ? This is with glib 2.84.0 as testcase

bilelmoussaoui:
Did the changes to ignore the attribute in https://github.com/gtk-rs/gir/pull/1645, other changes could be done separately later to not block other packages forever & we can backport it easily.

PR #1675, opened 2025-09-18
Update rustdoc-stripper to 0.1.19

jbicha:
This dependency used the "git" version [many years ago](https://github.com/gtk-rs/gir/commit/7011eeb), but we can just use a released version instead now. This also helps Debian's build tools.

~I didn't touch Cargo.lock~


bilelmoussaoui:
Please update Cargo.lock

sdroege:
Just running `cargo check` after this change should do that already

sdroege (approved):

PR #1705, opened 2026-01-24
Support static libraries

willwwarner:
This does what is necessary for static library bindings to work.
This is used by [libgnome-games-support-rs](https://gitlab.gnome.org/World/Rust/libgnome-games-support-rs).

### DIFF
--- a/src/codegen/sys/lib_.rs
+++ b/src/codegen/sys/lib_.rs
@@ -1,6 +1,6 @@
 use std::{
     fs,
-    io::{Error, Result, Write},
+    io::{Result, Write},
 };
 
 use log::info;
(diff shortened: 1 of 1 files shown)


willwwarner:
> > Even for static libraries, you want the -sys crate to "link" it so that the static library ends up being linked into the final build result.
> 
> Yes, I removed the linking step because I wanted to get my bindings to build without the subproject. I will fix that.

I have removed that part of this PR.

willwwarner:
> you want the -sys crate to "link" it so that the static library ends up being linked into the final build result.

I did some investigating a while ago, and IIRC, libgnome-games-support never creates a pkg-config file, so it needs to link a different way than is built in to gir, but making that optional is out of scope for this.

PR #1715, opened 2026-02-16
Use gir-parser

A6GibKm:
This is able already to generate graphene's bindings without any difference to the current version with `python generator.py graphene`.

As far as I know, the remaining issues are related to the read_array and read_type fns. These can be seen on e.g. pango bindings. See https://github.com/gtk-rs/gir/pull/1577#issuecomment-3905049921.

Continuation of https://github.com/gtk-rs/gir/pull/1577.


src/analysis/function_parameters.rs
```diff
@@ -477,16 +482,16 @@ fn detect_length<'a>(
 }
 
 fn is_length(par: &library::Parameter) -> bool {
```
bilelmoussaoui, line 484: This is something we could add to upstream gir-parser btw.

bilelmoussaoui:
I will try to have a look at this tomorrow and hopefully we can land it

src/analysis/function_parameters.rs
bilelmoussaoui: Done that in https://github.com/bilelmoussaoui/gir-parser/commit/bad66f5e4d95f295df0ae204dde8090b53596e67

bilelmoussaoui:
Can you split the commits up to the one that switches to using gir-parser? we can merge them already

bilelmoussaoui:
while at it, i think https://github.com/gtk-rs/gir/blob/c316bb9fc346731f0a36103a32e6d0898191919b/src/library.rs#L137-L155 are not used and can be removed/replaced with plain bools like for Nullable.

bilelmoussaoui:
Closing in favour of the new attempt

PR #1718, opened 2026-03-14
Fix two clippy warnings in generated code.

Ytsejam76:
1. Emit `#[derive(Default)]` with a `#[default]` attribute on the variant
instead of a manual `impl Default` block. Falls back to the manual impl
when the default variant has a version or `cfg` condition.

2. `GObject` constructors often return a base class rather than the concrete
type, which triggers `clippy::new_ret_no_self`. Suppress the warning on
generated `new()` methods when the return type differs from `Self`.

### DIFF
--- a/src/codegen/enums.rs
+++ b/src/codegen/enums.rs
@@ -115,6 +115,19 @@ fn generate_enum(
         });
     }
 
+    // Check if we can use #[derive(Default)] instead of a manual impl.
+    // This is possible when the default member has no version or cfg condition.
+    let derive_default_member = config.default_value.as_ref().and_then(|default_value| {
+        let m = members
+            .iter()
+            .find(|m| m.name == enum_member_name(default_value))?;
+        if m.version.is_none() && m.cfg_condition.is_none() {
+            Some(m.name.clone())
+        } else {
+            None
+        }
+    });
+
     cfg_deprecated(
         w,
         env,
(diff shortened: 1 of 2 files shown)


bilelmoussaoui:
Could you share a branch showing the changes on a repo like https://github.com/gtk-rs/gtk-rs-core or gtk4-rs?

Ytsejam76:
> Could you share a branch showing the changes on a repo like https://github.com/gtk-rs/gtk-rs-core or gtk4-rs?

Sure, this should be it: [wip/clippy_warnings_2026](https://github.com/Ytsejam76/gtk-rs-core/commit/7b9f67ceb2a522112857749104dd2a08825cd614)

bilelmoussaoui:
Seems like it doesn't change anything in gtk-rs-core at least?

Ytsejam76:
> Seems like it doesn't change anything in gtk-rs-core at least?

It seems so. Perhaps those warnings weren't there in the first place. On my codebase it does change the bindings and removes the warnings. 
I will try to see what happens with gtk4-rs

Ytsejam76:
> Seems like it doesn't change anything in gtk-rs-core at least?

I don't see any change on gtk4-rs either: https://github.com/Ytsejam76/gtk4-rs

But, if I’m not mistaken, cargo clippy didn’t report any warnings before either.

bilelmoussaoui:
Weird.

PR #1719, opened 2026-03-19
postprocessing: handle missing c:type for CArray fields

jpwhiting:


book/book.toml
```diff
@@ -1,7 +1,6 @@
 [book]
 authors = ["The gtk-rs Project Developers"]
 language = "en"
-multilingual = false
```
sdroege, line 4: Can you move this into a separate PR?

src/library_postprocessing.rs
```diff
@@ -328,13 +328,22 @@ impl Library {
...
                                 actions.push((tid, fid, Action::SetCType(array_c_type)));
                                 continue;
                             }
+                            if let Type::CArray(..) = field_type {
+                                // Flexible arrays / VLA-like fields may not carry a field-level
+                                // c:type in GIR even when the inner type is known.
+                                // We set a non-pointer sentinel to satisfy the field c:type
+                                // requirement in sys field codegen.
+                                let array_c_type = "c_array".to_owned();
+                                actions.push((tid, fid, Action::SetCType(array_c_type)));
```
sdroege, line 344: Can you show the diff this causes in the output? I assume for gtk-rs-core / gtk4-rs there's nothing, and for GStreamer it generates `gst_analytics::Tensor` correctly now?

book/book.toml
jpwhiting: Yep, moved here: https://github.com/gtk-rs/gir/pull/1720

src/library_postprocessing.rs
jpwhiting: I checked this by running gir before and after the patch (c316bb9f vs b7c64019) on GStreamer configs. The generated code diff is effectively nil (only versions.txt generator hash changed), but behavior changes from an error to a handled fallback: before it errored on Field Tensor::dims is missing c:type, now it warns and assumes c_array. Also, GstAnalytics.Tensor is configured as manual in both normal and sys Gir configs, so this patch doesn’t newly auto-generate that type; it makes postprocessing more robust for missing field-level c:type on C arrays.

sdroege: Makes sense. It still can't be generated anyway

sdroege: What happens if we were trying to generate it now, what's the output (or error)?

jpwhiting: No errors, but manual is still needed for these reasons.

Generated high-level file gstreamer-rs/gstreamer-analytics/src/auto/tensor.rs contains the key Tensor APIs as commented/unimplemented (new_simple, check_type, dims, set_simple) because CArray is still unimplemented in codegen output for these signatures.

Generated sys layout in lib.rs maps dims as a single size_t field, while the current manual modeling in manual.rs provides the flexible trailing-array behavior needed by the wrapper.

Current manual API in tensor.rs remains the only complete/ergonomic one (dims() slice, new_simple, etc.).

jpwhiting: Oh, that wasn't the answer to your question. One moment I'll try without this change and get the current output.

`[WARN  libgir::library_postprocessing] Field Tensor::dims missing c:type assumed to be c_array` is what you get with the Tensor not being manual with the current gir code.

sdroege: That also didn't answer my question but I can just check that myself :)

```
... 361 #[derive(Copy, Clone)]
... 362 #[repr(C)]
... 363 pub struct GstTensor {
... 364     pub id: glib::GQuark,
... 365     pub layout: GstTensorLayout,
... 366     pub data_type: GstTensorDataType,
... 367     pub data: *mut gst::GstBuffer,
... 368     pub dims_order: GstTensorDimOrder,
... 369     pub num_dims: size_t,
... 370     pub dims: size_t,
... 371 }
```

So that's not very correct unfortunately. But closer than before, not too much missing.

PR #1731, opened 2026-06-12
codegen: Fix version guards when a type has multiple interfaces imple…

bilelmoussaoui:
…mentations

Fixes an issue when generating gtk4-rs

cc @sdroege


bilelmoussaoui:
The ci will fail because we need the gtk4-rs pr to be merged as well

sdroege (approved):

PR #1733, opened 2026-06-13
Use gir-parser

bilelmoussaoui:
Generates valid code with 0 diff in gtk-rs-core at least.

Take over from previous attempts at #1577 #1715


sdroege:
Sad that it's only `+2.262-2.723`, I hoped for more code to disappear :)

But let's get this in then if it doesn't cause any differences in here anymore so it gets wider testing?

sdroege (approved):

bilelmoussaoui:
This doesn't get rid of the existing structs, it just feeds them from gir_parser for now. The problem is that gir by itself contains a lot of weird decisions and cleaning those up would be necessary before getting rid of more code :)

sdroege:
Probably makes more sense to do that in a second step anyway

PR #1735, opened 2026-06-15
parser: Parse both union/fields at the same time

bilelmoussaoui:
To perserve their order

@sdroege should fix the gstreamer issue


sdroege:
Still moves unions to the front

sdroege (approved):

PR #1747, opened 2026-07-21
codegen: generate system-deps 8 for sys crates

calvaris:
system-deps was released to crates.io as major 8.0.0 (adds the prebuilt binary bundle feature). The generated sys crates' [build-dependencies] must follow so the whole gtk-rs graph can unify on a single system-deps major; the entry point used by the generated build.rs, system_deps::Config::new().probe(), is unchanged between 7 and 8.


thiblahute:
@sdroege Could you check that one please?

PR #1753, opened 2026-08-30
For different namespaces, compare the parent version against the exte…

sdroege:
…rnal library's minimum version instead

And if no minimum version is configured then assume a guard is not needed.

----

CC @A6GibKm


sdroege:
Breaks sourceview5 again.

sdroege:
Let's do this differently by configuring the min version in gstreamer.

